PR #5, opened 2016-02-05
Readme branch

krisma:
Readme has been modified to include instruction on loading the project from Xcode.


README.md
```diff
@@ -35,6 +35,14 @@ Installation is now complete. You can view the current state of the application
 
     $ ionic emulate ios --target="iPhone-6"
 
+Note that sometimes the IOS emulator doesn't work when called from command line, then you can simply load
```
shankari: Can you put a section header here, something like "Troubleshooting"?

app-settings.json
```diff
@@ -0,0 +1,28 @@
+[
```
shankari: This is an autogenerated file, should not be checked in.

shankari:
There's a bunch of autogenerated files this PR (app-settings.json, .DS_Store).
Please remove them
    $ git rm --cached <file>
and commit again.

Committing from the same branch will update the same pull request.
It would be super awesome if you could edit the `.gitignore` to exclude them.
https://help.github.com/articles/ignoring-files/

PR #9, opened 2016-02-12
Kris branch

krisma:
I modified the summary and consent view. It's confirmed that summary looks good. I changed consent accordingly but then I was unable to load in emulator (I tried 10 times at least, for summary, it sometimes took 5 times reload but eventually would work). So it's not guaranteed that consent.html looks good enough. Also the bug stopped me from proceeding any further(login.html and popup window on "I refuse").


shankari:
I agree that the development process is annoying with the multiple reloads, so I spent a bunch of time refactoring the existing plugins to hopefully make things better. In particular, I moved the native setup code into the `pluginInitialize` methods, and added promises to ensure that all the javascript setup calls are complete before starting to load the UI.

This has made the situation much better for me - when I run from xcode, I have had everything load correctly 10/10 times. Unfortunately, when I run from the command line, it still fails fairly quickly (second or third time when I tested).

I wonder if this is related to the automatic icon generation - running from the command line regenerates the icons automatically and running from xcode does not. And somebody claimed that when they put the correct icons in, their white screen problem went away.

So maybe one of the tasks for next week can be to finalize an icon, and then we can add the icon and see if the errors go away.

Let's see if you can preview the consent with my changes tomorrow and then I can merge the change.

PR #18, opened 2016-02-21
Kris branch

krisma:
Including: style changes in intro, consent, and login.
Not including: popup view that user puts in their google account, since I could find that part of code.
Issues: I noticed that for device older than or equal to iPhone5, the user will be brought to intro every time he or she starts the app. This happens at least to me.
Future: I may customize the tab bar and setting view. Partly because since I can't log in in developer mode, the tabs other than setting are not shown.


shankari:
Thanks for the pull request!

Some comments:
- Can you send screenshots of the new screens? The code seems fine, but since these are all cosmetic changes, a picture is worth a thousand words.
- Our app won't work for models older than iPhone5s anyway because they don't have the M7 co-processor and don't support activity detection.
- the popup view that puts the user into their google account is a native screen that is part of the google open source framework. I don't think we should muck with it
- Changing the css files was a quick way to get you familiar with the code. But I don't think that making the settings screen beautiful is the current highest priority since we do not think that the users will interact with it a lot. Instead, I think that you should focus on the travel diary and result screens, and the ability to mark whether the trips are correct or not.

Thoughts?

krisma:
1. Screenshot. 
Sure I'll do that next time. I didn't do much change after the last time I showed you on the meeting.
2. Focus
I agree that the more important part is diary and result. However I still can't get to the screen I saw on your device. When I login with my gmail, I get JavaScript level error.
Can you help with that?
3. Mark trip correctness
One possible way to do this is to allow user to mark the wrong one with one delete button. And if the back end is restful, just remove the trip. To avoid mis-click we either asks for confirmation on front end, or mark it as "to be removed" by adding an attribute on the back end and remove all marked record every night with cron job.

shankari:
One last question before I merge:

1. Did you intend to remove all those entries from the .gitignore? I am pretty sure that the platform directory should still be ignored. Can you explain further?
2. Ah. You basically need to run the server on your laptop so that you can do end to end testing. Let me update the README to reflect that
3. So we need to think a bit more carefully about the trip correctness.
 * At a high level, since this is a research project, we never want to delete anything. Instead, if something is incorrect, we should store it with an "incorrect" tag of some sort so that it is part of the ongoing dataset.
 * I am not sure that there is one kind of "wrong" trip. For example:
    * a trip may have the wrong mode
    * it may have wrong start and end points
    * it may have the wrong trajectory, e.g. zigzags

Ideally, we would want to know what was wrong, and have some way to indicate what the correct thing was. This may be less important in the case of trajectories, where you can just look at them and see what is wrong, and is more important in the case of modes where we have no idea what happened.

Also, we should not delete a trip just because it is wrong - for e.g. if the mode is wrong, the rest of the trip is still perfectly good.

So I think that it would be good to capture the following about each "wrong" trip - why it is wrong and what is the correct thing? Users should be able to specify this in as few clicks as possible.

I'm updating the README now - ETA for completion is tonight :)

krisma:
1. Uh my bad. I didn't mean to change gitignore since I learned from my last mistake and now I only manually git add whatever I changed. But I thought that .gitignore will be merged with the older one as well so I'll be only adding terms to it instead of deleting.

2. Thanks for the update. I got a project due tonight. Hopefully I'll try that tomorrow before our meeting. 

3. I got your point. I suggested deleting since I tend to choose the way that's easier to implement. Now I have a new suggestion due to the points you mentioned. We could have a report button, for example question mark icon in yellow color, and when the user click that button we show a popup letting user choose between: wrong mode, wrong start end point. And for wrong mode we can have another question asking what the right one is in the same popup. For the wrong start end point we show a map instead and ask the user to pin out the right ones by first bringing the map camera to the wrong start end point. One example is the app I saw today which pay its users to take survey about their preference. I didn't dig into its business model but the UI looks like what we can learn from.

shankari:
> Uh my bad. I didn't mean to change gitignore since I learned from my last mistake and now I only manually git add whatever I changed. But I thought that .gitignore will be merged with the older one as well so I'll be only adding terms to it instead of deleting. 

Just do
```
$ git checkout .gitignore
```
and commit + push again.

I don't want to merge this `.gitignore` since it is incorrect.

shankari:
> Thanks for the update. I got a project due tonight. Hopefully I'll try that tomorrow before our meeting.

Good luck on the project!

> We could have a report button, for example question mark icon in yellow color, and when the user click that button we show a popup letting user choose between: wrong mode, wrong start end point. And for wrong mode we can have another question asking what the right one is in the same popup. For the wrong start end point we show a map instead and ask the user to pin out the right ones by first bringing the map camera to the wrong start end point. 

Yeah that's closer to what I was thinking. We should think through the various things that can go wrong and figure out how to collect the correct version for them. Also, we want to collect red/green (good/bad) reports at various points.

PR #44, opened 2016-06-21
Temp branch

krisma:
Implemented: Pull to refresh
Improved: font-size for list and detail view changed to "em" unit from "px". Text size is fine tested on iPhone 4s. (No 5c model available in simulator)
To be done: Leaflet doesn't load correctly when the size is specified in percentage instead of real # of pixels so I still need to figure out how to achieve responsive resizing for map. 

screen shot 2016-06-21 at 12 33 16 am
screen shot 2016-06-21 at 12 33 10 am


shankari:
According to this:
http://leafletjs.com/examples/mobile.html
% is supported at least for height.

Another workaround is to read the browser width and set the width based on that in javascript.
https://www.drupal.org/node/2086539

Or try applying this patch, since we have a patched leaflet anyway. You'll have to find where that code is in the concatenated leaflet.js that we include.
https://www.drupal.org/files/issues/map-height-relative-2086539-9.patch

shankari:
ok to test

shankari:
Or you could try searching/posting an issue in the leaflet repo.
It looks like the maintainer is really responsive.

www/js/diary/list.js
```diff
@@ -155,7 +155,9 @@ angular.module('emission.main.diary.list',['ui-leaflet', 'nvd3ChartDirectives',
     */
 
     $scope.refresh = function() {
+      if ($ionicScrollDelegate.getScrollPosition().top < 5) {
```
shankari: Can you add an issue to investigate this further? It seems really wrong that it is required.

www/templates/diary/detail.html
```diff
@@ -3,20 +3,20 @@
...
-                <div style="font-size: 13px; padding-top: 5px; padding-bottom: 5px;">{{getFormattedTime(tripgj.data.properties.end_ts)}}
+                <div style="font-size: 1em; padding-top: 5px; padding-bottom: 5px;">{{getFormattedTime(tripgj.data.properties.end_ts)}}
                 </div>
             </div>
 
             <div class="col-70">
 
-                <div class="item item-icon-left" href="#" style="font-size: 13px; padding-top: 5px; padding-bottom: 5px;padding-left: 25px; margin-top: 0; margin-bottom: {{getEarlierOrLater(tripgj.data.properties.start_ts, tripgj.data.id) == ''? '0':'20px'}};">
+                <div class="item item-icon-left" href="#" style="font-size: 1em; padding-top: 5px; padding-bottom: 5px;padding-left: 25px; margin-top: 0; margin-bottom: {{getEarlierOrLater(tripgj.data.properties.start_ts, tripgj.data.id) == ''? '0':'20px'}};">
```
shankari: shouldn't the padding also be in `em` instead of `px`?

www/js/main.js
```diff
@@ -50,19 +40,320 @@ angular.module('emission.main', ['emission.main.diary', 'emission.main.common',
...
+          controller: 'logCtrl'
+        }
+    }
+
   });
 
   $ionicConfigProvider.tabs.style('standard')
   $ionicConfigProvider.tabs.position('bottom');
 })
+.controller('appCtrl', function($scope, $ionicModal, $timeout) {
```
shankari: I think that this is going backwards. We don't want to have giant files that have all the controllers smushed in. I think that you should keep the controllers in their original file and include the module here.

www/js/main.js
```diff
@@ -50,19 +40,320 @@ angular.module('emission.main', ['emission.main.diary', 'emission.main.common',
...
+          controller: 'mapCtrl'
+        }
+      }    
+  })
+  .state('root.main.log', {
+    url: '/log',
+    views: {
+        'main-control': {
+          templateUrl: 'templates/recent/log.html',
+          controller: 'logCtrl'
```
shankari: These controllers don't have to be in the same file - they can also be in the a module that you include.

www/js/main.js
```diff
@@ -50,19 +40,320 @@ angular.module('emission.main', ['emission.main.diary', 'emission.main.common',
         controller: 'ControlCtrl'
       }
     }
+  })
+  .state('root.main.sensed', {
```
shankari: it is not wrong to have the states here, but it is not strictly necessary.
you can navigate to the root.main.recent.sensed state from the main template.

www/js/main.js
shankari: Keeping the original code will also make this change much simpler.

www/css/main.diary.css
```diff
@@ -37,8 +37,8 @@
   color:whitesmoke !important;
 }
 .angular-leaflet-map {
-    width: 200px;
-    height: 150px;
+  height: 150px;
```
shankari: consider using window width to dynamically resize instead

www/js/recent.js
```diff
@@ -1,313 +0,0 @@
-angular.module('emission.main.recent', ['ngCordova', 'emission.services'])
```
shankari: Again, I think that this should be retained and we should just navigate to the states from the HTML. It's bad to blow up main.js when it is not necessary. Lots of small files are better than one humongous file.

www/templates/control/main-control.html
```diff
@@ -8,47 +8,71 @@
...
+      <div style="background-color: white; height: 50px; margin-top: 0.5px;">
+        <div style="padding: 15px 15px; float: left;">Check sensed data</div>
+        <div ng-click="showSensed()" ng-style="getIconButtonStyle('CCCCCC')"><i class="ion-ios-arrow-right" ng-style="getIconStyle()"></i></div>
+      </div>   
+      <div style="background-color: white; height: 50px; margin-top: 0.5px;">
+        <div style="padding: 15px 15px; float: left;">Check map</div>
+        <div ng-click="showMap()" ng-style="getIconButtonStyle('CCCCCC')"><i class="ion-ios-arrow-right" ng-style="getIconStyle()"></i></div>
+      </div>   
+      <div style="background-color: white; height: 50px; margin-top: 0.5px;">
+        <div style="padding: 15px 15px; float: left;">Collection</div>
```
shankari: Why are we using more `px`? Shouldn't we be using `em` or relative widths instead?

www/templates/control/log.html
```diff
@@ -0,0 +1,18 @@
+<ion-view style="" title="Log">
```
shankari: This is not required, since `log.html` is still inside `recent`, right?

shankari:
overall:
- the real goal of this change was to shift to relative sizes so that it would scale on various form factors. The fonts now scale, but the other elements, notably the padding, don't. That should be fixed.
- we shouldn't need to move the code from `recent.js` into `main.js`. That just makes `main.js` super large and hard to work with. You can rename `recent.js` if you want, but I think that the code there now should remain there, and you should just go to the correct URLs from the HTML.

krisma:
I put recent back. For relative size, even em unit is converted into px (1em = 1 base font size, say 14 in Ionic) when interpreted so it's not that responsive unless we use @media to specify base font size for each screen size. Also, em is usually used for font size. Eventually, I could write a python script that turns every px into em. But that doesn't work for now because there are more elements to implement, and em is a larger scale that is hard to control. (like 1 px difference equals 0.07em)

shankari:
I'm going to merge this once you fix the state stuff, but I'd really like us to spend some time setting up the structure of the css correctly. Right now, the padding is scattered through the files, colors are scattered through the files, and so on, so it is hard to change them. I really liked your idea of defining a color swatch, but I think that you should make it more structured by putting the swatches into a file, giving them descriptive names (highlight, etc) and then using the names in the HTML.

Otherwise, we'll have to spend a bunch of time at the end fixing and tweaking everything and checking that we haven't missed any spot

www/js/main.js
```diff
@@ -50,19 +40,59 @@ angular.module('emission.main', ['emission.main.diary', 'emission.main.common',
...
+    url: "/sensed",
+    views: {
+      'main-control': {
+        templateUrl: "templates/recent/sensedData.html",
+        controller: 'sensedDataCtrl'
+      }
+    }
+  })
+  .state('root.main.map', {
```
shankari: Can you explain why we need to move the states in here? Now that recent.js is restored, the states are there. Why can't the buttons just navigate to 'root.main.recent.log' instead of 'root.main.log'?

krisma:
Now when I run I have bug like this. 

Error: window.cordova.plugins.BEMServerSync.getConfig is not a function. (In 'window.cordova.plugins.BEMServerSync.getConfig()', 'window.cordova.plugins.BEMServerSync.getConfig' is undefined)

Other parts are good and I manually recovered the style in main control. If the bug is resolved, the style should come back.

Final list view
screen shot 2016-06-24 at 9 18 39 pm

shankari:
You need to do "ionic state reset" again for it to see the getConfig.
I changed native plugins this time as well!

krisma:
the desired UI, that when users click the time tag the earlier/later thing jumps out. Turned out that it's super complicated to implement as angular takes fixed amount of height of list-item (aka not adjustable in run time), when I add one more element (aka change display from none to block), the height cannot be modified, so the cards get stuffed together as the margin is occupied. As a compromise, I put the earlier/later thing there permanently.

shankari:
Fine. That was just a suggestion. If it is too hard to implement, feel free
to modify to something that is.
We should really move on from optimizing list view even more to
implementing new functionality.

krisma:
Yeah I already pushed. The last commit "chaotic". If I wake up early enough tomorrow I'll test the sync stuff

www/js/main.js
```diff
@@ -360,28 +350,13 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
     $ionicPopover.fromTemplateUrl('templates/control/main-collect-settings.html', {
         scope: $scope
     }).then(function(popover) {
-        $scope.collectSettingsPopup = popover;
```
shankari: This seems like a wrong change from the merge. This should be `collectSettingsPopup` not `settingsPopup`

www/js/main.js
```diff
@@ -360,28 +350,13 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
-        scope: $scope
-    }).then(function(popover) {
-        $scope.syncSettingsPopup = popover;
+    $ionicPopover.fromTemplateUrl('templates/control/main-sync-settings.html', {        
+        scope: $scope       
+    }).then(function(popover) {     
+        $scope.syncSettingsPopup = popover;     
     });
-
-    $scope.trackingOn = function() {
```
shankari: Did you intend to remove this as part of the merge?

www/templates/control/main-control.html
```diff
@@ -10,60 +10,41 @@
...
-        <div class="control-list-text">Check map</div>
+      <div style="background-color: white; height: 50px; margin-top: 0.5px;">
+        <div style="padding: 15px 15px; float: left;">Check map</div>
         <div ng-click="showMap()" ng-style="getIconButtonStyle('CCCCCC')"><i class="ion-ios-arrow-right" ng-style="getIconStyle()"></i></div>
       </div>   
-      <div class="control-list-item">
-        <div class="control-list-text">Collection</div>
+      <div style="background-color: white; height: 50px; margin-top: 0.5px;">
+        <div style="padding: 15px 15px; float: left;">Collection</div>
         <div ng-click="editConfig($event)" ng-style="getIconButtonStyle('CCCCCC')"><i class="ion-edit" ng-style="getIconStyle()"></i></div>
```
shankari: Note that this was changed to `editCollectionConfig` as part of my changes. You copied the code over so you need to change this as well.

www/js/main.js
```diff
@@ -278,39 +275,35 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
+    $scope.saveAndReloadCollectSettingsPopover = function(result) {
         console.log("new config = "+$scope.settings.collect.new_config);
         if (result == true) {
             window.cordova.plugins.BEMDataCollection.setConfig($scope.settings.collect.new_config)
             .then($scope.getCollectionSettings);
         }
     };
 
-    $scope.saveAndReloadCollectSettingsPopover = function() {
+    $scope.saveAndReloadSettingsPopover = function() {
```
shankari: This is wrong too - `saveAndReloadCollectSettingsPopover` is correct.

www/js/main.js
```diff
@@ -278,39 +275,35 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
             .then($scope.getCollectionSettings);
-        $scope.collectSettingsPopup.hide();
-    };
-
-    $scope.saveAndReloadSyncSettingsPopover = function() {
-        console.log("new config = "+$scope.settings.sync.new_config);
-        window.cordova.plugins.BEMServerSync.setConfig($scope.settings.sync.new_config)
-            .then($scope.getSyncSettings);
-        $scope.syncSettingsPopup.hide();
+        $scope.settingsPopup.hide();
```
shankari: This too.

shankari:
I think that the merge got screwed up - it looks like you picked HEAD instead of master for most of the changes, so a lot of the changes were overwritten

krisma:
Latest changes include:
Now by default the users see light theme, since many people like the way it was. However, on the meeting we already showed people that dark one. In order to make my work last week at least mean something, the users can now choose dark theme in their setting. To implement that, a model consists of appPreference and $rootScope is introduced in some controllers, hope that won't bring more bug?

www/js/main.js
```diff
@@ -275,31 +307,34 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
-        $scope.syncSettingsPopup.show($event);              
+
+    $scope.editSyncConfig = function($event) {
+        $scope.settings.sync.new_config = JSON.parse(JSON.stringify($scope.settings.sync.config));
+        console.log("settings popup = "+$scope.syncSettingsPopup);
+        $scope.syncSettingsPopup.show($event);
     }
-    $scope.saveAndReloadCollectSettingsPopover = function(result) {
+    
+    $scope.saveAndReloadSettingsPopup = function(result) {
```
shankari: This is incorrect. Should be CollectSettings.

shankari: Hm, are we even using this? Or just the popover?
Did the pre-merge code actually have `saveAndReloadCollectSettingsPopover` AND `saveAndReloadSettingsPopover`?

www/templates/control/main-control.html
```diff
@@ -10,45 +10,68 @@
...
 
-  <ion-content style="background-color: #262A30;">
+  <ion-content>
 <!--     <div style="text-align: center;">
       <img src="img/avatar_1.png" ng-style="getAvatarStyle()">
     </div> -->
     
-    <div style="background-color: white; height: 50px;">
-      <div style="padding: 15px 15px; float: left;">{{settings.auth.email}}</div>
+    <div class="control-list-item">
```
shankari: Awesome to allow choice between the two. I think that doing that will encourage you to make the settings configurable rather than hard-coded.

www/templates/diary/list.html
```diff
@@ -1,19 +1,17 @@
-<ion-view nav-title="Trip Diary" style="background-color: #303030;">
+<ion-view nav-title="Trip Diary" ng-class="ionViewBackgroundClass()">
 	<ion-nav-title class="dark-color">
 		<div class="buttons" style="text-align:center; background-color: transparent !important;" id="toget">
-			<button class="button button-icon icon ion-arrow-left-b" style="color: rgba(51,224,187, 0.7); margin-top: 4px; background-color: transparent !important;" ng-click="prevDay()">
+			<button class="button button-icon icon ion-arrow-left-b date-picker-arrow" ng-click="prevDay()">
```
shankari: nice wrt css-ifying

www/templates/diary/list.html
```diff
@@ -1,19 +1,17 @@
-<ion-view nav-title="Trip Diary" style="background-color: #303030;">
+<ion-view nav-title="Trip Diary" ng-class="ionViewBackgroundClass()">
```
shankari: Nice css-ify stuff

www/templates/control/main-control.html
```diff
@@ -10,45 +10,68 @@
...
+      </label>
+    </div>
+    <div class="control-list-item">
+      <div class="control-list-text">Force sync</div>
+      <div ng-click="forceSync()" ng-style="getIconButtonStyle('D6C780')"><i class="ion-android-sync" ng-style="getIconStyle()"></i></div>
+    </div>   
+    <div class="control-list-item">
+      <div class="control-list-text">Dark theme</div>
+
+      <label ng-click="willUseDarkTheme()" class="toggle control-list-toggle">
```
shankari: For next week, I think that you can do ng-model instead.

shankari:
Merging now.
If I have any other comments, will file issues for them, or if they are super urgent, will just address them

www/templates/control/main-control.html
krisma: Yeah otherwise it's gonna take forever to change theme

Issue #53, opened 2016-06-26
Parsing boolean values in the JSON returned for config objects seems to be flaky on iOS


shankari:
I am not sure why this is happening, but if I install the app from the emulator, then even if the default values have duty cycling set, they are returned at "1" instead of true. But I install directly from xcode, it appears to be true.

Need some more investigation.
This is fairly high priority since @juemura ran into the case in which duty cycling was turned off and so her battery was draining rapidly. But I checked the timeline, and she only updated the sensor_config once, which was probably the time that I set it for her. Does this mean that the default doesn't actually work? Need some more debugging...

shankari:
```
In [896]: list(edb.get_timeseries_db().find({'user_id': <redacted>, 'metadata.key': 'config/sensor_config'}))
Out[896]: 
[{u'_id': ObjectId('57649d2e4a9bba50e3f85f17'),
  u'data': {u'accuracy': -1,
   u'accuracy_threshold': 200,
   u'android_geofence_responsiveness': -1,
   u'filter_distance': 50,
   u'filter_time': -1,
   u'geofence_radius': 100,
   u'ios_use_remote_push_for_sync': 1,
   u'ios_use_visit_notifications_for_detection': 1,
   u'is_duty_cycling': 1,
   u'simulate_user_interaction': 0,
   u'trip_end_stationary_mins': 10},
  u'metadata': {u'key': u'config/sensor_config',
   u'platform': u'ios',
   u'plugin': u'none',
   u'read_ts': 0,
   u'time_zone': u'America/Los_Angeles',
   u'type': u'rw-document',
   u'write_fmt_time': u'2016-06-17T15:41:55.435798-07:00',
   u'write_ts': 1466203315.435798},
  u'user_id': <redacted>}]
```

shankari:
Bump up the priority on this since it looks like the value got automatically reset as well. This is concerning and should be investigated.

shankari:
> it looks like the value got automatically reset as well

No. The visit check was wrong. See https://github.com/e-mission/e-mission-data-collection/issues/118.
This is no longer super high priority.

PR #55, opened 2016-06-28
Bugs fixed

krisma:
Improvements:
1. Responsive sizing. Now although we are still using a combo of px, em, and percentage, it's the best design I could come up with due to the complications. Leaflet has to accept width and height in number of px and can't be overridden by ng-class or ng-style. So we use media to fit different screens. Now list view should look nice on all devices.
2. Back button colors are right now.
3. In detail view there is a date now, like the one in date picker in list view. 
4. Frequent functions added to control view.
5. Missing functions are back. (the server ip is now in sync tab install of having it's own tab, because the original design has a button on the right side og a tab, while the ip doesn't need any operation would be better put inside the list)
6. Heatmap loading state added.


www/css/main.diary.css
```diff
@@ -36,9 +36,37 @@
   background:#7570b3!important;
   color:whitesmoke !important;
 }
-.angular-leaflet-map {
-  width: 300px;
-  height: 150px;
+
+
+/* leaflet */
+/* ----------- iPhone 5 and 5S ----------- */
```
shankari: what about android phones?
this is not an iOS-only app

I know that leaflet doesn't support `ng-class` etc but you should be able to use the window-size to set the size, using javascript if necessary.

Here are three suggestions that I provided earlier.

According to this:
http://leafletjs.com/examples/mobile.html
% is supported at least for height.

Another workaround is to read the browser width and set the width based on that in javascript.
https://www.drupal.org/node/2086539

Or try applying this patch, since we have a patched leaflet anyway. You'll have to find where that code is in the concatenated leaflet.js that we include.
https://www.drupal.org/files/issues/map-height-relative-2086539-9.patch

www/js/diary/detail.js
```diff
@@ -9,7 +9,10 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'nvd3ChartDirectives'
     JSON.stringify($stateParams));
 
   $scope.mapCtrl = {};
-
+  $scope.getFormattedDate = function(ts) {
```
shankari: Why isn't this in the service?

shankari: It should be common code, right?

www/js/diary/list.js
```diff
@@ -76,10 +76,26 @@ angular.module('emission.main.diary.list',['ui-leaflet', 'nvd3ChartDirectives',
       return ($scope.dark_theme)? "ion-view-background-dark" : "ion-view-background"; 
     }
     $scope.datePickerClass = function() {
-
     }
     $scope.listCardClass = function() {
-      return ($scope.dark_theme)? "list card list-card-dark" : "list card list-card"; 
+      if ($window.screen.width <= 320) {
```
shankari: Again, this looks like the three sizes for iphone. What happens about android phones?

www/js/diary/list.js
```diff
@@ -100,14 +116,16 @@ angular.module('emission.main.diary.list',['ui-leaflet', 'nvd3ChartDirectives',
...
+      to: new Date(),
       mondayFirst: true,  //Optional
       templateType: 'popup', //Optional
       showTodayButton: 'true', //Optional
       modalHeaderColor: 'bar-positive', //Optional
       modalFooterColor: 'bar-positive', //Optional
       callback: $scope.setCurrDay, //Mandatory
-      dateFormat: 'dd-MMMM-yyyy', //Optional
-      closeOnSelect: true, //Optional
+      dateFormat: 'dd MMMM yyyy', //Optional
```
shankari: should we change this to be yyyy MMMM dd, similar to the button that launches the selector?

www/js/diary/services.js
```diff
@@ -35,12 +35,12 @@ angular.module('emission.main.diary.services', ['emission.services'])
         'text-align': 'center',
         'background-color': '#33e0bb',
         'color': 'black',
-        'width': '48px',
+        'width': '18%',
```
shankari: Great to see that this is changed to %
But the colors are still hardcoded here.
Why can't all this be in a CSS style instead of being in javascript?

If it does have to be in javascript, it should be in a separate config module instead of being mixed in with the main functionality.

www/js/diary/services.js
```diff
@@ -85,12 +85,12 @@ angular.module('emission.main.diary.services', ['emission.services'])
         'text-align': 'center',
         'background-color': '#ff5251',
         'color': 'black',
-        'width': '48px',
+        'width': '18%',
         'height': '16px',
```
shankari: still see some `px` here!

www/js/main.js
```diff
@@ -96,8 +96,29 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
+    $scope.emailLog = ControlHelper.emailLog;
     $scope.dark_theme = $rootScope.dark_theme;
+
+    
+    $scope.toggleLowAccuracy = function() {
+        if ($scope.settings.collect.config == null) {
+            return false; // config not loaded when loading ui, set default as false
+        } else {
+            var accuracy = $scope.settings.collect.config.accuracy; 
+            return accuracy >= 100;
```
shankari: this is iOS specific - you need to handle both iOS and android correctly

www/js/main.js
```diff
@@ -96,8 +96,29 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
+        }
+    }
+    $scope.willUseLowAccuracy = function() {
+        var accuracy = $scope.settings.collect.config.accuracy;
+        $scope.settings.collect.new_config = JSON.parse(JSON.stringify($scope.settings.collect.config));
+        if (accuracy < 100) {
+            $scope.settings.collect.new_config.accuracy = 100;
```
shankari: this is iOS specific - you need to handle both iOS and android correctly

www/js/main.js
```diff
@@ -96,8 +96,29 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
 
-.controller('ControlCtrl', function($scope, $window, $ionicScrollDelegate, $state, $ionicPopup, $ionicActionSheet, $ionicPopover, $rootScope) {
+.controller('ControlCtrl', function($scope, $window, $ionicScrollDelegate, $state, $ionicPopup, $ionicActionSheet, $ionicPopover, $rootScope, ControlHelper) {
+    $scope.emailLog = ControlHelper.emailLog;
     $scope.dark_theme = $rootScope.dark_theme;
+
+    
+    $scope.toggleLowAccuracy = function() {
+        if ($scope.settings.collect.config == null) {
+            return false; // config not loaded when loading ui, set default as false
```
shankari: I am not sure that this is the right thing to do. This means that the low accuracy will return `false` even the user set it before. This is going to cause a lot of confusion - see bugs like https://github.com/e-mission/e-mission-phone/issues/53

Isn't the profile screen basically the rebranded control screen? The control screen loads all the settings on startup. If it doesn't right now, it should. Otherwise, we will get bugs in which it will look like the user is not logged in although they are and so on.

www/js/main.js
shankari: Note that the value for android is not quite this simple.

www/templates/control/main-collect-settings.html
```diff
@@ -2,8 +2,7 @@
 <ion-content style="margin: 0; padding: 0;">
 <ion-button-bar>
 
-    <button class="button button-active button-block" ng-click="saveAndReloadCollectionSettingsPopover()" style="margin: 0; color: #fff; background-color: #95BCA2; border-bottom-left-radius: 0;
-border-bottom-right-radius: 0;">Save</button>
+    <button class="button button-active button-block" ng-click="saveAndReloadCollectSettingsPopover()" style="margin: 0; color: #fff; background-color: #95BCA2; border-bottom-left-radius: 0; border-bottom-right-radius: 0;">Save</button>
```
shankari: `saveAndReloadCollectionSettingsPopover` is correct. Please don't revert my change :)

I fixed this bug as part of the last release. Did you pull?
https://github.com/e-mission/e-mission-phone/pull/54/commits/50dec5e1c38c38943c8d03418e469a97e6584b0a

www/templates/control/main-control.html
```diff
@@ -29,20 +29,33 @@
       </label>
     </div>
     <div class="control-list-item">
-      <div class="control-list-text">Force sync</div>
-      <div ng-click="forceSync()" ng-style="getIconButtonStyle('D6C780')"><i class="ion-android-sync" ng-style="getIconStyle()"></i></div>
-    </div>   
+      <div class="control-list-text">Low accuracy</div>
+      <label ng-click="willUseLowAccuracy()" class="toggle control-list-toggle">
```
shankari: you should be able to use ng-model instead. I had this as feedback for some other toggle in the control screen as well.

www/templates/diary/list.html
```diff
@@ -100,7 +101,7 @@
 
 
                 <div style="padding: 0.5em;">
-                    <div ng-class="listLocationClass()" href="#" style="background-color: transparent; font-size: 0.8em; padding-top: 5px; padding-bottom: 5px; padding-left: 30px; margin-top: 0; margin-bottom: {{getEarlierOrLater(tripgj.data.properties.start_ts, tripgj.data.id) == ''? '0':'20px'}};">
+                    <div ng-class="listLocationClass()" href="#" style="background-color: transparent; font-size: 0.8em; padding-top: 5px; padding-bottom: 5px; padding-left: 30px; margin-top: 0; margin-bottom: 0;">
```
shankari: if you have an `ng-class`, do you still need the style?

shankari:
Would also like to see screenshots of changed screens.

At a high level, I feel like there's still too many references to hardcoded pixels. The goal is not to look good on all the iPhones, but to look good on all iPhones *and* android phones, and android phones have a really wide variety of form factors.

shankari:
I'm not going to merge this now, and will proceed with my changes (to the common screen).
Hopefully there won't be too much overlap.
We can deal with the merge in person on Friday...

www/js/main.js
shankari: Here's where the screen (used to be?) refreshed
https://github.com/e-mission/e-mission-phone/blob/master/www/js/main.js#L384

www/js/diary/list.js
krisma: what do you mean by selector? datepicker?

shankari: yes

www/js/diary/services.js
krisma: I don't think it's necessary to make this taller in larger screen because it just show the time, and as long as the time is clear to see, it should be fine. What do u think?

www/js/main.js
krisma: Do u want me to debug on this? I tested that they are not loaded on first call. However, the reason that I dare to set a default value is that I noticed the toggle function is called lots of times, and in the later calls, the config is loaded and the toggle is set to the right state. I'm only using this default to avoid null pointer exception.

shankari: the toggle function is only called to change the value, right? so then who calls it lots of times?

krisma: uh I mean the toggle function (toggleLowAccurafcy), returns bool, which is used in ng-check, to set toggle to the correct state. Another function (willUseLowAccuracy) is actually changing the state, as u mentioned. I mean the bool function is called multiple times, by someone I don't know. Maybe it's the scope.watch that ionic used to handle ng stuff?

www/templates/control/main-control.html
krisma: This one I actually need to call BEMDataCollection stuff to update the config. I could, of course, bind the ng-model to a setter function, but then it would be the same to this function. For the toggle in dark theme, I also need to have a function call for appPreference fetching.

krisma:
Bug fixed. Review again plz.

shankari:
Screenshots?
Also, if we are going to keep ng-class and javascript styling, please move the styling code into a separate module, independent of the functionality. In MVC, it is good practice to keep the M, V and C separate, and this seems to be munging V and C

www/js/diary/services.js
shankari: I don't have a strong opinion. Was just commenting on the general px v/s em v/s % debate.

shankari:
Once you address the last 2 comments (right above), I think that this version is good to go.
I am looking forward to seeing the results of your experiment with sass and the flexible grid layout

www/js/main.js
```diff
@@ -100,46 +100,60 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
+            for (var k in $scope.settings.collect.accuracyOptions) {
+                if ($scope.settings.collect.accuracyOptions[k] == accuracy) {
+                    v = k;
+                    break;
+                } 
+            }
             if ($scope.isIOS()) {
-                var accuracy = $scope.settings.collect.config.accuracy; 
-                return accuracy >= 100;
+                return v != "kCLLocationAccuracyBestForNavigation" && v != "kCLLocationAccuracyBest";
```
shankari: && v != kCLLocationAccuracyTenMeters

www/js/main.js
```diff
@@ -100,46 +100,60 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
             } else if ($scope.isAndroid()) {
-                return accuracy != "PRIORITY_HIGH_ACCURACY";
+                return v != "PRIORITY_HIGH_ACCURACY";
             }
 
         }
     }
-    $scope.willUseLowAccuracy = function() {
+    $scope.setLowAccuracy = function() {
```
shankari: this should actually be toggle, since you are switching the value

shankari:
I don't see that you have moved the styling code out into a separate module to be consistent with MVC guidelines. This was the second set of changes that I recommended last night.

> Also, if we are going to keep ng-class and javascript styling, please move the styling code into a separate module, independent of the functionality. In MVC, it is good practice to keep the M, V and C separate, and this seems to be munging V and C

www/js/main.js
```diff
@@ -100,46 +100,60 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
...
         var accuracy = $scope.settings.collect.config.accuracy;
         $scope.settings.collect.new_config = JSON.parse(JSON.stringify($scope.settings.collect.config));
+        var v;
+        for (var k in $scope.settings.collect.accuracyOptions) {
```
shankari: This is re-used - pull out into a separate function.

shankari: You could also consider just prepopulating the string value when the config is read (in `getSettings`)

www/js/diary/services.js
```diff
@@ -28,6 +28,7 @@ angular.module('emission.main.diary.services', ['emission.services'])
     return d;
   }
   dh.isCommon = function(id) {
+    return true;
```
shankari: Need to remove hardcoded values!

krisma:
the most recent commit has them removed

www/css/style.css
```diff
@@ -210,9 +223,11 @@ button.button.ng-binding i.icon.ion-edit {
   font-size: 12px;
 }
 button.button.back-button.buttons.button-clear.header-item {
-  color: #fff;
+  color: #303030; opacity: 0.7;
```
shankari: next week: move colors into a separate css file with names? That way, it will be easier to reuse colors and have a consistent theme

www/js/diary/list.js
```diff
@@ -76,12 +76,33 @@ angular.module('emission.main.diary.list',['ui-leaflet', 'nvd3ChartDirectives',
...
+      if (margin == 0) {
+        return ($scope.dark_theme)? "col-50 list-col-left-dark" : "col-50 list-col-left";   
+      } else {
+        return ($scope.dark_theme)? "col-50 list-col-left-margin-dark" : "col-50 list-col-left-margin"; 
+      }
+    }
+    $scope.listColRightClass = function() {
+      return ($scope.dark_theme)? "col-50 list-col-right-dark" : "col-50 list-col-right"; 
+    }
+    $scope.differentCommon = function(tripgj) {
```
shankari: I'm not sure what you are trying to do there, and there are no screenshots, but I'm just going to merge now.

www/js/main.js
```diff
@@ -435,8 +476,8 @@ angular.module('emission.main', ['emission.main.recent', 'emission.main.diary',
             'height': '100%', 
             'background-color': '#' + color,
             'color': '#fff',
-            'padding': '15px 15px',
-            'width': '50px'
+            'padding-top': '16px',
```
shankari: for next week, `getAvatarStyle`, `getButtonStyle` and `getIconButtonStyle` should be moved to css, probably with @media sizing? Also `getAvatarStyle` is obsolete, I think

www/templates/control/main-control.html
```diff
@@ -51,6 +64,10 @@
     <!-- Begin developer zone -->
     <div class="" ng-show="collectionExpanded()">
       <div class="control-list-item">
+        <div class="control-list-text">Refresh</div>
+        <div ng-click="refreshScreen()" ng-style="getIconButtonStyle('D6C780')"><i class="ion-refresh" ng-style="getIconStyle()"></i></div>
```
shankari: nit: color in CSS?

PR #58, opened 2016-06-30
Restructure the common trips screen to be more complex

shankari:
The default view is still a map, but we also have the option of seeing a list
or details of the places.

Similar to the recent view, the selection between screens is using a sidebar
menu. I would have liked to use tabs, but ionic tabs cannot be used inside a
content, or at least, I could not get it to work.

http://ionicframework.com/docs/api/directive/ionTabs/

> Note: do not place ion-tabs inside of an ion-content element; it has been known to cause a certain CSS bug.

The actual functionality in list and detail have not yet been implemented, but
I wanted to get the structure in place first for easy reviewing.

### DIFF
--- a/www/js/common.js
+++ b/www/js/common.js
@@ -1,100 +1,45 @@
 angular.module('emission.main.common',['ui-leaflet',
                                       'ionic-datepicker',
+                                      'emission.main.common.map',
+                                      'emission.main.common.list',
+                                      'emission.main.common.detail',
                                       'emission.main.common.services',
                                       'emission.services'])
+.config(function($stateProvider, $ionicConfigProvider, $urlRouterProvider) {
+  $stateProvider
+
+  .state('root.main.common.map', {
+    url: '/map',
+    views: {
+      'menuContent': {
+        templateUrl: 'templates/common/map.html',
+        controller: 'CommonMapCtrl'
+      }
+    }
+  })
+
+  .state('root.main.common.list', {
+    url: '/list',
+    views: {
+      'menuContent': {
+        templateUrl: 'templates/common/list.html',
+        controller: 'CommonListCtrl'
+      }
+    }
+  })
+
+  .state('root.main.common.place-detail', {
+    url: '/place/:placeId',
+    views: {
+        'menuContent': {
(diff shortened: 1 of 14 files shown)


shankari:
A lot of the changes are just whitespace, or moving code around.

shankari:
Screenshots:

| Map | Menu | List
|----------------------|---------------------|-------------------------|
common_map | common_menu | common_list

shankari:
@sunil07t, @krisma, review code?
I'm going to actually implement list and detail for now, so there's some time to review, but I will merge later tonight.

krisma:
I think it's ready to merge. My branch is also ready to merge. I haven't tested on android since the emulator is not working but we can test tmr after the meeting. The code "seems to be working".

www/js/common.js
```diff
@@ -1,100 +1,45 @@
...
-          name = name + ", " + address["city"];
-        } else if (address["town"]) {
-          name = name + ", " + address["town"];
-        } else if (address["county"]) {
-            name = name + ", " + address["county"];
+    }
+  })
+
+  .state('root.main.common.place-detail', {
+    url: '/place/:placeId',
```
sunil07t: I do not thing this is required but instead of '/place/:placeId' we can use '/map/:placeId'

shankari: Why is that better than `place/:placeId`?
Note that I will soon be checking in a similar structure for trips as well.

sunil07t: Are we going to have a list of places and when we press on an item from that list it will direct to a more detailed information about the place?

shankari: Yes! And similarly for trips. Here's an example screenshot.

places_list

shankari: We could also add maps and other visual representations to make this prettier, but I'll leave that for you/Kris to deal with.

sunil07t: Oh that is also similar to the edit functionality of the goals list. So I think if we use  '/place/:placeId'  and name the list url '/place' just to make the code easy to read.

shankari: I'd already changed the list URL to `/place` locally but haven't pushed yet!

PR #61, opened 2016-07-05
Test style

krisma:
removed .idea


www/js/styles.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+angular.module('emission.styles', [])
+
+.service('StylesHelper', function($window, $scope, $rootScope) {
```
shankari: The problem is with the arguments passed in here. I am not sure which of them it is - you can figure out by trial and error.

This works.
```
.service('StylesHelper', function() { 
    this.listCardClass = function() { 
        return "list card list-card-dark list-card-sm";
    }
});
```

This doesn't.

```
.service('StylesHelper', function($window, $scope, $rootScope) {
    this.listCardClass = function() {
        return "list card list-card-dark list-card-sm";
    }
});
```

Maybe given that services are bound to scopes, you cannot pass in `$scope`? Note the error with the path `$scopeProvider <- $scope <- StylesHelper`

shankari:
This has now been debugged.

PR #62, opened 2016-07-06
Added goals tab

sunil07t:
Habitica avatar and stats UI and todo list functionality.


www/index.html
```diff
@@ -4,9 +4,12 @@
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
     <meta http-equiv="Content-Security-Policy"
-        content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu 'unsafe-inline';
+        content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://habitica-assets.s3.amazonaws.com https://www.habitica.com
```
shankari: Fine for now - need to decide whether to use the standard habitica server or our copy

www/index.html
```diff
@@ -4,9 +4,12 @@
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
     <meta http-equiv="Content-Security-Policy"
-        content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu 'unsafe-inline';
+        content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://habitica-assets.s3.amazonaws.com https://www.habitica.com 
+        https://habitica.com/api/v3/members/4f369eef-aed4-4408-bcbf-b34896daf7e3 'unsafe-inline';
```
shankari: Do you need this? Isn't it subsumed by the `https://www.habitica.com`?

The reason I ask is that it looks like this is the API to read data/assets for member `4f369eef-aed4-4408-bcbf-b34896daf7e3`. Since each user will have their own habitica UUID, this is not a scalable solution.

So it would be nice to know whether it is necessary or not.

www/index.html
```diff
@@ -4,9 +4,12 @@
...
             style-src 'self' 'unsafe-inline';
-            img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com">
+            img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com https://habitica-assets.s3.amazonaws.com https://www.habitica.com 
+            https://habitica.com/api/v3/members/4f369eef-aed4-4408-bcbf-b34896daf7e3
```
shankari: Ditto.

www/templates/main-goals.html
```diff
@@ -0,0 +1,11 @@
+<ion-view view-title="Goals">
+<ion-content class="has-header" overflow-scroll="true" padding="true">
+<h3 style="align:center;">
+    My goal is {{goal}}
+    </h3>
+    <label class="item item-input">
+    <input  ng-model="data.userGoal" type="text" placeholder="Goal">
+    </label>
+    <button ng-click="saveGoal()" class="button button-assertive button-block" ng-click="getPopRoute()">Submit</button>
```
krisma: Y does it have two ng-click?

www/templates/main-goals.html
```diff
@@ -0,0 +1,68 @@
...
+            </div>
+          </div>
+        </a>
+      </div>
+        <div class="center" style="margin-top:-5px;">
+          <img src="img/ic_header_healer.png" style="height:40px; width:40px;" class="center">
+          <h2 style="font-size:12px; text-align:center; margin:0px">{{user.data.profile.name}}</h2>
+          <p style="font-size:10px; text-align:center;">{{user.data.stats.class}}</p>
+        </div>
+      <ion-list can-swipe="true">
```
shankari: what happens when you swipe?

www/templates/main-goals.html
```diff
@@ -0,0 +1,68 @@
...
+        <ion-item class="item" ng-repeat="goal in goals" style="border:1px;">
+          <div class="row">
+          <lable class="checkbox">
```
shankari: does this really work? Shouldn't it be `label`?

www/templates/main-goals.html
```diff
@@ -0,0 +1,68 @@
...
+          <!--div class="col col-20" style="display:flex; justify-content:flex-end; align-items:center;">
+          <button class="button button-clear" style="font-size:2rem !important">
+            <i icon="ion-checkmark-circled"></i>
+          </button>
+          </div-->
+          <div class="col col-80" style="padding-left: 5% !important;">
+          <h2>{{goal.name}}</h2>
+          <p style="font-size: 10px">{{goal.note}}</p>
+          </div>
+          <ion-option-button class="button-assertive" ng-click="removeGoal(goal)">Delete
```
shankari: Why not use the standard directive options - `show-delete`, `ion-delete-button`, etc?

http://ionicframework.com/docs/api/directive/ionList/

shankari:
High level:
- screenshots?
- did you try copying the screens from the habitica UI? How did that work?

sunil07t:
screen shot 2016-07-05 at 9 37 56 pm
screen shot 2016-07-05 at 9 38 29 pm

www/templates/main-goals.html
sunil07t: I removed this line because ion-option-button (line 56) takes care of the swipe to show the delete option.

www/templates/main-goals.html
sunil07t: Sorry, I am not sure what line are you referring to.

www/templates/main-goals.html
sunil07t: I think the checkbox did not work with this when I tried it first but I changed the checkbox css for swipe delete, so I can can try it for this again. Which one do you prefer?

www/templates/main-goals.html
sunil07t: This was an old commit

www/index.html
sunil07t: This is necessary, I am using this just to get a members data that does not require authentication. Using this until we have the backend api functionality.

www/templates/main-goals.html
shankari: I'm confused. Doesn't it still say `can-swipe="true"'? What happens when you swipe?

www/templates/main-goals.html
shankari: `<lable class="checkbox">`

www/templates/main-goals.html
shankari: I'm a bit confused. I assumed that you would use.
Is there a reason why you would handle it on your own?
I would prefer using the standard directives as much as possible because ionic is likely to make additional improvements going forward...

`<ion-list show-delete="shouldShowDelete">
   <ion-item>
  ...
    <ion-delete-button class="ion-minus-circled"
                       ng-click="removeGoal(goal)">
    </ion-delete-button>
   </ion-item>
</ion-list>
```

**Comment by shankari** [comment]
Also, any comments on copying over the screens directly from habitica? Does that work instead of building our own?

The screens are at:
https://github.com/HabitRPG/habitrpg/tree/develop/website/views

File: www/templates/main-goals.html (hunk 5 above)
Comment (sunil07t) [inline www/templates/main-goals.html]: This is default true so I removed that line but I have not commit yet. This allows to show the options when an iteam in a list is swiped. Right now it shows the delete option.

File: www/templates/main-goals.html (hunk 6 above)
Comment (sunil07t) [inline www/templates/main-goals.html]: Oh thanks for catching that! For some reason it was still working.

**Comment by sunil07t** [comment]
i will build the party screen referring to their layout instead of copying over directly.

**Comment by shankari** [comment]
I would like to see the results of an experiment of just copying over directly. The advantage then is that we can easily add other screens by copying them over instead of editing them. Maybe do that for the next screen?

**Comment by shankari** [comment]
I will merge this once the integration with our server is complete. Hopefully, we can demo that this Friday.
Let's discuss what we should show in the demo at today's meeting.

File: package.json [hunk 8]
```diff
@@ -77,6 +79,10 @@
       },
       "locator": "cordova-plugin-crosswalk-webview",
       "id": "cordova-plugin-crosswalk-webview"
+    },
+    {
+      "locator": "../me.apla.cordova.app-preferences",
```
shankari: I think that @apla has fixed this now. We should try again, but at least, we should not check in the relative path.

shankari:
whoa! there are 4589 changed files!
I guess that's all the pngs needed to generate the sprite?
Now that I see that, I think that maybe the png export was good after all :)
At any rate, I'll configure the s3 stuff on Fri so that we have the option to go back to it if we want...

www/templates/main-goals.html
```diff
@@ -54,6 +55,7 @@
               <span ng-show="profile.items.currentPet" class="current-pet Pet-{{profile.items.currentPet}}"></span>
             </div>
           </figure>
+          <!--Habitica jade to html code ends here-->
```
shankari: How does a `<div>` start have a `</figure>` end? Are you sure that the comments are in the correct place?

www/js/diary/services.js
```diff
@@ -400,7 +400,7 @@ angular.module('emission.main.diary.services', ['emission.services'])
      }, function(error) {
        window.Logger.log(window.Logger.LEVEL_INFO,
         "while reading data for "+day
-        +" from server, error = "+JSON.stringify(error));
+        +" from server, error = "+getTimelineForDay);
```
shankari: I don't think that this change is correct - please revert.

www/js/goals.js
```diff
@@ -45,10 +45,26 @@ angular.module('emission.main.goals',[])
...
-	var UUID= '4f369eef-aed4-4408-bcbf-b34896daf7e3';
+	var callOpts = {'method': 'GET', 'method_url': "/api/v3/user",
+                    'method_args': null};
+    //callOpts = {'method': 'GET', 'method_url': "/export/avatar-",
+    				//'method_args': };
+
+    
+	CommHelper.habiticaProxy(callOpts, function(response){
+		 $scope.user = response.data;
+		 console.log(error);
```
shankari: why are you logging `error` here?

www/js/main.js
```diff
@@ -50,15 +50,19 @@ angular.module('emission.main', ['emission.main.recent',
...
-    controller: 'GoalsCtrl'
-    })
+    views: {
+      'main-goals': {
+        templateUrl: 'templates/main-goals.html',
+        controller: 'GoalsCtrl'
+      }
+    }
+  })
   .state('root.main.signup', {
```
shankari: I'm going to merge this as-is for now, but instead of a state for the signup, why not use a popup or a popover? That will ensure that the habitica module encapsulates all the habitica information and it is not spread across other places.

e.g. http://ionicframework.com/docs/api/service/$ionicPopup/

shankari: That should also solve your problem with the ordering of the screens.

www/js/services.js
```diff
@@ -24,7 +24,7 @@ angular.module('emission.services', [])
      * Other fields can be added easily and the server can be modified at the same time.
      */
     this.habiticaRegister = function(regConfig, successCallback, errorCallback) {
-        var dateString = date.startOf('day').format('YYYY-MM-DD');
+        //var dateString = date.startOf('day').format('YYYY-MM-DD');
```
shankari: You can just remove these - no need to comment them out

www/js/services.js
```diff
@@ -45,7 +45,7 @@ angular.module('emission.services', [])
      * ....
      */
     this.habiticaProxy = function(callOpts, successCallback, errorCallback) {
-        var dateString = date.startOf('day').format('YYYY-MM-DD');
+        //var dateString = date.startOf('day').format('YYYY-MM-DD');
```
shankari: Ditto.

www/templates/main-goals.html
```diff
@@ -33,23 +33,25 @@
...
+            <img src="img/ic_header_gold.png" style="height:15px; width:15px;float: left;"><p style="font-size: 10px; float: left; margin: 0 10% 0 2%;">0</p>
+            <img src="img/ic_header_silver.png" style="height:15px; width:15px;float: left;"><p style="font-size: 10px; float: left; margin: 0 10% 0 2%;">0</p>
+            <img src="img/ic_header_gem.png" style="height:15px; width:15px;float: left;"><p style="font-size: 10px; float: left; margin: 0 0 0 2%;">0</p>
         </div>
-        <div class="col col-50" style="margin-top:-5px;">
-          <img src="img/ic_header_healer.png" style="height:40px; width:40px;" class="center">
-          <h2 style="font-size:12px; text-align:center; margin:0px">{{user.data.profile.name}}</h2>
-          <p style="font-size:10px; text-align:center;">{{user.data.stats.class}}</p>
+        <div class="col col-40" style="margin-top:-5px;">
+          <img src="img/ic_header_healer.png" style="height:30px; width:30px;" class="center">
```
shankari: I'm going to ask the same question as I always ask Kris. Is it possible to make this be relative, or to put it into css?

www/templates/main-goals.html
shankari: Also, can you add a short README  that describes the steps to export the habitica jade to HTML?
- load the page from the server
- click on the avatar
- .....

shankari:
The diff is too large for me to see all the files, so I've put in as many comments as I can.
I need to pull the change locally anyway to resolve merge conflicts so will look at the rest of the changes then.

www/templates/main-goals.html
shankari: Also does this mean that we now need to run `gulp` for `e-mission` too?
If so, that needs to be put into the README

PR #63, opened 2016-07-07
Add a new call to register habitica users

shankari:
Uses the newly fixed communication code in
https://github.com/e-mission/cordova-server-communication/pull/12
to make a user authenticated HTTP POST call to the server with the habitica username


shankari:
@sunil07t, @juemura, this is the function you should call on the phone to register.
Corresponding server side call
https://github.com/e-mission/e-mission-server/pull/306

PR #65, opened 2016-07-08
Common trips + distribution

krisma:
Added feature: 
1.common trips similar to diary trips.
2.distribution of start hour in that weekday. click on trip item to show graph

Removed feature:
1. cplace detail remove for now, it has no more content that cplace list

UNTESTED:
1. the list item will expand to show graph when user clicks on it, however, it has unknown behavior if there are more than 1 trip. (my server has one ctrip only) but the code looks correct.

Issues:
1. Chart should be able to be responsive. But I need to figure that out next week
2. at the very bottom of the common trip page, if the list item is expanded, it seems that the scroll doesn't make enough space for the expanded item.

### DIFF
--- a/www/js/common/map.js
+++ b/www/js/common/map.js
@@ -1,7 +1,7 @@
 angular.module('emission.main.common.map',['ionic-datepicker',
                                       'emission.main.common.services',
                                       'emission.services',
-                                      'emission.main.diary.services'])
+                                      'emission.main.diary.services', 'nvd3'])
 
 .controller("CommonMapCtrl", function($window, $scope, $rootScope, $ionicPlatform, $state,
                                     CommonGraph, Config, DiaryHelper) {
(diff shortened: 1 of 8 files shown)


www/js/common/services.js
```diff
@@ -208,7 +209,66 @@ angular.module('emission.main.common.services', [])
           "features": places.concat(trips)
         };
     };
+    var getDisplayNameForTrip = function(common_trip) {
```
shankari: Should just change `getDisplayName` to take a `cobject`?
Or even to return a promise that you use to set the value for whatever object.
But you can fix this later.

shankari: Filed https://github.com/e-mission/e-mission-phone/issues/66

www/js/common/map.js
```diff
@@ -44,17 +47,79 @@ angular.module('emission.main.common.map',['ionic-datepicker',
...
+
+  /*
+   * if given group is the selected group, deselect it
+   * else, select the given group
+   */
+  $scope.toggleGroup = function(group) {
+    group.show = !group.show;
+    if (group.show) {
+      var vals = [];
+      var probs = group.probabilites[group.start_times[0].weekday];
```
shankari: I am not sure what you are doing here, but it seems incorrect - it looks like you are returning the probability of the first start time. I see no evidence that the start times are sorted by most popular -> least popular. You need to use `binArray` to get the distribution of start times.

Unless there is an nvd3 `histogram` that will do the computation for you? I think that there's at least a raw d3 `histogram`.
https://github.com/d3/d3-array/blob/master/README.md#histograms

shankari: https://github.com/e-mission/e-mission-phone/issues/67

shankari:
> 1. cplace detail remove for now, it has no more content that cplace list

Actually cplace detail shows the list of successors. I am not sure that all the successors to a common place have a common trip for them. Need to explore the model further.

shankari:
I'm going to merge this for now so that I can build on it, and have filed issues for later cleanup.

shankari:
> cplace detail remove for now, it has no more content that cplace list

Now that I've pulled and installed, I seriously disagree with this. The cplace detail would show the list of successors, and most importantly, if you clicked on any of them, you would be able to go to that detail and so on.

So you could navigate through the common places graph from node to node. The new functionality just shows a list of places and removes the navigation, which was the most interesting/important part.

shankari:
we have to restore the navigation before I release this weekend.
we can wait until next week for the other cleanup.

shankari:
also, why was the number of successors removed?

PR #69, opened 2016-07-14
Port the stats screen over to the phone

shankari:
simulator screen shot jul 13 2016 7 46 32 pm

### DIFF
--- a/www/js/metrics.js
+++ b/www/js/metrics.js
@@ -1,0 +1,184 @@
+'use strict';
+
+angular.module('emission.main.metrics',['nvd3', 'emission.services'])
+
+.controller('MetricsCtrl', function($scope, $ionicActionSheet, CommHelper) {
+    $scope.options = {
+        chart: {
+            type: 'multiBarChart',
+            height: 450,
+            margin : {
+                top: 20,
+                right: 20,
+                bottom: 40,
+                left: 55
+            },
+            showValues: true,
+            stacked: false,
+            x: function(d){ return d[0]; },
+            y: function(d){ return d[1]; },
+            /*
+            average: function(d) { 
+                var vals = d.values.map(function(item){
+                    return item[1];
+                });
+                return d3.mean(vals);
+            },
+            */
+
+            color: d3.scale.category10().range(),
+            // duration: 300,
+            useInteractiveGuideline: true,
+            // clipVoronoi: false,
+
+            xAxis: {
+                axisLabel: 'Date',
+                tickFormat: function(d) {
+                    return d3.time.format('%y-%m-%d')(new Date(d * 1000))
+                },
+                showMaxMin: false,
+                staggerLabels: true
+            },
+            yAxis: {
+              axisLabel: "Number",
+              axisLabelDistance: -10
+            },
+        }
+    };
+
+    $scope.data = [];
+
+    $scope.getMetrics = function() {
+      var data = {
+        freq: 'DAILY',
+        start_time: $scope.selectCtrl.fromDate,
+        end_time: $scope.selectCtrl.toDate,
+        metric: $scope.selectCtrl.metric
+      };
+      console.log("Sending data "+JSON.stringify(data));
+      CommHelper.getMetrics("local_date", data, function(response) {
+        if (angular.isDefined(response.user_metrics)) {
+          console.log("Got aggregate result "+response.user_metrics.length);
+          $scope.$apply(function() {
+              $scope.showCharts(response.user_metrics)
+          });
+        } else {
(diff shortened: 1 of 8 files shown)


www/js/services.js
```diff
@@ -48,6 +48,15 @@ angular.module('emission.services', [])
         var dateString = date.startOf('day').format('YYYY-MM-DD');
         window.cordova.plugins.BEMServerComm.postUserPersonalData("/habiticaProxy", "callOpts", callOpts, successCallback, errorCallback);
     };
+
+    this.getMetrics = function(timeType, metrics_query, successCallback, errorCallback) {
```
sunil07t: This method is like habiticaproxy method right?

www/templates/main-metrics.html
```diff
@@ -0,0 +1,60 @@
...
+
+      <!--
+      <em-local-date date="selectCtrl.fromDate"></em-local-date>
+      <em-local-date date="selectCtrl.toDate"></em-local-date>
+      -->
+      <!--
+        <button style="" class="button button-stable button-block  icon ion-checkmark">Valid</button>
+      -->
+    </div>
+    <nvd3 options="options" data="data"></nvd3>
```
sunil07t: This is pretty cool, just a line of code in HTML for the nvd to show up

www/js/services.js
shankari: Yup. Another example of making an authenticated call to the server.

www/templates/main-metrics.html
shankari: Yeah I've been pretty happy with angular-nvd3 so far

www/js/metrics.js
```diff
@@ -0,0 +1,184 @@
...
+            useInteractiveGuideline: true,
+            // clipVoronoi: false,
+
+            xAxis: {
+                axisLabel: 'Date',
+                tickFormat: function(d) {
+                    return d3.time.format('%y-%m-%d')(new Date(d * 1000))
+                },
+                showMaxMin: false,
+                staggerLabels: true
```
sunil07t: Is there other way instead of stagger like make the font smaller or the month, day and year to only show day and month and show the year only once on the first label.

www/js/metrics.js
```diff
@@ -0,0 +1,184 @@
...
+
+    $scope.showCharts = function(agg_metrics) {
+        var mode_bins = {};
+        agg_metrics.forEach(function(metric) {
+            for (var field in metric) {
+                // TODO: Consider creating a prefix such as M_ to signal
+                // modes. Is that really less fragile than caps, though?
+                // Here, we check if the string is all upper case by 
+                // converting it to upper case and seeing if it is changed
+                if (field == field.toUpperCase()) {
```
sunil07t: Why do we need to check if the field is upper case? What happens if it is not uppercase?

www/templates/main-metrics.html
```diff
@@ -0,0 +1,60 @@
+<ion-view view-title="Metrics">
+  <ion-nav-bar class="bar-stable">
+    <ion-nav-back-button></ion-nav-back-button>
+    <ion-nav-buttons side="right">
+      <button class="button button-assertive button-block"
+              ng-click="getMetrics()">Get!</button>
```
sunil07t: Can we move the Get button on the content instead of nav-bar the lower half of the button is floating below the navbar.

www/js/metrics.js
shankari: Yes, it is possible to control the formatting of the ticks.
e.g. from 
https://krispo.github.io/angular-nvd3/#/cumulativeLineChart

```
function (d) {
    "use strict";
    return d3.time.format('%m/%d/%y')(new Date(d))
}
```

Would be harder to maintain state in the ticks, but I assume it can be done.

www/js/metrics.js
shankari: This is related to the data format of the JSON.

As we can see from:
https://github.com/e-mission/e-mission-server/blob/master/emission/core/wrapper/modestattimesummary.py

there is a prop for each mode with the name in upper case, so 'ON_FOOT', 'BICYCLING', etc
the other properties are in lower case. So I use lower case v/s upper case to distinguish between modes and other properties like the timestamp.

www/templates/main-metrics.html
shankari: I tried to have it as part of the content back when I made the heatmap screen, but the problem is that everything else comes in twos while the button is single. Putting it in the content creates a new col which is a problem when we already have very little width.

Maybe we could move it under the metric name? Anyway, my goal was to create a functional and not too ugly UI.

My hope is that now some of the you can work on the design - we will reuse it for heatmap, metrics and potentially for other searches (trips, common trips) in the future.

PR #71, opened 2016-07-15
Integrate with ionic deploy

shankari:
- Add ionic platform to bower requirements
- Add deploy plugin to package.json
- New `app_id` generated and stored in `ionic.project` by `ionic io init`
- Dependency on `'ionic.service.core'` added to the app
- Refactor the SplashCtrl into two service to keep the controller light and to
  allow methods to be invoked from multiple controllers
- Add an invocation for checkAndUpdate to the startup
- Add a button + function that can check for the update to the `CtrlController`
- Implement an UpdateCheck service that will
    - display the release notes if an update is available and ask the user if she wants to accept the update
    - if so, apply the update
    
Testing done:
    
```
$ ionic emulate ios
```
    
Edit `index.html` and change the title bar color to green
    
```
$ ionic upload --note "green title"
```

- Go to the ionic dashboard and deploy the app to dev
- Click on the check for updates button in the profile screen
- Update was detected
- Apply
- App reloads
- Background is green

Final note: I don't think that I am using promises the right way here. I want to rework this to use promises properly, but I am not sure that I have the time to do so. I'm going to spend some time working through the promise stuff again because I think that it can make the code much nicer. I also want to create a promisified usercache, and it will help with that as well.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -10,7 +10,8 @@
     "leaflet": "^0.7.7",
     "ui-leaflet": "^1.0.0",
     "Leaflet.awesome-markers": "^2.0.2",
-    "ionic-datepicker": "^0.9.0"
+    "ionic-datepicker": "^0.9.0",
+    "ionic-platform-web-client": "^0.7.1"
   },
   "resolutions": {
     "angular": "1.5.3",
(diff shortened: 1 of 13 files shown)


shankari:
If the deploy check consistently fails on real devices, we need to comment back the check in the state change

Issue #66, opened 2016-07-08
Refactor getDisplayName to be generic


shankari:
There are now three minor variations of `getDisplayName` - one for the standard diary and two for common trip and place views. Rewrite to a generic version. I would suggest passing in a function to extract lat/lng (in addition to the object) and return a promise with the name that can be used to set the value in the object.

shankari:
@krisma, here's another one for cleanup

PR #73, opened 2016-07-16
charts

krisma:
Added: start time distribution chart, duration distribution chart, speed chart
simulator screen shot jul 16 2016 2 08 39 am
simulator screen shot jul 16 2016 2 09 00 am
simulator screen shot jul 16 2016 2 08 58 am


krisma:
I don't know why it has conflicts. I merged before checkout to new branch

shankari:
> I don't know why it has conflicts. I merged before checkout to new branch

When did you checkout new branch? I merged both my changes for ionic deploy and Sunil's changes for the goal tab yesterday.

shankari:
Graphs look great.
Couple of comments:
- You need a label for the y-axis in the speed graph. I think that it is m/s
- since we are short of space, you can skip entries that are below the min value and above the max value. You should retain 0 entries in the middle. So you can remove 18-23 but should retain 14.

Taking a look at the code now.

shankari:
Do you have a label for the x axis in the start time histogram? I can't see from the screenshots.

shankari:
I'm a little surprised to see such a wide distribution in the trip durations (`5h` versus `13h`). Not for this week, but for next week, I think that it would be cool to be able to click on those outliers show a list of (not-common) trips that match them, so that we can see whether they make sense or whether we should ignore them.

We may need to add new server-side API methods to support this. Let's talk about the best way to do this.

shankari:
And in terms of interactivity, for the speed profile in the detail view, if I click on a section, can it highlight the appropriate range in the graph?

I'll make an issue with a list of cool visualizations and we can prioritize them wrt other changes next week.

www/js/common/services.js
```diff
@@ -119,7 +126,107 @@ angular.module('emission.main.common.services', [])
...
+        }
+        if (address["city"]) {
+          name = name + ", " + address["city"];
+        } else if (address["town"]) {
+          name = name + ", " + address["town"];
+        } else if (address["county"]) {
+          name = name + ", " + address["county"];
+        }
+        console.log("got response, setting display name to "+name);
+        switch (mode) {
```
shankari: as we discussed, make this return a promise instead
#66

www/js/common/trip-detail.js
```diff
@@ -0,0 +1,83 @@
...
+    nv.utils.windowResize(chart1.update);
+    nv.addGraph(chart1);
+    
+    
+    var distributionDuration = $scope.trip.durations;
+    var valsDuration = [];
+    var nob = Math.floor(Math.sqrt(distributionDuration.length)); // # of bins using square-root choice
+    var mMax = Math.max(...distributionDuration);
+    var mMin = Math.min(...distributionDuration);
+    var valsDuration = d3.layout.histogram().bins(nob).range([mMin, mMax])(distributionDuration);
```
shankari: if the range is already [mMin, mMax] then why do we see the values from 18-23 in the screenshot?

shankari: ah nvm, this is the distribution histogram.

www/js/common/trip-detail.js
```diff
@@ -0,0 +1,83 @@
...
+                                        CommonGraph, $window) {
+    $scope.tripId = $stateParams.tripId;
+    $scope.trip = CommonGraph.data.cTripId2ObjMap[$scope.tripId];
+    var distributionStartHour = $scope.trip.common_hour_distribution;
+    var valsStartHour = [];
+    for (var i = 0; i < 24; i++) {
+      if (distributionStartHour.get(i)) {
+        valsStartHour.push({'label': i + ':00', 'value': distributionStartHour.get(i) });
+      } else {
+        valsStartHour.push({'label': i + ':00', 'value': 0 });
```
shankari: Ah this is why. you need to get max and min appropriately and then stop pushing here.

www/js/common/trip-detail.js
```diff
@@ -0,0 +1,83 @@
...
+        .call(chart1)
+        .attr('width', 350).attr('height', 1000);
+
+    nv.utils.windowResize(chart1.update);
+    nv.addGraph(chart1);
+    
+    
+    var distributionDuration = $scope.trip.durations;
+    var valsDuration = [];
+    var nob = Math.floor(Math.sqrt(distributionDuration.length)); // # of bins using square-root choice
```
shankari: why square root? Please document for others to understand.

www/js/diary/detail.js
```diff
@@ -32,4 +32,44 @@ angular.module('emission.main.diary.detail',['ui-leaflet',
   $scope.tripgj = DiaryHelper.directiveForTrip($scope.trip);
   
   console.log("trip.start_place = " + JSON.stringify($scope.trip.start_place));
+
+  var data  = [];
+  var totalDistance = 0;
+  for (var s in $scope.tripgj.sections) {
+    for (var p in $scope.tripgj.sections[s].properties.distances) {
+      totalDistance += $scope.tripgj.sections[s].properties.distances[p];
```
shankari: I thought that my feedback was to have speed versus time. You get the time from the points, which are ~ 30 secs apart.

PR #74, opened 2016-07-17
Chartspr

krisma:
Charts optimized.
0. Overlap resolved.
1. Now the bar width is constant. I estimate a linear function for bar height. d3 charts use transform:translate to control size so height and width are Nan. I have to estimate.
2. Head and tail 0s are eliminated.
simulator screen shot jul 17 2016 5 14 18 am
simulator screen shot jul 17 2016 5 14 12 am
simulator screen shot jul 17 2016 5 14 08 am

### DIFF
--- a/.io-config.json
+++ b/.io-config.json
@@ -1,1 +1,0 @@
-{"app_id":"66d1ea4e","api_key":"b29c0fee3007e42d1d16a3f547c469865283eff9089fb0d8"}
--- a/.io-config.json
+++ b/.io-config.json
@@ -1,0 +1,1 @@
+{"app_id":"66d1ea4e","api_key":"b29c0fee3007e42d1d16a3f547c469865283eff9089fb0d8"}
--- a/ionic.project
+++ b/ionic.project
@@ -1,6 +1,6 @@
 {
   "name": "e-mission-phone",
-  "app_id": "9a6e5f95",
+  "app_id": "66d1ea4e",
   "browsers": [
     {
       "platform": "android",
--- a/ionic.project
+++ b/ionic.project
@@ -1,6 +1,6 @@
 {
   "name": "e-mission-phone",
-  "app_id": "66d1ea4e",
+  "app_id": "9a6e5f95",
   "browsers": [
     {
       "platform": "android",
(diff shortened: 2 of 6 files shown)


krisma:
Y axis label added. It requires a fixed amount of left margin to give space to the label. I used 65 and only tested on iPhone6 
screen shot 2016-07-17 at 5 37 06 am

.io-config.json
```diff
@@ -0,0 +1 @@
+{"app_id":"66d1ea4e","api_key":"b29c0fee3007e42d1d16a3f547c469865283eff9089fb0d8"}
```
shankari: while these may be necessary for you to get everything to run, please don't include them as part of the pull request. We have a particular `app_id` and `api_key` for our project on the ionic platform and overwriting it means that ionic deploy won't work any more.

ionic.project
```diff
@@ -1,6 +1,6 @@
 {
   "name": "e-mission-phone",
-  "app_id": "9a6e5f95",
+  "app_id": "66d1ea4e",
```
shankari: this too

www/index.html
```diff
@@ -26,7 +26,8 @@
 
     <!-- ionic/angularjs js -->
     <script src="lib/ionic/js/ionic.bundle.js"></script>
-<script src="lib/ionic-platform-web-client/dist/ionic.io.bundle.min.js"></script>
+    <script src="lib/ionic-service-core/ionic-core.js"></script>
```
shankari: is this required? it works for me without it.

www/js/common/trip-detail.js
```diff
@@ -11,14 +11,29 @@ angular.module('emission.main.common.trip-detail',['ui-leaflet',
     $scope.tripId = $stateParams.tripId;
     $scope.trip = CommonGraph.data.cTripId2ObjMap[$scope.tripId];
     var distributionStartHour = $scope.trip.common_hour_distribution;
-    var valsStartHour = [];
+    var valsStartHour = []; 
+    var state = 1; // state machine
```
shankari: cool @state machine

shankari:
just to clarify, I am waiting for the following files to be reverted before merging and releasing.

- `.io-config.json`
- ` ionic.project`
- `www/index.html`

PR #84, opened 2016-07-19
Added Party and Challenge tabs on Goals screen

sunil07t:
Work in progress, ignore HTML for now and need to make animations smoother.


package.json
```diff
@@ -79,10 +79,6 @@
       },
       "locator": "cordova-plugin-crosswalk-webview",
       "id": "cordova-plugin-crosswalk-webview"
-    },
-    {
-      "locator": "../me.apla.cordova.app-preferences",
```
shankari: nit: you don't want to replace this completely - just with a reference to the original appPreference.

www/js/goals.js
```diff
@@ -61,38 +81,239 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize'])
...
-			console.log("error");
+			console.log("User profile error");
+		});
+	};
+
+	userInfo();
+
+	var callOpts = {'method': 'GET', 'method_url': "/api/v3/tasks/user",
+                    'method_args': null};
+    var tasks;
```
shankari: nit: fix indentation for readability

www/js/goals.js
```diff
@@ -61,37 +81,334 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize'])
 		});
 		CommHelper.habiticaRegister(regConfig, function(response) {
 			console.log("Success!")
-			$scope.screen = response.success;
-			console.log(response);
-			$window.location.reload();
+			userInfo();
```
shankari: nit: for readability, I would use something like `fetchUserInfo()` or `getUserInfo()` to indicate that it actually returns the `userInfo`

www/js/goals.js
```diff
@@ -61,37 +81,334 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize'])
...
 		});
 
-	var callOpts = {'method': 'GET', 'method_url': "/api/v3/user",
+
+	$scope.refreshUserInfo = function(){
+		console.log("Refreshing user information");
+		userInfo();
+	};
+
+    var userInfo = function(){
```
shankari: nit: indentation

www/templates/main-goals.html
```diff
@@ -73,45 +79,148 @@
...
-            <div class="col col-80" style="padding-left: 5% !important;">
-            <h2>{{goal.name}}</h2>
-            <p style="font-size: 10px">{{goal.note}}</p>
-            </div>
-            <ion-option-button class="button-assertive" ng-click="removeGoal(goal)">Delete
-            </ion-option-button>
+      <div class="row" id="costume-tab">
+        <div class="col">
+          <button id="challenge" class="button" ng-class="{'active': isActive}" ng-click="activeButton()" type="button">Challenges</button>
+          <button id="party" class="button" ng-class="{'active': isActiveP}" ng-click="partyBotton()" type="button">Party</button>
```
shankari: nit: minor typos such as `Botton`. Please do a quick spellcheck and fix in the next round.

www/js/goals.js
```diff
@@ -61,37 +81,334 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize'])
...
  
+
     $ionicLoading.show({
 			template: '<ion-spinner icon="bubbles" class="costume"></ion-spinner>'
 		});
 
-	var callOpts = {'method': 'GET', 'method_url': "/api/v3/user",
+
+	$scope.refreshUserInfo = function(){
+		console.log("Refreshing user information");
```
shankari: This is really the only substantive comment for the PR. 
It looks like `refreshUserInfo` refreshes (duh) the user info.
But the bugs that we saw yesterday were around refreshing party state, party member list, challenge information, etc.

How do I refresh all of those?

My suggestion is to make this `refreshInfo` and have it call all the functions - `userInfo`, `userTask`, `getMembers`, etc. and then you can just call `refreshInfo()` when the page loads instead of calling `userInfo()`, `userTask()`, `getMembers()` etc

www/js/goals.js
```diff
@@ -61,37 +81,340 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize'])
...
+		getUserTask();
+		getChallenges();
+	};
+
+	refreshInfo();
+
+	$scope.refreshPage = function() {
+		console.log("Refreshing page");
+		refreshInfo();
+		$scope.$broadcast('scroll.refreshComplete');
```
shankari: I think that you should only broadcast when all the refreshing is complete. Note that `refreshInfo()` is not synchronous

www/js/goals.js
```diff
@@ -61,37 +81,340 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize'])
...
     $ionicLoading.show({
 			template: '<ion-spinner icon="bubbles" class="costume"></ion-spinner>'
+	});
+
+	var getUserInfo = function(){
+		var callOpts = {'method': 'GET', 'method_url': "/api/v3/user",
+	                    'method_args': null};
+
+		CommHelper.habiticaProxy(callOpts, function(response){
+			$scope.screen = response.success;
```
shankari: I am not sure that this works the way that you want. Note that the `habiticaProxy` method is asynchronous. Let us assume that you invoke it when the controller is instantiated. It is not going to be true immediately - in particular, when the HTML is loading and the value of the variable is evaluted in `ng-if`.

Then, when the async call does complete and you set it, angular will not re-read the variable unless you indicate that it has changed. You are not doing `$scope.$apply()` so I don't see why that would happen.

Are you sure that this works? Not just by reloading from safari, but from actually relaunching the app in the emulator.

shankari: Maybe I can see a demo at 10pm.

Issue #43, opened 2016-06-20
Start tracking only after the user has accepted the consent document


shankari:

shankari:
Actually, this requires some fiddling with the plugin code so assigning to myself for now.

shankari:
Since we are building a platform, and we anticipate that we may need to amend the consent document from time to time, we want to tie this to the version of the consent document.

So the desired behavior is:
- on the phone:
  - if the consented version is not the current version, pop up the consent document
  - else go on to the last selected view
- on the server:
  - if the consented version is not the current version, generate a local notification asking the user to consent
  - if the consented version is the current version, start tracking as usual

This means that the requirements are:
- we need to need to access both the current version and the consented version from both native code and javascript
- since it runs at startup, it would be preferable if the version checking in javascript did not use native plugins. Note that the app preference plugin that we used for this before had issues with using localStorage instead of real native preferences (https://github.com/e-mission/e-mission-phone/issues/56, https://github.com/e-mission/e-mission-data-collection/issues/88)
- it would also be preferable for us to have a record of when the user consented to which version. This is just good for record keeping in general and will also allow us to ignore/discard any data collection before consent was obtained.

Given the considerations above, I plan to have separate and parallel storage in javascript and native code.

- for native code, it looks like other plugins have their preferences in `config.xml` like `<preference name="DisallowOverscroll" value="true"/>`.

  The value is then read on android like `if (preferences.getBoolean("DisallowOverscroll", false)) {` and on iOS like `BOOL bounceAllowed = !([settings cordovaBoolSettingForKey:@"DisallowOverscroll" defaultValue:NO]);`

  Since our preference is for a plugin instead of for the main webview, we can't use that directly on iOS. So we use this example from the LocalStorage plugin instead `self.commandDelegate.settings[[@"BackupWebStorage" lowercaseString]]];`
- for javascript, we just create a consentConfig.js that has a json object with the values and the corresponding consent documents. This does mean that we have to replicate the consent version number multiple times, but it's only two times, so hopefully it is not too bad. Later, it would be good to convert this to a `build_hook` instead.

Then, the flow will be:
- before consent: plugin initialize reads value and generates notification
- before consent: app launch pops up consent screen
- disagree: goes to aggregate screen
- before consent: going to any page with user information pops up the consent screen again
- consent: saves consented value to both localStorage and to db as a message
- after consent: plugin initialize does actual initialization
- after consent: all screens work without popups

shankari:
[...]

PR #87, opened 2016-07-22
Switch the startup storage to localStorage + support re-consent

shankari:
The goal is to remove all native code from the splash screen in order to allow
startup to proceed more quickly.

This should help us with issues like https://github.com/e-mission/e-mission-phone/issues/56 in which preferences don't appear to be read correctly on startup.


shankari:
This is fixes the javascript parts of
https://github.com/e-mission/e-mission-phone/issues/43
and
https://github.com/e-mission/e-mission-phone/issues/56

PR #90, opened 2016-07-25
Goal tab improvements.

sunil07t:
Better UI and functionality.  Show only positive or negative for certain goal, larger stats and bar size and promisified the habitica API.


shankari:
ok to test

Issue #92, opened 2016-07-26
ControlCtrl is extremely verbose now and should be restructured


shankari:
From a software engineering perspective, `ControlCtrl` is now way too verbose and complicated. It should be restructured into a proper MVC pattern.

https://github.com/e-mission/e-mission-phone/pull/88

either @krisma or @sunil07t can do this

shankari:
If you want to see an example, see my code for the splash screen, now split between `controllers.js` and the two javascript files in `splash`. Also note how I hide details of the storage implementation in the service. This makes it easier to change the storage later without changing all the code.

shankari:
This is also true for `metrics.js`, which is currently 934 lines of code, and includes at least 2 fairly big factories.

sunil07t:
I can look into ControlCtrl to start with and that way I can learn the functionality of this controller too.

shankari:
Also, instead of wrapping four promisifed wrappers for `CommHelper.getMetrics`, the function should itself be promisified.

PR #104, opened 2016-08-09
controlCtrl and metrics.js cleaning

sunil07t:
Local storage implementation is in the factory instead of control, calls to server are in the service.js, style codes are in css instead of HTML and controlCtrl is in separate file to have a proper MVC pattern. Also, relocated factory from metrics.js to metrics-factory.js and implemented promise wrapper to all the controlCtrl functions that calls to the server.  Fix for issue #92.


www/js/control.js
```diff
@@ -0,0 +1,431 @@
...
+                $scope.forceTransition(button.transition);
+                return true;
+            }
+        });
+    };
+    $scope.editCollectionConfig = function($event) {
+        $scope.settings.collect.new_config = JSON.parse(JSON.stringify($scope.settings.collect.config));
+        console.log("settings popup = "+$scope.collectSettingsPopup);
+        $scope.collectSettingsPopup.show($event);
+        /*
```
shankari: I think you can remove this since we haven't used it for a while

www/js/control.js
```diff
@@ -0,0 +1,431 @@
...
+        UserCalorieData.delete();
+        $ionicPopup.alert({template: 'User data erased.'});
+
+    }
+    $scope.parseState = function(state) {
+        if (state) {
+            return state.substring(6);
+        }
+    }
+    $scope.toggleCollection = function() {
```
shankari: Can we rename this? It sounds like we are _toggling_ (turning on and off) collection.

www/js/control.js
```diff
@@ -0,0 +1,431 @@
...
+                    v = k;
+                    break;
+                }
+            }
+            if ($scope.isIOS()) {
+                return v != "kCLLocationAccuracyBestForNavigation" && v != "kCLLocationAccuracyBest" && v != "kCLLocationAccuracyTenMeters";
+            } else if ($scope.isAndroid()) {
+                return v != "PRIORITY_HIGH_ACCURACY";
+            }
+
```
shankari: maybe add an `else` here and popup an "unsupported" alert, just to make sure that we remember to extend this if we support other platforms.

www/js/control.js
```diff
@@ -0,0 +1,431 @@
...
+                $scope.settings.collect.new_config.accuracy = $scope.settings.collect.accuracyOptions["kCLLocationAccuracyHundredMeters"];
+            } else if ($scope.isAndroid()) {
+                $scope.settings.collect.new_config.accuracy = $scope.settings.collect.accuracyOptions["PRIORITY_BALANCED_POWER_ACCURACY"];
+            }
+        }
+        ControlHelper.dataCollectionSetConfig($scope.settings.collect.new_config)
+        .then(function(){
+            console.log("setConfig Sucess");
+        }, function(err){
+            console.log("setCongif Error: " + err);
```
shankari: nit: typo

www/js/control.js
```diff
@@ -0,0 +1,431 @@
...
+        if ($scope.dark_theme) {
+            $rootScope.dark_theme = false;
+            $scope.dark_theme = false;
+            StartPrefs.setDefaultTheme(null);
+        } else {
+            $rootScope.dark_theme = true;
+            $scope.dark_theme = true;
+            StartPrefs.setDefaultTheme('dark_theme');
+        }
+        $ionicPopup.alert({template: 'Restart the app to see all changes!'})
```
shankari: You should be able to reload the app automatically (recall that all of this is HTML after all. According to 
http://stackoverflow.com/questions/21885518/angularjs-reload-page

`$window.location.reload();` should work

www/js/localstorage.js
```diff
@@ -0,0 +1,36 @@
+'user strick';
+
+angular.module('emission.main.localstorage',['angularLocalStorage'])
```
shankari: nice! I also have wrappers for the local storage, but I create modules based on where they are used. With that pattern, you would use a CalorieCalc module instead, which would encapsulate load, store and calculation.

www/js/main.js
```diff
@@ -2,15 +2,13 @@
...
                                  'emission.main.diary',
+                                 'emission.main.control',
                                  'emission.main.goals',
                                  'emission.main.common',
                                  'emission.main.heatmap',
                                  'emission.main.metrics',
                                  'ngCordova',
-                                 'emission.services',
-                                 'emission.splash.updatecheck',
-                                 'emission.splash.startprefs',
```
shankari: yay!

www/js/metrics-factory.js
```diff
@@ -0,0 +1,157 @@
+'use strict';
+
+angular.module('emission.main.metrics.factory', [])
+
+.factory('FootprintHelper', function() {
```
shankari: nice!

www/js/metrics-factory.js
```diff
@@ -0,0 +1,157 @@
...
+      return readable(footprint.train * mtokm(distance)) + ' ~ ' + readable(footprint.car * mtokm(distance));
+    } else {
+      return readable(footprint[mode] * mtokm(distance));
+    }
+  }
+  return fh;
+})
+
+.factory('CalorieHelper', function(){
+  Number.prototype.between = function (min, max) {
```
shankari: nice! I've generally been putting load/store functions into the factory instead of having a separate module for load/store like you do with the user data.

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
                "Email cancel reported, seems to be an error on android");
         });
     }
+
+    this.dataCollectionSetConfig = function(config) {
+      return new Promise(function(resolve, reject) {
```
shankari: I don't think you need these - the data collection interface is already promisified.

https://github.com/e-mission/e-mission-data-collection/blob/master/www/datacollection.js#L39

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+      });
+    };
+
+    this.dataCollectionGetConfig = function() {
+      return new Promise(function(reslove, reject) {
+          reslove(window.cordova.plugins.BEMDataCollection.getConfig());
+      });
+    };
+    this.serverSyncSetConfig = function(config) {
+      return new Promise(function(resolve, reject) {
```
shankari: these are not yet promisified, but since they are our plugins, we should just change our plugin to be promisified.

You just make the change to this repo:
https://github.com/e-mission/cordova-server-sync

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+    };
+
+    this.serverSyncGetConfig = function() {
+      return new Promise(function(reslove, reject) {
+          reslove(window.cordova.plugins.BEMServerSync.getConfig());
+      });
+    };
+
+    this.getAccuracyOptions = function() {
+      return new Promise(function(reslove, reject) {
```
shankari: should already be a promise

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+          reslove(window.cordova.plugins.BEMDataCollection.getAccuracyOptions());
+      });
+    };
+
+    this.getUserEmail = function() {
+      return new Promise(function(reslove, reject) {
+        window.cordova.plugins.BEMJWTAuth.getUserEmail(function(response) {
```
shankari: need to convert to a promise

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+            reslove(response);
+        }, function(error) {
+            reject(error);
+        });
+      });
+    };
+
+    this.getState = function() {
+      return new Promise(function(reslove, reject) {
+        window.cordova.plugins.BEMDataCollection.getState(function(response) {
```
shankari: should already be a promise

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+            reslove(response);
+        }, function(error) {
+            reject(error);
+        });
+      });
+    };
+
+    this.getSettings = function() {
+      return new Promise(function(reslove, reject) {
+        window.cordova.plugins.BEMConnectionSettings.getSettings(function(response) {
```
shankari: Need to promisify interface.

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+            reslove(response);
```
shankari: typo. Does this even work?

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+        }, function(error) {
+            reject(error);
+        });
+      });
+    };
+
+    this.forceTransition = function(transition) {
+      return new Promise(function(resolve, reject) {
+        resolve(window.cordova.plugins.BEMDataCollection.forceTransition(transition));
```
shankari: should already be a promise

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+      });
+    };
+
+    this.forceSync = function() {
+      return new Promise(function(resolve, reject) {
```
shankari: change the plugin to make it a promise

www/js/services.js
```diff
@@ -102,6 +102,91 @@ angular.module('emission.services', [])
...
+        window.cordova.plugins.BEMServerSync.forceSync(function(response) {
+          resolve(response);
+        }, function(error) {
+           reject(error); 
+        });
+      });
+    };
+
+    this.getDocument = function() {
+      return new Promise(function(resolve, reject) {
```
shankari: usercache should already be a promise I think

www/templates/control/main-control.html
```diff
@@ -1,29 +1,14 @@
...
-
   <ion-content>
-<!--     <div style="text-align: center;">
-      <img src="img/avatar_1.png" ng-style="getAvatarStyle()">
-    </div> -->
-
     <div class="control-list-item">
       <div class="control-list-text">{{settings.auth.email}}</div>
-      <div ng-click="returnToIntro()" ng-style="getIconButtonStyle('E0A1A1')"><i class="icon ion-log-out" ng-style="getIconStyle()"></i></div>
+      <div ng-click="returnToIntro()" id ="signout" class="icon-button-style"><i class="icon ion-log-out"></i></div>
```
shankari: not sure that this should be called "signout", because it really doesn't sign you out. If you click on this and then kill the app and then restart it, for example, I think that the prior login will still work.

www/templates/control/main-control.html
```diff
@@ -1,29 +1,14 @@
...
     </div>
     <div class="control-list-item">
       <div class="control-list-text">Tracking</div>
       <label ng-click="userStartStopTracking()" class="toggle control-list-toggle">
         <input type="checkbox" ng-checked="trackingOn()">
-        <div class="track" style="border-color: #2bb096; background-color: #2bb096;">
+        <div class="track toggle-color">
```
shankari: nice! wrt css changes

www/js/main.js
```diff
@@ -144,476 +133,4 @@ angular.module('emission.main', ['emission.main.recent',
...
-            'text-align': 'center',
-            'float': 'right',
-            'height': '100%',
-            'background-color': '#' + color,
-            'color': '#fff',
-            'padding': '15px 15px',
-            'width': + ($window.screen.width * 0.25).toString() + 'px'
-        }
-    }
-    $scope.getIconButtonStyle = function(color) {
```
shankari: I'm glad that this got removed :)

www/js/control.js
```diff
@@ -0,0 +1,431 @@
...
+    };
+
+    $scope.getCollectionSettings = function() {
+        var promiseList = []
+        ControlHelper.dataCollectionGetConfig().then(function(response){
+            promiseList.push(response);
+        }).then(function(){
+            ControlHelper.getAccuracyOptions().then(function(response){
+                promiseList.push(response);
+            }).then(function(){
```
shankari: I am not sure that this is the best way to do this. With this, you are making the first call and then the second call and then dealing with the results. But the second call doesn't need any data from the first call, so it can execute in parallel with it.

To do that, I would put the two promises into the list and then use `Promise.all()` to wait for both of them to be complete and then use the results. Look at https://github.com/e-mission/cordova-usercache/blob/master/www/usercache.js#L125 for an example.

shankari:
I really liked the refactoring.
Note that the data collection interface is already using promises.
The other plugins don't, but since I wrote a lot of the plugins, we can just convert the interfaces into promises. That and the `.all()` are the only two substantive changes - the others are mainly nits.

www/js/localstorage.js
sunil07t: Should I make this module specific to calorie data or this broad local storage module better so we can add factories to this module later.

www/js/services.js
sunil07t: it only worked because my parameter deceleration also had a typo.

www/js/services.js
sunil07t: Do I promisify the plugin for BEMConnections?

shankari: Good point. There should not be many, but some simple grepping can help figure out if there are.

shankari: e.g. ServerSync is only used in control

```
bash-3.2$ grep -r BEMServerSync www/js
www/js/main.js:        promiseList.push(window.cordova.plugins.BEMServerSync.getConfig());
www/js/main.js:            window.cordova.plugins.BEMServerSync.forceSync(function(result) {
www/js/main.js:        window.cordova.plugins.BEMServerSync.setConfig($scope.settings.sync.new_config)
```

Auth is used in one place other than control

```
bash-3.2$ grep -r BEMJWTAuth www/js
www/js/intro.js:    window.cordova.plugins.BEMJWTAuth.signIn(function(userEmail) {
www/js/main.js:            window.cordova.plugins.BEMJWTAuth.getJWT(function(result) {
www/js/main.js:        window.cordova.plugins.BEMJWTAuth.getUserEmail(function(result) {
www/js/main.js:            window.cordova.plugins.BEMJWTAuth.getJWT(function(result) {
```

Ditto for `ConnectionSettings`

```
bash-3.2$ grep -r BEMConnectionSettings www/js
www/js/intro.js:    window.cordova.plugins.BEMConnectionSettings.getSettings(function(settings) {
www/js/main.js:        window.cordova.plugins.BEMConnectionSettings.getSettings(function(result) {
www/js/services.js:        window.cordova.plugins.BEMConnectionSettings.getSettings(
```

www/js/localstorage.js
shankari: If we follow my pattern, I would have moved this functionality into `CalorieCalc` (see consent code for example). We would not have such a module (which is why we haven't had one so far). 

The main reason for reading and writing in the place where it is used, as opposed to a separate module is the encapsulation of the data model. If we wanted to change the name of one of the fields for the calorie calculation, for e.g. - `heightUnit` to `height_unit`, we would need to change it in both `CalorieCalc` and here. But if we move this function into the `CalorieCalc`, we only need to change it in one place.

So if somebody new wants to touch this code 5 years later, they won't have to know that they need to modify two places.

www/js/services.js
shankari: Would be nice to, yes.

shankari:
Looks good so far!

www/js/services.js
sunil07t: I am trying to promisify the plugins but for the ones with callback like getState()
https://github.com/e-mission/e-mission-data-collection/blob/master/www/datacollection.js#L48
the success and error callback parameter is passed into exec() and promises depends on the response of the callback so is it possible to make this plugin return a promise?

shankari: Everything in `datacollection.js` is already a promise except for this
function (that I probably forgot).

You can use `getAccuracyOptions` right above it as a template for how to
convert it - basically, the `resolve` function is the `successCallback` and
the `reject` function is the `errorCallback`

Shankari

shankari:
Looks good. Only issue left is the `Promise.all()`, I think

shankari:
Looks pretty good modulo README clarifications.
were you going to make the `getMetrics()` change as well or defer that until the dashboard redesign?

sunil07t:
I will make the changes for getMetrics() and readme

shankari:
I've merged all the changes to the other plugins, so you should be able to test those changes too, after doing `ionic state restore`

README.md
```diff
@@ -129,3 +129,44 @@ $ git pull upstream master
...
+
+	3. Copy the gulp-sprites.js file from habitrpg to emission
+
+		$ cp -r habitrpg/tasks/gulp-sprites.js e-mission-phone/www/js/tasks/
+
+	4. Add the following line to e-mission-phone/gulpfule.js to sycn the gulp-sprites.js file
+
+		require('glob').sync('/www/tasks/gulp-*').forEach(require);
+
+	5. Copy the 3 folders from habitrpg/common- css, dist and img and paste it to e-mission-phone/www/common
```
shankari: by this, you just mean copy the folders, right, 

e.g. 

```
$ cp habitrpg/common/css e-mission-phone/www/common
```

sunil07t: Yes, I will push the changes

PR #105, opened 2016-08-09
Newbranch

krisma:
Fixed local date in timeline.
Fixed color in detail view.
Fixed border width.


.io-config.json
```diff
@@ -0,0 +1 @@
+{"app_id":"7c252161","api_key":"3e2dc53de89d0586e307db28007ff543062bb392f1094d75"}
```
shankari: This file is not supposed to be checked in.

shankari: Please remove it from the PR. You can also add it to `.gitignore`

bower.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "angular-nvd3": "^1.0.7",
     "string.startsWith": "*",
-    "angularLocalStorage": "^0.3.2"
+    "angularLocalStorage": "^0.3.2",
+    "angular-local-storage": "^0.3.0"
```
shankari: why do we need the second `angular-local-storage` plugin?

ionic.project
```diff
@@ -1,11 +0,0 @@
-{
-  "name": "e-mission-phone",
```
shankari: Why delete this file?

package.json
```diff
@@ -1,94 +0,0 @@
-{
-  "name": "emission",
```
shankari: this should definitely not be deleted.

www/js/diary/list.js
```diff
@@ -117,6 +117,12 @@ angular.module('emission.main.diary.list',['ui-leaflet',
           readAndUpdateForDay(moment(val));
         }
     }
+    $scope.localTimeString = function(dt) {
+      var hr = ((dt.hour > 12))? dt.hour - 12 : dt.hour;
+      var post = ((dt.hour >= 12))? " pm" : " am";
+      var min = (dt.minute.toString().length == 1)? "0" + dt.minute.toString() : dt.minute.toString();
```
shankari: I wonder if it is easier to continue using the timestamp but with the timezone from the `local_dt`. Care to share your rationale behind this choice?

shankari:
Changes seem fairly straightforward. Basically, fix the files included in the PR and then I can merge.

krisma:
fixed

www/js/diary/list.js
krisma: I think hr+min is more straight-forward. I was about to use 24hr syntax for less code, but then thinking about users would notice the change, I parse it to 12hr.

PR #112, opened 2016-08-14
End of summer milestone issue #102 fix

### LINKED ISSUE #102: Add progress bar for ionic deploy
The current app size is pretty big and is taking a long time to download using `ionic deploy`. So long that the user thinks that nothing is happening, and tries to start the download again. There is support for getting progress updates, which we currently use to print out a message. We should convert this to a progress bar instead.

```
  var applyUpdate = function() {
    deploy.update().then(function(res) {
       'Ionic Deploy: Update Success! ', res);
      var reloadAlert = $ionicPopup.alert("Update done, reloading...");
      reloadAlert.then(function(res) {
        deploy.load();
      });
    }, function(err) {
       console.log('Ionic Deploy: Update error! ', err);
       window.Logger.log(window.Logger.LEVEL_ERROR,
         'Ionic Deploy: Update error! '+ res);
    }, function(prog) {
      console.log('Ionic Deploy: Progress... ', prog);
    });
  };
```

This should be as easy as updating a scope variable in the function instead. The only catch is that since this is a service, we need to use the `$rootScope` instead of `$scope`

sunil07t:
Until now- control button opacity bug, minor readme changes and progress bar during channel UI download. Fixes #102


www/js/splash/updatecheck.js
```diff
@@ -45,6 +45,10 @@ angular.module('emission.splash.updatecheck', ['angularLocalStorage'])
          'Ionic Deploy: Update error! '+ res);
     }, function(prog) {
       console.log('Ionic Deploy: Progress... ', prog);
+      $rootScope.progress = prog;
+      $rootScope.isDownloading = true;
+    }).then(function(){
+      $rootScope.isDownloading = false;
```
shankari: shouldn't this be in the existing `then` implementation above? I don't think that the implementation above returns a promise, does it?

www/templates/control/main-control.html
```diff
@@ -1,8 +1,11 @@
 <ion-view view-title="Profile" ng-class="ionViewBackgroundClass()">
+   <div>
+    <progress ng-if="isDownloading" class=" has-header download" value="{{progress}}" max="100"></progress>
+  </div>
   <ion-content>
     <div class="control-list-item">
       <div class="control-list-text">{{settings.auth.email}}</div>
-      <div ng-click="returnToIntro()" id ="switch-user" class="icon-button-style"><i class="icon ion-log-out"></i></div>
+      <div ng-click="returnToIntro()" id ="switch-user" class="control-icon-button"><i class="icon ion-log-out"></i></div>
```
shankari: Yeah I wondered if this should have the new class as well, but @krisma's change didn't include it.

shankari:
So you ended up using the standard `<progress` instead of a library?

www/js/splash/updatecheck.js
shankari: also, we want to set isDownloading = false in the error case as well, right?

sunil07t:
Yes the progress has more flexibility and using a function thats already available instead of adding a dependency is usually better.

www/js/splash/updatecheck.js
sunil07t: I though this would be promise because update() has a .then() function but this might also be ionic's function. 
The isDownloading is true in the progress callback so I think we do need to set isDownloading to false if there is error during the progress.

shankari: update does return a promise, but the `then` of that promise doesn't return a promise. In fact, it doesn't return anything (no return statement). Given that, I don't think that the return value of update().then() is itself `thenable`

shankari:
just to clarify - I am waiting for a fix for the `isDownloading` before the merge

Issue #107, opened 2016-08-09
Design and implement a "dashboard"


shankari:
Feedback to @juemura - users are confused when they load the app and they can't see anything.
Idea: should start with a "dashboard".

- dashboard is similar to current summary
- dashboard need not have complex search
- dashboard does not need to be part of metrics
- calorie estimation and carbon footprint are both useful but need more explanation

shankari:
The long-term design for this should be using the "Visual Display of Quantitative Information" by Tufte. Will add it to the reading list for CS 199

shankari:
I would like to see a plan for this before it is implemented.

shankari:
I wasn't able to find a calorie burn/day recommendation. The CDC website is all about balancing calorie intake and burn. And in order to find calorie intake, we will need to track their food too, which I don't really want to build in at this point (there are a ton of apps that track food, and it is not clear that we want to spend our time doing something that has been done to death).

One fairly cheap way to relate the number to something real is to convert it to a tangible treat like a chocolate chip cookie. I have a link to the USDA nutrition information in #80 - pick a treat that you think will be relatable (and that you like) and express the burn in terms of that (e.g. 2 chocolate chip cookies). Or if you want to stretch, pick a set of treats (cookie/brownie/donut/ice cream) and allow the user to choose which one to use :)

Or if you want to be super duper aggressive, build in a way to look up the USDA database and really pick their "treat food equivalent" ([greek yoghurt](https://ndb.nal.usda.gov/ndb/foods/show/216?manu=&fgcd=)? [paneer tikka masala](https://ndb.nal.usda.gov/ndb/foods/show/9027?manu=&fgcd=)? [lamb brain](https://ndb.nal.usda.gov/ndb/foods/show/5286?manu=&fgcd=)?) Unfortunately, the USDA database, as expected, doesn't cover things that are not "normal american food" very well.

shankari:
In order to have a responsive dashboard, we can also compute a document that contains all the data for the current dashboard on the server and send it over as part of every sync (similar to what we do with the timeline). Then, the dashboard would load quickly from the cache, and we would make calls to the server if the user changed anything in the dashboard.

we would compute the document as part of the intake pipeline. @juemura knows how to do this but she is busy now. Let's first see how responsive it is with just making the call to the server for now, and I can show one of @krisma or @sunil07t how to do this once the basic redesign is done.

shankari:
dashboard:
- data for the past 7 days
- footprint card
[...]

Issue #93, opened 2016-07-27
Bug to track cleanup issues from the new metrics screen


shankari:
### Issue 1: units for the mode summaries change, but the values are always the same. I think that they are based on distance ###

| Count | Distance | Duration
| ------ | --------- | ---------
| ![simulator screen shot jul 26 2016 6 57 29 pm](https://cloud.githubusercontent.com/assets/2423263/17161523/25ff00c8-5363-11e6-83d4-03864e46b83e.png) | ![simulator screen shot jul 26 2016 6 58 11 pm](https://cloud.githubusercontent.com/assets/2423263/17161522/25fd59e4-5363-11e6-9420-d036921f9909.png) | ![simulator screen shot jul 26 2016 7 00 38 pm](https://cloud.githubusercontent.com/assets/2423263/17161540/48ca169c-5363-11e6-9ddc-cea523ab5268.png) |

shankari:
I think that this is true for the chart too - no matter which metric you select, the value for the 23rd/24th is always 28.7.

| Distance |   Count |
|---------|-----------|
| ![simulator screen shot jul 26 2016 7 04 14 pm](https://cloud.githubusercontent.com/assets/2423263/17161708/7522c3d2-5364-11e6-9ccb-12bad0d9c1fc.png) | ![simulator screen shot jul 26 2016 7 04 38 pm](https://cloud.githubusercontent.com/assets/2423263/17161709/7525a070-5364-11e6-95a7-3b37bd65fab5.png) |

shankari:
This worked when I did some testing earlier, but does not appear to work now. In particular, I have lots and lots of distance in vehicle, but the carbon footprint shown in 0.

| Chart             |    Summary              |
| ------------- | -------------------|
| ![simulator screen shot aug 9 2016 9 11 26 pm](https://cloud.githubusercontent.com/assets/2423263/17541637/38958a86-5e76-11e6-92ce-d98d4ab1f54a.png) | ![simulator screen shot aug 9 2016 9 12 16 pm](https://cloud.githubusercontent.com/assets/2423263/17541638/38958ac2-5e76-11e6-83aa-e598dee977ea.png) |

shankari:
Ok I think I know what's going on. It looks like we use the metric from the chart view to calculate the views from the summary. This works well if the metric is DISTANCE but does not work if it is anything else (e.g. count, duration).

| COUNT            | DURATION     | DISTANCE    |
| -------------- | ------------- | ------------- |
[...]

PR #113, opened 2016-08-22
New Dashboard to fix issue #107 and #93

sunil07t:
3 cards: Footprint, Calories and Metrics

screen shot 2016-08-22 at 10 33 54 am

screen shot 2016-08-22 at 10 34 11 am

screen shot 2016-08-22 at 10 34 27 am

screen shot 2016-08-22 at 10 34 38 am

screen shot 2016-08-22 at 10 34 56 am

Footprint shows last weeks data and compares it with the previous week to get the change in percentage. When the card is extended it show: Optimal, Average, Worst, Last week, CA 2035, CA 2050.

Calories shows the users calories from last week and compares it to number of chocolate chip cookies (default). Users can change the food to either banana, vanilla ice cream or keep the default chocolate chip cookie. The selected food is store to local storage so the users can see that instead of default. Also shows the average and last weeks calories when the care is expanded.

Slide cards that shows the modes with different metrics. Meter is converted to km and seconds is converted to minutes.


shankari:
@sunil07t this looks awesome!
BTW, are you logged in as me or as Prof. Culler?
This person is doing pretty well wrt calories...

shankari:
Do we have a comparison with last week for calories as well?

shankari:
I see that we do. It seems like we could add a trend comparison (% over week) to calories as well

shankari:
I am not sure that I trust the calorie calculations - 100 bananas?

sunil07t:
I am logged into your account. There is one day you walked a lot that is the reason the calories is so high. Should I add the percentage comparison for calories too? Also, for carbon footprint trend I used the average of the range of car and train, is there a better way than comparing the average?

shankari:
> Should I add the percentage comparison for calories too? 

If it is easy to add (and it sounds like it is), it would be great to add it.

> Also, for carbon footprint trend I used the average of the range of car and train, is there a better > way than comparing the average?

we have a carbon footprint range, but it is not separated by car/train. So I am not sure I understand what this is... I'll look at the code to check

sunil07t:
The calories are high because I set the users weight as kg instead of lb. It not just because of the trips.

shankari:
Aha!!

www/js/metrics.js
```diff
@@ -785,4 +1004,45 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
+      $scope.expandedc = true
+    }
+  }
+  $scope.checkCalorieCardExpanded = function() {
+        return ($scope.expandedc)? "icon ion-chevron-up" : "icon ion-chevron-down";
+  }
+  var expandFootprintDOM = function() {
+    var div = document.getElementById('dashboard-footprint');
+    if (div.style.height == '470px'){ 
+      div.style.height = '140px'
```
shankari: nit: I think that you are supposed to use a directive to implement this

www/js/metrics.js
```diff
@@ -301,69 +420,127 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
-            $scope.carbonData.aggrCarbon.push({key: aggrCarbonData[i].key, values: FootprintHelper.getFootprint(aggrCarbonData[i].values, aggrCarbonData[i].key)});
             if (aggrCarbonData[i].key === "IN_VEHICLE") {
               $scope.carbonData.aggrVehicleRange = FootprintHelper.getFootprintRaw(aggrCarbonData[i].values, aggrCarbonData[i].key);
+              $scope.carbonData.aggrCarbon = FootprintHelper.getFootprint(aggrCarbonData[i].values, aggrCarbonData[i].key);
             }
           }
         }
+        $scope.summaryData.defaultSummary = $scope.summaryData.userSummary;
+        var change = "";
+        var lastWeekCarAndTrain = lastWeekCarbon.replace(/ /g,'').split("~");
```
shankari: Is this really the best way to do this? string manipulation? since we're calculating the values locally, can't we just keep them as integers? In general, you want to avoid string manipulation (with regular expressions!) if you can because:
a) it is expensive to convert back and forth
b) it introduces unnecessary fragility where it is not necessary

www/js/metrics.js
```diff
@@ -301,69 +420,127 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
+        var twoWeekAgoCarAndTrain = twoWeeksAgoCarbon.replace(/ /g,'').split("~");
+        var calculation = (((parseInt(lastWeekCarAndTrain[0]) + parseInt(lastWeekCarAndTrain[1])) / 2)
+                          / ((parseInt(twoWeekAgoCarAndTrain[0]) + parseInt(twoWeekAgoCarAndTrain[1])) / 2))
```
shankari: ditt wrt parsing from string

PR #117, opened 2016-08-29
User visible enhancements to the dashboard Fixes #116

### LINKED ISSUE #116: User visible enhancements to the dashboard
- Better error handling for the metrics. Right now, if there is any error from the server, then the `$ionicLoading` screen stays stuck forever. I ran into this when I tried switching from the local server to the main e-mission. In order to reproduce this locally, change `emission/net/api/cfc_webapp.py` to be like so:

```
@post('/result/metrics/<time_type>')
def summarize_metrics(time_type):
    raise HTTPError(403, "token is valid, but no account found for user")
    if 'user' in request.json:
        user_uuid = getUUID(request)
    else:
        user_uuid = None
...
```

and restart the webserver and the app. You will be stuck on the loading screen forever. We should handle errors from the server by implementing a `catch` in addition to `then` so that an incorrect query doesn't hose the entire app.

- support an option to refresh the summary screen (pull to refresh). Very annoying if there's something wrong (like the above) and you just want to pull the most recent data but there is NO way to do so.

- Hide the comparison to the past week if this is a summary of the users current query. Otherwise, since we don't have the past week's data, we end up with NaN% and Infinity % (see below).

- Also, why is the average for the calories NaN but the average for the carbon footprint can be calculated.

| Hide comparison for custom query | Difference in average | 
| --------------------- | ------------------- |
| simulator screen shot aug 23 2016 10 26 37 pm | simulator screen shot aug 23 2016 10 27 19 pm |

sunil07t:
Might have to change the UI for the chart button option for better looks.

### DIFF
--- a/www/js/metrics.js
+++ b/www/js/metrics.js
@@ -6,7 +6,7 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
                                     CommHelper, $window, $ionicPopup,
                                     FootprintHelper, CalorieCal, $ionicModal, $timeout, storage,
                                     $ionicScrollDelegate) {
-    
+
     var lastWeekQuery = true;
     var first = true;
     var lastWeekCalories = 0;
@@ -109,6 +109,12 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
     }
     $scope.weightToggleRightClass = function() {
       return $scope.userData.weightUnit == 0? "unit-toggle-right-active hvcenter" : "unit-toggle-right hvcenter";
+    }
+    $scope.currentQueryForCalorie = function() {
+      return $scope.uictrl.current == "Last week"? "user-calorie-percentage" : "user-calorie-no-percentage";
+    }
+    $scope.currentQueryForCarbon = function() {
+      return $scope.uictrl.current == "Last week"? "user-carbon-percentage" : "user-carbon-no-percentage";
     }
     $scope.showRange = function() {
       if ($scope.uictrl.showFilter) {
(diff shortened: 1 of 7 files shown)


www/js/metrics.js
```diff
@@ -1026,23 +1003,13 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
-      div.style.height = '140px'
-      $ionicScrollDelegate.scrollTop()
-    } else { 
-      div.style.height = '470px'
-    }
+
+  $scope.changeFootprintCardHeight = function() {
+        return ($scope.expandedf)? "expanded-footprint-card" : "small-footprint-card";
   }
-  var expandCalorieDOM = function() {
```
shankari: Is this deleted? Why?

shankari: is it because it is using the dom directly?
I think that was one of my comments - so if that is why and you are using CSS instead, that's great!

www/js/metrics.js
```diff
@@ -906,11 +888,8 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
   $scope.selectCtrl = {}
   initSelect(); 
   $timeout(function() {
-    getDataOfTwoWeeksAgo();
-  }, 1)
-  $timeout(function() {
     getData();
-  }, 100)
+  }, 1)
```
shankari: you were getting the data after 100secs!??

www/js/metrics.js
```diff
@@ -260,26 +306,23 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
       } else if (mode === 'timestamp') { // timestamp range
-        if(twoWeeksAgo){
-          var tempFrom = moment2Timestamp(moment().day(-14)); 
-          var tempTo = moment2Timestamp(moment().day(-8)); 
-          twoWeeksAgo = false; 
-        } else if(lastWeekQuery) {
-          var tempFrom = moment2Timestamp(moment().day(-7)); // Last week's Sunday
-          var tempTo = moment2Timestamp(moment().day(-1)); // Last week's Satruday
-          lastWeekQuery = false; // Only get last week's data once
+        if(lastWeekQuery) {
```
shankari: nit: name change to lastTwoWeeksQuery?

www/js/metrics.js
```diff
@@ -805,83 +850,20 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
-
-      };
-      $scope.showca2020 = false;
-      $scope.showca2035 = false;
-
-    }
-    $scope.shouldshowca2020 = function() {
-      return $scope.showca2020;
-    }
-    $scope.shouldshowca2035 = function() {
```
shankari: Good comment cleanup!

www/js/metrics.js
sunil07t: Its after 1 millisecond just so the the dates of last 2 week are set to $scope.

www/js/metrics.js
sunil07t: Yes implemented CSS instead of this.

PR #120, opened 2016-09-07
Adds arrows on menus and on diary card to indicate clickability

Fernando-Abreu:
Below are some screenshots showing the screen before and after adding the arrows. Their object is to indicate some areas os the screen are clickable. 

Before
screen shot 2016-09-07 at 12 20 12 am

After
screen shot 2016-09-06 at 3 19 05 am

Before
screen shot 2016-09-07 at 12 19 48 am

After
screen shot 2016-09-06 at 3 15 19 am


www/css/style.css
```diff
@@ -806,9 +806,26 @@ button.button.back-button.buttons.button-clear.header-item {
         top: 310px;
         line-height: 16px;
 }
+
+.start-end-addresses-container {
+  display: inline-block;
+  width: calc(100% - 25px);
```
shankari: nit: why 25px instead of 32 px?

shankari:
Looks good overall.

You also need to add arrows to 
- `www/templates/common/place-list.html`
- `www/templates/common/trip-list.html`

I think that's all that's left - here's the complete list of details.

```
bash-3.2$ find www/templates/ -name \*detail\*
www/templates//chat-detail.html            (part of example, can ignore)
www/templates//common/place-detail.html    (need to handle by modifying place-list.html)
www/templates//common/trip-detail.html     (need to handle by modifying trip-list.html)
www/templates//diary/detail.html           (handled by modifying diary/list.html)
www/templates//main-detail.html            (obsolete, replaced by diary/detail.html)
```

shankari:
@krisma @sunil07t if you have time, maybe you can review this (fairly simple) change as well

www/css/style.css
Fernando-Abreu: To leave more space for the start and end addresses. 
I guess it could get a bit tight the space for addresses in small screen smartphones. Do you agree? Or you think we should leave more space between the addresses and the arrow?

sunil07t:
Every thing is fine. I just found out that the titles in place-detail, trip-detail, detail and main-detail do not display. @Fernando-Abreu can you please fix this issue in this pull request. The titles do not display with view-title or title attribute, this should help you
https://www.raymondcamden.com/2015/12/18/is-your-ionic-view-title-not-updating/

Fernando-Abreu:
@sunil07t on my emulator the only titles that weren't displaying were place-list and trip-list

www/templates/common/trip-list.html
```diff
@@ -1,6 +1,6 @@
-<ion-view nav-title="Trips list">
-  <ion-nav-title class="dark-color">
-    <!--
+<ion-view title="Trips list">
+<!--   <ion-nav-title class="dark-color">
```
shankari: just so I understand what is happening here - we want the title to be "Trips list". And it is right there in as `title="Trips list"`. But the `<ion-nav-title>` overrides the`title` param and then is all blank anyway. So you re-introduce it here.

Looks good!

sunil07t:
@Fernando-Abreu I am not quite sure how to generate common pattern in the emulator to display these screens. You can display trip-detail and place-detail screen after you click the item in the list in place-list and trip-list screen. I saw it on the phone and I think trip-detail and place-detail can be fixed just like place-list and trip-list.

PR #126, opened 2016-09-17
Url

krisma:
CustomURL plugin added.
Logic built.

### DIFF
--- a/config.xml
+++ b/config.xml
@@ -79,5 +79,8 @@
     <splash src="resources/android/splash/drawable-port-xxhdpi-screen.png" density="port-xxhdpi"/>
     <splash src="resources/android/splash/drawable-port-xxxhdpi-screen.png" density="port-xxxhdpi"/>
   </platform>
+  <plugin name="cordova-plugin-customurlscheme" spec="~4.2.0">
+      <variable name="URL_SCHEME" value="emission" />
+  </plugin>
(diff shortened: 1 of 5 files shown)


config.xml
```diff
@@ -79,5 +79,8 @@
     <splash src="resources/android/splash/drawable-port-xxhdpi-screen.png" density="port-xxhdpi"/>
     <splash src="resources/android/splash/drawable-port-xxxhdpi-screen.png" density="port-xxxhdpi"/>
   </platform>
+  <plugin name="cordova-plugin-customurlscheme" spec="~4.2.0">
```
shankari: This can be removed.

www/js/app.js
```diff
@@ -69,3 +69,43 @@ angular.module('emission', ['ionic','ionic.service.core', 'emission.controllers'
...
+  var paramsList = params.split('&');
+  var rtn = {route: route};
+  for (var i = 0; i < paramsList.length; i++) {
+    rtn[paramsList[i].split('=')[0]] = paramsList[i].split('=')[1];
+  }
+  return rtn;
+}
+function handleOpenURL(url) {
+  setTimeout(function() {
+    var kvList = parseURL(url);
```
shankari: Please add a new alert here that shows the parsed components (addr, route, paramsList)

PR #128, opened 2016-09-17
Setup tutorial framework 

Fernando-Abreu:
On this PR I setup a tutorial framework called nzTour. What it does is that you simple pass a list of objects containing elements and phrases, and then it attach to them like the demo below.
I also created an example of it on the diary tab. 
It is only displayed the first time the user visits the tab. But he/she can choose to see it again by click on the help icon at the navbar.

j0tynscr9i

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -23,6 +23,8 @@
     "string.startsWith": "*",
     "angularLocalStorage": "^0.3.2",
     "ng-walkthrough": "^1.0.0",
-    "animate.css": "^3.5.2"
+    "animate.css": "^3.5.2",
+    "nz-tour": "^1.2.1",
+    "jquery": "^3.1.0"
   }
 }
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -2,14 +2,14 @@ angular.module('emission.main.diary.list',['ui-leaflet',
                                       'ionic-datepicker',
                                       'emission.main.common.services',
                                       'emission.services',
-                                      'ng-walkthrough'])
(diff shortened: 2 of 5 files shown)


bower.json
```diff
@@ -23,6 +23,8 @@
     "string.startsWith": "*",
     "angularLocalStorage": "^0.3.2",
     "ng-walkthrough": "^1.0.0",
-    "animate.css": "^3.5.2"
+    "animate.css": "^3.5.2",
+    "nz-tour": "^1.2.1",
+    "jquery": "^3.1.0"
```
shankari: Why do we need jquery? I've been trying to avoid using jquery because mixing and matching two different frameworks can lead to unexpected effects. It is not clear to me what functionality you are using from jquery. Is it a dependency of nz-tour?

shankari: nvm, I see that it is a dependency.

www/js/diary/list.js
```diff
@@ -280,6 +278,38 @@ angular.module('emission.main.diary.list',['ui-leaflet',
 
     };
 
+    // Tour steps
+    var tour = window.tour = {
```
shankari: why do you need to set `window.tour` to the tour? In general, angular discourages you from using window and encourages the use of `$scope` instead.

But I don't even see you using `window.tour` anywhere...

bower.json
Fernando-Abreu: That was the reason indeed

www/js/diary/list.js
Fernando-Abreu: I don't need it, just removed this.

PR #133, opened 2016-09-24
Add a new popup to view the credentials

shankari:
And add a button for it in the game screen. The credentials are only viewable,
not editable.


shankari:
| New button | Launches creds | 
| ----- | ------ |
| simulator screen shot sep 24 2016 12 27 21 am | simulator screen shot sep 24 2016 12 27 03 am  |

PR #137, opened 2016-09-26
Onboarding

juemura:
When I pulled from upstream to get the survey, it wasn't working (the survey wasn't loading). I started a fresh project and it worked, so I copied over my changes little by little and have been debugging, but there's one thing I still haven't been able to fix. If the user says yes to Bic2Cal, after the survey closes the game signup screen isn't loaded. I want to fix this before copying over the code for the tutorial, but I need to get some sleep now. Will try again tomorrow. Lmk if you know the fix.
Thanks!


juemura:
hm, looking at all these messy commits makes me want to fork again and start fresh...

README.md
```diff
@@ -37,10 +55,19 @@ Install all javascript components using bower
 $ bower update
 ``
 
+``
+$ ionic io init
+``
+
+``
+$ ionic state reset
```
shankari: I don't think you need reset while installing from a fresh repo (I know you had to do reset but I think that was the obsolete upstream). 

I think that the general rule of thumb is:
- fresh install (`ionic state restore`)
- pull into existing repo when plugins have changed (`ionic state reset`)

.io-config.json
```diff
@@ -0,0 +1 @@
+{"app_id":"b6a57699","api_key":"acbf10269e89184d103dcb94e36b989a24dfd84d5c4ffbac","dev_push":true}
```
shankari: Also, this is in the `.gitignore`. Surprised that it even showed up in the diff.

www/js/goals.js
```diff
@@ -591,7 +591,8 @@ angular.module('emission.main.goals',['emission.services', 'ngSanitize', 'ngAnim
       $rootScope.$on('$cordovaInAppBrowser:exit', function(e, event) {
         console.log("exiting, event = "+JSON.stringify(event));
       });
-  	}
+      $scope.refreshPage();
```
shankari: Line 584 above, where I check to see if we are navigating to bic2cal is the one that triggers the close of the inAppBrowser, which in turn generates the exit event. If you want to refresh the page on exit, you need to put it inside the `$on`

PR #140, opened 2016-09-29
Preb2crelease

juemura:
I commented out the div, so the user won't see the b2c stuff, but left the js.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,6 @@ $ git clone <your repo URL>
 $ cd e-mission-phone
 ``
 
-Restore platforms and plugins. This is a good time to get a cup of your favorite beverage
-
-``
-$ ionic state restore
-``
-
 Enable platform hooks, including http on iOS9
 
 ``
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -37,6 +35,28 @@ Install all javascript components using bower
...
 ``
 
+Create a remote to pull updates from upstream
+
+``
+$ git remote add upstream https://github.com/e-mission/e-mission-phone.git
+``
+
+``
+$ npm install
```
shankari: nit: I don't think that this is technically required. You had the issues with the glob, but that was before you pulled from upstream. I have never had to do this to get it to work, and I've pulled a fresh repo from upstream recently to get it to work.

gulpfile.js
```diff
@@ -50,4 +50,4 @@ gulp.task('git-check', function(done) {
   done();
 });
 
-require('glob').sync('/www/tasks/gulp-*').forEach(require);
\ No newline at end of file
+//require('glob').sync('/www/tasks/gulp-*').forEach(require);
```
shankari: Ditto.

PR #143, opened 2016-09-30
Bump up version number + fix two week query

shankari:
Two weeks is 14 days before today, not two calendar weeks.
Unlike the building dashboard, we don't want to be slower than we already are.


shankari:
Note that iOS has been bumped up twice because of https://github.com/e-mission/e-mission-data-collection/pull/133

Issue #148, opened 2016-10-02
Dashboard does not work if user has data for some days but not two weeks


shankari:
The current check for user and aggregate information in the dashboard assumes either the user has no data, or has the full two weeks of data. But this can be incorrect when the user has been using the app for less than 2 weeks, i.e in the days right after sign up.

e.g. note that the user has 7 days worth of data while the aggregate has 14 days.

<img width="863" alt="screen shot 2016-10-02 at 1 42 09 pm" src="https://cloud.githubusercontent.com/assets/2423263/19023565/183a1b76-88a6-11e6-83ad-f556b94e963e.png">

shankari:
Examples of results showing this behavior - can be unzipped and used to test the fix.
[results_dump.json.zip](https://github.com/e-mission/e-mission-phone/files/505242/results_dump.json.zip)

shankari:
@sunil07t this needs to be fixed right away.

shankari:
Here's an example of how the slice fails in this case. twoWeeksAgoDuration has length 8, userDuration has length 0. So all metrics are zero.

```
> twoWeeksAgoDuration
< [Object, Object, Object, Object, Object, Object] (6) = $8
> userDuration
< [] (0) = $9
```

shankari:
| No data  | No Data | 
| ----- | ----- |
| ![simulator screen shot oct 2 2016 3 17 01 pm](https://cloud.githubusercontent.com/assets/2423263/19024207/609da1b4-88b3-11e6-9e1e-7510b8ef4bb6.png) | ![simulator screen shot oct 2 2016 3 17 09 pm](https://cloud.githubusercontent.com/assets/2423263/19024208/60a1b7b8-88b3-11e6-8deb-621cd3b6a916.png) |

shankari:
Fixed by #149 and #151

PR #149, opened 2016-10-03
Dashboard loads users data depending on date, fixes issue #148

sunil07t:
- Use 7th day ago from today to determind the split of 2 weeks

### DIFF
--- a/www/js/metrics.js
+++ b/www/js/metrics.js
@@ -327,8 +327,8 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
         };
       } else if (mode === 'timestamp') { // timestamp range
         if(lastTwoWeeksQuery) {
-          var tempFrom = moment2Timestamp(moment().utc().subtract(14, 'days'));
-          var tempTo = moment2Timestamp(moment().utc().subtract(1, 'days'))
+          var tempFrom = moment2Timestamp(moment().utc().startOf('day').subtract(14, 'days'));
+          var tempTo = moment2Timestamp(moment().utc().startOf('day').subtract(1, 'days'))
           lastTwoWeeksQuery = false; // Only get last week's data once          
         } else {
           var tempFrom = moment2Timestamp($scope.selectCtrl.fromDateTimestamp);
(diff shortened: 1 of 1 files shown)


www/js/metrics.js
```diff
@@ -443,6 +456,11 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
           var aggCount = results[2].aggregate_metrics;
           var aggDistance = results[3].aggregate_metrics;
         }
+        console.log(moment().utc().startOf('day'));
```
shankari: nit: remove debug comments

www/js/metrics.js
```diff
@@ -421,14 +421,27 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
+              userDistance.push(results[3].user_metrics[i]);  
+            } else {
+              twoWeeksAgoDuration.push(results[0].user_metrics[i]);
+              twoWeeksAgoMedianSpeed.push(results[1].user_metrics[i]);
+              twoWeeksAgoDistance.push(results[3].user_metrics[i]);
+            }
+          }
           var aggDuration = results[0].aggregate_metrics.slice(7);
           var aggMedianSpeed = results[1].aggregate_metrics.slice(7);
           var aggCount = results[2].aggregate_metrics.slice(7);
```
shankari: why not do something similar for the aggregate metrics as well? I agree this is less important because the aggregate metrics are unlikely to be missing, but it doesn't cost anything to handle that corner case as well.

www/js/metrics.js
```diff
@@ -443,6 +456,11 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
           var aggCount = results[2].aggregate_metrics;
```
shankari: I have never really understood how the user values work here. if `results[0].user_metrics.length == 0`, then results[0].user_metrics is an empty array. So what does it mean to set the value `var userDuration = results[0].user_metrics;` here? Why do we need to treat first and not first differently?

www/js/metrics.js
shankari: Just to clarify - please do not attempt to fix this now. This is documenting it for later cleanup.

www/js/metrics.js
shankari: concretely, now that we check timestamps, if the user has no recent data, we will get an empty array anyway

sunil07t: This checks if the data is empty and if its empty non of the calculation will be triggered. The calculation will only run once. In the else case `var userDuration = results[0].user_metrics;` is empty but it does't matter because if its empty the UI will say No Summary Data. In line 615 `first` is set to `false` after 1st run. So now whenever the user selects costume data `var userDuration = results[0].user_metrics;` in the else case is used. Both initial call for last week data and user's costume call have same vars but twoWeeksAgo** is only set in first case.

shankari:
How did you test this?
Did you get with my data? Now I get this screenshot.

| Nan + infinity | Plenty of data | 
| ------- | ------ |
| simulator screen shot oct 3 2016 3 28 22 am | simulator screen shot oct 3 2016 3 31 01 am |

shankari:
Doesn't happen if I reload the page. Only on the first load.

shankari:
that's because we set first=false after first load

Issue #146, opened 2016-09-30
Join group does not work when app is already launched


shankari:
Original speculation was that function `handleOpenURL(url)` wasn't invoked if the app was already running. However, there was no evidence of this in the documentation for the plugin
https://github.com/EddyVerbruggen/Custom-URL-scheme
and bug reports around issues with cold start clearly indicated that it worked perfectly otherwise.

So I connected the debugger to a running instance and found that the function
was in fact called. The reason that we don't go to the game screen is that the
function merely parses the URL and sets some values in local storage.

```
function handleOpenURL(url) {
  setTimeout(function() {
    var kvList = parseURL(url);
    // There are 3 types of users in total
    
    if (kvList.route == 'join') {
      var INTRO_DONE_KEY = 'intro_done';
      var HABITICA_REGISTERED_KEY = 'habitica_registered';
      var REFERRAL_NAVIGATION_KEY = 'referral_navigation';
      var REFERRED_KEY = 'referred';
      var REFERRED_GROUP_ID = 'referred_group_id';
      var REFERRED_USER_ID = 'referred_user_id';
      window.localStorage.setItem(REFERRED_KEY, true);
      window.localStorage.setItem(REFERRED_GROUP_ID, kvList['groupid']);
      window.localStorage.setItem(REFERRED_USER_ID, kvList['userid']);
      window.localStorage.setItem(REFERRAL_NAVIGATION_KEY, 'goals');
    }
}, 0);
```

If this is done on warm start, then we invoke the `start_prefs` code where we check for the presence of a referrer URL and move to `goals` in that case. But if we are already in some state, we won't launch start_prefs again until the app is killed.

```
    startprefs.getNextState = function() {
      return startprefs.getPendingOnboardingState().then(function(result){
        var REFERRAL_NAVIGATION_KEY = 'referral_navigation';
        if (result == null) {
          var temp = getReferralNavigation();
          storage.remove(REFERRAL_NAVIGATION_KEY);
          if (temp == 'goals') {
            return 'root.main.goals';
          } else {
            return 'root.main.metrics';
          }
        } else {
          return result;
        }
      });
```

So this works, but provides an unnecessarily shitty user experience.

shankari:
@krisma @juemura @sunil07t this just shows that phone FSM needs serious work, and needs to be properly documented as well so we know what we are doing and what we are supposed to do.

shankari:
[...]

PR #150, opened 2016-10-03
Finally fix invites for once and for all

shankari:
Fix all the issues encountered in the tests.
Referral now works for:
- cold start
- warm start when the user has not signed up for habitica
- warm start when the user has signed into habitica but is on a different page
- warm start when the user has signed in to habitica and is on the same page

All of them appear to work. I am now going to declare victory :)


shankari:
This fixes #146

PR #154, opened 2016-10-04
Tutorial

juemura:
Added/edited tutorial in Goals, Heatmap, and Diary.


www/js/goals.js
```diff
@@ -2,9 +2,10 @@
 
 angular.module('emission.main.goals',['emission.services', 'emission.plugin.logger',
                 'ngSanitize', 'ngAnimate',
-                'emission.splash.referral', 'angularLocalStorage'])
+                'emission.splash.referral', 'angularLocalStorage',
+                'ng-walkthrough', 'nzTour'])
 
-.controller('GoalsCtrl', function(CommHelper, $state, $ionicLoading, $scope, $rootScope, $ionicModal,
+.controller('GoalsCtrl', function(CommHelper, $state, $ionicLoading, $scope, $rootScope, $ionicModal, nzTour, Config,
```
shankari: nit: why do we need to add the `Config` service here? Don't seem to use it in the rest of the code....

www/js/heatmap.js
```diff
@@ -183,4 +184,45 @@ angular.module('emission.main.heatmap',['ui-leaflet', 'emission.services'])
...
+
+
+  // Tour steps
+  var tour = {
+    config: {
+
+    },
+    steps: [{
+      target: '.datepicker',
+      content: 'This heatmap shows the aggregate data for all E-mission users. Select the dates you want to see, and filter by hours of the day (24h format) and days of the week. For example, if you enter 16 and 19 in the last field, and select Monday and Friday, you\'ll see the Heatmap filtered to show the traffic on weekdays between 4pm and 7pm.'
```
shankari: you can also filter by day of month - so if you pick year = 2016, month = blank, day = 0 - 10, hour = 16-19, dayofweek = Mon - Fri, you should see the points for all weekdays between 4pm and 7pm on the first 10 days of the month for the year 2016.

www/js/heatmap.js
```diff
@@ -183,4 +184,45 @@ angular.module('emission.main.heatmap',['ui-leaflet', 'emission.services'])
...
+    },
+    {
+      target: '.heatmap-mode-button',
```
juemura: it has to be a class, I tried with id and it didn't work. That's why I added classes to some elements.

www/js/heatmap.js
```diff
@@ -183,4 +184,45 @@ angular.module('emission.main.heatmap',['ui-leaflet', 'emission.services'])
...
+
+  var startWalkthrough = function () {
+    nzTour.start(tour);
+  };
+
+  /*
+  * Checks if it is the first time the user has loaded the heatmap tab. If it is then
+  * show a walkthrough and store the info that the user has seen the tutorial.
+  */
+  var checkHeatmapTutorialDone = function () {
```
juemura: I agree. Just figured it was better to get something in place that worked rather than spending a bunch more time with this. It'll get easier as I get more used to the phone code.

www/js/goals.js
juemura: it didn't, I copied that over by mistake.

www/js/heatmap.js
juemura: I'm assuming we don't want people to leave year/month in blank because it would be a lot of data and it would be super slow. Also, since year/month/day work as a range I figured it was better to abstract that detail to the general user because it's really confusing.

PR #157, opened 2016-10-04
Fix ionic deploy update progress to really work

shankari:
The prior solution required the introduction of a progress element in every
screen - otherwise the transparent overlay meant that the progress bar was not
actually visible on (say) the trip list screen.

This solution moves the progress bar into a `$ionicPopup`. This ensures that:
- all the code can be isolated to the update check class
- the user visually sees that the update is happening, and does not kill the app. and if she does, she understands why it no longer works :)
- there is a clearer indication when the progress is done and the screen is re-loaded

### DIFF
--- a/www/js/splash/updatecheck.js
+++ b/www/js/splash/updatecheck.js
@@ -32,6 +32,12 @@ angular.module('emission.splash.updatecheck', ['angularLocalStorage'])
   var deploy = new Ionic.Deploy();
 
   var applyUpdate = function() {
+    $ionicPopup.show({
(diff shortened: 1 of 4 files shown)


shankari:
| screenshots    | and more     |
| -------------- | ------------ |
| simulator screen shot oct 4 2016 1 53 31 pm | simulator screen shot oct 4 2016 1 53 39 pm |
| simulator screen shot oct 4 2016 1 53 46 pm | simulator screen shot oct 4 2016 1 54 13 pm |

PR #161, opened 2016-10-05
Bosses arts are visible on the phone UI

sunil07t:


shankari:
Happy to pull the bosses art, but I'm worried about the refresh change. Do we really need to do it? Was it broken before? How did you test it to make sure that it works?

Even changes that look simple can break things, specially when we do not have any automated tests, so I am reluctant to merge changes that don't have a strong reason to make it in.

sunil07t:
I added questContent() in the getUserInfo() because the if($scope.inQuest) in refreshInfo() would run before $scope.inQuest is set to true, but now that questContent() is in the getUserInfo() it only runs after $scope.inQuest is set to true. Also, the refresh boolean makes the questContent() run only on refresh so it does not get called when other functions like score up or down, that calls getUserInfo(), get called. getContent() gets all the content/items available in habitica.

shankari:
> How did you test it to make sure that it works?

sunil07t:
I stared a quest with a test users and went to the emulator to see if the initial load of the screen would show all the quest content in the party tab and it did. Before the quest contents would not show up and users had to refresh again to make it appear on the UI.

shankari:
ok merging now and doing another UI deploy

Issue #119, opened 2016-09-05
Support the "air" mode on the phone


shankari:
As part of https://github.com/e-mission/e-mission-server/issues/380, we have added support for AIR modes. The problem is that the phone currently does not support AIR - there's no icon for it, so we need to add a new icon everywhere.

Note that I had argued strongly against using icons for just this reason - every time we add a new mode, and once we have user-specific modes, every time a user adds a new user specific mode, we need to add a new icon.

Switching to a non-icon based representation is a bigger design challenge, so for now, we will probably just add another icon for AIR.

shankari:
And while we are doing this, we might as well support UNKNOWN as well.

shankari:
UNKNOWN shows up on older iPhones (without M7 chip) and on some short trips where we didn't get enough activity points.

shankari:
@sunil07t has supported AIR_OR_HSR. Still need to add support for UNKNOWN.

PR #165, opened 2016-10-06
supports UNKNOWN trips on diary Fixes issue #119

sunil07t:


shankari:
This is the size of change that I think is appropriate for a last minute update to a release.

shankari:
This is the size of change that I think is appropriate for a last minute update to a release.

PR #166, opened 2016-10-06
UI changes in goals screen

sunil07t:
- Hide the nav button in Goals screen when the user is not signed up
- Thin border to separate the habits in the list
	- Had to add “no-border” id to some of the items  in diary screen and goal screen for items those do not need border
- Changed Goals screen Sign up button and form
- Avatar background is white if there is not background

### DIFF
--- a/www/build/app.css
+++ b/www/build/app.css
@@ -6569,18 +6569,18 @@ textarea.option-content {
   padding-top: 0em;
 }
 .herobox.noBackgroundImage {
-  background: #c5d3d6;
+  background: #fff;
 }
 .herobox.noBackgroundImage:hover,
 .herobox.noBackgroundImage:focus {
-  background: #cad7da;
+  background: #fff;
 }
 .herobox.noBackgroundImage.isUser {
-  background: #dae3e5;
+  background: #fff;
 }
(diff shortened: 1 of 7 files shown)


shankari:
@sunil07t is the thought that these changes should go into this release? Or should we wait until the next release to merge them?

shankari:
I would be tempted to wait until the next release.
Don't want to break anything at this delicate point.

sunil07t:
Old and New Sign up screen
currently sign up
sign up changed

sunil07t:
Old and New game sign up screen
currently sign up
sign up changed

sunil07t:
Old and new habit list and avatar background
old list
list changed

sunil07t:
The dark background is in the current version after viewing the detail screen of the trip the back ground gets dark, this is solved in this change. Also the later two pictures do no have borders on the trips, only when the habit list items have border.
currently
no border after change 2
no border after change

sunil07t:
These are just minor css and html changes. You can include these in either release but they are quit important like the background change and nav button showing on the sing up screen.

sunil07t:
I will close this PR since these changes are included in the leaderboard PR.

PR #187, opened 2016-11-29
iab experiment

krisma:


www/js/goals.js
```diff
@@ -129,14 +141,118 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
+              location: 'yes',
+              clearcache: 'yes',
+              toolbar: 'yes'
+            };
 
+
+
+          $rootScope.$on('$cordovaInAppBrowser:loadstart', function(e, event){
+            // insert Javascript via code / file
+            $cordovaInAppBrowser.executeScript({
```
shankari: Also, where is the callback for the `executeScript`? The raw IAB has a callback function. You are using the angular wrapper, so presumably it returns a promise. Where's the `then` for the promise?

www/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://berkeley.qualtrics.com https://jfe-cdn.qualtrics.com emission:;
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://berkeley.qualtrics.com https://jfe-cdn.qualtrics.com emission: https://em-game.eecs.berkeley.edu;
```
shankari: Do your really need this?
https://github.com/apache/cordova-plugin-inappbrowser#urls-that-are-not-white-listed

> The InAppBrowser is not subject to the whitelist, nor is opening links in the system browser.

krisma:
the reason that it didn't work before is that the javascript is executed when the page is loading. after changing the loadingstart to loadingstop, it worked.

www/index.html
krisma: um right. so since we are using _blank (iab not cordova webview), we don't need this.

www/js/goals.js
```diff
@@ -129,14 +141,118 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
+              code: "localStorage.setItem('habit-mobile-settings', '" + JSON.stringify(settings) + "');" 
+              + "alert('a');window.location.href = 'https://em-game.eecs.berkeley.edu/#/tasks';"
+            });
+
+          });
+        $cordovaInAppBrowser.open('https://em-game.eecs.berkeley.edu/#/tasks', '_blank', options)
+          .then(function(event) {
+            // success
```
krisma: the tutorial suggested us to use handler like rootScope.on

krisma: and after I changed that to loading stop, I think the result should be the same

www/css/style.css
```diff
@@ -1691,4 +1691,10 @@ html, body, #chart2, svg {
 }
 .full-toggle-right:active {
   opacity: 0.7;
-}
\ No newline at end of file
+}
+#iframe {
```
shankari: You don't need the iframe any more since you are using the iab, right?

www/js/goals.js
```diff
@@ -6,7 +6,7 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
                 'ng-walkthrough', 'nzTour'])
 
 .controller('GoalsCtrl', function(CommHelper, $state, $ionicLoading, $scope, $rootScope, $ionicModal, nzTour,
-                                $window, $http, $ionicGesture, $ionicPopup, $timeout, storage, ReferralHandler, ReferHelper, Logger){
+                                $window, $http, $ionicGesture, $ionicPopup, $timeout, storage, ReferralHandler, ReferHelper, Logger, $cordovaInAppBrowser){
```
shankari: nit: can you put the standard services (e.g. $cordovaInAppBrowser) before our services (e.g. ReferralHandler)

www/js/goals.js
```diff
@@ -129,7 +130,35 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
         showDelete: false
     };*/
+    $scope.toBrowser = function() {
+            var options = {
+              location: 'yes',
+              clearcache: 'yes',
+              toolbar: 'yes'
+            };
+
+            var settings = localStorage.getItem("habit-mobile-settings")
```
shankari: where is `habit-mobile-settings` set?
I don't see you setting it anywhere in this PR...

shankari:
tasks page loads over and over with this error

screen shot 2016-12-13 at 11 20 51 pm

krisma:
I had some quick debug last night and figured out:
1. Why did it work on my side?
When I hardcode the token into local storage with iframe, it stayed there. I remove the hard code but did not clean up local storage.
2. How to fix?
Turned out that I still can't find the token on client side. I thought it worked for unknown reason, and it's just because it's hardcoded. I tried again to find token by inspecting any response from CommHelper.habiticaProxy but still can't find the token. Under "profile" key there is only username. I just need the token (although I worked on this a couple of times).

shankari:
@krisma 
Thanks for the quick response! The token is definitely available on the server side (we use it to make the proxy calls). So maybe you can add a call to the server to get the token (something like `getToken`)?

krisma:
I'll try to find that proxy call tmr after my final, but after that I will be heading China so I'm not sure if I can finish.

shankari:
You can't find it - you would have to write it. Can you work on it from China? You will be working on the paper writing anyway, right?

krisma:
iab experiment done.
result.encoding, result._content = 'utf8', json.dumps(temp) does the magic

PR #189, opened 2016-12-01
Changes to release version 1.6.0

shankari:
This includes:
- switching all URLs back to production
- restoring autoload on the heatmap page
- real phones don't generate click events with the correct targets - deal with that by:
   - refactoring the code to work with a set of points
   - registering a listener on the map directly

And last but not least, bumping up the version number!


shankari:
dom structure showing the svg and what the target is on real phones.
Note that we cannot just pass the event down to all children because there are multiple children and we don't know which is the valid section.

screen shot 2016-11-30 at 12 43 03 pm

PR #190, opened 2016-12-01
Restore a cancel button to the actionsheet

shankari:
Without this button, the android actionsheet is missing a cancel button.
Supposedly this is consistent with the material design guidelines, but I think
that it is confusing, and I don't want to end up with a bunch of spurious
incidents.

So, let's restore the cancel button back manually.


shankari:
| Before | After |
| ------- | ----- |
| ugly_android_actionsheet | better_android_actionsheet |

shankari:
Further documentation on the ugliness.
https://forum.ionicframework.com/t/actionsheets-android-ugly-styling-need-help/18462

PR #193, opened 2017-01-04
B2csurvey

juemura:
The survey works but I'm still having trouble with two things: Copying the id to the clipboard and automatically popping up the survey if the user hasn't seen it yet. I'll point to these things in the code, let me know if you have suggestions.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -29,8 +29,6 @@
     -->
 
     <!-- ionic/angularjs js -->
-    <script src="lib/ngCordova/dist/ng-cordova.min.js"></script>
-    <script src="cordova.js"></script>
     <script src="lib/jquery/dist/jquery.min.js"></script>
     <script src="lib/ionic/js/ionic.bundle.js"></script>
     <script src="lib/ionic-platform-web-client/dist/ionic.io.bundle.min.js"></script>
--- a/www/index.html
+++ b/www/index.html
@@ -98,6 +98,10 @@
     <script src="lib/nvd3/build/nv.d3.min.js"></script>
     <script src="lib/angular-nvd3/dist/angular-nvd3.min.js"></script>
     <script src="lib/nz-tour/dist/nz-tour.min.js"></script>
+    <script src="lib/angular/angular.min.js"></script>
+    <script src="lib/clipboard/dist/clipboard.min.js"></script>
+    <script src="lib/ngclipboard/dist/ngclipboard.min.js"></script>
+
     <link rel="stylesheet" href="lib/nvd3/build/nv.d3.css">
   </head>
 
--- a/www/index.html
+++ b/www/index.html
@@ -98,7 +98,6 @@
     <script src="lib/nvd3/build/nv.d3.min.js"></script>
     <script src="lib/angular-nvd3/dist/angular-nvd3.min.js"></script>
     <script src="lib/nz-tour/dist/nz-tour.min.js"></script>
-
     <link rel="stylesheet" href="lib/nvd3/build/nv.d3.css">
   </head>
 
--- a/www/index.html
+++ b/www/index.html
@@ -29,6 +29,8 @@
     -->
 
     <!-- ionic/angularjs js -->
+    <script src="lib/ngCordova/dist/ng-cordova.min.js"></script>
+    <script src="cordova.js"></script>
     <script src="lib/jquery/dist/jquery.min.js"></script>
     <script src="lib/ionic/js/ionic.bundle.js"></script>
     <script src="lib/ionic-platform-web-client/dist/ionic.io.bundle.min.js"></script>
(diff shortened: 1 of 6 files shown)


www/js/goals.js
```diff
@@ -864,6 +871,81 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
+        console.log("Showing user id");
+        $ionicPopup.show({
+          title: 'Bic2Cal Survey',
+          templateUrl: 'templates/goals/uid.html',
+          scope: $scope,
+            buttons: [{
+              text: 'Copy user id and open survey',
+              type: 'button-positive',
+              onTap: function(e) {
+                $cordovaClipboard.copy($scope.profile.id).then(function () {
```
juemura: here, `$cordovaClipboard.copy($scope.profile.id)` should be copying the id to the clipboard when the user clicks, but it doesn't. You can manually copy and paste it, but I can't figure out why this isn't working.

www/js/goals.js
```diff
@@ -891,20 +973,22 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
         startWalkthrough();
         storage.set(GOALS_DONE_KEY, true);
       }
-    };
+    };*/
 
-    $scope.startWalkthrough = function () {
-      startWalkthrough();
-    }
+    //checkSurveyDone();
```
juemura: I had to comment out the call to checkSurveyDone() because it makes the app crash when I go to the Goals tab. Let me know if you can see what I'm doing wrong...

www/index.html
```diff
@@ -29,6 +29,8 @@
     -->
 
     <!-- ionic/angularjs js -->
+    <script src="lib/ngCordova/dist/ng-cordova.min.js"></script>
```
shankari: I don't think you need to do this - we use multiple ngCordova features already.

www/index.html
```diff
@@ -29,6 +29,8 @@
...
+    <script src="cordova.js"></script>
```
shankari: We already use cordova.js. Why add it a second time?

www/js/goals.js
shankari: I have never used cordovaClipboard before, so I don't have any suggestions. You just need to debug. What happens if you only call `$cordovaClipboard.copy`? What if you try to copy something else?...

www/js/goals.js
shankari: What is the crash? Just saying that "it crashes" doesn't give any indication of what the error is...

www/js/goals.js
```diff
@@ -864,6 +871,81 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
+
+    $scope.startSurvey = function () {
+      showUserId();
+    }
+
+    var checkSurveyDone = function () {
+      console.log("Checking if user already completed survey");
+      var SURVEY_DONE_KEY = 'survey_done';
+      var surveyDone = storage.get(SURVEY_DONE_KEY);
+      if (!surveylDone) {
```
shankari: spelling mistake. Are you sure this works?

www/js/goals.js
juemura: I tried all that, and nothing gets copied to the clipboard. But now I tried again (without changing absolutely nothing), and it works! haha

juemura:
I fixed some minor things and it should be working properly now. I want to work on the scripts and on the actual paper, but once that's done I'll try again to:
1. make the uid be automatically inserted into the survey
2. use the actual emission uid instead of the habitica one.

shankari:
I will deploy this.
Do you want to add a new button that launches the survey even after the first time, in case people don't have time to finish the survey right then but want to wait until later?

www/js/goals.js
```diff
@@ -864,6 +872,77 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
+          scope: $scope,
+            buttons: [{
+              text: 'Ok',
+              type: 'button-positive',
+              onTap: function(e) {
+                 startSurvey();
+              }
+            }]
+        });
+        storage.set(SURVEY_DONE_KEY, true);
```
shankari: Maybe you should set this after the survey is actually done...

www/js/goals.js
```diff
@@ -864,6 +872,77 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
...
+    var checkSurveyDone = function () {
+      console.log("Checking if user already completed survey");
+      var SURVEY_DONE_KEY = 'survey_done';
+      var surveyDone = storage.get(SURVEY_DONE_KEY);
+      if (!surveyDone) {
+        $ionicPopup.show({
+          title: 'Bic2Cal Survey: Please take this 15-minute survey to provide feedback for our research.',
+          scope: $scope,
+            buttons: [{
+              text: 'Ok',
```
shankari: Maybe you should add a button for "later"

shankari:
Merging for now, we can tweak the flow later.

shankari:
@juemura did you test this on both 9.2 and 10.1? The UUID did not appear to populate on 10.1...

PR #215, opened 2017-03-20
Display unprocessed (draft) data in the diary as well

### LINKED ISSUE #214: Display unprocessed data in the diary
There is a requirement to display unprocessed data in the diary. This has multiple benefits:
- People will see that it is "working" immediately. This is particularly important since we have currently set the time interval for the pipeline to 12 hours!!
- Instead of moving to a new state for the incident reporting, we can simply go to the existing diary screen, which should fix many issues with moving to the correct screen
- It will also fix the issue with having multiple trips and have the incident reporting screen only work for one (the last one on android, since it is replaced every time, and any visible one on iOS, since once we go to the incident reporting screen, all prior popups disappear)
- Joan's group wants it

shankari:
This fixes #214.
Details in there.
High level design:
- read transitions after trip end
- walk through transitions and make trips
- for each trip, pull raw data
- convert this into a trip geojson
- add it to the existing geojson objects

The hardest part of this was the fiddly work of getting the geojson format consistent with the values sent from the server. But now we can potentially send only the trips instead of the full timeline for the day from the server to the client.

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -934,7 +934,7 @@ button.button.back-button.buttons.button-clear.header-item {
   background-color: #ffffff; 
 }
 .bg-unprocessed {
-  background-color: #22ee11; 
+  background-color: #9eb2aa; 
 }
 .list-card-sm {
   width: 95%;
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -924,11 +924,17 @@ button.button.back-button.buttons.button-clear.header-item {
   right: 16px;
   top: 16px;
 }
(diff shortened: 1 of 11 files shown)


shankari:
The light green background indicates that it is a draft. We also show "draft" to indicate this. I think that the button to re-run the pipeline should be at the top, because we will only support running the whole pipeline, not the pipeline for a particular trip.
 
| ios | android |
| ---- | ------- |
| screen shot 2017-03-20 at 2 45 17 pm | draft_trips_android |

PR #226, opened 2017-05-18
Restructure settings dialogs

shankari:


shankari:
For the record, the css changes in https://github.com/e-mission/e-mission-phone/pull/226/commits/9da7a834d1458d5e85bc50eb36e50483843326bc don't really work. They just make `toggle-color` unusable, so that it doesn't mess anything up.

It turns out that the default background color is green so this works for now. We should read up on how best to do this - maybe by using SCSS?

PR #231, opened 2017-06-09
removed Javascript includes to support iOS version 8.4

sunil07t:


shankari:
tested on iOS 8.4, works

PR #223, opened 2017-05-12
Adding the current screen

agarlin:
Add a screen that will be displayed while a trip is in progress

PR #234, opened 2017-06-27
Clean current screen

sunil07t:
- Fixed issues from PR #223 
- Changed current screen UI
- Speed in mph and direction
- Incidents are stored in cache
- Using pacman gif as the coordinate point to test change in direction and distance
- Moved current state config from main.js to diary.js because this version of current is in diary tab

Screenshot of the current screen UI
screen shot 2017-06-27 at 6 13 20 am


www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+    }).then(function(result) {
+      console.log(result);
+      var coordinates = result.map(function(locWrapper, index, locList) {
+        return [locWrapper.data.longitude, locWrapper.data.latitude];
+      });
+      var both = [result[0].data.longitude, result[0].data.latitude];
+      var bearing = Math.round(result[0].data.bearing);
+      console.log(bearing);
+      startTimeFn(result[result.length - 1].data.ts);
+      msToMph(result[0].data.sensed_speed);
```
shankari: This is not always guaranteed to be present, specially if location is generated from wifi instead of GPS. Might want to calculate it from the first two points if it is not present.

www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+            lng: both[0],
+            zoom: 15
+          }
+        },
+        markers: {
+          hereMarker: {
+            lat: both[1],
+            lng: both[0],
+            icon: {
+              iconUrl: 'img/test.gif',
```
shankari: nit: rename `test.gif` to something else

www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+    });
+    console.log($scope.mapCtrl);
+  };
+
+  $scope.$on('leafletDirectiveMap.current.resize', function(event, data) {
+        console.log("current/map received resize event, invalidating map size");
+        data.leafletObject.invalidateSize();
+  });
+   
+  var addSafe = function(marker, ts, map) {
```
shankari: Please don't copy-paste this from `post-trip-manual.js`. It is a service for a reason. Import it and use the methods in it. That's what I told AnnaMarie as well in her comments.

www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+    var value = {
+      loc: marker.toGeoJSON().geometry,
+      ts: ts,
+      stress: stressLevel
+    };
+    console.log("marker value", value);
+    $window.cordova.plugins.BEMUserCache.putMessage(MANUAL_INCIDENT, value);
+  };
+
+  var getLocalIncidents = function() {
```
shankari: It is not sufficient to get local incidents. Local incidents are only incidents that *this* user has entered (and which have not yet been pushed to the server). We want to show the user all incidents in this area, including those entered by other users. That's what makes it useful, right?

So AnnaMarie's original query was partially correct. You might want to only pull incidents from the past x time (maybe a slider to control that) so that the data is not obsolete.

You should combine the local incidents with the incidents from all users and show them on the same map.

www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+          marker.setStyle({color: 'red'});
+        } else {
+          marker.setStyle({color: 'green'});
+        }
+        _map.addLayer(marker);
+      });  
+    });
+  };
+
+  var cancelTempEntry = function(marker, ts, map) {
```
shankari: use from service, do not copy.

www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+    map.removeLayer(marker);
+  };
+
+  var addmarker = function(map, ts, llng) {
```
shankari: use from service, do not copy

www/js/diary/current.js
```diff
@@ -0,0 +1,224 @@
...
+    var hours = date.getHours();
+    var minutes = date.getMinutes();
+    var amOrPm = hours < 12 ? 'AM' : 'PM';
+    hours = hours % 12;
+    hours = hours ? hours : 12;
+    minutes = minutes < 10 ? '0'+ minutes : minutes;
+    $scope.startTime = hours + ':' + minutes + ' ' + amOrPm;
+  };
+
+  var msToMph = function(ms) {
```
sunil07t: I will change this to km/h but should we change to Mph for the app in the future?

www/js/diary/current.js
sunil07t: addSafeEntry, addSafeEntry, and cancelTempEntry function in post-trip-manual.js takes event as one of the argument but it is not used inside these functions. Also the call form current.js also does not need to pass the event. Should I remove the event argument from these functions in post-trip-manual.js

sunil07t: Also these functions are internal function. Should I make them part of the factory?

www/js/diary/current.js
shankari: Maybe offer people a choice between miles and km? Ideally, when people customize the UI, they would change this - a US-only one would use mph and the rest of the world would use kmph. But if you feel strongly about this, and it is not too much work, you can have it be controlled by a slider in the profile.

www/js/diary/current.js
shankari: 1. yeah I'm fine with removing the event
2. yeah you can make them be part of the factory. But I still don't see why you need to call them directly. why not just call `showSheet` from the factory?

sunil07t: `showSheet` needs featureArray, the current trip does not have this array.

shankari: you can pass in an empty featureArray. It won't harm anything - we just add the new feature to it.

shankari:
can you upload a new screenshot?

www/index.html
```diff
@@ -24,6 +24,8 @@
     <link href="common/dist/sprites/habitrpg-shared.css" rel="stylesheet">
     <link href="common/img/sprites/spritesmith/quests/bosses/quest.bosses.css" rel="stylesheet">
     <link rel="stylesheet" href="lib/ng-walkthrough/css/ng-walkthrough.css">
+    <link href="lib/angularjs-slider/dist/rzslider.css"/ rel="stylesheet">
```
shankari: doesn't ionic have a slider? why is this better?

www/js/diary/current.js
```diff
@@ -23,11 +29,47 @@
   });
 
   angular.extend($scope.mapCtrl.defaults, Config.getMapTiles());
-
   console.log("mapCtrl", $scope.mapCtrl);
 
-  var date = new Date();
-  var time = date.toTimeString().split(' ')[0].split(':');
+  $scope.verticalSlider = {
```
shankari: can address later: add a slider or an option in the profile window for how quickly to poll the server for incidents

www/js/diary/current.js
```diff
@@ -37,39 +79,60 @@
...
-    $scope.startTime = hours + ':' + minutes + ' ' + amOrPm;
+    return hours + ':' + minutes + ' ' + amOrPm;
+  };
+
+  var deg2rad = function(deg) {
+    return deg * (Math.PI/180);
+  };
+
+  // Haversine formula
+  var getMeterfromLglat = function(curr_lglat, last_lglat) {
```
shankari: you can use Leaflet `distanceTo` instead of re-implementing. I think that is what I use.

www/js/diary/current.js
```diff
@@ -37,39 +79,60 @@
...
     var index = Math.floor((degree / 45) + 0.5) % 8;
-    $scope.currentDirection = directions[index];
+    return directions[index];
   };
 
   var refreshTrip = function() {
-    leafletData.getMap('current').then(function(map) {
-      return db.getAllSensorData(BACKGROUND_LOCATION);
-    }).then(function(result) {
+    db.getAllSensorData(BACKGROUND_LOCATION).then(function(result) {
```
shankari: can address later: maybe try with filtered location instead of general location

www/js/diary/current.js
```diff
@@ -83,7 +146,7 @@
             lat: both[1],
             lng: both[0],
             icon: {
-              iconUrl: 'img/test.gif',
+              iconUrl: 'img/pacman.gif',
```
shankari: cool idea for pacman!

www/js/diary/current.js
```diff
@@ -110,115 +173,95 @@
...
+            PostTripManualMarker.showSheet($scope.features, latlng, ts, marker, _map);
     })
     .catch(function(error) {
       console.log("error while getting map current from leafletData");
     });
   };
 
+  $scope.$watch('features', function(newVal, oldVal){
+    if(newVal.length === 1) {
+      addIncidentLayer(newVal[0].properties.stress, marker, _map);
```
shankari: what does this do? Doesn't `showSheet` add the marker to the map anyway?

www/js/diary/current.js
```diff
@@ -110,115 +173,95 @@
...
+    gettingIncidents = setTimeout(getIncidentsLoop, 1000*60); //refresh every minute
+  };
 
-  $scope.$on('$ionicView.leave', function() {
-    clearTimeout(mapRunning);
-  });
+  leafletData.getMap('current').then(function(map) {
+    _map = map;
+    _map.removeControl(_map.zoomControl);
+    $scope.$on('$ionicView.enter', function() {
```
shankari: why are these inside `getMap`?
they don't need the map, and isn't it bad if you register for the enter event a little late? Maybe it will be fired before you register for it...

www/index.html
sunil07t: This slider has more options and it is easier to customize the slider css.

www/js/diary/current.js
sunil07t: `showSheet` does not add the marker to the map, it just stores the data locally.

www/js/diary/current.js
sunil07t: Current screen screenshot

sunil07t:
Current screen screenshot
screen shot 2017-06-29 at 2 22 33 pm

shankari:
What about the `getMap`? I think that is what I most want to see explained/changed before I merge

sunil07t:
Current screen gif
currentscreen

PR #239, opened 2017-07-06
Minor fixes on current and goal screen

sunil07t:
- Check newVal is defined 
- Using leaflet distanceTo to check distance between latlng
- Fixed walkthrough fix map button
- Added border to sign up text field
- Removed survey button from goal screen

### DIFF
--- a/www/js/diary/current.js
+++ b/www/js/diary/current.js
@@ -113,12 +113,7 @@
       var last_ts = result[1].data.ts;
       var bearing = Math.round(result[0].data.bearing);
       $scope.startTime = startTimeFn(result[result.length - 1].data.ts);
-      //if(angular.isDefined(result[0].data.sensed_speed)){
-        //$scope.currSpeedInKmh = Math.round(result[0].data.sensed_speed * 3.6); // mps to kmph
-      //} else {
-        // This speed is and average (not accurate) of current point and last point so it is not instantaneous rate of change
       $scope.currSpeedInKmh = getSpeed(both, last_both, curr_ts, last_ts);
-      //}
       $scope.currentDirection = degreeToDirection(bearing);
       angular.extend($scope.mapCtrl, { 
         defaults : {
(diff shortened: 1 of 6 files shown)


www/js/diary/current.js
```diff
@@ -82,31 +82,13 @@
...
-    var d = R * c;
-    return d;
-  };
-
   var getSpeed = function(curr_lglat, last_lglat, curr_ts, last_ts) {
-    distance = getMeterfromLglat(curr_lglat, last_lglat);
+    var curr_latlng = L.latLng(curr_lglat[1],curr_lglat[0]);
+    var last_latlng = L.latLng(last_lglat[1],last_lglat[0]);
+    var meters = curr_latlng.distanceTo(last_latlng);
+    console.log("Distance To", meters);
```
shankari: nit/fyi: using the Logger service will log to the database log that can be emailed in case we find bugs. Might want to do that at least for important logs.

shankari: `www/js/plugin/logger.js`

www/js/diary/current.js
```diff
@@ -131,12 +113,7 @@
       var last_ts = result[1].data.ts;
       var bearing = Math.round(result[0].data.bearing);
       $scope.startTime = startTimeFn(result[result.length - 1].data.ts);
-      if(angular.isDefined(result[0].data.sensed_speed)){
-        $scope.currSpeedInKmh = Math.round(result[0].data.sensed_speed * 3.6); // mps to kmph
```
shankari: gave up on `sensed_speed`?

www/templates/main-goals.html
```diff
@@ -1,10 +1,4 @@
 <ion-view view-title="Goals">
-  <ion-nav-buttons side="left">
-   </ion-nav-buttons>
-  <ion-nav-buttons side="right">
-      <button class="button button-icon ion-clipboard" ng-click="startSurvey()">Survey</button>
-      <button ng-if="screen" class="button button-icon ion-help" ng-click="startWalkthrough()"></button>
```
shankari: why remove the walkthrough as well?
having help is still useful...

PR #240, opened 2017-07-18
Changes to phone code for shake detect

ariasfelipe:
Additions and updates to files to create shake detection notifications. Something may be missing here for the notifications to work. The configurations/settings for potential_incident are not set.


www/index.html
```diff
@@ -82,6 +82,7 @@
     <script src="js/incident/post-trip-manual.js"></script>
     <script src="js/incident/post-trip-prompt.js"></script>
     <script src="js/incident/post-trip-map-display.js"></script>
+    <script src="js/incident/potential-incident-prompt.js"></script>
```
shankari: where is this file? I don't see it...

www/js/controllers.js
```diff
@@ -4,6 +4,7 @@ angular.module('emission.controllers', ['emission.splash.updatecheck',
                                         'emission.splash.startprefs',
                                         'emission.splash.referral',
                                         'emission.splash.pushnotify',
+                                        'emission.incident.potentialincident.prompt',
```
shankari: this is not sufficient, you need to inject the service as well.
See `PostTripAutoPrompt` below for an example - otherwise the init code will never be invoked.

www/js/incident/potential-incident-prompt.js
```diff
@@ -0,0 +1,250 @@
...
+          }
+        });
+    });
+    $window.cordova.plugins.notification.local.on('click', function (notification, state, data) {
+      // alert("clicked, no action");
+      if (!checkCategory(notification)) {
+          Logger.log("notification "+notification+" is not an incident report, returning...");
+          return;
+      }
+      cleanDataIfNecessary(notification, state, data);
```
shankari: we should report a positive incident on click as well, I think

PR #243, opened 2017-07-20
Removed tabs and buttons from profile UI

sunil07t:
Screenshots of the Trips and Profile tab
screen shot 2017-07-20 at 2 32 58 pm
screen shot 2017-07-20 at 2 33 46 pm

### DIFF
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -378,32 +378,6 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       $state.go('root.main.detail');
     };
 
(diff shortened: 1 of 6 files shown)


www/js/diary/list.js
```diff
@@ -378,32 +378,6 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       $state.go('root.main.detail');
     };
 
-    $scope.redirect = function(){
```
shankari: Hm, longer-term, we should do a better structure for the current screen. Shouldn't have to touch diary to fix this.

PR #249, opened 2017-07-28
Changed destination to purpose, color of the buttons and added uber/l…

sunil07t:
…yft in taxi text


www/js/diary/detail.js
```diff
@@ -221,17 +221,17 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
...
     } else {
-      checkOtherOption($scope.chosen.destination);
+      checkOtherOption($scope.chosen.purpose);
     }
   };
 
   $scope.chooseMode = function (){
     if($scope.chosen.mode != "other_mode"){
-      alert($scope.chosen.mode); // store Destination here
+      // store mode here
```
shankari: nit: this needs to be fixed :)

PR #250, opened 2017-07-29
Moved choose mode and choose purpose buttons to diary list screen

sunil07t:


www/js/diary/list.js
```diff
@@ -374,10 +374,113 @@ angular.module('emission.main.diary.list',['ui-leaflet',
         readAndUpdateForDay(nextDay);
     };
 
-    $scope.toDetail = function() {
-      $state.go('root.main.detail');
+    $scope.toDetail = function(param) {
```
shankari: why do you need this change?

www/templates/diary/list.html
```diff
@@ -33,17 +33,18 @@
         MUST test on android first to make sure that the scrolling works.
         -->
 
-		<div collection-repeat="tripgj in data.currDayTripWrappers" item-height="differentCommon(tripgj)? '348px':'308px'">
+		<div collection-repeat="tripgj in data.currDayTripWrappers" item-height="differentCommon(tripgj)? '413px':'365px'">
            <div class="start-time-tag">{{localTimeString (tripgj.data.properties.start_local_dt)}}</div>
             <div  style="padding-left: 19%;">
 
-			 <ion-item style="background-color: transparent;" id="no-border" class="list-item" href="#/root/main/diary/{{tripgj.data.id}}">
```
shankari: why remove the `href` from here?

sunil07t: The buttons are inside `ion-item` so I had to make separate div for buttons and the redirect. If both are just inside `ion-item` the click will direct to trip details and the buttons will not work properly.

shankari: can't you leave the href on the non-button div? Just curious so I can understand how href works.

sunil07t:
Changes:
- Moved the choose mode and choose purpose buttons to diary list card. 
- The popover buttons are little bigger than last version so they are easier to press.
- Had to change the height of the cards to show the buttons. 
- After one of the options from the list is chosen the popover will disappear.

Testing done: 
- The popover shows up on the diary list screen
- Spacing and height of the cards in the diary screen are responsive
screen shot 2017-07-29 at 4 20 34 pm
screen shot 2017-07-29 at 4 20 44 pm
screen shot 2017-07-29 at 4 20 52 pm

www/templates/diary/list.html
sunil07t: Actually, the href did not work when I tested it. So I changed it to `stage.go()` with params

shankari: so you left the href on the part of the item that does not have the button?
e.g. `<div ng-click="toDetail(tripgj.data.id)">` ->  `<div href="#/root/main/diary/{{tripgj.data.id}}">`
doesn't work?

sunil07t: This does not work

PR #254, opened 2017-07-31
Add support for a trip-specific qualtrics survey

shankari:
Add support for launching a qualtrics survey with auto-completable start and
end times as well. This allows the survey results to be associated with trips
later.

We should really refactor `uuid_insert` and `time_insert` into a single file.
But given the tight schedule here, I copy to avoid regressions.
This should be fixed later.


shankari:
Testing done:
1. changed the post-trip-prompt code to launch the survey (code not included here)
1. ended a trip
1. launched the survey.

| Prompt | Survey | 
| --------| -------|
| simulator screen shot jul 31 2017 7 31 33 pm | simulator screen shot jul 31 2017 7 31 43 pm |

shankari:
@sunil07t here are the changes to support trip-specific surveys

PR #256, opened 2017-08-04
Show choose mode and purpose prompt after the trip with the trip map

sunil07t:
Changes done:
- Decreased the height of the map to fit the prompt
- Used slider to prompt choose mode and choose purpose.
- Unable to slide the slider unless user picks a mode and presses continue 
- After a mode is chosen, the slider slides to second prompt
- Store the mode and purpose only after user presses continue or done, not when they click one of the options in the radio list.
- The radio list is scrollable
- Walkthrough tells the user that the radio list is scrollable 

Testing done on both Android and iOS devices:
- Walkthrough highlights the radio list
- Mode and purpose stored to a variable, test done with console log 
- Tested on phones with different screen size

Screenshots:
screen shot 2017-08-03 at 7 22 45 pm
screen shot 2017-08-03 at 7 45 07 pm
screen shot 2017-08-03 at 7 45 25 pm
screen shot 2017-08-03 at 7 45 35 pm
screen shot 2017-08-03 at 8 03 18 pm

### DIFF
--- a/www/templates/control/main-control.html
+++ b/www/templates/control/main-control.html
@@ -93,13 +93,13 @@
     </div-->
 
 
-    <!--div class="control-list-item">
+    <div class="control-list-item">
       <div class="control-list-text">Developer zone</div>
       <div ng-click="expandDeveloperZone()" class="control-icon-button"><i ng-class="getExpandButtonClass()" id="expandButton"></i></div>
-    </div-->
+    </div>
 
     <!-- Begin developer zone -->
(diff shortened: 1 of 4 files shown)


www/templates/control/main-control.html
```diff
@@ -93,13 +93,13 @@
     </div-->
 
 
-    <!--div class="control-list-item">
```
shankari: this is not correct - you checked in the changes to re-enable the developer zone (presumably for testing)
you need to revert them before I can push.

PR #260, opened 2017-08-08
Minor changes for cci

sunil07t:
Changes done:
- Changed km to mi and m/s to mph in diary screen
- Used ionic `hide-back-button` to hide the back button in about screen
- Added Transit transfer to purpose 

Testing done:
- Calculated the km to mi shown on the emulator
- Back button shows up in diary details but is hidden in about screen after completing the prompt


www/templates/diary/detail.html
```diff
@@ -33,7 +33,7 @@
         <div style="height: 15px;"></div>
         <div class="row" style="padding-top: 10px !important; padding-bottom: 10px !important; padding-left: 20px !important; padding-right: 20px !important;">
           <div class="col" style="width: 30%; margin-top: 0; font-size: 1.1em; font-weight: thin;">
-          <p style="font-size: 1.1em; color: #cccccc; margin-bottom: 0;">Distance</p>{{getFormattedDistance(trip.properties.distance)}} km</div>
+          <p style="font-size: 1.1em; color: #cccccc; margin-bottom: 0;">Distance</p>{{getFormattedDistance(trip.properties.distance)}} mi</div>
```
shankari: doesn't the detail view also have a graph of the speed so far?
does that work correctly with this?

sunil07t: Yes, the graph was in m/s which is now mph https://github.com/e-mission/e-mission-phone/pull/260/files#diff-30e353982546b2071becab3038f1bc08R89

PR #271, opened 2017-08-27
Restore cci branch changes

shankari:
I inadvertently merged changes from `interscity` to `master`, in 98f76e324167525513ea6c1cdfc3e4a8f150b463, when I intended the reverse.  Since we branched `interscity` from `cci-berkeley`, this means that changes from `cci-berkley` before 137933c0186978bb9bb62a95ebd1931fb904e351 were pulled into master.

I am still not sure why this happened. The best explanation I can think of is that I pulled upstream `master` into my local `interscity` and then instead of pushing to `origin interscity`, I pushed back to `upstream master`.  

Since this was unintentional, I reverted that change. But that ended up reverting all changes to the `cci` branch before 137933c0186978bb9bb62a95ebd1931fb904e351.

This commit, and 702de2fca40ef04d88b8b5f1b0998fbe1f48fd7b, restore those changes. I generated this diff by generating a patch with the reverted changes and applying it to the current top.

```
$ git diff ff3ee48a8ae3db977857111cd672fcb8239791be 137933c > /tmp/restore_cci_changes.patch
```

The difference between the result and the correct change to the cci branch (02e511b64c18d6465aa94d71844aaa516386b774), should only be the changes made to the master in that time.

```
$ git diff 02e511b64c18d6465aa94d71844aaa516386b774 > /tmp/real_changes_from_main.patch
```

That check caught a few more missing pieces, but finally, the diff (attached) looks good.
And that is what this commit is.


shankari:
Diff is:
[real_changes_from_main.patch.gz](https://github.com/e-mission/e-mission-phone/files/1254174/real_changes_from_main.patch.gz)

Resulting UI is:

| About              | Trips          | Config |
|--------------|------------|--------|
| simulator screen shot aug 26 2017 11 10 42 pm | simulator screen shot aug 26 2017 11 10 20 pm  | simulator screen shot aug 26 2017 11 07 51 pm |

PR #278, opened 2017-08-31
fix notification registering error

Andrew-Tan:
:)))))


Andrew-Tan:
Still need to tweak to not to directly jump to survey every time the trip stops, but it is fine for now.

PR #282, opened 2017-09-15
Show modes and purpose on list screen, not completed but to test

sunil07t:
The PR is to test. Work in progress.
- The list of mode and purpose saved locally get overridden when choosing mode from prompt screen.

- Test done
    - Picked the modes from list screen for couple trips which populated the local mode and purpose list. 
    - Then simulated a trip from control screen and to go to notification/prompt screen to chose the mode and purpose for the simulated trip.
    - Added a break point on list screen to see the purpose list and it only has one purpose in the list, the one that was picked from prompt screen, previous purposes were overridden.


shankari:
I added the attached patch to debug the number of items retrieved.
[print_list_length.patch.txt](https://github.com/e-mission/e-mission-phone/files/1308096/print_list_length.patch.txt)

And I didn't see this behavior the first time, but I did see it the second time.
What was the difference between the two?
There was a push to the server in between.

```
[Log] Modes list length = 4 (console-via-logger.js, line 173)
[Log] Modes list length = 4 (console-via-logger.js, line 173)
[Log] Modes list length = 5 (console-via-logger.js, line 173)
[Log] Modes list length = 5 (console-via-logger.js, line 173)
[Log] Modes list length = 5 (console-via-logger.js, line 173)
[Log] Modes list length = 1 (console-via-logger.js, line 173)
```

You are reading only from the cache `BEMUserCache` so you don't see values that have been pushed to the server. You need to use `UnifiedDataLoader`, just like we do for (say) transitions.

```
       return UnifiedDataLoader.getUnifiedMessagesForInterval("statemachine/transition", tq)
        .then(function(transitionList) {
....
```

PR #287, opened 2017-09-22
Added survey to about page, fixed minor changes

jenny3248:


www/js/cci-about/about.js
```diff
@@ -19,5 +24,7 @@ angular.module('emission.main.cci-about', ['emission.plugin.logger'])
            window.Logger.log(window.Logger.LEVEL_INFO,
                "CCI email cancel reported, seems to be an error on android");
         });
-    };
+  };
+  var SURVEY_DONE_KEY = 'survey_done';
```
shankari: you only need this if you want the survey to pop-up automatically. And then you would set the `SURVEY_DONE_KEY` to true once the user finished the survey. I commented this out for the goals screen since we didn't want to popup a survey.

Original code for the popup is here:
https://github.com/e-mission/e-mission-phone/commit/6d86ce5fb6ed3dfa385b19e02ea9d574e2345755#diff-c3e554e73846808d6fb5cf0538e649c5 (line 933 onwards)

www/js/goals.js
```diff
@@ -896,8 +896,8 @@ angular.module('emission.main.goals',['emission.services', 'emission.plugin.logg
 
     $scope.startSurvey = function () {
       // (URL, elementID)
-      SurveyLaunch.startSurvey('https://berkeley.qualtrics.com/SE/?SID=SV_5pzFk7JnMkfWBw1', 'QR~QID2');
-      // startSurvey();
+      SurveyLaunch.startSurvey('https://berkeley.qualtrics.com/jfe/form/SV_eQBjPXx10yaAScl', 'QR~QID3');
```
shankari: you don't need this since you never show the goals screen.

shankari: You can revert this change.

PR #291, opened 2017-10-01
Changed consent form and other minor changes

jenny3248:
I am still working on getting the app running on my computer.


www/templates/cci-about/about.html
```diff
@@ -7,7 +7,7 @@
 		<p style="font-size: 15px;text-align: center;padding: 5px 10px 5px 10px;">
 			If this is your first time logging in please complete the following survey:
 		</p>
-		<button class="button button-icon ion-clipboard" ng-click="startSurvey()">Survey
+		<button class="button button-icon ion-clipboard" ng-click="startSurvey()" class="button icon icon-left ion-email cci-contact-button">Survey
```
shankari: This has both `ion-clipboard` and `ion-email`? Don't you want one or the other?

ionic.config.json
```diff
@@ -0,0 +1,11 @@
+{
```
shankari: Please don't check in this file - it is specific to my compiled version in the app store. Other installations should have their own app id.

shankari: you can remove it using `git rm --cached ionic.config.json`

www/templates/intro/summary.html
```diff
@@ -1,14 +1,12 @@
...
 	<div class="intro-text">
-	This version of E-mission is designed to:
 	<div class="intro-space"></div>
-	<li> This version of E-Mission is designed to record your travel behavior for a period of four days in order to calculate your Vehicle Miles Travelled (VMT). The application will collect trip location data, travel speed, and estimate mode of travel such as walking, biking, or vehicle. All data collected is used solely for the purpose of research for the research study at UC Berkeley’s Center for Community Innovation, entitled “Assessing the Travel Demand and Co-Benefit Impacts of Affordable Transit-Oriented Development.”
-	</ol>
+	This version of E-Mission is designed to record your travel behavior for a period of four days in order to calculate your Vehicle Miles Travelled (VMT). The application will collect trip location data, travel speed, and estimate mode of travel such as walking, biking, or vehicle. All data collected is used solely for the purpose of research for the research study at UC Berkeley’s Center for Community Innovation, entitled “Assessing the Travel Demand and Co-Benefit Impacts of Affordable Transit-Oriented Development.”
```
shankari: This is not a code change, but the summary says four days and the consent says one week

shankari:
@jenny3248, why did you close this?

PR #295, opened 2017-10-17
changed login and about page

jenny3248:


www/templates/cci-about/about.html
```diff
@@ -7,7 +7,7 @@
 		<p style="font-size: 15px;text-align: center;padding: 5px 10px 5px 10px;">
 			If this is your first time logging in please complete the following survey:
 		</p>
-		<button class="button button-icon ion-clipboard" ng-click="startSurvey()">Survey
+		<button id="call-cci" class="button icon icon-left ion-clipboard cci-contact-button" ng-click="startSurvey()">Survey
```
shankari: nit: the id for the elements should be unique, so you should use something other than `call-cci` for the id

PR #305, opened 2017-12-24
Ground truth other user's input option debugged

sunil07t:
**Problem**
When the user presses other option, the popup text field UI waits for user to input their own mode or purpose. The `choosePurpose` and `chooseMode` function checks if the user pressed other and takes them to pop up. However, in `choosePurpose` and `chooseMode` functions `closeModePopover` and `closePurposePopover` runs in asynchronous and in these 2 function `$scope.draftPurpose = angular.undefined;` makes the draft object undefined. Which causes an error when a user adds their input in the other popup text field because the save button tap event uses `storeMode` and `storePurpose` to store the ground truth and it uses $scope.draftPurpose for the timestamp. 

**Solution**
So the solution adds a boolean that does not `$scope.draftPurpose = angular.undefined;` in `closeModePopover` and `closePurposePopover` if the user pressed other option because `storeMode` and `storePurpose` still needs `$scope.draftPurpose` object. `$scope.draftPurpose = angular.undefined;` is done inside `storeMode` and `storePurpose` for other option.


sunil07t:
Closed because the PR was done to wrong branch

PR #306, opened 2017-12-24
Ground truth other option debugged

sunil07t:
**Problem**
When the user presses other option, the popup text field UI waits for user to input their own mode or purpose. The `choosePurpose` and `chooseMode` function checks if the user pressed other and takes them to pop up. However, in `choosePurpose` and `chooseMode` functions `closeModePopover` and `closePurposePopover` runs in asynchronous while checking if user pressed other. In these 2 function `$scope.draftPurpose = angular.undefined` makes the draft object undefined. Which causes an error when a user adds their input in the other popup text field because the save button tap event uses `storeMode` and `storePurpose` to store the ground truth and it uses `$scope.draftPurpose` for the timestamp.

**Solution**
So the solution adds a boolean that does not `$scope.draftPurpose = angular.undefined` in `closeModePopover` and `closePurposePopover` if the user pressed other option because `storeMode` and `storePurpose` still needs `$scope.draftPurpose` object. `$scope.draftPurpose = angular.undefined` is done inside `storeMode` and `storePurpose` for other option.


www/js/diary/list.js
```diff
@@ -445,19 +448,23 @@ angular.module('emission.main.diary.list',['ui-leaflet',
    };
 
   $scope.choosePurpose = function() {
+    var isOther = false
     if($scope.chosen.purpose != "other_purpose"){
-      $scope.storePurpose($scope.chosen.purpose);
+      $scope.storePurpose($scope.chosen.purpose, isOther);
```
shankari: why not just put `true` or `false` here directly instead of having an `isOther` variable?

www/js/diary/list.js
```diff
@@ -487,16 +494,20 @@ angular.module('emission.main.diary.list',['ui-leaflet',
    {text:'Religious', value:'religious'},
    {text:'Other',value:'other_purpose'}];
 
-   $scope.storeMode = function(mode_val) {
+   $scope.storeMode = function(mode_val, isOther) {
       $scope.draftMode.label = mode_val;
       Logger.log("in storeMode, after setting mode_val = "+mode_val+", draftMode = "+JSON.stringify($scope.draftMode));
       $window.cordova.plugins.BEMUserCache.putMessage(MODE_CONFIRM_KEY, $scope.draftMode);
+      if(isOther = true)
```
shankari: isn't `isOther = true` assignment? Shouldn't it be `isOther == true` for equality?

www/js/diary/list.js
```diff
@@ -487,16 +494,20 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
       $window.cordova.plugins.BEMUserCache.putMessage(MODE_CONFIRM_KEY, $scope.draftMode);
+      if(isOther = true) 
+        $scope.draftPurpose = angular.undefined;
    }
 
-   $scope.storePurpose = function(purpose_val) {
+   $scope.storePurpose = function(purpose_val, isOther) {
       $scope.draftPurpose.label = purpose_val;
       Logger.log("in storePurpose, after setting purpose_val = "+purpose_val+", draftPurpose = "+JSON.stringify($scope.draftPurpose));
       $window.cordova.plugins.BEMUserCache.putMessage(PURPOSE_CONFIRM_KEY, $scope.draftPurpose);
```
shankari: doesn't this store the `draftPurpose` (which is `undefined` in this case) into the database?

sunil07t: This gives an error because $scope.draftPurpose.label is undefined so it is not possible to assign purpose_val to it, this is the bug.

www/js/diary/list.js
sunil07t: Sorry, I will make this change. This worked because it was just `$scope.draftPurpose = angular.undefined` after every store.

www/js/diary/list.js
sunil07t: I added isOther variable for consistent temporary names to make it easy follow why it passes a boolean.

www/js/diary/list.js
shankari: @sunil07t But how does your change fix this? you pass in `isOther`, but you don't actually check it until the last line.

sunil07t: This check is just to make `draftPurpose` undefined

sunil07t: This is just to make the `draftPurpose` undefined

shankari: I am sorry, I still don't understand how this works.

```
$scope.storePurpose = function(purpose_val, isOther) {
<----- here, $scope.draftPurpose is undefined
       $scope.draftPurpose.label = purpose_val; <--- so why doesn't this try to set a subfield of undefined?
       Logger.log("in storePurpose, after setting purpose_val = "+purpose_val+", draftPurpose = "+JSON.stringify($scope.draftPurpose));
```

Did you re-test after making the boolean check? Can you explain to me how it works?

sunil07t: Yes, I re-tested it after fixing the boolean check. The boolean check just makes the `draftPurpose` undefined at the end because the closePopup functions will not make `draftPurpose` undefined and that is what we want. The closePopup functions makes `draftPurpose` undefined before the user has a chances to add the label.

sunil07t: However, we need closePopover function to close the tab when the user presses one of the options in the popover list.

PR #307, opened 2017-12-28
Show the changes to the UI after user chooses mode/purpose on list screen

sunil07t:
Before the user could choose the mode/purpose but it did not show up on the screen. This change will show the selected mode/purpose on the UI. Following are the screenshots of the UI
screen shot 2017-12-27 at 11 07 20 pm
screen shot 2017-12-27 at 11 07 34 pm

### DIFF
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -14,9 +14,10 @@ angular.module('emission.main.diary.list',['ui-leaflet',
                                     $ionicActionSheet,
                                     ionicDatePicker,
                                     leafletData, Timeline, CommonGraph, DiaryHelper,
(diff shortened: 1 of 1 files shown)


atton16:
When user answer to the prompt, will the answer get synced after the **next** trip has ended?

shankari:
>  When user answer to the prompt, will the answer get synced after the next trip has ended?

In general, yes. It depends on when the user provides the answer. The trip end is typically detected ~ 5-10 mins after the user actually stops moving; we need that time to ensure that stops are not treated as trip ends. The answers to the responses are stored in the local database, just like all other data collected on the phone. The goal is to make the data collection robust to connection failures - we should be able to collect data even when the user is underground, has no connectivity...

PR #310, opened 2018-01-08
Create the new branch by pulling from interscity

shankari:


shankari:
+ switch ground truth

PR #314, opened 2018-01-16
Fix issue with bad location points

shankari:
Apparently, we sometimes get bad location points, with `data.ts` significantly
different from `metadata.write_ts`.
https://github.com/e-mission/e-mission-server/issues/532#issuecomment-357801072

This leads to weird trips
https://github.com/e-mission/e-mission-server/issues/532#issuecomment-357791873
and can be fixed by filtering points that don't fit within the transitions
https://github.com/e-mission/e-mission-server/issues/532#issuecomment-357834736

### DIFF
--- a/www/js/diary/services.js
+++ b/www/js/diary/services.js
@@ -703,13 +703,20 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
             return undefined;
           }
           var sortedLocationList = locationList.sort(tsEntrySort);
-          var tripStartPoint = sortedLocationList[0];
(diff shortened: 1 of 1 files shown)


shankari:
After fix, trips are:
9:19 - 9:32
9:34 - 9:56
5:28 - 5:49
5:54 - 6:46
7:02 - 7:50

shankari:
| Beginning of day | End of day |
| -----| ---- |
| simulator screen shot - iphone 8 plus - 2018-01-15 at 18 28 50 | simulator screen shot - iphone 8 plus - 2018-01-15 at 18 29 12 |

PR #341, opened 2018-02-04
Add a button that launches the access map in an iab window

shankari:
UW team can experiment with `"_blank"`, `"_system"`, `"_self"` etc to see which UI:
a) works
b) feels intuitive


shankari:
Works on iOS
[in_app_browser.mov.gz](https://github.com/e-mission/e-mission-phone/files/1692389/in_app_browser.mov.gz)

Andrew-Tan:
Thank you for your effort! Currently considering two mechanism:
1. Use in-app browser
2. Open external browser with the link

shankari:
@Andrew-Tan LMK what works! FYI using `"_system"` with the iab is supposed to open the page in the system browser instead of in a webview, so it should be easy to test out all the options :)

Andrew-Tan:
Great! Thanks, will try that out.

shankari:
closing this since @Andrew-Tan is already testing it out separately.

PR #344, opened 2018-02-07
Urap 2017 information

jackcsullivan:
Demo information code.


www/js/diary/list.js
```diff
@@ -172,16 +172,26 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
-          $scope.data = Timeline.data;
-          $scope.datepickerObject.inputDate = Timeline.data.currDay.toDate();
-          $scope.data.currDayTrips.forEach(function(trip, index, array) {
-              PostTripManualMarker.addUnpushedIncidents(trip);
-          });
-          $scope.data.currDayTripWrappers = Timeline.data.currDayTrips.map(
-            DiaryHelper.directiveForTrip);
-          $ionicScrollDelegate.scrollTop(true);
-      });
+      if ($rootScope.tripTimelineUpdate == true) {
```
shankari: When I tested it, I added a new `$scope.$on(Timeline.UPDATE_DONE, function(event, args) {` to `metrics.js` and put this new code (that invokes `Timeline.updateForDay($rootScope.recentTripDate.startOf('day'));...`) in there.

shankari: You should do something like that too because otherwise, you call the update when it is done, which seems wrong.

www/js/metrics.js
```diff
@@ -1315,9 +1313,13 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
   }
 
-  $scope.openRecentTrip = function(tripId, startTs) {
-      Timeline.updateForDay(moment(startTs).startOf('day'));
-      window.location.href = "#/root/main/diary/" + tripId;
+  $scope.openRecentTrip = function(tripId, trip_ts) {
+    $rootScope.recentTripID = tripId;
+    $rootScope.recentTripDate = moment(trip_ts * 1000);
+    $rootScope.tripTimelineUpdate = true;
+    $state.go('root.main.diary').then(function() {
```
shankari: As I said, this will only work if the recent trips are from today

> if you test that against a user who does have data for today, and replace `currTrips[0].id` with the id of the recent trip, it should work

shankari: My original suggestion was

> it is actually a relevant use case that there are no trips over (say) the weekend, so you may want to load the timeline (in the background) for the days that correspond to the recent trips

But looking at the code again, we can only have the timeline for one day at a time right now

So a better short-term fix might be to have the list view take in the day to load, and the invoke `$state.go` with the correct params.

The long-term fix would be to have the timeline cache the data for multiple days instead of just the current day and then load multiple timeline days in the background.

Or have the timeline work for a time range instead of on a day by day basis.

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.1.1",
   "description": "Automatic tracker",
   "dependencies": {
+    "@ionic-native/launch-navigator": "^4.5.3",
```
shankari: Is this a new plugin? New plugin stuff uses cordova directly, which puts this into config.xml

www/js/control/general-settings.js
```diff
@@ -105,6 +105,34 @@ angular.module('emission.main.control',['emission.services',
             $state.reload();
         }
     }
+    $scope.showUsernamePopup = function() {
+    $scope.data = {};
+
+    var usernamePopup = $ionicPopup.show({
+      template: '<input type="userEmail" ng-model="data.wifi">',
```
shankari: use a different variable name for clarity

www/js/intro.js
```diff
@@ -58,6 +58,10 @@ angular.module('emission.intro', ['emission.splash.startprefs',
       }
     });
   };
+  $scope.startSurvey = function () {
+      SurveyLaunch.startSurvey('https://docs.google.com/forms/d/e/1FAIpQLSd47sf_0bieu81-HtXO5PV3EmOdUfDjWE5xrLSzQ-1hVrgLgQ/viewform?usp=sf_link', 'QR~QID3');
+      startSurvey();
```
shankari: what is this `startSurvey()` for?

www/js/intro.js
```diff
@@ -95,6 +100,35 @@ angular.module('emission.intro', ['emission.splash.startprefs',
         $scope.finish();
     });
   };
+  $scope.showUsernamePopup = function() {
+  $scope.data = {};
+
+  var usernamePopup = $ionicPopup.show({
+    template: '<input type="userEmail" ng-model="data.wifi">',
```
shankari: nit: fix variable name for clarity

www/js/main.js
```diff
@@ -50,6 +50,16 @@ angular.module('emission.main', ['emission.main.recent',
       }
     }
   })
+  
+  .state('root.main.leaderboard', {
+    url: '/leaderboard',
+    views: {
+      'leaderboard': {
+        templateUrl: 'templates/leaderboard.html',
+        controller: 'MetricsCtrl'
```
shankari: why not have a separate controller for the leaderboard instead of glumping everything into `MetricsCtrl` which is big enough already?

www/templates/intro/consent.html
```diff
@@ -1,113 +1,105 @@
 <ion-content>
 <div class="consent-title">
-<center><h4>E-Mission: Data driven carbon emission reduction</h4></center>
+<center><h4>ECO-ESCORT: Data driven carbon emission reduction</h4></center>
```
shankari: Fix NAME!!!

www/templates/intro/consent.html
```diff
@@ -1,113 +1,105 @@
...
-<h3>Introduction and Purpose</h3>
-This app is part of a travel behavior research platform developed at the
-University of California, Berkeley. The platform development is being led by <a
-href="http://www.cs.berkeley.edu/~shankari">K. Shankari</a> under the
-supervision of <a href="http://www.cs.berkeley.edu/~culler/">Prof. David
-Culler</a>. Our goal is to collect information from users about travel behavior
-across all modes, and use the aggregate information for better transportation
-infrastructure planning.
+<h3>Introduction</h3>
+Our names are Andy Yan, Jack Sullivan, Jesse Zhang, Kyle Tse, Simon Behar, John Leyden and
```
shankari: Fix Simon's name

www/templates/intro/consent.html
```diff
@@ -1,113 +1,105 @@
...
+
 
 <h3>Questions</h3>
 If you have any questions about this research, please feel free to contact us.
-For the quickest response, you can send email to the entire team at <a
-href="mailto:e-mission@lists.eecs.berkeley.edu">e-mission@lists.eecs.berkeley.edu</a>.
-If you want to contact only the lead researcher or the PI, you can use their
-profiles above.
+For the quickest response, you can send email to the entire team at
+<a href='emissionberkeley@gmail.com'>emissionberkeley@gmail.com</a>.
```
shankari: Switch to study email

www/templates/intro/consent.html
```diff
@@ -1,113 +1,105 @@
...
+and want to participate in our study.
 
 <div class="row" style="margin-top: 20px;">
-  <div class="col" style="margin-right: 5px">
-    <button class="button button-block button-assertive button-outline" ng-click="disagree()" id="consent-button">I Refuse</button>
-  </div>
-  <div class="col" style="margin-left: 5px">
-    <button class="button button-block button-balanced button-outline" ng-click="agree()" id="consent-button">I Agree</button>
+  <div class="col">
+    <button class="button button-block button-balanced button-outline" ng-click="agree()" id="consent-button">Agree</button>
```
shankari: IRB didn't require an "I disagree" button? They were very picky when I first filed the protocol.

www/templates/main.html
```diff
@@ -5,42 +5,49 @@
 -->
 <ion-view>
 <ion-tabs ng-class="tabsCustomClass()">
-
-  <!-- Diary Tab -->
-  <ion-tab title="Diary" icon="ion-map" href="#/root/main/diary">
+  <!-- TODO: make a new leaderboards tab, and make a polar bear tab that shows if they're on emotion -->
```
shankari: Can remove TODO!

shankari: Also, you don't need to check in packages-lock.json and should use the icon font instead of separate icons to avoid duplication/extra space.

www/js/diary/list.js
```diff
@@ -172,16 +172,26 @@ angular.module('emission.main.diary.list',['ui-leaflet',
 
     $scope.$on(Timeline.UPDATE_DONE, function(event, args) {
       console.log("Got timeline update done event with args "+JSON.stringify(args));
-      $scope.$apply(function() {
```
shankari: in `list.js` line 53, read the day to update from the `$stateParams`, falling back to `moment().startOf('day')` only if there is no parameter specified. `readAndUpdateForDay(moment().startOf('day'));`

www/js/metrics.js
```diff
@@ -1157,5 +1313,13 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
         return ($scope.expandedc)? "expanded-calorie-card" : "small-calorie-card";
   }
 
-  
+  $scope.openRecentTrip = function(tripId, trip_ts) {
+    $rootScope.recentTripID = tripId;
+    $rootScope.recentTripDate = moment(trip_ts * 1000);
+    $rootScope.tripTimelineUpdate = true;
+    $state.go('root.main.diary').then(function() {
```
shankari: Then, call this `$state.go('root.main.diary')` with stateParams, e.g.
`$state.go('root.main.diary', {'date': moment(trip_ts * 1000).startOf('day')})`  or something like that.
Combined with the change in `list.js`, it will ensure that the timeline for the correct day is loaded.
And then when you get the TIMELINE_DONE event, you can switch to the detail view with stateParams = tripid and everything should work.

www/js/intro.js
jesbu1: This is for launching the Initial Survey after they log in.

www/templates/intro/consent.html
jesbu1: No, we took out the "releasing data after a year" thing

www/js/intro.js
shankari: Yeah but isn't the `SurveyLaunch.startSurvey` enough for that?
why do you need a separate `startSurvey`?

www/js/control/general-settings.js
```diff
@@ -290,9 +293,18 @@ angular.module('emission.main.control',['emission.services',
...
     $scope.getUsername = function() {
-      CommHelper.getUsername().then(function(results){
-        $scope.settings.username = results['username']
-      })
+      var username = localStorage.getItem('username');
+      if (username != null) {
+        $scope.settings.username = username;
+      } else {
+        CommHelper.getUsername().then(function(results){
+          ionic.show
```
shankari: what does this line do?

www/js/intro.js
jesbu1: Oh huh, I think I was going to add extra functionality to that method but moved it to the username one. 

Will change.

www/js/intro.js
```diff
@@ -104,8 +108,8 @@ angular.module('emission.intro', ['emission.splash.startprefs',
   $scope.data = {};
 
   var usernamePopup = $ionicPopup.show({
-    template: '<input type="userEmail" ng-model="data.wifi">',
-    title: 'Create a new username/Edit your username',
+    template: '<input type="userEmail" ng-model="data.wifi">', //This is wifi because i copoied pasted code, afraid to change it (dont fix what ain't broke!)
```
shankari: Refactor to a single service instead of duplicating code in control and intro?

www/js/control/general-settings.js
jesbu1: If for some reason the username is not yet stored in local storage (like for when people are upgrading versions to the new version that stores it into localStorage) then it's going to retrieve the username from the server to display in settings.

shankari: yeah but what does `ionic.show` with no arguments do?

jesbu1: OH no idea how that got there, definitely shouldn't be there.

package.json
jesbu1: Don't actually need that i think I looked into that before I knew about survey launch

www/templates/intro/consent.html
jesbu1: Not sure what you mean by this

config.xml
```diff
@@ -13,7 +9,9 @@
   </author>
   <content src="index.html"/>
   <access origin="*"/>
-  <allow-navigation href="https://em-game.eecs.berkeley.edu/*" />
+  <allow-navigation href="https://em-game.eecs.berkeley.edu/*"/>
+  <allow-navigation href="http://10.142.36.203:8100"/>
```
shankari: Do you really need this `allow-navigation`? I think we only used it to access the em-game website directly from the app.

config.xml
```diff
@@ -22,11 +20,11 @@
...
-  <preference name="xwalkVersion" value="19+" />
-  <preference name="xwalkCommandLine" value="--disable-pull-to-refresh-effect" />
-  <preference name="xwalkMode" value="embedded" />
-  <preference name="xwalkMultipleApk" value="true" />
-  <preference name="emSensorDataCollectionProtocolApprovalDate" value="2016-07-14" />
+  <preference name="xwalkVersion" value="19+"/>
+  <preference name="xwalkCommandLine" value="--disable-pull-to-refresh-effect"/>
+  <preference name="xwalkMode" value="embedded"/>
+  <preference name="xwalkMultipleApk" value="true"/>
+  <preference name="emSensorDataCollectionProtocolApprovalDate" value="2016-07-14"/>
```
shankari: Are all of these real changes? It looks like they are primarily related to removing whitespace...

www/js/metrics.js
```diff
@@ -1316,10 +1316,17 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
   $scope.openRecentTrip = function(tripId, trip_ts) {
     $rootScope.recentTripID = tripId;
     $rootScope.recentTripDate = moment(trip_ts * 1000);
-    $rootScope.tripTimelineUpdate = true;
-    $state.go('root.main.diary').then(function() {
+    $rootScope.recentTripDetailLoad = true;
+    $state.go('root.main.diary', {'date': moment(trip_ts * 1000).startOf('day'), "trip_id": tripId}).then(function() {
```
shankari: I don't think you need the `trip_id` here

shankari: also, nit: can replace `moment(trip_ts * 1000)` with `$rootScope.recentTripDate`

www/js/metrics.js
```diff
@@ -1316,10 +1316,17 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
       console.log("finished going to the list view, moving to the detail view now");
     });
   }
 
+  $scope.$on(Timeline.UPDATE_DONE, function(event, args) {
+    if ($rootScope.recentTripDetailLoad == true) {
+      $state.go('root.main.diary-detail', {tripId: $rootScope.recentTripID})
+      $rootScope.recentTripDetailLoad = false;
```
shankari: nit: should also reset `$rootScope.recentTripID` after going to the detail page.

shankari:
High level changes:
- `config.xml` merge is incorrect - you need to end up with version 2.5.0, which is the version in master
- remember to remove everything in `www/common` since you don't use habitica and it is very big

www/templates/intro/consent.html
shankari: Simon is no longer part of the project, no? But I guess we can't actually change this because the consent document has to be unchanged. Bummer.

shankari:
`config.xml` changes look good.
Just waiting for the correct fix for the notification disable before I merge.

PR #351, opened 2018-02-16
ionic.config.json updated (control)

jackcsullivan:


www/js/intro.js
```diff
@@ -84,6 +84,9 @@ angular.module('emission.intro', ['emission.splash.startprefs',
 
   $scope.login = function() {
     window.cordova.plugins.BEMJWTAuth.signIn().then(function(userEmail) {
+      CommHelper.updateUser({
+        client: "urap-2017-information"
```
shankari: this is control. you don't want to set the client to information.

PR #353, opened 2018-02-17
Auto survey close

jesbu1:


www/js/survey/launch.js
```diff
@@ -64,8 +64,14 @@ angular.module('emission.survey.launch', ['emission.services',
                                                          endFmtTimeElementId,
                                                          startTs,
                                                          endTs) {
+      var options = {
```
shankari: Did this get deleted while resolving a merge conflict or something?

www/js/survey/launch.js
```diff
@@ -64,8 +64,14 @@ angular.module('emission.survey.launch', ['emission.services',
...
+        location: 'no',
+        clearcache: 'no',
+        toolbar: 'yes'
+      };
+
       // THIS LINE FOR inAppBrowser
-      $cordovaInAppBrowser.open(url, '_blank', surveylaunch.options)
+      $cordovaInAppBrowser.open(url, '_blank', options)
```
shankari: why revert this?

www/js/survey/launch.js
```diff
@@ -107,19 +113,23 @@ angular.module('emission.survey.launch', ['emission.services',
...
         console.log("started loading, event = "+JSON.stringify(event));
+
+
+      });
+      $rootScope.$on('$cordovaInAppBrowser:loadstop', function(e, event) {
+        console.log("stopped loading, event = "+JSON.stringify(event));
+        if (String(event.url).includes('formResponse')) {
+            $cordovaInAppBrowser.close();
+        }
+        replace_uuid(uuidElementId);
```
shankari: so you decided to replace the uuid on load stop instead of load start?
Why? Did you test to confirm that it worked?

www/js/survey/launch.js
jesbu1: I think this and the thing above might have just gotten deleted, I can put it back.

www/js/survey/launch.js
jesbu1: Yes, I just decided to move all of the logic to loadstop.
The auto-close is on load stop because when it's in load start the survey actually doesn't get submitted. 

So I just decided to put replace_uuid to load stop also so all modifications to the inAppBrowser function happen after the loading stops (and it is working).

PR #356, opened 2018-02-23
 Fix polar bear rendering, start screen, initial camera positioning

AndyLc:


www/js/bear.js
```diff
@@ -49,13 +49,14 @@ angular.module('emission.main.bear',['nvd3', 'emission.services', 'ionic-datepic
           response.otherBears[key]['size'] = response.otherBears[key]['size'] * scale;
           $scope.otherBears.push(response.otherBears[key]);
         }
+        $state.reload();
```
shankari: Typical fix is to put the updates inside `$scope.$apply`, not reload the page. Have you tested the change?
http://jimhoskins.com/2012/12/17/angularjs-and-apply.html

www/js/intro.js
```diff
@@ -45,7 +45,7 @@ angular.module('emission.intro', ['emission.splash.startprefs',
   };
 
   $scope.disagree = function() {
-    $state.go('root.main.heatmap');
+    $state.go('root.main.bear');
```
shankari: this is actually irrelevant since you don't have the "i disagree" button

shankari:
Have you tested with `reload`? I really think that `$apply` is the recommended option instead.

PR #357, opened 2018-02-23
Urap 2017 control

jesbu1:


www/templates/main.html
```diff
@@ -22,9 +22,9 @@
 
 
   <!-- Dashboard Tab -->
-  <!-- <ion-tab title="Dashboard" icon="ion-ios-analytics" href="#/root/main/metrics">
+  <ion-tab title="Dashboard" icon="ion-ios-analytics" href="#/root/main/metrics">
```
shankari: nit: weren't we going to change title from "Dashboard"?

PR #364, opened 2018-02-26
added new page in intro, and stop sign

jesbu1:


www/templates/intro/intro.html
```diff
@@ -3,6 +3,9 @@
       <ion-slide>
         <ng-include src="'templates/intro/summary.html'"></ng-include>
       </ion-slide>
+      <ion-slide>
+        <ng-include src="'templates/intro/study-text.html'"></ng-include>
```
shankari: should it go before or after the summary?

shankari: I feel like before the summary is better, because other studies may want to have other summaries.
You could change the summary too, to say something like "TripAware is a study ...."
Nobody said that the summary has to be the e-mission summary.

www/templates/intro/study-text.html
```diff
@@ -0,0 +1,20 @@
...
+	<!-- <div class="intro-space"></div> -->
+	<div class="intro-title">
+	<center><ion-icon></center>
+	</div>
+	<!-- <div class="intro-space"></div> -->
+	<div class="intro-text">
+	<center><img src="img/stop.png"/ style="width: 80%; height: auto;"></center>
+	<div class="intro-space"></div>
+	<center><b>If you are joining a study, please return to your study instructions and proceed with
+	the second step.</b></center>
```
shankari: Maybe "and complete the second step before proceeding"?
I know I said that not to include the text about "have completed the second step", but I feel like this small change may sound more like the other step is not to download a new UI but to do something else (maybe survey!) and then come back to the same UI.
I am kind of paranoid about being rejected from app store review...

shankari:
If you can get the changes done by around our noon meeting, I can submit the versions to stores today afternoon. That will allow us to have some time to respond and resolve app store feedback (if any) before Friday.

PR #365, opened 2018-02-26
Urap 2017 control

tsekylekobie:
Made changes based on feedback. Also made the diary the same as the info and emotion versions of the app

### DIFF
--- a/www/js/diary/detail.js
+++ b/www/js/diary/detail.js
@@ -2,7 +2,7 @@
 angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
                                       'nvd3', 'angularLocalStorage',
                                       'emission.services', 'emission.plugin.logger',
-                                      'emission.stats.clientstats'
+                                      'emission.stats.clientstats',
                                       'emission.incident.posttrip.manual'])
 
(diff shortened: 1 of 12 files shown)


www/js/diary/detail.js
```diff
@@ -58,7 +65,9 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
   $scope.allModes = DiaryHelper.allModes;
   $scope.trip = Timeline.getTrip($stateParams.tripId);
   $scope.getKmph = DiaryHelper.getKmph;
-  $scope.getFormattedDistance = DiaryHelper.getFormattedDistance;
+  $scope.getFormattedDistance = function(v) { //converts m to miles
+      return Math.round(v / 1609.34 * 100) / 100;
```
shankari: Shouldn't there be a function for this instead of copy-pasting it everywhere?

www/js/metrics.js
```diff
@@ -1160,11 +1160,25 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
     getData();
   }, 1)
 
+  $scope.sendStats = function() {
```
shankari: why do we need a `sendStats` and why do we record `OPENED_APP` in it?

jesbu1: Forgot to remove this

PR #366, opened 2018-02-26
New background and more

jackcsullivan:


config.xml
```diff
@@ -112,6 +112,5 @@
     </plugin>
     <plugin name="edu.berkeley.eecs.emission.cordova.auth" spec="https://github.com/e-mission/cordova-jwt-auth.git" />
     <preference name="SplashShowOnlyFirstTime" value="false" />
-    <engine name="ios" spec="^4.5.4" />
```
shankari: Why remove the ios engine from `config.xml`?

www/js/metrics.js
```diff
@@ -1030,10 +1030,14 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
       return Math.round(v / 1609.34 * 100) / 100;
     }
 
-    $scope.roundCarbon = function(val) {
+    var roundCarbon = function(val) {
```
shankari: why change `$scope` to `var`?

jesbu1: It's called within metrics.js now (not html), no need to be scope var

PR #412, opened 2018-05-12
Minor, non-user-visible changes to get phonegap serve to work

shankari:
Summary of changes:
- Change index.html so that the phonegap rewrites (https://github.com/phonegap/connect-phonegap) works
- Make native calls only after the platform has initialized
- Handle weird behavior of files loaded from the content sync location


shankari:
Merging this now - will update README after some checking against a fresh repository.

PR #431, opened 2018-06-12
Translated some views to spanish and added purpose and mode selection

sebapastore:
The views translated are the ones in diary, incident and a little part of profile. Added mode and purpose selection inside the list of trips and in the notification after the a trip ends


www/js/intro.js
```diff
@@ -117,7 +117,14 @@ angular.module('emission.intro', ['emission.splash.startprefs',
     // this is not a promise, so we don't need to use .then
     StartPrefs.markIntroDone();
```
shankari: can you also, either here, or in `markIntroDone()`, set the user channel to `megumi-brt`?
Since you are using a shared server, that will let me figure out which users are for your study.
Similar to:
https://github.com/e-mission/e-mission-phone/blob/urap-2017-emotion/www/js/intro.js#L90

shankari: alternatively, you can submit a pull request to master that updates the user client to the stored channel, which is a generic solution and will not require each channel to make this change

www/js/intro.js
```diff
@@ -117,7 +117,14 @@ angular.module('emission.intro', ['emission.splash.startprefs',
...
     $scope.getIntroBox().slide(0);
-    StartPrefs.loadPreferredScreen();
+    // StartPrefs.loadPreferredScreen();
+    $state.go('root.main.diary');
```
shankari: why did you need this change?

www/js/main.js
```diff
@@ -8,7 +8,8 @@ angular.module('emission.main', ['emission.main.recent',
                                  'emission.main.heatmap',
                                  'emission.main.metrics',
                                  'emission.incident.posttrip.map',
-                                 'emission.services'])
+                                 'emission.services',
+                                 'emission.main.survey',])
```
shankari: nit: syntax error?

www/js/survey.js
```diff
@@ -0,0 +1,17 @@
...
+angular.module('emission.main.survey', [])
+
+.controller('SurveyCtrl', function($scope, SurveyLaunch) {
+  $scope.$on('$ionicView.afterEnter', function() {
+    var surveyUrl = 'https://berkeley.qualtrics.com/jfe/form/SV_afnJgB5brT0zCo5';
+    var uuidEelementId = 'QR~QID56';
+    SurveyLaunch.startSurvey(surveyUrl, uuidEelementId);
+  })
+
+  $scope.startSurvey = function () {
```
shankari: not really my problem, but if you are already displaying the survey on `afterEnter`, then why do you need a button?

sebapastore:
I made changes in:
- Disagree return to intro
- Put back loadPreferredScreen()
- Added the code to save the client to the server (Based in the example. Let me know if that is what you asked).
- Delete redundant Survey view

shankari:
You have addressed all the high level issues, but not the smaller ones, like the incident reporting. So I will just deploy whatever you have so far, and you can submit fixes after megumi tests it out.

PR #432, opened 2018-06-21
Bug correction in main.js

sebapastore:
Deleted "emission.main.survey" from main.js because it doesn't exists anymore


shankari:
@sebapastore merged and deployed. Should be live in ~ 5 minutes - try again then!

PR #434, opened 2018-06-21
Add config changes

asiripanich:
Add clientID for android and iOS

### DIFF
--- a/www/templates/intro/consent.html
+++ b/www/templates/intro/consent.html
@@ -1,16 +1,16 @@
 <ion-content>
 <div class="consent-title">
-<center><h4>E-Mission: Data driven carbon emission reduction</h4></center>
+<center><h4>My Cool App: Get suggestions on your commute</h4></center>
 </div>
 <div class="consent-text">
 <h3>Introduction and Purpose</h3>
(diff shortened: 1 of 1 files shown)


www/json/connectionConfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "connectUrl": "https://emission-app.byamarin.com",
+    "android": {
+        "auth": {
+            "method": "google-authutil",
+            "clientID": "1047828012325-nv86fbmhks8h94bdc3bsvid1r48lajor.apps.googleusercontent.com"
```
shankari: In general, you don't want to check in/push the `connectionConfig.json` to the pull request.
If you do, everybody can see the `clientID`, which is bad, because it allows people not associated with you to upload data to your servers.

- if you are using the pre-built native apps and only configuring the UI, you need to get the keys from me. Since I already have the keys in this case, there is no need to check them in
- if you are building your own native app, you don't need to submit a pull request to `e-mission-phone`. You can decide what to do with your fork, but I would strongly recommend against checking in keys, for the reasons listed above.

PR #437, opened 2018-06-25
Changes in main-control and bug correction in select "other" mode or purpose

sebapastore:


shankari:
@sebapastore I will merge this, but none of them appear to be related to the error with marking consent, which seems like the most urgent issue now. Were you able to investigate that further?

PR #440, opened 2018-07-04
Remove heatmap tab and game tab

asiripanich:


shankari:
@asiripanich change should be live in ~ 5 minutes

PR #442, opened 2018-07-04
Pull from master

shankari:
Looks like we dealt with the fallout from merging to master already, and the current set of changes is pretty small. HOWEVER, the merge will revert the CCI changes. So I should either skip the merge, or revert + re-apply.


shankari:
On the one hand, I don't really have to do this.
On the other hand, it is not good for the `cci-berkeley` branch to diverge too much from `master`, and if I don't merge it now, merging it later will just be a giant pain.

Merging it now using the following process:
- download diff
- merge
- remove changes from diff that I want to retain
- apply reversed diff
- commit

shankari:
merged through https://github.com/e-mission/e-mission-phone/commit/a4b76a23b521d0c10ecb5a2811ca0ef023622f5c
Restored changes through https://github.com/e-mission/e-mission-phone/commit/2b251a0e589279cdd67e19ee4857f1592cc214a0

Unsure why the merged commits are not detected automatically.
Will close manually tonight.

shankari:
Ah because there was one commit missing :)

PR #448, opened 2018-07-07
Update login.html

unaissait:
Changed the auth page text as follows :
UC Berkeley -> IISc Bangalore
CalNet ID -> iiscwlan ID
"Login via google" -> "Login using a google account"


shankari:
Closing since we have hired people already.

PR #451, opened 2018-07-10
Changes in the Consent content

sebapastore:


shankari:
@sebapastore deployed, new version should be live in ~ 10 mins

PR #457, opened 2018-08-05
Code deployment for OpenToAll version of E-mission

Andrew-Tan:


Andrew-Tan:
Done resolving conflicts.

Andrew-Tan:
connectionConfig.json:
```json
{
  "connectUrl": "https://emission.open-to-all.com",
  "android": {
    "auth": {
      "method": "openid-authutil",
      "discoveryURI": "https://accounts.open-to-all.com/auth/realms/OpenToAll/",
      "clientID": "emission-mobile-dev",
      "scope": "openid email profile offline_access"
    }
  },
  "ios": {
    "auth": {
      "method": "openid-authutil",
      "discoveryURI": "https://accounts.open-to-all.com/auth/realms/OpenToAll/",
      "clientID": "emission-mobile-dev"
    }
  }
}
```

www/js/diary/list.js
```diff
@@ -39,15 +39,6 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       readAndUpdateForDay($rootScope.barDetailDate);
       $rootScope.barDetail = false;
     };
-    if($rootScope.displayingIncident == true) {
```
shankari: when you are testing, if you see weird issues with page loading, you may want to restore this.

www/js/incident/post-trip-prompt.js
```diff
@@ -150,70 +140,18 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
-                    }
-                });
-            }
-        }).catch(function(error) {
-            $ionicPopup.alert({
-                title: "Error while muting notifications for trip end. Try again later.",
-                template: JSON.stringify(error)
-            });
-        });
+      } else if (data.identifier == 'BIKE') {
```
shankari: what do you want to do if users pick `WALK` or `BIKE`? Currently, you do nothing. Is that what you want?

www/templates/intro/consent.html
```diff
@@ -1,111 +1,174 @@
...
+  terms used in this Privacy Policy have the same meanings as in our Terms and
+  Conditions, accessible at <a
+  href="http://www.accessmap.io">http://www.accessmap.io</a>
+  <br/>
+
+<h3>Information Collection And Use</h3>
+  While using our Service, we may ask you to provide us with certain personally
+  identifiable information that can be used to contact or identify you.
+  Personally identifiable information may include, but is not limited to, your
+  email address, name, phone number, postal address, other information
```
shankari: Shouldn't you add something about collecting location information in the background? Or is it covered by the "not limited to" clause?

www/js/incident/post-trip-prompt.js
Andrew-Tan: I don't remember I've changed this. Might be an issue when doing merging. What should I do to this?

www/templates/intro/consent.html
Andrew-Tan: Got it. Will state it explicitly.

www/js/diary/list.js
Andrew-Tan: I see. Might be an issue created when merging with other branches previously

www/js/incident/post-trip-prompt.js
shankari: @Andrew-Tan what *do* you want to do when the user chooses `'BIKE'` or `'WALK'` from the prompt at the end of the trip? Note that the default prompt says "Snooze" or "Mute". But the CCI group changed it to `'BIKE'` and `'WALK'` because they wanted to have one-click confirmation for the most common types of trips. You probably got this as part of the CCI merge.

You still want to get feedback on the trips, not just allow people to confirm mode. So maybe you don't want to change the buttons, or change them to something else?

You could go back to `'SNOOZE'` and `'MUTE'` and change it later based on the first level of feedback.

Current `'SNOOZE'`/`'MUTE'` functionality is in the master 
https://github.com/e-mission/e-mission-phone/blob/master/www/js/incident/post-trip-prompt.js#L23

shankari: or you could just leave it as is for now. Just tell your initial users that the buttons don't work, and ask for feedback on what they should say.

www/js/incident/post-trip-prompt.js
```diff
@@ -5,39 +5,39 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
+      selMessage = "Tap to rate your experience.";
     }
     return selMessage;
   };
 
   var getTripEndReportNotification = function() {
     var actions = [{
-       identifier: 'MUTE',
-       title: 'Mute',
+       identifier: 'WALK',
```
shankari: nit: you need to revert these as well

www/js/incident/post-trip-prompt.js
```diff
@@ -46,13 +46,14 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
 
     var reportNotifyConfig = {
       id: REPORT,
-      title: "Incident to report?",
+      title: "Did you use AccessMap to get your there?",
```
shankari: nit: spelling/grammar?

www/js/incident/post-trip-prompt.js
```diff
@@ -46,13 +46,14 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
       text: reportMessage(ionic.Platform.platform()),
       icon: 'file://img/icon.png',
       smallIcon: 'res://ic_mood_question.png',
       sound: null,
-      actions: actions,
-      category: REPORT_INCIDENT_TEXT,
+      // actions: actions,
+      actions: [],
+      // category: CHOOSE_MODE_TEXT,
```
shankari: I am not sure this will work without a category. Make sure to test on both iOS and android.
Original change is
https://github.com/e-mission/e-mission-phone/commit/129d4bbe43af63d044f28afe2659aee290e9a125

config.cordovabuild.xml
```diff
@@ -112,6 +112,8 @@
     </plugin>
     <plugin name="edu.berkeley.eecs.emission.cordova.auth" spec="https://github.com/e-mission/cordova-jwt-auth.git" />
     <preference name="SplashShowOnlyFirstTime" value="false" />
+    <allow-navigation href="https://accessmap.io/*" />
+    <allow-navigation href="https://emission-accessmap.open-to-all.com/*" />
```
shankari: I hope this is not required because `config.xml` is part of the native code and using channels can only apply native code changes.

shankari:
Just deployed, should be active in ~ 10 minutes.
Follow the instructions at
https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=opentoall&clear_usercache=true&clear_local_storage=true
to test

PR #462, opened 2018-08-21
Updated readme with e-mission-docs

jackcsullivan:
Updated old wiki link and added wiki section at end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,11 @@ e-mission phone app
 
 This is the phone component of the e-mission system.
 
+Additional Documentation
+---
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -298,3 +298,8 @@ $ git pull upstream master
 $ git push origin master
 $ git branch -d mybranch
 ```
+
+## Wiki: ##
+----------
+All wikis have been moved to their own repository [e-mission-docs](https://github.com/e-mission/e-mission-docs). E-mission phone specific wikis can be found here:
```
shankari: Should this be at the end or at the beginning?
Also, maybe instead of "all wikis", say "additional documentation", since the new repository is not really a wiki any more :)

PR #463, opened 2018-09-05
Slight UI tweak and bugs fixes

Andrew-Tan:
Some bug fix and UI tweak.

For radio buttons, the long text's tail becomes "...", and I'm still not able to find a way to do word wrap...


shankari:
LGTM!

shankari:
Deployed change to the `opentoall` channel; should be live in ~ 20 mins.

PR #467, opened 2018-09-24
Uiwork

trevor-wu:
- updated consent.html to fix formatting errors
- uploaded configuration.json which points to Emma's server


www/json/connectionConfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "connectUrl": "http://52.91.65.149:8080",
+    "android": {
+        "auth": {
+            "method": "token_list",
```
shankari: there is no `token_list` method supported on the phone. See https://github.com/e-mission/e-mission-phone/pull/465 for examples

www/json/connectionConfig.json
```diff
@@ -0,0 +1,15 @@
...
+            "clientID": "8QU4GCAZ"
```
shankari: dunno why the `clientID` is set, where you got the value from and what it is supposed to mean

www/json/connectionConfig.json
```diff
@@ -2,14 +2,14 @@
     "connectUrl": "http://52.91.65.149:8080",
     "android": {
         "auth": {
-            "method": "token_list",
-            "clientID": "8QU4GCAZ"
+            "method": "prompted-auth",
+            "prompt": "8QU4GCAZ"
```
shankari: why is your prompt `8QU4GCAZ`? Which language is that in?

PR #469, opened 2018-09-25
Megumi brt ground truth

sebapastore:
This in an attempt to improve Ground Truth based on https://github.com/e-mission/e-mission-phone/pull/308.

Despite the attempt, we belive that the UX isn't satisfactory, and I belive that the approach must be rethought. Anyway, the code to the comunity if is usefull for someone.


shankari:
@sebapastore thank you for the contribution! Do you have thoughts on a better approach for the UX that you can also add for the record? Does the current UX actually work?

Issue #443, opened 2018-07-05
Errors prompt re-login


asiripanich:
After a few hours of leaving the app on the background on my phone, I received these errors when I launched the app to check. It went back to the login step which I had done when I first installed the app.  

Any idea what might be the cause of these errors?
![image001](https://user-images.githubusercontent.com/17020181/42298743-d6053fe4-804a-11e8-92ad-9bd98c861ac9.png)

![image002](https://user-images.githubusercontent.com/17020181/42298813-3224c9d4-804b-11e8-80af-3774d939a9ed.png)

shankari:
@asiripanich, can you upload your logs too? (Profile -> Email log). You can email the log to yourself and upload it here, or you can email directly to me for greater privacy.

shankari:
Others have reported this issue before, but they have not been developers, so I have not been able to get the logs in time.

Current speculation is that this is happens because of apps that are "offloaded" when you are short of space.
www.iphonehacks.com/2017/08/how-to-automatically-remove-unused-apps-free-up-storage-space-iphone-ipad-ios-11.html

So can you comment on how much space you have available? Also the logs will help me debug exactly what is not getting restored when the app is restored again.

asiripanich:
@shankari 

Sent my log file. 

The offload option has never been enable on my phone (iOS 11.4, iPhoneX). So that might not be the cause for my case. 

Space available on my phone: 54.4 GB of 64 GB used.

shankari:
A similar issue was reported by @manuelsantanap while testing on emTripLog.
Looking at his logs, it looks like the underlying issue was with sharedPreferences not being stored properly (https://github.com/e-mission/e-mission-base/issues/3#issuecomment-411586627). This is an issue with the ionic deploy plugin.

Let's use this issue to track that fix

shankari:
I quickly looked through these logs to see if this was the same issue, and unfortunately, there are no such logs.

```
$ grep "A newer version" /tmp/loggerDB.amarin.unable_to_login.withdate.log | wc -l
       0
```

However, this is a not a debug build, so we really will not expect it to be.

shankari:
[...]

PR #470, opened 2018-09-25
Fix client reset on start

shankari:
Main change is to switch to native storage, consistent with https://github.com/e-mission/e-mission-phone/issues/443#issuecomment-423066729

Bonus fixes:
- Move native code in other screens to after platform.ready
- Clarify the prompted-auth example

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -677,56 +677,7 @@ background-image: none !important; }
   margin: 1rem;
   position: relative;
 }
-/* Dark theme */
-/*
-.tab-nav{
-  background-color: #212121 !important; background-size: 0 !important;
-}
-ion-header-bar {
-  background-color: #212121 !important;
-}
-.ion-view-background{
-  background-color: #303030 !important;
-}
-.tabs-custom-dark > .tabs,
-.tabs.tabs-custom-dark {
-  border-color: #5D3A23;
-  background-color: #5D3A23;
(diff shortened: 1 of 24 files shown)


shankari:
Dark theme is now gone for good

```
$ grep -r dark www/js
C02KT61MFFT0:e-mission-phone shankari$ grep -r dark www/templates/
www/templates//common/place-list.html:<!--   <ion-nav-title class="dark-color">
www/templates//common/trip-list.html:<!--   <ion-nav-title class="dark-color">
www/templates//diary/list.html:        <div class="buttons dark-color" style="text-align:center; background-color: transparent !important;" id="toget">
```

shankari:
Other modules also stored data in local storage. Moved almost all of them to native storage.
Exceptions:
- referral, because that is essentially session storage and can be recreated by re-clicking the invite link
- walkthrough, because the consequences if it occasionally gets deleted are pretty minor

I also tried to move out the walkthrough into its own module for clarity but that seems to subtly interfere with the timing, so moved it back.

shankari:
Future improvements may include:
- moving everything to native storage
- really refactoring stuff properly into services and factories - e.g. the display code for metrics, the walkthrough code, etc
- making sure that all startup code is bunched properly and invoked when the platform is ready instead of being strewn all over the place in the body of the controller.

PR #473, opened 2018-10-10
UI Changes as of 10/9

sam11b:
Can now scroll horizontally through different recommendations on Recommend tab
Background is now same color as other parts of the app
Button color has changed from green to white/same color as other parts of app
Making yelp star image a little smaller in an effort so that it doesn't bug out on phones
Bolded certain texts

PR #474, opened 2018-10-10
Greentrip UI Work

trevor-wu:
-Contains Samantha's changes which include:
-color changes
-scrolling functionality
-bug fixes


shankari:
This is identical to #473 
3 merge conflicts, extra added files etc.
Are you looking at the pull requests before you create them? :)

sam11b:
I committed the merge conflict changes so there should be none left. Sorry about that.

sam11b:
this should also be from a new separate branch.

shankari:
Ok so the "files changed" is now fine and the merge conflicts are gone, but the old commits still show up.

Recall that I squashed the changes while merging last time, so directly merging from `gtui` to a new branch doesn't really fix the issues. Your changes are pretty small, why not just apply a patch to a brand new branch checked out from master?

I really don't want all the reverts (e.g. `corrected '.shxowChart' to '.showChart'` into the main branch).

shankari:
I'd be open to squashing commits and merging again.
But then you can't revert individual commits from this branch. I am not sure if that is a problem for you.

Next time, my advice is again to **start work** on a new branch. It will make your life much easier.

sam11b:
Trevor and I talked and I don't think there should be any reason that we would need to go back to individual commits. Next time we will work off a new branch off of greentrip and not gtui. Apologies.

shankari:
Deployed to the channel. You should get the update within ~ 10 mins

PR #475, opened 2018-10-11
UI wasn't merged correctly

sam11b:
Doubled text eliminated


shankari:
Just deployed to the channel.
Should be live in ~ 10 mins

PR #490, opened 2018-10-29
JSON pull changes, still requires backend integration

sam11b:
Added to recommendations.js JSON pull request. Upon Vanessa's integration, should display a recommendation in the first recommendation box. Otherwise, sets to previous sample ("Cafe and Beverages"). Eventually, this will use a for loop for every recommendation given by Vanessa's backend. Takes away Trevor's HTML sample request.

### DIFF
--- a/www/js/recommendations.js
+++ b/www/js/recommendations.js
@@ -32,6 +32,7 @@ Ask Shankari about this Wednesday
       $scope.currentDisplay.style.display = "block";
     };
 
+    //waiting on Vanessa integration 
     $scope.getSuggestion = function() {
       return new Promise(function(resolve, reject) {
         var msgFiller = function(message) {
--- a/www/js/recommendations.js
+++ b/www/js/recommendations.js
@@ -19,6 +19,19 @@ Ask Shankari about this Wednesday
 
 .controller('RecommendationsCtrl', function(CommHelper, $state, $ionicLoading, $scope, $rootScope, $ionicModal, nzTour,
                                 $window, $http, $ionicPopup, $timeout, storage, ReferralHandler, ReferHelper, Logger, $cordovaInAppBrowser, SurveyLaunch) {
+
+    $scope.currentDisplay = "";
+    //Check and see if this is right with Vanessa
+    //x = window.cordova.plugins.BEMServerComm.pushGetJSON("/suggestion_sys", msgFiller, resolve, reject);
+    //for now using sample
+    $http.get('json/sampleRecommendation.json').then(function(result) {
+        var name = JSON.stringify(result.data.name);
+        var mode = JSON.stringify(result.data.mode);
+        document.getElementById("message_1").innerHTML = name.replace(/\"/g, "");
+        document.getElementById("mode_1").innerHTML = mode.replace(/\"/g, "");
+      }
+    )
+
     $scope.clickDirections = function(id) {
       $scope.currentDisplay = document.getElementById("mapModal" + id);
       $scope.currentDisplay.style.display = "block";
@@ -30,17 +43,6 @@ Ask Shankari about this Wednesday
     $scope.clickReview = function(id) {
       $scope.currentDisplay = document.getElementById("reviewModal" + id);
       $scope.currentDisplay.style.display = "block";
-    };
-
-    //waiting on Vanessa integration 
-    $scope.getSuggestion = function() {
-      return new Promise(function(resolve, reject) {
-        var msgFiller = function(message) {
-          console.log("THIS IS THE SUGGESTION" + JSON.stringify(message));
-          document.getElementById("0").innerHTML = JSON.stringify(message);
-        }
-        window.cordova.plugins.BEMServerComm.pushGetJSON("/suggestion_sys", msgFiller, resolve, reject);
-      });
     };
 
 });
--- a/www/js/recommendations.js
+++ b/www/js/recommendations.js
@@ -6,15 +6,19 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
                 'emission.splash.referral', 'angularLocalStorage',
                 'ng-walkthrough', 'nzTour'])
 
+/*
+Ask Shankari about this Wednesday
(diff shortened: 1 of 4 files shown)


www/js/recommendations.js
```diff
@@ -6,15 +6,19 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
                 'emission.splash.referral', 'angularLocalStorage',
                 'ng-walkthrough', 'nzTour'])
 
+/*
+Ask Shankari about this Wednesday
```
shankari: why wait until Wednesday!?

www/js/recommendations.js
```diff
@@ -6,15 +6,19 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
...
+.service('CommHelper', function($http) {
```
shankari: `CommHelper` is already defined in `services.js`. Unsure why you think you need to redefine it here?

www/js/recommendations.js
shankari: not every discussion needs to happen in person. In fact, for a lot of open source projects, no discussions will happen in person.

www/js/recommendations.js
```diff
@@ -27,13 +31,16 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
...
-    }
+
+    //waiting on Vanessa integration 
+    $scope.getSuggestion = function() {
+      return new Promise(function(resolve, reject) {
+        var msgFiller = function(message) {
+          console.log("THIS IS THE SUGGESTION" + JSON.stringify(message));
+          document.getElementById("0").innerHTML = JSON.stringify(message);
+        }
+        window.cordova.plugins.BEMServerComm.pushGetJSON("/suggestion_sys", msgFiller, resolve, reject);
```
shankari: The general design pattern is to define this in `CommHelper` so that we have a list of all the remote calls in one place.

shankari:
```
$ grep -rl pushGetJSON www/js
www/js/services.js
```

www/js/recommendations.js
```diff
@@ -27,13 +31,16 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
...
-      var y2 = Math.pow(yDif, 2);
-      var dist = Math.sqrt(x2 + y2);
-      return Math.round(dist * 10) / 10 + " miles away";
-    }
+
+    //waiting on Vanessa integration 
+    $scope.getSuggestion = function() {
+      return new Promise(function(resolve, reject) {
+        var msgFiller = function(message) {
+          console.log("THIS IS THE SUGGESTION" + JSON.stringify(message));
```
shankari: Not sure what this is doing. `msgFiller` is supposed to fill in the message to send *to* the server. Why would you send the suggestion to the server?

www/templates/main-recommendations.html
```diff
@@ -8,22 +7,23 @@
      <link rel="stylesheet" href="css/main-recommend.css">
      <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.3.1/css/all.css">
      <link rel="stylesheet" type="text/css" href="slick/slick.css"/>
-  <ion-content class="has-header" overflow-scroll="true" padding="false" ng-controller="RecommendationsCtrl">
```
shankari: We generally do this using states.

www/templates/main-recommendations.html
```diff
@@ -8,22 +7,23 @@
...
        <h4>
-       <p> {{recs.name}} </p>
-         <p> {{recs.address}} </p>
-           <p> {{getDistance(recs.coord[0], recs.coord[1])}}</p>
-           <img src="img/small/small_{{recs.stars}}.png" widsth="75" />
+
+       <p> Ernie's Espresso </p>
+         <p>1234 Sesame St. </p>
+           <p> 0.2 Miles Away </p>
+           <img src="img/small/small_4.png" width="75" />
```
shankari: I don't see how the suggestion will show up correctly from Vanessa. You just removed all the template code. How will you format the received suggestion?

sam11b:
Added html template back. Vanessa told me that currently the recommendations only include a message and a mode so that has been added to the template as well. Will uncomment and use that JSON rather than the sampleRecommendations.json upon full integration.

www/js/recommendations.js
```diff
@@ -19,6 +19,19 @@ Ask Shankari about this Wednesday
...
                                 $window, $http, $ionicPopup, $timeout, storage, ReferralHandler, ReferHelper, Logger, $cordovaInAppBrowser, SurveyLaunch) {
+
+    $scope.currentDisplay = "";
+    //Check and see if this is right with Vanessa
+    //x = window.cordova.plugins.BEMServerComm.pushGetJSON("/suggestion_sys", msgFiller, resolve, reject);
+    //for now using sample
+    $http.get('json/sampleRecommendation.json').then(function(result) {
+        var name = JSON.stringify(result.data.name);
+        var mode = JSON.stringify(result.data.mode);
+        document.getElementById("message_1").innerHTML = name.replace(/\"/g, "");
```
shankari: This is the wrong way to set HTML values in angular.
It is slow and ugly. @trevorwu's approach, with the `{{}}` is the right approach.

www/js/recommendations.js
```diff
@@ -19,6 +19,19 @@ Ask Shankari about this Wednesday
 
 .controller('RecommendationsCtrl', function(CommHelper, $state, $ionicLoading, $scope, $rootScope, $ionicModal, nzTour,
                                 $window, $http, $ionicPopup, $timeout, storage, ReferralHandler, ReferHelper, Logger, $cordovaInAppBrowser, SurveyLaunch) {
+
+    $scope.currentDisplay = "";
+    //Check and see if this is right with Vanessa
+    //x = window.cordova.plugins.BEMServerComm.pushGetJSON("/suggestion_sys", msgFiller, resolve, reject);
```
shankari: when is the integration going to get done?!

PR #494, opened 2018-10-31
Update rk-unsw branch

asiripanich:
We are trying to see how the end-trip surveys work on our deployed version.


asiripanich:
@atton16  could you help me resolve the conflicts? Thanks.

www/js/app.js
```diff
@@ -64,6 +61,87 @@ angular.module('emission', ['ionic',
             Logger.log("error "+JSON.stringify(err)+" reading or setting defaults, giving up");
         });
     });
+
+    Logger.log("Setting collection settings to default");
+    var androidDefaultCollectionConfig = {
```
shankari: You probably don't want to merge this change. It is not required, and can cause an issue on iOS if people click on the customization link twice without going through onboarding (https://github.com/e-mission/e-mission-phone/commit/df672bd7433d87893d1b8c08373e0d63d3a0e0b5). I removed it from the `opentoall` branch, will also remove from `cci-berkeley`

www/js/diary/list.js
```diff
@@ -37,15 +40,6 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       readAndUpdateForDay($rootScope.barDetailDate);
       $rootScope.barDetail = false;
     };
-    if($rootScope.displayingIncident == true) {
-      if (angular.isDefined(Timeline.data.currDay)) {
-          // page was already loaded, reload it automatically
-          readAndUpdateForDay(Timeline.data.currDay);
```
shankari: I will merge this for now, but note that you may want to restore it after testing since the trip end surveys were removed during the pilot in `cci-berkeley`

www/js/incident/post-trip-prompt.js
```diff
@@ -121,99 +121,34 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
-                });
-            }
-        }).catch(function(error) {
-            $ionicPopup.alert({
-                title: "Error while muting notifications for trip end. Try again later.",
-                template: JSON.stringify(error)
-            });
-        });
+      } else if (data.identifier == 'BIKE') {
+        // store mode here
```
shankari: you should store `WALK` and `BIKE` `mode_confirm` objects here

www/js/incident/post-trip-prompt.js
```diff
@@ -224,39 +159,46 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
 
+  /*
+   * Unregister the post-trip notification for cci-berkeley
+   * because they want to make it opt-in. Since the base app registers by
+   * default, we need to explicitly unregister, not just comment out.
+   */
   $ionicPlatform.ready().then(function() {
-    ptap.registerTripEnd();
-    ptap.registerUserResponse();
+    ptap.unregisterTripEnd();
```
shankari: this is why you don't want to merge blindly :) Read the docs!

www/js/app.js
atton16: Do you mean the `androidDefaultCollectionConfig` part or `iosDefaultCollectionConfig` or both?

www/js/incident/post-trip-prompt.js
atton16: I think we can leave this one for now. Also, I am not sure how to store mode. I cannot spot codes related to mode storing.

www/js/diary/list.js
atton16: I will revert this one back to what it was. Thank you.

www/js/incident/post-trip-prompt.js
atton16: I will also revert this one back to what it was. Thank you.

www/js/incident/post-trip-prompt.js
shankari: mode is stored in the popup - look at `$scope.storeMode` for example

www/js/incident/post-trip-prompt.js
```diff
@@ -224,36 +159,43 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
           return;
       }
       cleanDataIfNecessary(notification, state, data);
       displayCompletedTrip(notification, state, data);
     });
-  }
+  };
 
+  /*
+   * Unregister the post-trip notification for cci-berkeley
```
shankari: nit: remove the comment as well since it is no longer applicable.

shankari:

PR #497, opened 2018-11-13
UI changes so that there is only one recommend box and the update button matches the one on the Big Data tab

sam11b:
The angular adaptation still needs to be completed. The recommendation now matches the style of the Big Data tab and has an update button in the corner. There is no longer a direction button but reviews can still be pulled up


www/js/recommendations.js
```diff
@@ -46,12 +27,12 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
 
     $scope.clickSuggestion = function(id) {
```
shankari: Duplicate code in`CommHelper.getSuggestion()` - one in `clickSuggestion` and one in general at controller load? You should really consolidate to one - copy-pasting code is generally bad because you tend to not keep the two copies in sync. For example, you added the error handling to this call, but forgot to add it to the other one

sam11b:
took out the duplicate function written in the controller load.

shankari:

PR #498, opened 2018-11-15
made code more angular

trevor-wu:
made code more angular, changes from last week merged with Sam's UI change

PR #499, opened 2018-11-15
Made Language from #498 more professional

trevor-wu:
Angularization and merger with Sam's UI from #498 also included.


shankari:

PR #500, opened 2018-11-15
removed JSON.stringify in recommendations.js

sam11b:
Recommendations will no longer have double quotes surrounding them


shankari:

PR #506, opened 2018-12-24
Changes made in mode and purpose

deepalics0044:


www/json/connectionConfig.physical_device2localhost.json.sample
```diff
@@ -2,13 +2,13 @@
     "connectUrl": "http://<your laptop ip>:8080",
     "android": {
         "auth": {
-            "method": "prompted-auth",
+            "method": "dummy-dev",
```
shankari: Intentional?

www/json/connectionConfig.physical_device2localhost.json.sample
```diff
@@ -2,13 +2,13 @@
...
             "clientID": "ignored"
         }
     },
     "ios": {
         "auth": {
-            "method": "prompted-auth",
+            "method": "dummy-dev",
```
shankari: Intentional?

shankari:
@deepalics0044 is there going to be an update to this PR?

deepalics0044:
Yes ma'am

shankari:
@deepalics0044 you still have changes in 4 files, including the deletion of `www/json/connectionConfig.openid.json.sample` and ` www/json/connectionConfig.token_list.json.sample`. Those are primarily sample files, so theoretically I can merge them, but I am concerned that having unrelated changes will cause you to have a merge conflict the next time you pull from master.

Is there a reason you did not revert those changes either?

deepalics0044:
Can you please merge the changes, we need it fast.

shankari:
@deepalics0044 ok, but I would strongly recommend fixing the deletes so that you don't run into further problems down the road.

shankari:

PR #508, opened 2019-01-02
Editable mode & purpose confirm

atton16:


shankari:
> The reason that we did not use the command you gave me is that our programmer reported that the command not found so we will investigate into that matter after this.

https://github.com/e-mission/e-mission-phone/blob/master/www/js/services.js#L232

shankari:
> We knows this limitation and we are working on it.

ok, I am merging this for now. @asiripanich, bewarned while testing...

shankari:
@atton16 @asiripanich deployed, should be live in ~ 10-20 mins

asiripanich:
@shankari  Thanks heaps!

PR #511, opened 2019-01-09
Xcode10 workaround

jf87:
Note: compile via cordova CLI still fails, but at least it compiles in Xcode.


shankari:
Some of the changes are really minor formatting changes, let's squash before merging. Thanks to @jf87 for contributing.

PR #512, opened 2019-01-09
New Individual Trip Function and Other Changes

sam11b:
New individual trip function in place for implementation. Other minor changes include "Distanz" -> "Distance", ionic loading appears after the refresh button on the recommend page is clicked, grid layout display for individual suggestions on the details page

### DIFF
--- a/www/js/diary/detail.js
+++ b/www/js/diary/detail.js
@@ -4,7 +4,7 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
                                       'emission.services', 'emission.plugin.logger',
                                       'emission.incident.posttrip.manual'])
 
-.controller("DiaryDetailCtrl", function($scope, $rootScope, $window, $stateParams, $ionicActionSheet,
+.controller("DiaryDetailCtrl", function($scope, $rootScope, $window, $stateParams, $ionicActionSheet, $ionicLoading,
                                         leafletData, leafletMapEvents, nzTour, KVStore,
                                         Logger, Timeline, DiaryHelper, Config,
                                         CommHelper, PostTripManualMarker) {
@@ -47,6 +47,25 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
       $scope.$broadcast('invalidateSize');
   };
 
+  $scope.getIndividualSuggestion = function() {
+    var tripId = $stateParams.tripId;
+    console.log(tripId);
+    var uuID = $stateParams.uuID;
+    console.log(uuID);
+    //TO DO: call function to get individual suggestion depending on tripID
+    $ionicLoading.show({
+        template: 'Loading...'
+        });
+    CommHelper.getIndividualSuggestion().then(function(result) {
+      console.log(result);
+      $ionicLoading.hide();
+      $scope.name = result.message;
+      $scope.mode = result.method;
+    }).catch(function(err) {
+      console.log("Error while getting individual suggestion" + err);
+    });
+  };
+
   $scope.getFormattedDate = DiaryHelper.getFormattedDate;
   $scope.arrowColor = DiaryHelper.arrowColor;
   $scope.parseEarlierOrLater = DiaryHelper.parseEarlierOrLater;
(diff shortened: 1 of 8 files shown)


www/templates/main-recommendations.html
```diff
@@ -34,6 +35,15 @@ <h3>Reviews</h3>
          <button class="button button-block button-balanced button-outline back" ng-click="clickX()">Back</button>
      </div>
    </div>
+   <div id="tripSelection" class="modal">
+     <div class="modal-content">
+       <div align="left">
+       <h3>Past Cleaned Trips</h3>
+       Cleaned trips will be pulled and shown here.
```
shankari: is there a reason why this can't be integrated directly with the diary? Is it the distinction of cleaned versus draft trips or ???

shankari: because it looks like you already have a button in the detail

sam11b:
Sorry that was an old version of main-recommendations.html that I had added accidentally. I just changed the file to be the newer version. You are right, the past clean trips testing stuff is covered in the new get suggestions button on each detail page of each trip. Apologies.

www/js/services.js
```diff
@@ -42,6 +42,16 @@ angular.module('emission.services', ['emission.plugin.logger'])
       });
     };
 
+    this.getSpecificSuggestion = function() {
+      return new Promise(function(resolve, reject) {
+        var msgFiller = function(mesage) {
+          // donothing
+        }
+        window.cordova.plugins.BEMServerComm.pushGetJSON("/suggestion_sys", msgFiller, resolve, reject);
+        // add trip locations where Vanessa releases them
```
shankari: will this be filled in after the server starts working?

www/js/diary/detail.js
```diff
@@ -47,6 +47,25 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
       $scope.$broadcast('invalidateSize');
   };
 
+  $scope.getIndividualSuggestion = function() {
+    var tripId = $stateParams.tripId;
+    console.log(tripId);
+    var uuID = $stateParams.uuID;
```
shankari: are you filling this in somewhere? in general, the uuid does not need to be specified, it is implicit in the authentication used

shankari: and you need to make sure that variables get put into the stateParams. stateParams is like scope, it is just a basic hash that you need to populate with your keys and values.

sam11b:
Sorry, the old stateParam variables were only being logged on the console to make sure the information was available for the new function Vanessa wrote. The comment about trip releases was outdated. Is there something specific you want from a "test update". Vanessa should be releasing a new pull request soon with debugging changes.

shankari:
> Is there something specific you want from a "test update"? 

In my very first comment, I asked for "Can you please add some details of the testing done?"
but I haven't seen any test reports.

Have you tested it? 
- if you have, please indicate testing done
- if you have not, please do the testing and provide a report
- if you want to postpone the testing for some reason, please justify it

Basically, code review is for style/consistency, not correctness. I am not a computer :)

You still need to do actual testing to ensure that the code works as expected, and you should document that testing so that when things break later, you can go back and see what you did and did not test.

sam11b:
I added a text file explaining the testing we are currently doing and changes that have been made. Let me know if it needs to be more extensive.

testing_update
```diff
@@ -0,0 +1,5 @@
+Testing Update as of 1/19/2019
+
+Some people got recommendations but for those who weren't located in large cities, the API call resulted in a key value error. This was combatted by integrating the ProPublica API (allowing for unlimited calls using zip codes). The key value error was due to towns or a call resulting in only a zip code and not an actual city. For this reason, there was an error in giving back a suggestion. If trips are not cleaned, no suggestion will appear until the trip has been cleaned and a mode of transportation is associated with it. The suggestion system has been updated to guesstimate endpoints, since in some areas it can be difficult to pinpoint a restaurant (malls, parking lots, etc).
```
shankari: this is pretty close to what I was expecting.
Maybe with steps like:
- trip to coffee shop in Milpitas; suggestion generated to ....
- trip to friend's home in OC suburb; suggestion generated although it shouldn't be
....

shankari: also, have you tested this end to end? If the testing is happening in an ipython notebook, then how is the UI talking to the changed server?

sam11b: Yes sorry. To clarify, end-to-end its displaying properly unless there is an error on the backend, in which case the screen simply displays "loading" continuously. If that is the case, we specifically have been debugging those individual trips in the ipython notebook. Once the debugging is completed, the trips seem to show up normally. I added some sample tests and suggestions.

shankari:
@sam11b ok, will merge this at the same time as https://github.com/e-mission/e-mission-server/pull/625
Or does it make sense to merge sooner?

sam11b:
I think it would be easier to merge this first so that Vanessa can make sure everything is aligned. Thanks!

PR #513, opened 2019-01-10
Switch to the old build system so that cordova can continue to work

shankari:
This is the e-mission-phone equivalent of
https://github.com/e-mission/e-mission-base/pull/13

With this change, I still cannot build from the command line - the build
terminates after

```
Building project: /Users/shankari/e-mission/e-mission-phone/platforms/ios/emission.xcworkspace
	Configuration: Debug
	Platform: emulator
Build settings from command line:
...
i    PODS_BUILD_DIR = $BUILD_DIR
    PODS_CONFIGURATION_BUILD_DIR = $PODS_BUILD_DIR/$(CONFIGURATION)$(EFFECTIVE_PLATFORM_NAME)
    PODS_PODFILE_DIR_PATH = ${SRCROOT}/.
    PODS_ROOT = ${SRCROOT}/Pods
    SWIFT_OBJC_BRIDGING_HEADER = $(PROJECT_DIR)/$(PROJECT_NAME)/Bridging-Header.h

(emission) C02KT61MFFT0:e-mission-phone shankari$ ^Ap^An
```

I am not quite sure why since the base app builds fine. But this does build
properly using the Xcode UI. Will experiment with this some more, otherwise
just document in the `README` and move on

### DIFF
--- a/build.json
+++ b/build.json
@@ -1,0 +1,9 @@
+{
+  "ios": {
+    "release": {
+      "buildFlag": [
+        "-UseModernBuildSystem=0"
+      ]
+    }
+  }
+}


shankari:
@jf87 I am hoping this can replace #511 
I **think** this worked for me before, but *only on this repo* the command line still doesn't work
xcode UI works fine 
I will compare the settings for this repo and the working repos to see if I can figure anything out

jf87:
Yes, this is basically equivalent to providing the build flag inline.
I just had another related issue.
I needed to change the configuration in xcode a well in order to deploy to the actual phone:
https://stackoverflow.com/questions/52424462/xcode-10-a-valid-provisioning-profile-for-this-executable-was-not-found

shankari:
I don't recall having run into the issue with the provisioning profile; there is a comment that claims that it only happens on Mojave, not High Sierra (https://stackoverflow.com/a/53931732/4040267). I haven't yet updated to Mojave. So are you saying you had to do https://stackoverflow.com/a/52528662/4040267?

I just opened one of the projects in xcode and it already has the build system set to "Legacy". Doesn't running the build with legacy system flag already set that value? Otherwise, presumably even the build won't work...

jf87:
Yes, I am on Mojave. I needed to change additionally in Xcode settings like in the stackoverflow suggestions.

shankari:
ok so I narrowed down the issue in 

> With this change, I still cannot build from the command line - the build terminates after

to an inability to correctly parse the `build.json`

In particular, the behavior that I observe is:
- `cordova build ios`

```
Reading build config file:
Building project: ....xcworkspace
Build settings from configuration file '...build-debug.xcconfig':
<dies>
```

- `cordova build ios --buildConfig build.json`

```
Reading build config file:
Building project: ....xcworkspace
Build settings from configuration file '...build-debug.xcconfig':
<dies>
```
- `cordova build ios --buildFlag="-UseModernBuildSystem=0"`

```
Reading build config file:
Building project: .../emission.xcworkspace
Adding xcodebuildArg: -UseModernBuildSystem=0
User defaults from command line:
    UseModernBuildSystem = 0
Build settings from command line:
    CONFIGURATION_BUILD_DIR = ...
    SDKROOT = ...
    SHARED_PRECOMPS_DIR = ...

Build settings from configuration file '...build-debug.xcconfig':
...
Prepare build
note: Using legacy build system

=== BUILD TARGET CordovaLib OF PROJECT CordovaLib WITH CONFIGURATION Debug ===
=== BUILD TARGET emission OF PROJECT emission WITH CONFIGURATION Debug ===
** BUILD SUCCEEDED **
```

But the `build.json` is exactly the same one as in the base, and it works fine there

```
$ diff e-mission-base/build.json e-mission-phone/build.json
```

I am a bit out of my depth here; filing an issue on `cordova-ios`. In the meanwhile, I will document the command line as a workaround (e.g. merge a modified version of #511

shankari:
Note that https://github.com/e-mission/e-mission-phone/pull/513#issuecomment-457483071 is true even for a freshly checked out build.

shankari:
Merging this to make the `cordova-ios` issue easier to report.

PR #516, opened 2019-01-12
Greentrip ui16

trevor-wu:
- Added API calls to yelp
- more testing needed


shankari:
I don't see any obvious correctness issues, but note that I am not a computer, so you should really test it to make sure that it works. Did you try it at all? Does it work? In general, you submit tested code for code review with a note saying how you tested it, and I review it for style, consistency, gotchas etc.

On that note, I do see a small performance issue that you might want to address at some point. You currently read the API keys from the json file on every request.
- If you anticipate making a lot of requests each time the app is launched, you want probably want to read the API keys when the page is loaded and then reuse them
- if you anticipate only one or two calls which may or may not happen, then your current structure is correct because you only read the data when you need it -> no wasted effort

PR #517, opened 2019-01-13
Rk unsw

atton16:
Hi we have made some changes as follows:
- Fixed bugs related to mode/purpose confirm mechanics
- Get back Dashboard tab
- Disable Trip-end Prompt
- Display only analyzed trips on Diary tab

### DIFF
--- a/www/js/controllers.js
+++ b/www/js/controllers.js
@@ -5,8 +5,7 @@ angular.module('emission.controllers', ['emission.splash.updatecheck',
                                         'emission.splash.pushnotify',
                                         'emission.splash.localnotify',
                                         'emission.survey.launch',
-                                        'emission.stats.clientstats',
-                                        'emission.incident.posttrip.prompt'])
+                                        'emission.stats.clientstats',])
 
 .controller('RootCtrl', function($scope) {})
 
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -172,12 +172,16 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       return $window.cordova.plugins.BEMUserCache.getAllMessages(MODE_CONFIRM_KEY, false).then(function(modes) {
         Logger.log("Modes stored locally" + JSON.stringify(modes));
         var tripMode = {};
+        var found = false;
         if(modes.length > 0) {
           modes.forEach(function(mode) {
             if ((trip.properties.start_ts == mode.start_ts) &&
                  (trip.properties.end_ts == mode.end_ts)) {
-              tripMode = mode;
-              Logger.log("trip" + JSON.stringify(trip)+ "mode" + JSON.stringify(tripMode));
+              if (!found) {
+                tripMode = mode;
+                Logger.log("trip" + JSON.stringify(trip)+ "mode" + JSON.stringify(tripMode));
+                found = true;
+              }
             }
           });
         }
@@ -189,12 +193,16 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       return $window.cordova.plugins.BEMUserCache.getAllMessages(PURPOSE_CONFIRM_KEY, false).then(function(purposeList) {
         Logger.log("Purpose stored locally" + JSON.stringify(purposeList));
         var tripPurpose = {};
+        var found = false;
         if(purposeList.length > 0) {
           purposeList.forEach(function(purpose) {
             if ((trip.properties.start_ts == purpose.start_ts) &&
                  (trip.properties.end_ts == purpose.end_ts)) {
-              tripPurpose = purpose;
-              Logger.log("trip" + JSON.stringify(trip) + "purpose" + JSON.stringify(tripPurpose));
+                  if (!found) {
+                    tripPurpose = purpose;
+                    Logger.log("trip" + JSON.stringify(trip) + "purpose" + JSON.stringify(tripPurpose));
+                    found = true;
+                  }
             }
           });
         } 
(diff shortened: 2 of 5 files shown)


www/js/controllers.js
```diff
@@ -5,8 +5,7 @@ angular.module('emission.controllers', ['emission.splash.updatecheck',
                                         'emission.splash.pushnotify',
                                         'emission.splash.localnotify',
                                         'emission.survey.launch',
-                                        'emission.stats.clientstats',
-                                        'emission.incident.posttrip.prompt'])
+                                        'emission.stats.clientstats',])
```
shankari: does this work syntactically in spite of the trailing comma?

shankari:
@atton16 @asiripanich merged and deployed. should be live in ~ 10 mins

www/js/controllers.js
atton16: Yes, trailing comma is syntactically corrected.

www/js/diary/list.js
```diff
@@ -172,12 +172,16 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
         var tripMode = {};
+        var found = false;
         if(modes.length > 0) {
           modes.forEach(function(mode) {
             if ((trip.properties.start_ts == mode.start_ts) &&
                  (trip.properties.end_ts == mode.end_ts)) {
-              tripMode = mode;
-              Logger.log("trip" + JSON.stringify(trip)+ "mode" + JSON.stringify(tripMode));
+              if (!found) {
+                tripMode = mode;
```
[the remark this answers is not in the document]
atton16: It appears that `BEMUserCache.getAllMessages` return the `latest` entry first. As this commit is a bug fix, so we simply inverse the logic. The initial commit selected the last item of the loop and results in incorrect value.

www/js/controllers.js
atton16: I would normally do this since it helps avoid potential bugs.

asiripanich:
> @asiripanich I am not going to provide input on your UX choices because the whole point of the open source platform is that you can experiment with what works for you. I will just provide some feedback on what I have seen other groups do in similar areas:
> 
> * Get back Dashboard tab: This is surprising because most groups did not want to inadvertently influence behavior during the study (the Hawthorne effect?)

We have found some errors from enabling this feature. So we will disable it in the next PR. 

> * Disable Trip-end Prompt: yeah, CCI disabled the trip-end prompt too. Miriam (the PI) really, really liked it, but some of the users in their pilot hated it. We talked about having an opt-in option as part of the onboarding, but ran out of time. I would really encourage you to explore an opt-in option during onboarding.

Will consider this option. 

> * Display only analyzed trips on Diary tab: I'd encourage you to be cautious about this. In the bic2cal study, people expected some kind of instant feedback at the end of a trip that their trip had _registered_. If they didn't see that, they thought that the app was not working. Given the latency in transmitting data + running the pipeline, it might be better to leave draft trips, or at least warn people about some delay in seeing the results (like the old moves app used to).

I don't want users to answer the trip-end questions for draft trips as they can be very annoying sometimes (very often when I wait to cross the app registers that my trip has ended.). In the next PR we will take out the trip-end questions from draft trips and bring draft trips back to Diary tab.

shankari:
> We have found some errors from enabling this feature. So we will disable it in the next PR.

what were the errors?

shankari:
> very often when I wait to cross the app registers that my trip has ended.

Is this the reason you are disabling trip-end notifications? Because instead of just disabling them, we could add some simple (configurable) checks before the trip end notification was displayed. I've thought about this for a while, but haven't had a sense of what the checks should be. Thoughts?

Also, if you feel that the processed trips are cleaner, another option is to generate the notifications from the server as a push notification. This will have some latency from the end of the trip, but you will know that there is a processed trip ready for the users to provide input for.

asiripanich:
Hi @shankari, sorry for super duper late response to this. I lost track of this thread. Anyway, since I've found it via the email exchange you had with my supervisor here is my response. :p 

> > very often when I wait to cross the app registers that my trip has ended.
> 
> @asiripanich Is this the reason you are disabling trip-end notifications? Because instead of just disabling them, we could add some simple (configurable) checks before the trip end notification was displayed. I've thought about this for a while, but haven't had a sense of what the checks should be. Thoughts?

Yes, this is part of the reason we decided to disable the trip-end notifications. I'm not so sure what sort of checks we need to prevent this from happening. On a related matter, how Emission handle trip segmentation, I think there should be a way for user to segment the trip or add more info into the trip. For example, if I go to buy lunch at the cafeteria it is not so obvious where I made the stop but instead it registered the whole trip as me going somewhere and return to where I initially started. Does a solution for this exists in Emission that I may not know of?

shankari:
I am opening a new issue for trip end notification improvements because I think that Ipsita also had some thoughts around it. The changes would need to be made in the native plugin. They are not very hard, and I am happy to make them if we can come up with a clear specification of what the changes should be, and whether they should apply to android or iOS.

shankari:
> On a related matter, how Emission handle trip segmentation, I think there should be a way for user to segment the trip or add more info into the trip. For example, if I go to buy lunch at the cafeteria it is not so obvious where I made the stop but instead it registered the whole trip as me going somewhere and return to where I initially started. Does a solution for this exists in Emission that I may not know of?

Again, there is a very very preliminary PR that @sunil07t worked on earlier that would allow you to break trips and select the mode for each section of the trip (https://github.com/e-mission/e-mission-phone/pull/308). This was then revised by the team from Paraguay (https://github.com/e-mission/e-mission-phone/pull/469).

I have not tested it and I don't know if it works. Take it with a large grain of salt. If you want to pursue this feature further, I would suggest opening a new issue to track the discussion around it. Step 1 being testing the proposed PRs to see if they actually work.

shankari:
I am actually opening a new issue for the trip editing as well because when I open https://github.com/e-mission/e-mission-phone/pull/469 I see a reference to an issue from @PatGendre too.

Issue #509, opened 2019-01-07
"Downloading UI-only update" freezes if the screen has been locked during the update 


asiripanich:
If my screen goes dark (auto-lock) during an UI update the downloading bar will be frozen when I unlock my phone to check. Is this a bug ?

Device info: iPhone X running iOS 12

shankari:
we use the ionic deploy libraries for UI updates, so I don't know the code as well as e-mission. But it doesn't seem that surprising, iOS will typically send the app to the background when the screen locks. If you are not moving, we are not engaged with one of the supported background modes, so iOS will suspend the app. Then, when you unlock the screen, it will resume the app, but the UI may or may not pick up correctly from there.

I can try to file an issue for this against ionic, but we are on an older version of the deploy plugin since I want to retain support for older versions of iOS (https://github.com/orgs/e-mission/teams/em-workshop-2018/discussions/16)

shankari:
@asiripanich is this a blocking issue? if you kill the app and restart it, you should be able to trigger the download again.

shankari:
@asiripanich hm, first can you confirm that if you kill the app and restart it, you can trigger the download again and it succeeds? I want to get a sense of how bad this is...

shankari:
@asiripanich just to confirm, the steps I am looking for are:
- start UI download
- lock screen midway
- unlock screen -> download bar is frozen
- kill app
- re-launch app
- Profile -> check for updates to check if the update is done

asiripanich:
@shankari the screen has to be on the whole time for the update to be completed. Usually, if the download bar is frozen, I would tell the user to kill the app and re-launch it again but then make sure the screen is on until the download is completed.

shankari:
@asiripanich maybe I should be more clear. I am asking for your help in testing since I haven't actually tested (or thought to test) this use case. This is how users can help open source projects :)

So if you followed the steps listed in https://github.com/e-mission/e-mission-phone/issues/509#issuecomment-452355862, would the download be completed when the app relaunched? In other words, is it a problem with the download or just the UI representation of the download in the progress bar?

asiripanich:
@shankari my app has already been updated. Will send another PR sometime this week and will try the steps above and report back.

shankari:
@asiripanich new PR merged and deployed. Chance to test this again?

asiripanich:
[...]

PR #518, opened 2019-01-14
Rk unsw

atton16:
- Put back unprocessed trips
- Conditionally display mode/purpose confirm only when the trip is processed
- Disable Dashboard tab (again)

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -1804,9 +1804,9 @@ span.rz-bar-wrapper {
   width: 230px;
 }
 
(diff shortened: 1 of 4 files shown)


shankari:
@atton16 @asiripanich merged and deployed. Should be live in ~ 10 mins.
@asiripanich, you can use this update to test #509

asiripanich:
@shankari Thanks for accepting the PR. About the issue, I will update as soon as I finish with my TRB presentations.

PR #520, opened 2019-01-22
api calling function implemented

trevor-wu:
Api key function for getting reviews


www/js/recommendations.js
```diff
@@ -12,15 +12,42 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
 
     $scope.name = "Cannot Retrieve Suggestion";
     $scope.mode = "Cannot Retrieve Mode";
+    $scope.yelp = {
+      "headers": {
+    		"Authorization": "Bearer your-api-key",
```
shankari: I'm pretty sure that the `your-api-key` here needs to be replaced with an actual API key

www/js/recommendations.js
```diff
@@ -12,15 +12,42 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
...
+      $http({
+        "async": true,
+        "crossDomain": true,
+        "url": "https://api.yelp.com/v3/businesses/yalis-stanley-hall-cafe-berkeley/reviews",
+        "method": "GET",
+        "headers": $scope.yelp.headers
+      }).then(function(res) {
+        $scope.revs = res.data.reviews;
+      });
+      $http({
```
shankari: why make this call twice?

www/js/recommendations.js
shankari: And given `yelpfusion.json`, do you even need this here?

PR #521, opened 2019-01-23
Adding more questions in each trip

deepalics0044:


www/js/diary/list.js
```diff
@@ -411,12 +413,79 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
+      $scope.parkingPopover.show($event);
+   };
+
+  var closeParkingPopover = function($event, isOther) {
+      if(isOther == false)
+        $scope.draftParking = angular.undefined;
+      Logger.log("in closeParkingPopover, setting draftParking = "+JSON.stringify($scope.draftParking));
+      $scope.parkingPopover.hide($event);
+   };
+
```
shankari: nit: In a later PR, might want to remove unnecessary whitespace

www/js/diary/list.js
```diff
@@ -512,6 +666,28 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
+	
+	
+	
+	
+	
+	
+	 $scope.quickrideOptions = [
+  
+  
+   {text: 'Click here',value:'other_quickride'}];
```
shankari: So you only want "other" as the option? I guess that works?! but I wonder why you want to have a separate button then.

PR #522, opened 2019-01-29
Rk unsw

atton16:


www/templates/diary/list.html
```diff
@@ -87,10 +85,10 @@
                 </div>
              </div>
              <div ng-if="isAnalyzed(tripgj)" class="row" style="padding-left: 5px;padding-right: 5px;">   
-                <div class="col-50" style="text-align: center;font-size: 14px;font-weight: 600;" ng-if="checkMode(tripgj)">
```
shankari: not sure what this is supposed to do

shankari:
@atton16 @asiripanich merged and deployed, should be live in ~ 10 mins

PR #523, opened 2019-01-29
Fixing function pathway and singleTrip function call

sam11b:
With regards to the image of the star images, I believe Trevor should be doing this in his pull request since his API call includes valid star images.


shankari:
merged + deployed, should be live in ~ 10 mins

PR #524, opened 2019-02-01
Addressed issues in Pull Request #520

trevor-wu:
- Reduced api calls down to 1
- integrated with server changes
- renamed star files so the number doesn't have to be stringified and rounded
- kept api key in separate json


www/js/recommendations.js
```diff
@@ -12,15 +12,33 @@ angular.module('emission.main.recommendations',['emission.services', 'emission.p
...
       $scope.currentDisplay.style.display = "block";
+      $ionicLoading.show({
+        template: 'Loading Reviews...'
+        });
+      $http({
+        "async": true,
+        "crossDomain": true,
+        "url": "https://api.yelp.com/v3/businesses/"+$scope.bid+"/reviews",
+        "method": "GET",
+        "headers": $scope.yelp.headers
```
shankari: this can technically have a race condition in which you try to read the yelp headers before they are initialized as part of the `$http` request above. I suspect this is hard to hit in your case because the header reading is nearly instantaneous, clearly faster than any human can press the review button.

I am going to merge this for now, but for the record, you should really understand asychronous programming and promises better :)

trevor-wu:
screen shot 2019-02-01 at 12 58 32 pm
Screenshot from the simulator showing that the api call works

shankari:
These changes are fairly small, so I'm going to squash before merging. Please take this into account while starting your next PR

PR #526, opened 2019-02-08
changed url for config file

trevor-wu:
www/json/connectionConfig.json


www/json/connectionConfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "connectUrl": "http://52.91.65.149:8080",
+    "connectUrl": "http://52.91.65.149:80",
```
shankari: nit" technically, you don't need to specify the port, because the default port for http is 80

shankari:
merged and deployed, should be live in ~ 10 mins.
Please test, test, test before demo tomorrow.

PR #528, opened 2019-02-20
Rk unsw

atton16:
Add support for Remote mode confirm / purpose confirm data

### DIFF
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -299,7 +299,10 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         });
         if (modes.length > 0) {
           var mode = modes[modes.length - 1];
-          $scope.chosen.mode.label = mode.data.label;
+          var opt = $scope.modeOptions.filter(o => o.value === mode.data.label);
+          if (opt.length > 0) {
+            $scope.chosen.mode.label = opt[0].text;
+          }
           hasMode = true;
         }
       }
@@ -333,7 +336,10 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         });
         if (purposes.length > 0) {
           var purpose = purposes[purposes.length - 1];
-          $scope.chosen.purpose.label = purpose.data.label;
+          var opt = $scope.purposeOptions.filter(o => o.value === purpose.data.label);
+          if (opt.length > 0) {
+            $scope.chosen.purpose.label = opt[0].text;
+          }
           hasPurpose = true;
         }
       }
@@ -645,7 +651,8 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
             );
           });
           if (unified_modes.length > 0) {
-            $scope.selected.mode = unified_modes[unified_modes.length - 1];
+            var unified_mode = unified_modes[unified_modes.length - 1];
+            $scope.selected.mode.label = unified_mode.data.label;
           }
         } else {
           $scope.selected.mode = mode;
@@ -692,7 +699,8 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
             );
           });
           if (unified_purposes.length > 0) {
-            $scope.selected.purpose = unified_purposes[unified_purposes.length - 1];
+            var unified_purpose = unified_purposes[unified_purposes.length - 1];
+            $scope.selected.purpose.label = unified_purpose.data.label;
           }
         } else {
           $scope.selected.purpose = purpose;
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -19,7 +19,6 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
     console.log("controller DiaryListCtrl called");
     var MODE_CONFIRM_KEY = "manual/mode_confirm";
     var PURPOSE_CONFIRM_KEY = "manual/purpose_confirm";
-    $scope.unifiedResults = null;
 
     // Add option
 
@@ -290,8 +289,8 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         }
       });
       // Fallback to unified Results
-      if (!hasMode && $scope.unifiedResults !== null) {
-        var modes = $scope.unifiedResults.modes.filter(mode => {
+      if (!hasMode && $scope.data.unifiedConfirmsResults !== null) {
+        var modes = $scope.data.unifiedConfirmsResults.modes.filter(mode => {
           return (
             mode.data.start_ts === trip.data.properties.start_ts &&
             mode.data.end_ts === trip.data.properties.end_ts
@@ -327,8 +326,8 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         }
       });
       // Fallback to unified Results
-      if (!hasPurpose && $scope.unifiedResults !== null) {
-        var purposes = $scope.unifiedResults.purposes.filter(purpose => {
+      if (!hasPurpose && $scope.data.unifiedConfirmsResults !== null) {
+        var purposes = $scope.data.unifiedConfirmsResults.purposes.filter(purpose => {
           return (
             purpose.data.start_ts === trip.data.properties.start_ts &&
             purpose.data.end_ts === trip.data.properties.end_ts
@@ -352,34 +351,17 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         $scope.data = Timeline.data;
         $scope.datepickerObject.inputDate = Timeline.data.currDay.toDate();
         
-        function routineTasks () {
-          $scope.data.currDayTrips.forEach(function (trip, index, array) {
-            addUnpushedMode(trip);
-            addUnpushedPurpose(trip);
-            PostTripManualMarker.addUnpushedIncidents(trip);
-          });
-          $scope.data.currDayTripWrappers = Timeline.data.currDayTrips.map(
-            DiaryHelper.directiveForTrip);
-          // $scope.data.currAnalysedDayTripWrappers = $scope.data.currDayTripWrappers.filter(function(el){
-          //   return el.data.id.indexOf('unprocessed_') === -1;
-          // });
-          $ionicScrollDelegate.scrollTop(true);
-        }
-        if ($scope.unifiedResults === null) {
-          var tq = { key: 'write_ts', startTs: 0, endTs: moment().endOf('day').unix(), };
-          Promise.all([
-            UnifiedDataLoader.getUnifiedMessagesForInterval(MODE_CONFIRM_KEY, tq),
-            UnifiedDataLoader.getUnifiedMessagesForInterval(PURPOSE_CONFIRM_KEY, tq)
-          ]).then(results => {
-            $scope.unifiedResults = {
-              modes: results[0],
-              purposes: results[1],
-            };
-            routineTasks();
-          });
-          return;
-        }
-        routineTasks();
+        $scope.data.currDayTrips.forEach(function (trip, index, array) {
+          addUnpushedMode(trip);
+          addUnpushedPurpose(trip);
+          PostTripManualMarker.addUnpushedIncidents(trip);
+        });
+        $scope.data.currDayTripWrappers = Timeline.data.currDayTrips.map(
+          DiaryHelper.directiveForTrip);
+        // $scope.data.currAnalysedDayTripWrappers = $scope.data.currDayTripWrappers.filter(function(el){
+        //   return el.data.id.indexOf('unprocessed_') === -1;
+        // });
+        $ionicScrollDelegate.scrollTop(true);
       });
     });
 
@@ -643,8 +625,8 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         }
       };
       getLocalTripMode(fakeTrip).then(function (mode) {
-        if (!mode.label && $scope.unifiedResults !== null) {
-          var unified_modes = $scope.unifiedResults.modes.filter(unified_mode => {
+        if (!mode.label && $scope.data.unifiedConfirmsResults !== null) {
+          var unified_modes = $scope.data.unifiedConfirmsResults.modes.filter(unified_mode => {
             return (
               unified_mode.data.start_ts === start_ts &&
               unified_mode.data.end_ts === end_ts
@@ -691,8 +673,8 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
         }
       };
       getLocalTripPurpose(fakeTrip).then(function (purpose) {
-        if (!purpose.label && $scope.unifiedResults !== null) {
-          var unified_purposes = $scope.unifiedResults.purposes.filter(unified_purpose => {
+        if (!purpose.label && $scope.data.unifiedConfirmsResults !== null) {
+          var unified_purposes = $scope.data.unifiedConfirmsResults.purposes.filter(unified_purpose => {
             return (
               unified_purpose.data.start_ts === start_ts &&
               unified_purpose.data.end_ts === end_ts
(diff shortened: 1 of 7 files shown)


atton16:
Regardless of what we have discussed on the "Interactive trip-end survey" issue, this PR is working on dev-app.

asiripanich:
@shankari Hi Shankari could you please give us your feedback on this?

shankari:
@asiripanich sorry for the delay, I was focusing on writing for the past three days since I had to submit a chapter to a thesis writing workshop that I am part of. In general, during this semester, I will be working on code on Monday and Tuesday, in Berkeley + meetings on Wednesday, and writing on Thursday, Friday and Saturday.

www/js/diary/list.js
```diff
@@ -622,43 +636,63 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
...
+    $scope.selected = {
+      mode: {
+        label: ''
+      },
+      purpose: {
+        label: ''
+      }
+    };
+    /**
+     * Store currently chosen value
```
shankari: Can you add some more documentation on `chosen` versus `selected`? Is the thought that the user first chooses the mode and then does something else to `select` it? Or vice versa?

www/js/diary/list.js
```diff
@@ -315,17 +345,35 @@ angular.module('emission.main.diary.list', ['ui-leaflet',
...
+            DiaryHelper.directiveForTrip);
+          // $scope.data.currAnalysedDayTripWrappers = $scope.data.currDayTripWrappers.filter(function(el){
+          //   return el.data.id.indexOf('unprocessed_') === -1;
+          // });
+          $ionicScrollDelegate.scrollTop(true);
+        }
+        if ($scope.unifiedResults === null) {
+          var tq = { key: 'write_ts', startTs: 0, endTs: moment().endOf('day').unix(), };
+          Promise.all([
+            UnifiedDataLoader.getUnifiedMessagesForInterval(MODE_CONFIRM_KEY, tq),
```
shankari: The `UnifiedDataLoader` loads data from both the local cache and the server 
https://github.com/e-mission/e-mission-phone/blob/master/www/js/services.js#L233

so falling back to the `UnifiedDataLoader` if the local call fails (e.g. `getLocalTripMode().then....`) makes an extra, unnecessary call to the local database. You should be able to just use the `unifiedResults` directly.

shankari:
@asiripanich can you give me a sense of the timeline for your survey? I want a strike a balance between long-term cleanliness of code and short-term utility for you.

asiripanich:
> @asiripanich can you give me a sense of the timeline for your survey? I want a strike a balance between long-term cleanliness of code and short-term utility for you.

Hi Shankari, from my last discussion with Taha there isn't a clear timeline (not in the next month anyway) for us regarding deployment of this to real participants yet. 

But if there will be significant changes to the codebase that could potentially impact our branch please keep us in the loop.

shankari:
@asiripanich @atton16 my turn to remind you to respond to my question :)

atton16:
> The changes to the HTML files are pretty straightforward. They mainly seem to be about storing the chosen mode as an object instead of a string.
> 
> But the `list.js` changes are really hard to review. The vast majority of the changes seem to be whitespace changes but there are so many of them that it is hard to see what the relevant functionality changes are.
> 
> Is there a reason why you wanted to reformat `list.js`? Large scale reformatting can make following the history trickier (e.g. all these lines will now be marked as modified by you), and so I avoid it unless the previous code is completely terrible. And the formatting of the previous code didn't seem to be that terrible to me. Is it violating a javascript standard that I am not aware of?

The reason for formatting the list.js is that there are a lot of indentation that is misplaced which makes code reading very difficult and the IDE cannot collapse the code properly.

www/js/diary/list.js
atton16: `selected` is for display purpose only, this is to display the currently selected option on the list.

`chosen` is for storing the actual value.

The previous code was like
```
$scope.mode = '';
$scope.purpose = '';
$scope.chosen = {
  mode: '',
  purpose: '',
}
```

Now I am refactor it to be more appropriate by moving `$scope.mode` and `$scope.purpose` to `$scope.chosen`. And `$scope.chosen` to `$scope.selected` for readability and structure.

www/js/diary/list.js
atton16: This is for compatibility with the existing phone that is running.

www/js/diary/list.js
shankari: ok, but can you add this as a comment to the code? Not everybody is going to go and find the pull request related to a change.

www/js/diary/list.js
shankari: Not sure what you mean by this. You are currently merging to `rk-unsw` and you are the only ones using that branch. So compatibility is not an immediate issue.

wrt to other branches, the PR that you got the original change from is unmerged, so nobody else is using it either.

Given that this is essentially the first implementation of this change, I would rather that we did it right wrt performance instead of prioritizing compatibility with an implementation that nobody is using.

Also, at least from your code, the right way doesn't seem to be that much harder than your way, and will make the final implementation significantly easier to understand. Please let me know if you disagree and what the concrete issues with using the right way are.

shankari:
@atton16 `list.js` explanation makes sense; I use vim to edit code so I haven't encountered it, but I know that I am in the minority.  Thank you for splitting that out into a separate commit.

Just out of curiosity, which IDE are you using?

www/js/diary/list.js
shankari: concretely, this line

> selected is for display purpose only, this is to display the currently selected option on the list.

atton16:
> @atton16 `list.js` explanation makes sense; I use vim to edit code so I haven't encountered it, but I know that I am in the minority. Thank you for splitting that out into a separate commit.
> 
> Just out of curiosity, which IDE are you using?

I am using VSCode for heavy duty work like when I am working on the project. For quick editing, I am uring Sublime Text.

atton16:
As for what we are trying to achieve, we are trying to display to remotely saved responses on the buttons and the popover options. [See Buttons Flow Chart](http://drive.attawit.me/s/o6CbsDrk4Xg4Lj8).

The reason that we structure the code this way is
1. Compatibility with the older code as we did not modify the original logic.
2. Compatibility with the phones we are testing for both the older testers and new testers.
3. We are only fetching from remote server once during the timeline update.
    [See Timeline.UPDATE_DONE Flow Chart](http://drive.attawit.me/s/LMWSQwczsbyEk2E)

As for the reason behind this code structure is that, fetching data from remote server takes time and the `Buttons` and `Popovers` cannot wait that long (even 1second) to render the UI properly. So we went further to leverage `Timeline.UPDATE_DONE` event and retrieve the remotely stored values only once per user's lifetime. Our survey will strictly allow 1 login per participants. So this make senses.

Additionally, remote data is **only** required if users were previously logged in and somehow delete the app/logged out. So in the sense, we are trying to use as much of the local processes as possible and leave out the process of data saving (both locally and remotely) as is.

Hopefully, this make it clear to why we did what we did.

www/js/diary/list.js
atton16: Sure, done.

shankari:
> Additionally, remote data is only required if users were previously logged in and somehow delete the app/logged out. So in the sense, we are trying to use as much of the local processes as possible and leave out the process of data saving (both locally and remotely) as is.

Maybe I misunderstood what you are saying, but remote data is not only required if the user deleted the app. All local data is (or should be) pushed to the server when the trip ends.

Concretely, let us consider the following use case:
- user comes to work, completes a trip at 9am
- data is pushed, does not include trip and purpose
- user marks trip and purpose at noon
- user goes home, completes a trip at 6pm
- data is pushed, includes trip and purpose for the 9am trip but not for the 6pm trip
- user goes to work in the morning, completes a trip at 10am
- data is pushed, includes trip and purpose for the 6pm trip but not the 10am trip
- user views trip diary from the previous day; mode and purpose for both the 9am and 6pm trips have to be retrieved from the server

Is this not what you are seeing? Can you write out what you observe for this particular scenario?

shankari:
> So we went further to leverage Timeline.UPDATE_DONE event and retrieve the remotely stored values only once per user's lifetime. 

the `Timeline.UPDATE_DONE` event is generated on every update for a day. So if you load the data for yesterday, `Timeline.UPDATE_DONE` will be generated, if you load the data for the day before `Timeline.UPDATE_DONE` will be generated *again*. I don't see how this would be for the user's lifetime unless you plan to only collect data for one day.

shankari:
So to recap:
1. the compatibility arguments are not convincing since there is nobody actually using this code right now
1. the "only retrieve from remote server once during lifetime" is not convincing because with the current code, you will retrieve the messages for each day that you display. So assuming this is going to be a multi-day survey, you will retrieve data multiple times as the users fill in information for each day. You will even potentially retrieve data multiple times for the same day as the user takes trips through the day (see the 9am trip versus 6pm trip in the example above).

I think that the argument that you are making is actually:
1. @sunil07t's implementation retrieved data for each trip (e.g. https://github.com/e-mission/e-mission-phone/pull/307/files#diff-ddf1898e47aa5962e85291083584714eR168). So for a day with 5 trips, it would make 5 calls to the database. I think you are concerned about making that many remote calls to the server. I can see that concern, although the calls would be in parallel, and the range for each trip would be short, so presumably that would make the calls faster. Have you actually tested this? What is the empirical performance?
    - But then I don't see why the suggestion to only use the unified results doesn't work.
    - You have piggy-backed the unified results into the timeline load. But you have currently done so by launching the call after populating the trips, which is not consistent with the rest of the code, or with angular recommendations for what goes into controllers versus services.
    - You should launch the call to read the final mode and purpose in the timeline service at https://github.com/e-mission/e-mission-phone/blob/master/www/js/diary/services.js
    - That's where we make all other local and remote calls to read the timeline and generate the TIMELINE_DONE
    - You can launch this call in there as well, in parallel with the other calls, and fill in the mode and purpose objects (either from local or remote) into memory
    - Then, in the `TIMELINE_DONE` listener, you just populate them into the scope objects, just like the existing code

If you wanted to make minimal changes, then converting `window.cordova.plugins.BEMUserCache` -> `UnifiedDataLoader` would be the easy fix. But if you are going to mess with the timeline reading, then the above is the correct fix.

atton16:
> > Additionally, remote data is only required if users were previously logged in and somehow delete the app/logged out. So in the sense, we are trying to use as much of the local processes as possible and leave out the process of data saving (both locally and remotely) as is.
> 
> Maybe I misunderstood what you are saying, but remote data is not only required if the user deleted the app. All local data is (or should be) pushed to the server when the trip ends.
> 
> Concretely, let us consider the following use case:
> 
> * user comes to work, completes a trip at 9am
> * data is pushed, does not include mode and purpose
> * user marks mode and purpose for the 9am trip at noon
> * user goes home, completes a trip at 6pm
> * data is pushed, includes mode and purpose for the 9am trip but not for the 6pm trip
> * user marks mode and purpose for the 6pm trip at 10pm
> * user goes to work in the morning, completes a trip at 10am
> * data is pushed, includes mode and purpose for the 6pm trip but not the 10am trip
> * user views trip diary from the previous day; mode and purpose for both the 9am and 6pm trips have to be retrieved from the server
> 
> Is this not what you are seeing? Can you write out what you observe for this particular scenario?

I was trying to explain in the sense of the `phone` code. To put it simply, we only pull the old remote data once and use it as cache for displaying old responses. Once the old remote data is pulled, the UI can render buttons and popover smoothly. The local data is periodically synced out normally (we did not change it).

shankari:
> To put it simply, we only pull the old remote data once and use it as cache for displaying old responses. Once the old remote data is pulled, the UI can render buttons and popover smoothly. 

Just to make sure that we understand the same flow, when you pull the remote data and cache it, you do so in memory. So if the user loads the data for a new day, the cached values get overwritten. So you pull the old remote data once **for one day** at a time.

I agree with this approach in general, and that is what the timeline load does. But if you want to use this approach, you should load the data as part of the timeline, in a service, not in a controller.
https://stackoverflow.com/questions/27398741/angularjs-controller-vs-service

atton16:
> > To put it simply, we only pull the old remote data once and use it as cache for displaying old responses. Once the old remote data is pulled, the UI can render buttons and popover smoothly.
> 
> Just to make sure that we understand the same flow, when you pull the remote data and cache it, you do so in memory. So if the user loads the data for a new day, the cached values get overwritten. So you pull the old remote data once **for one day** at a time.
> 
> I agree with this approach in general, and that is what the timeline load does. But if you want to use this approach, you should load the data as part of the timeline, in a service, not in a controller.
> https://stackoverflow.com/questions/27398741/angularjs-controller-vs-service

At the moment, the logic behinds ensure that we pulls data once per lifetime.

shankari:
> At the moment, the logic behinds ensure that we pulls data once per lifetime.

Please explain how this is true with the use case from https://github.com/e-mission/e-mission-phone/pull/528#issuecomment-469414217

In even more detail:

-    user comes to work, completes a trip at 9am
-    data is pushed, does not include mode and purpose
-    user loads diary for the day, marks mode and purpose for the 9am trip at noon
-    user goes home, completes a trip at 6pm
-    data is pushed, includes mode and purpose for the 9am trip but not for the 6pm trip
-    user loads diary for the day, marks mode and purpose for the 6pm trip at 10pm
      - this will load the mode and purpose for the 9am trip from the server because it is not on the phone
-    user goes to work in the morning, completes a trip at 10am
-   data is pushed, includes mode and purpose for the 6pm trip but not the 10am trip
-   user loads diary for the day, marks mode and purpose for the 10am trip at noon
- user views trip diary from the previous day
      - mode and purpose for both the 9am and 6pm trips have to be retrieved from the server

I really don't see how in-memory data will remain for the lifetime of the app. I understand that you think that is true because you are piggy backing off `TIMELINE_DONE` but I don't think that `TIMELINE_DONE` means what you think it does.

shankari:
> As for Controller vs Service, usually we do a reusable non UI related code in service. The link you reference to also said it so. It also said in AngularJS concept here https://docs.angularjs.org/guide/concepts

But the timeline update **is** a reusable non UI related service. All the other timeline data is retrieved in the service and is mapped to UI elements when the timeline is done. It is inconsistent to have only the mode and purpose be retrieved in a UI-specific controller.

shankari:
So my suggestion is:
https://github.com/e-mission/e-mission-phone/blob/master/www/js/diary/services.js#L409
in the timeline.data variable, add a new field for `unifiedResults` so instead of `$scope.unifiedResults` you would have `timeline.data.unifiedResults`

populate `unifiedResults` in the same way, potentially launching it from `timeline.updateForDay` as a separate, parallel `Promise.all`
https://github.com/e-mission/e-mission-phone/blob/master/www/js/diary/services.js#L879

in the controller, when `TIMELINE_DONE`, then put the data from `timeline.data.unifiedResults` to `$scope`. then the buttons will all work off `$scope` just like the rest of the list UI

shankari:
This change looks fine to me.
Can you please add the testing done results here?

atton16:
> This change looks fine to me.
> Can you please add the testing done results here?

So here is the test procedure and results:
1. Remove `em-devapp` to ensure that all local cache is gone.
2. Using `test_july_22` user and data, navigate to `July 22, 2015`
3. We can see `Bike` on the mode for the first entry [Screenshot](http://drive.attawit.me/s/5L6jPC9cQi42bsM)
4. Override the response to `Drove Alone` [Screenshot](http://drive.attawit.me/s/5SJeb9kZoCYn8br)
5. Navigate to other date
6. Navigate back to `July 22, 2015`, the result is same as (4) [Screenshot](http://drive.attawit.me/s/5SJeb9kZoCYn8br)

shankari:
As we discussed, I am merging this as a temporary measure, but I would really like a fix for #334 as well for the long-term

shankari:

PR #532, opened 2019-03-05
Minor UI changes including getting rid of repeated information, and escaping out of loading button

sam11b:
Minor UI changes. Commented out repeated info on the diary page as per feedback from Professor Cassidy (wasn't sure if I should just delete it fully). Fixed loading button issue as brought up by Vanessa.


shankari:

PR #533, opened 2019-03-14
Bump up all the plugins for the migration to api 26

shankari:
Port of https://github.com/e-mission/e-mission-base/pull/15


shankari:
Everything looks consistent with https://github.com/e-mission/e-mission-base/pull/16

PR #534, opened 2019-03-18
fix(UI): Mode/Purpose confirm button turns green once answered

atton16:
This is a minor UI improvement. Mode/Purpose confirm button will turn to green once it is answered.

Screenshot: [http://drive.attawit.me/s/iB9LoT6zEx5Hwao](http://drive.attawit.me/s/iB9LoT6zEx5Hwao)


shankari:

PR #535, opened 2019-03-19
Greentrip ui23

trevor-wu:
- Double checked review system on simulator
- Added review system to trip diary, not enough trips on my end to test
- Changed yelpfusion.json to yelpfusion.json.sample
- Changed connectionConfig.json 's url to port 80


shankari:
closing the PR because of bad commit history. @trevor-wu please generate a good one.

PR #537, opened 2019-03-20
changes and fixes to review system

trevor-wu:
- Cleaned up log and commits
- Yelpfusion.json.sample no longer shows up as new file


shankari:
@trevor-wu can you please indicate the testing done?

shankari:
Please do so by tonight, I would like to merge this change today since we have already postponed this for 2 weeks

trevor-wu:
The api call was tested on the simulator and it does connect to Yelp using the API key I generated.

shankari:
`yelpfusion.json` customized, and deployed.
Should be live in ~ 10 mins

PR #538, opened 2019-03-22
changes in instructions and mode

deepalics0044:


shankari:
changes seem pretty straightforward. deployed, should be live in ~ 10 mins

PR #539, opened 2019-03-23
some more changes

deepalics0044:


www/templates/cci-about/about.html
```diff
@@ -7,7 +7,14 @@
...
 		</p>
 	
-	
+	  <div>
+	    <h1 class="signup-title" style="text-align:center">Enter serial no.</h1>
+	      <label class="item item-input">
+	        <input type="text" placeholder="Serial number" ng-model="user.username">
+	      </label>
+
+	    <button class="button button-positive button-full" ng-click="signup()">Submit</button>
```
shankari: wait, where's the javascript for `signup()`? This PR only has HTML changes.

shankari:
@deepalics0044 where is the javascript for the new button? Both the files here are html files only

deepalics0044:
Please check out the latest commit

shankari:
@deepalics0044 you cannot copy and paste `updateUser`. `updateUser is a standard function defined in `services.js`. As I said 

> you can search for examples of how it is used

which means **searching for how it is used**

e.g. https://github.com/e-mission/e-mission-phone/search?q=updateUser&unscoped_q=updateUser

and using it properly in the angular way

shankari:
@deepalics0044 @ipsita0012, I am freaked out about your impending release even if you are not, so I am going to tell you roughly what to do, although you should really read up on ionic and angular so that you can understand what you are doing. Please do not directly copy-paste the suggestion; this is just an aide to get you pointed in the right direction.

- you need to add the javascript in the javascript file (https://github.com/e-mission/e-mission-phone/blob/cistup-ec/www/js/cci-about/about.js) instead in a script tag
- the controller defined in the javascript file should import `CommHelper` (again look for examples of how this is done)
- you need to have a `$scope.` function that updates the user. You can look at the original CCI `$scope.startSurvey` for an example.
- the implementation of the `$scope` function should call `CommHelper` just like the other examples.

Please get this done ASAP. You have less than 48 hours for launch.

shankari:
@deepalics0044 @ipsita0012 when you send out the next set of changes, please indicate the testing that you did to ensure that it works. It is pretty clear that both the commits to add the functionality essentially didn't work. Code review is not a substitute for testing.

shankari:
ok, so move out the `connectionConfig.json` and connect to a server that you are running on your laptop. Have you at least made sure that the code loads without any errors in the javascript? Because the import is essentially wrong.

shankari:
@deepalics0044 testing done? I believe this is still not correct, and I don't want you to throw random tries at the code into the PR and have me tell you whether they do or do not work. Code review is **not** a substitute for testing.

deepalics0044:
Doing it right now.

shankari:
> Doing it right now.

Testing generally happens *before* you commit code not *after*. I will not look at any commits until you put in the testing information.

www/templates/intro/intro.html
```diff
@@ -1,11 +1,6 @@
 <ion-view hide-nav-bar="true" class="intro-view">
     <ion-slide-box delegate-handle="intro-box" on-slide-changed="slideChanged(index)" show-pager="false">
-      <ion-slide>
-        <ng-include src="'templates/intro/summary.html'"></ng-include>
-      </ion-slide>
-      <ion-slide>
-        <ng-include src="'templates/intro/consent.html'"></ng-include>
```
shankari: wait, did you intend to remove the summary and consent?
Why make changes to them then?

deepalics0044:
Yes it was a last minute change. We don't want the first two pages. As we
have already provided all the information covered in summary and consent.

shankari:

PR #549, opened 2019-04-08
Fix tracking toggle

shankari:
- Fix regression caused by
https://github.com/e-mission/e-mission-phone/commit/a99c4e8798b7cfa71db5b366c73bbf4f5f2f0237
and also simplify the resulting code

- Bump up the version numbers for the data collection plugin to get the fix for
https://github.com/e-mission/e-mission-data-collection/commit/9459e18376659c255defb5c02b3ad63b3d6dd91f

Testing done (in conjunction with https://github.com/e-mission/e-mission-data-collection/commit/9459e18376659c255defb5c02b3ad63b3d6dd91f):

    - Cleared logs
    - Toggled tracking
    - Ensured that refreshScreen was not called
    - Clicked on the button in the alert
    - Ensured that refreshScreen was called and the final state was consistent with
      the toggle


shankari:
This fixes https://github.com/e-mission/e-mission-docs/issues/367

PR #553, opened 2019-04-09
Merge the mode and purpose final code

shankari:
This fixes https://github.com/e-mission/e-mission-docs/issues/353

- Put the notification code into separate `tripconfirm`modules; there are two potential prompts and they can be configured in any combination
- Merge the buttons into the list view (list view now has support for both incidents and trip confirmation)
- switch the default prompt to tripconfirm
- change the post-trip prompt to be more meaningful

### DIFF
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -110,3 +110,34 @@ a.item-content {
   margin-left: 3px;
   margin-top: 3px;
 }
+.btn-mode-confirm-green,
+.btn-mode-confirm-green:hover,
+.btn-mode-confirm-green:active,
+.btn-purpose-confirm-green,
+.btn-purpose-confirm-green:hover,
+.btn-purpose-confirm-green:active {
+  background-color: #30A64A;
+  color: white;
+}
+.btn-mode-confirm-white,
+.btn-mode-confirm-white:hover,
+.btn-mode-confirm-white:active,
+.btn-purpose-confirm-white,
+.btn-purpose-confirm-white:hover,
+.btn-purpose-confirm-white:active {
+  background-color: #ddd;
+  color: #333;
+  font-size: 0.8em;
+}
+.btn-mode-confirm,
+.btn-purpose-confirm {
+  line-height: 30px;
+  min-height: 30px;
+  font-size: 0.8em !important;
+  width: 115px;
+  padding: 0;
+}
+.mode-confirm-container,
+.purpose-confirm-container {
+  margin-top: 5px;
+}
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -141,3 +141,13 @@ a.item-content {
 .purpose-confirm-container {
   margin-top: 5px;
 }
+
+.popover {
+  height: 297px;
+  width: 230px;
+}
+
+#diary-item {
+  padding: 0;
+  border-width: 0;
+}
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -905,7 +905,7 @@ button.button.back-button.buttons.button-clear.header-item {
         font-size: 9px;
         position: absolute;
         left: 1%;
-        top: 275px;
+        bottom: 15px;
         line-height: 16px;
 }
 .stop-time-tag-lower {
--- a/www/index.html
+++ b/www/index.html
@@ -74,6 +74,7 @@
     <script src="js/incident/post-trip-manual.js"></script>
     <script src="js/tripconfirm/post-trip-prompt.js"></script>
     <script src="js/tripconfirm/post-trip-map-display.js"></script>
+    <script src="js/tripconfirm/trip-confirm-services.js"></script>
     <script src="js/diary.js"></script>
     <script src="js/diary/list.js"></script>
     <script src="js/diary/current.js"></script>
--- a/www/index.html
+++ b/www/index.html
@@ -72,8 +72,8 @@
     <script src="js/main.js"></script>
     <script src="js/recent.js"></script>
     <script src="js/incident/post-trip-manual.js"></script>
-    <script src="js/incident/post-trip-prompt.js"></script>
-    <script src="js/incident/post-trip-map-display.js"></script>
+    <script src="js/tripconfirm/post-trip-prompt.js"></script>
+    <script src="js/tripconfirm/post-trip-map-display.js"></script>
     <script src="js/diary.js"></script>
     <script src="js/diary/list.js"></script>
     <script src="js/diary/current.js"></script>
--- a/www/js/common/map.js
+++ b/www/js/common/map.js
@@ -177,7 +177,7 @@ angular.module('emission.main.common.map',['emission.main.common.services',
         }
 
        console.log("got response, setting display name to "+name);
-       place_feature.properties.displayName = name;
+       place_feature.properties.display_name = name;
     };
 
 
--- a/www/js/common/services.js
+++ b/www/js/common/services.js
@@ -164,13 +164,13 @@ angular.module('emission.main.common.services', [])
         console.log("got response, setting display name to "+name);
         switch (mode) {
           case 'place':
-            obj.properties.displayName = name;
+            obj.properties.display_name = name;
             break;
           case 'cplace':
-            obj.displayName = name;
+            obj.display_name = name;
             break;
           case 'ctrip':
-            obj.start_displayName = name;
+            obj.start_display_name = name;
             break;
         }
 
@@ -193,7 +193,7 @@ angular.module('emission.main.common.services', [])
             }
         }
         console.log("got response, setting display name to "+name);
-        obj.end_displayName = name;
+        obj.end_display_name = name;
 
       };
       switch (mode) {
@@ -293,8 +293,8 @@ angular.module('emission.main.common.services', [])
         commonGraph.data.graph.common_places.forEach(function(cPlace, index, array) {
             commonGraph.data.cPlaceCountMap[cPlace._id.$oid] = cPlace.places.length;
             commonGraph.data.cPlaceId2ObjMap[cPlace._id.$oid] = cPlace;
-            if (angular.isDefined(cPlace.displayName)) {
-              console.log("For place "+cPlace.id+", already have displayName "+cPlace.displayName);
+            if (angular.isDefined(cPlace.display_name)) {
+              console.log("For place "+cPlace.id+", already have display_name "+cPlace.display_name);
             } else {
               console.log("Don't have display name for end place, going to query nominatim");
               commonGraph.getDisplayName('cplace', cPlace);
@@ -318,7 +318,7 @@ angular.module('emission.main.common.services', [])
                 "id": place._id.$oid,
                 "geometry": place.location,
                 "properties": {
-                    "displayName": place.displayName
+                    "display_name": place.display_name
                 }
             };
         });
--- a/www/js/controllers.js
+++ b/www/js/controllers.js
@@ -6,7 +6,7 @@ angular.module('emission.controllers', ['emission.splash.updatecheck',
                                         'emission.splash.localnotify',
                                         'emission.survey.launch',
                                         'emission.stats.clientstats',
-                                        'emission.incident.posttrip.prompt'])
+                                        'emission.tripconfirm.posttrip.prompt'])
 
 .controller('RootCtrl', function($scope) {})
 
--- a/www/js/diary/detail.js
+++ b/www/js/diary/detail.js
@@ -61,15 +61,11 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
   $scope.getFormattedDistance = DiaryHelper.getFormattedDistance;
   $scope.getSectionDetails = DiaryHelper.getSectionDetails;
   $scope.getFormattedTime = DiaryHelper.getFormattedTime;
+  $scope.getLocalTimeString = DiaryHelper.getLocalTimeString;
   $scope.getFormattedTimeRange = DiaryHelper.getFormattedTimeRange;
   $scope.getFormattedDuration = DiaryHelper.getFormattedDuration;
   $scope.getTripDetails = DiaryHelper.getTripDetails
-  $scope.tripgj = DiaryHelper.directiveForTrip($scope.trip);
-
-  $scope.getTripBackground = function() {
-     var ret_val = DiaryHelper.getTripBackground($scope.tripgj);
-     return ret_val;
-  }
+  $scope.tripgj = Timeline.getTripWrapper($stateParams.tripId);
 
   console.log("trip.start_place = " + JSON.stringify($scope.trip.start_place));
 
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -207,6 +207,14 @@ angular.module('emission.main.diary.list',['ui-leaflet',
       return false;
     };
 
+    $scope.explainDraft = function($event) {
+      $event.stopPropagation();
+      $ionicPopup.alert({
+        template: "This trip has not yet been analysed. If it stays in this state, please ask your sysadmin to check what is wrong."
+      });
+      // don't want to go to the detail screen
+    }
+
     $scope.isAnalyzed = function (trip) {
       return true;
     }
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -160,27 +160,6 @@ angular.module('emission.main.diary.list',['ui-leaflet',
     }
 
     /**
-     * Get all 'mode' for the trip from local cache (BEMUserCache)
-     */
-    var getLocalTripMode = function (trip) {
-      return $window.cordova.plugins.BEMUserCache.getAllMessages(MODE_CONFIRM_KEY, true).then(function (modes) {
-        Logger.log("Modes stored locally" + JSON.stringify(modes));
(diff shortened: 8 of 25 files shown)


shankari:
Testing done: https://github.com/e-mission/e-mission-docs/issues/353#issuecomment-481460050

shankari:
@deepalics0044 @PatGendre
Since this was a bit of a tricky change, I merged it to one of the channel that is essentially identical to master first for some additional testing.

Can you try out https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=martanetworksp18&clear_usercache=true&clear_local_storage=true for a little bit and let me know what you think?

I will plan on merging tomorrow morning.

deepalics0044:
Sure I'll test that :)

PatGendre:
@shankari thanks a lot!
I tested a little and it works well :-)

Note that for the first trip end, I had the app open and was asked the app to complete mode and purpose, but the notification was still active and I could re-enter the same info (the notification could be cancelled if the info is completed in the app; but this is definitely not a big problem).
Also, for the last txo trips, I had only one notification to complele with mode and purpose and it seems to me that both last 2 trips were completed with the same info, whereas I should have been asked twice. But I have to confirm this behaviour. Do you think this could be possible?
Finally, in the rk-unsw app, it was always possible to complete mode and/or purpose because there was a pair of buttons attached to the trip. I think it was a design decision from your part not to display the 2 buttons, however they could be convenient, possibly only on the detail.html trip diary screen, as this can be more practical for the user compared to answering to each notification, of which there can be many in a single day

As you explained in your comments, the manual/mode is not displayed in the maps as long as it was not processed by the pipeline but that it is not a big issue, your design decision of committing without solving any possible improvements was the right one as far as we are concerned.

shankari:
> Finally, in the rk-unsw app, it was always possible to complete mode and/or purpose because there was a pair of buttons attached to the trip. I think it was a design decision from your part not to display the 2 buttons, however they could be convenient, possibly only on the detail.html trip diary screen, as this can be more practical for the user compared to answering to each notification, of which there can be many in a single day

@PatGendre I did merge over the pair of buttons attached to each trip as well; see the description above "Merge the buttons into the list view" and details in the related issue. The rk-unsw team has disabled notifications for their app version and so can you (at the bottom of `www/js/tripconfirm/post-trip-prompt.js`). It really depends on whether you think that the notifications are useful or annoying :) With this version of the prompt, if the user thinks that the notifications are annoying, she can mute them and just classify the trips at the end of the day. If she thinks that they are useful reminders, she can keep them turned on.

>  Also, for the last txo trips, I had only one notification to complele with mode and purpose and it seems to me that both last 2 trips were completed with the same info, whereas I should have been asked twice. But I have to confirm this behaviour. Do you think this could be possible?

Don't think so, but no group has extensively tested the notification-based confirmation because everybody has been worried about getting users pissed off and deleting the app.

PatGendre:
Thanks, useful to know that the notification can be disabled, and you're right, it is a good idea to have added the mute/snooze buttons.

> "Merge the buttons into the list view"

Where do you configure the app so as to display the buttons? Is it in `www/js/diary/list.js` ?
Not to clear to me, sorry...

> The rk-unsw team has disabled notifications for their app version and so can you (at the bottom of www/js/tripconfirm/post-trip-prompt.js).

-> the file is in incident directory, not tripconfirm? `www/js/incident/post-trip-prompt.js`

I will try to test a little more...

shankari:
@PatGendre ah I realize why you didn't see the buttons in the list view; they are only displayed on cleaned trips 

```
             <div ng-if="isAnalyzed(tripgj)" class="row" style="padding-left: 5px;padding-right: 5px;">
                <div class="col-50" style="text-align: center;font-size: 14px;font-weight: 600;">
                Mode
                </div>
                <div class="col-50" style="text-align: center;font-size: 14px;font-weight: 600;">
                Purpose
                </div>
             </div>
```

and I'm not running the pipeline on the server just yet. And I think that they are displaying the buttons only for cleaned trips because they are matching up the mode and purpose using == instead of the more complex logic in `get_user_input_for_trip`.

So basically:
- the current version only works for people who are running the pipeline periodically through a cronjob and are not using trip-end notification
- it is fairly trivial to change it to work only for people who are NOT running the pipeline periodically and then it will also work for trip-end notification

ugh. This is why I decided to merge the change myself. I can do some hacks, but we are in this position because we did a bunch of hacks. Even if I don't refactor the code fully, I am going to:
- change the matching logic to the more complex `get_user_input_for_trip`
- enable the buttons for non-cleaned trips as well

I think that should fix it in terms of functionality at least for now

PatGendre:
ok thanks a lot, twice ;-)

shankari:
@PatGendre @deepalics0044 new changes are checked in and seem to work.
@MythriNagaraju you definitely want the changes to the list view.
You are welcome to use the changes to the post-trip prompt, but can comment them out (in `www/js/tripconfirm/post-trip-prompt.js`) if you don't.

PatGendre:
Great job!
A quick test, the buttons work but the app is very slow so I couldn't test purpose choice, and Other mode.
I attach a screenshot, the buttons are slightly off the frame, but this may be due to my old Android4 phone ;-)
Screenshot_2019-04-11-21-53-35

shankari:
>  A quick test, the buttons work but the app is very slow so I couldn't test purpose choice, and Other mode.

I strongly suspect this is because of the fact that checkMode and checkPurpose are called almost continuously as the UI is updated.

```
                    <div ng-if="checkMode(tripgj)" class="mode-confirm-container">
```

That is part of the bigger restructuring. I was going to defer that, but may move it up if others report the same behavior

shankari:
@deepalics0044 @MythriNagaraju can you provide feedback on the performance of the fix and how much I should prioritize additional changes?

PatGendre:
I confirm that the diary screen is much too slow, currently impossible to use.

jf87:
I just tried this. It seems to work fine for me (but I am also running the processing pipeline regularly on the server).

Would it be possible to also add the feature to let users tag the modes of single segments in the detailed view of a trip?

For example, below is a trip which consists in reality of walk, bus, metro, walk.

IMG_6938

IMG_6939

shankari:
@jf87 filed new issue https://github.com/e-mission/e-mission-docs/issues/374 for trip editing. 
Also, as an aside, which branch are you on? I notice that you have motorized section that is "secs" long, and I thought that I had eliminated those in the GIS based pipeline (https://github.com/e-mission/e-mission-server/pull/578)

If you are on the new pipeline, can you send me your timeline for the day so I can figure out why that is not fixed. The trouble with GIS-based mode inference is that you have to get the section segmentation to be perfect, otherwise it won't match up with bus stops and will be detected as car.

jf87:
@shankari This is still the pipeline from master branch. I'll try and merge with new GIS-based branch.
However, this trip editing feature is still relevant :-)

jf87:
I just got the new pipeline working, just FYI:

IMG_6940

IMG_6941

Anyway, I will be working on improving this during the next days.

shankari:
@jf87 

>  However, this trip editing feature is still relevant :-)

completely agree

> Anyway, I will be working on improving this during the next days.

You have my undying gratitude

shankari:
I have some tools to visualize various outputs better that I wrote for comparing the regular branch and the GIS branch and that I haven't checked in yet. If you are actually going to work on this, I will try to dig them up and prioritize checking them in.

shankari:
Testing done for all changes:
- Set a regular mode from the list view; ensure that it is updated and stays on page refresh
- Set regular mode/purpose from trip-end prompt; ensure they show up in the list view
- Set other mode from the list view; ensure that it is updated and stays on page refresh; ensure that the new mode shows up in the popup
- Set the other mode from the trip-end prompt; ensure that it shows up in the list view and in the popup

shankari:
@PatGendre @deepalics0044 @jf87 I think that this is now ready for re-testing.
A lot of the fixes were to the "Other" selection, so you might want to test that even if you typically do use the standard modes. And the performance should be a lot better.

PatGendre:
@shankari Thanks! For retesting, shall I use update the emtriplog app and use the greentrip or interscity UI?
Or shall I update the phone-app and deploy on our own server? (as @jf87 seems to have done?)

shankari:
@PatGendre et al The changes are deployed on the same testing channel as before
https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=martanetworksp18&clear_usercache=true&clear_local_storage=true

The one part that is sub-optimal is that the formatting of the buttons seems to be slightly off. I am not sure why - I ported over the HTML and css unchanged. If somebody who is more design-oriented than me wants to take a look at that, it would be great.

shankari:
ok so I tried to reproduce again by loading the same trip into rk-unsw and the current branch and they are identical. Based on this, and the way that the buttons moved when loading the Junipero Serra data, I suspect that this is related to the place name length causing a small amount of overflow.

| rk-unsw            | this branch        | this branch with selections |
|----------------|-----------------|----------------------------|
| Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 09 02 35 |  Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 09 06 42 | Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 09 07 13 |

shankari:
so on further investigation, the issue turned out to be with the draft mode. The `(draft)` text was placed in a row below the mode icon, so it moved everything down and caused the bottom of the buttons to get cut off.

I removed that text while connected in the emulator, and it worked. To fix it, rather than tweak it the spacing, I just removed the mode type and % which were always going to be UNKNOWN and 100% anyway, and replaced them with a single draft button. This made the draft nature much more clear, and provided an explanation of the draft mode when clicked.

| Draft text below mode | Draft text gone - spacing fixed! | Mode replaced with draft button (final) | Works in non-draft as well |
|---------------------|------------------|--------------------------|-----------------------|
| Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 09 22 45 | Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 09 36 50 | Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 12 50 09 | Simulator Screen Shot - iPhone 6 Plus - 2019-04-20 at 12 51 53 |

shankari:
@PatGendre et al, just pushed the new version to https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=martanetworksp18&clear_usercache=true&clear_local_storage=true

Of course, you can also rebuild your own custom app...

PatGendre:
@shankari Hi, it works fine, thanks a lot :-) Reponse time is good, buttons are nicely spaced, and the other field is Ok as well !

PatGendre:
@shankari I noticed a little problem with the "other" mode : when I first tried to select other, I had to give a name to the new "other" mode, and then this new mode is added to the list of other modes, and I could add a second one. However, I couldn't select the other mode : when selecting other, I have to fill in a new mode.

Same behavior for purpose : you can add a new purpose, it is added to the purpose list, but afterwards ou cannot select it in the list, you can only a new "other purpose".

Also, after playing 5' with the diary screen and pressing 10x the mode/purpose buttons, they don't seem to work any more, I click on mode or purpose and nothing happens.
Last question : do you run the pipeline on the martanetwork data? (my trips date back from April 10 are still in draft mode, by the way, the explanation when clicking on draft is a great idea!).

deepalics0044:
The above changes seems to work fine for me. Thank you.

As mentioned by  @PatGendre  when we enter other and further if we want to change, we see the previous other option in the list but cannot select it.

deepalics0044:
Do I need to swap files in my local folder with 'the set of files modified in this PR' in order to build an apk or send you a PR for changing my client.

shankari:
@deepalics0044 https://stackoverflow.com/questions/6743514/how-can-i-fetch-an-unmerged-pull-request-for-a-branch-i-dont-own

shankari:
@deepalics0044 @PatGendre @jf87 and anybody else who is testing this, I just merged the fixes to https://github.com/e-mission/e-mission-phone/pull/563 into the https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=martanetworksp18&clear_usercache=true&clear_local_storage=true channel as well. This is to allow @asiripanich to test out the embedded enketo forms before they merge.

So you may see a prompt for a new UI update. That is not a bug, there really is a new UI update. If you want to test out the new embedded forms, you can click on the "TEST SURVEY" button at the top of the metrics tab. The form was created using the kobotoolbox form builder (https://www.kobotoolbox.org/), which generates an xlsform (editable in excel!) -> xform -> enketo formatted json -> display.

Phew!

PatGendre:
@Shankari 
Hi, thanks for this update, however I still cannot select a mode in the other modes list.
Moreover, after using the diary screen for a while, the mode and purpose buttons become ineffective, as in my preceding comment. 
The survey form seems to work, including the draw route feature (although not very convenent on a phone).

shankari:
@PatGendre I did not actually fix the mode and purpose buttons. As you can see from my note, I merged another set of changes (https://github.com/e-mission/e-mission-phone/pull/563) to the same channel, and I just didn't want people to think that the UI update detection was broken.

I've spent this week primarily working on the embedded enketo forms.

The current form is just a sample/POC to illustrate how the forms can be embedded. @asiripanich will now create the real forms that the Australia team will use. I agree with you that the map prompts are not that useful, and can be inferred with much higher granularity from the sensed data anyway. IMHO, the survey should focus on data that can't be inferred from sensing, such as trip quality.

The reason that I spent time on this was to enable the basic embedding functionality in the platform so that we now have a survey builder. Any deployer who follows the steps above kobotoolbox form builder (https://www.kobotoolbox.org/), which generates an xlsform (editable in excel!) -> xform -> enketo formatted json -> display can get a form that they can display with one line of code and whose results they can manipulate before sending to the e-mission server.

PatGendre:
Ok, thanks! Definitely the survey functionality is useful too.
Actually I wasn't expecting the mode/purpose fix this week but tried it anyway.

shankari:
Here's the screenshot with the fix

shankari:
> Also, after playing 5' with the diary screen and pressing 10x the mode/purpose buttons, they don't seem to work any more, I click on mode or purpose and nothing happens.

I suspect this is just slowness in the trip screen due to sub-optimal coding by undergrads. I already made improvements to the performance once before, but there are still multiple calls to functions instead of $scope variables. Let me see if I can fix some more of them.

shankari:
Deployed everything upto now (mode + purpose fixes + (hopefully) performance fixes). I want to make a few more minor fixes tomorrow morning, but feel free to test right now.

PatGendre:
@shankari thanks a lot! It seems fixed:-) 
Both "other mode/purpose" selection and response time.

As mentioned by Attawit in  [issue 563](https://github.com/e-mission/e-mission-phone/pull/563#issuecomment-489320977) there is a small regression in the UI (e.g. date picker) but you explained it is due do CSS, not a big problem for us anyway.

PatGendre:
Currently the martanetwork app doesn't have a pipeline running : there may be a UI change induced by the mode/purpose feature. When displaying a validated trip instead of a draft trip, there will be a detected mode displayed and it should be shown in a different way than the mode defined by the user.  Or the detected mode could also be not displayed? Or selected as default value for the user-defined mode?

shankari:
> Currently the martanetwork app doesn't have a pipeline running

Yup, that channel also connects to the main e-mission-server server.

> When displaying a validated trip instead of a draft trip, there will be a detected mode displayed and it should be shown in a different way than the mode defined by the user. 

It is; see the screenshot in https://github.com/e-mission/e-mission-phone/pull/553#issuecomment-489391711 where the detected modes are displayed above the map (97% drive, 2% bike, 0% walk).

> Or selected as default value for the user-defined mode?

This is definitely possible, but we would need to figure out which mode to display in multi-modal trips.
But I am not going to make that change since I am not yet convinced that it is the right thing to do by default.

Individual deployments are welcome to change this; depending on what you find in your pilots, maybe you can submit a PR to merge it to master.

PatGendre:
Ok, you're right, the current UI is already Ok, and if we have requests from users we'll envisage other ways of displaying the mode.

shankari:
Testing for the detail screen, including working local time.

| List view                                |   Detail view                    |
|-----------------------------|-------------------------|
|Simulator Screen Shot - iPhone 7 - 2019-05-06 at 13 14 34 | Simulator Screen Shot - iPhone 7 - 2019-05-06 at 13 14 39 |
| Simulator Screen Shot - iPhone 7 - 2019-05-06 at 13 30 24 | Simulator Screen Shot - iPhone 7 - 2019-05-06 at 13 30 34 |

shankari:
with this, I believe I have addressed all known performance issues. The listview should be much snappier than before, in hopefully a more future-proof way as well.

Deployed new version; if this is good, will merge with master

PatGendre:
@shankari Thanks, as far as I tested, it works well and is definitely faster !
There is a small presentation defect with the datepicker, it is not blocking for us, it can be corrected in the css later on. I join the screeshot. 
Screenshot_2019-05-07-07-54-01

shankari:
@PatGendre that presentation defect is also from the survey CSS, which I will wait to merge to master. So when I merge this change and you update your apk, you should not see any presentation issues.

shankari:
@deepalics0044 also said that this works well in an email 

> In my opinion there is nothing wrong with the pull request.

so I am going ahead and finally merging to master. Woo!

PatGendre:
Ok, fine, thanks!

PR #557, opened 2019-04-14
Greentrip Control

ecvick85:
Updated control branch to only have diary, big data and profile with no suggestion system


config.xml
```diff
@@ -83,35 +83,5 @@
         <splash density="port-xxxhdpi" src="resources/android/splash/drawable-port-xxxhdpi-screen.png" />
     </platform>
     <icon src="resources/android/icon/drawable-xhdpi-icon.png" />
-    <plugin name="phonegap-plugin-push" spec="~1.9.2">
```
shankari: why are you removing these plugins from config.xml?

package.json
```diff
@@ -9,66 +9,8 @@
   "devDependencies": {
     "phonegap": "7.1.1"
   },
-  "cordova": {
-    "platforms": [
-      "ios",
-      "android"
-    ],
-    "plugins": {
-      "phonegap-plugin-push": {
```
shankari: and package.json?

ecvick85: I don't know why they were removed. I added them back in the most recent commit, and it still works while running locally.

shankari:
one final change - remove `npm-debug.log`

shankari:
Deployed this to a new channel with the same connection settings as the main greentrip channel since I have no evidence that the server has been reconfigured (@valin1? @b-i-l-l-c-a-o?)

https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=greentripcontrol&clear_usercache=true&clear_local_storage=true

@ecvick85 can you test out the deployed version?

PR #559, opened 2019-04-18
Updated Consent to match IRB

trevor-wu:
- also changed margins a bit to fix a weird bug
Screen Shot 2019-04-17 at 10 11 25 PM


shankari:
I don't think that the margin change in necessary, but I am not going to push back against it.

shankari:

PR #560, opened 2019-04-18
Consent Form + Login

trevor-wu:
Updated Consent Form to match IRB,
Updated Login so that it now displays Google instead of token list.

Screen Shot 2019-04-17 at 10 11 25 PM
Screen Shot 2019-04-17 at 9 56 00 PM
Screen Shot 2019-04-17 at 9 55 40 PM


shankari:
LGTM! Again, nice clean PR.

shankari:
deployed, should be live in ~ 10 minutes

PR #561, opened 2019-04-18
broke up button bar into 3 buttons, removed back button, review code for diary page

trevor-wu:
Screen Shot 2019-04-18 at 12 01 41 AM
Used Yali's reviews to show that review code works given a valid yelp-id. Currently don't have data that has valid suggestion.

### DIFF
--- a/www/js/diary/detail.js
+++ b/www/js/diary/detail.js
@@ -45,7 +45,7 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
   });
   $scope.name = "Cannot Retrieve Suggestion";
   $scope.mode = "Cannot Retrieve Mode";
-  $scope.bid = "yalis-stanley-hall-cafe-berkeley";
+  $scope.bid = "";
   $scope.stars = 4.5;
   $http.get('json/yelpfusion.json').then(function(result) {
         $scope.yelp = result.data;
--- a/www/js/diary/detail.js
+++ b/www/js/diary/detail.js
@@ -4,7 +4,7 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
                                       'emission.services', 'emission.plugin.logger',
                                       'emission.incident.posttrip.manual'])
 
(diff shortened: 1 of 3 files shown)


trevor-wu:
Screen Shot 2019-04-18 at 12 01 41 AM
Back button removed

www/js/diary/detail.js
```diff
@@ -42,11 +43,17 @@ angular.module('emission.main.diary.detail',['ui-leaflet', 'ng-walkthrough',
       console.log("diary/detail received resize event, invalidating map size");
       data.leafletObject.invalidateSize();
   });
-
+  $scope.name = "Cannot Retrieve Suggestion";
+  $scope.mode = "Cannot Retrieve Mode";
+  $scope.bid = "yalis-stanley-hall-cafe-berkeley";
```
shankari: I am glad you were able to test with a hardcoded value. But this code needs to be removed after testing! Otherwise, you will end up with a situation where your users see the yali's reviews even when they don't have a suggestion for yali's!

shankari:
still waiting for the test results:
> test to make sure that if there are no suggestions, there are no reviews.

trevor-wu:
Screen Shot 2019-04-18 at 10 25 22 PM

shankari:
@trevor-wu I'm sorry but now there is a broken image link below "Reviews" (the ? icon). That looks super unprofessional and should be fixed.

shankari:
This is why I ask for testing before submitting the PR!

trevor-wu:
Now when there are no reviews, no stars show up either. Review stars changed to be called by getReview function.
Screen Shot 2019-04-19 at 1 05 41 AM

shankari:
ok, so I believe now that this works if there is no suggestion, but did you check that it works if there is a suggestion? Does your change to set the rating along with the reviews actually work?

trevor-wu:
Screen Shot 2019-04-20 at 1 44 37 AM
Before getting suggestion
Screen Shot 2019-04-20 at 1 44 45 AM
After clicking get suggestion
Screen Shot 2019-04-20 at 1 44 59 AM
After clicking get review

shankari:
@trevor-wu LGTM!

PR #566, opened 2019-05-08
Unify the handling of error messages + fix error during force sync

shankari:
- Define a new function (displayError) that deals with undefined messages,
  formatting the error stack, etc.
- Replace all .catch(function(error)) calls with a call to displayError so that
  we can:
  - have a unified format that we can modify in one location
  - display meaningful error messages that should make it easier to figure out
    what is wrong.

This makes the error message from https://github.com/e-mission/e-mission-docs/issues/378
more meaningful


shankari:
| Before this change | After this change |
|-----|------|
| Simulator Screen Shot - iPhone 7 - 2019-05-07 at 21 04 20 | Simulator Screen Shot - iPhone 7 - 2019-05-07 at 21 03 03 |

PR #571, opened 2019-05-23
UI requested changes with distance color and new icon for recommendat…

sam11b:
New icon for recommendations tab is now speakerphone. Color for no distance data changed from yellow to green matching rest of app and easier to read.

### DIFF
--- a/www/templates/main-metrics.html
+++ b/www/templates/main-metrics.html
@@ -132,7 +132,7 @@
           <ion-slide id="modes">
             <div id="dashboard-bottom" class="card">
               <h4 class="dashboard-headers">Distance</h4>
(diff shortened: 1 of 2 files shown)


shankari:
@sam11b screenshots?

sam11b:
Screenshot_1558593012

shankari:
@sam11b I am merging this for now, but do you have a local server set up for additional testing in case it is needed? I don't see any data in this screenshot, so I assume not. I would strongly encourage you to have that ready and not wait to set it up under pressure.

shankari:
deployed, should be live in ~ 10 mins. Note that the current deployment also includes the merge from master - i.e. the basic mode and purpose changes.

shankari:
@sam11b do you plan to make similar changes to the control branch as well?

sam11b:
I can do that as well (specifically for the distance text color).

PR #572, opened 2019-05-23
Mode/Purpose Confirm Survey

atton16:
In this PR we are trying to implement:
1. Survey display for Mode & Purpose confirmation
2. Survey display with previous answer restoration (if any)
3. Dynamic button label display based on previously answered survey on both Mode Button and Purpose Button

At this point, I am struggle at Point (2). It appears that somehow `$window.cordova.plugins.BEMUserCache.getAllMessages` complicate things up.

If you look at the code in the file `js/survey/enketo-survey-services.js` line `55-64` never get executes.

Here is the output:
```
[phonegap] [console.log] EnketoSurvey.init()
[phonegap] [console.log] try to restore answer
[phonegap] [console.log] DEBUG:Answers stored locally[{"trip_properties":{"end_ts":1437582214.517,"start_ts":1437578093.881},"data":"<trip-end-survey_v5 xmlns:jr=\"http://openrosa.org/javarosa\" xmlns:odk=\"http://www.opendatakit.org/xforms\" xmlns:orx=\"http://openrosa.org/xforms\" id=\"trip-end-survey_v5\" version=\"veMfY6eZ2W9qoGhmfzJX8F\"><start>2019-05-23T17:19:43.089+07:00</start><end>2019-05-23T17:19:43.098+07:00</end><group_vs9ga59><trip_purpose_main>1</trip_purpose_main><trip_purpose_sub_category/></group_vs9ga59><group_ee9rk21><travel_mode_main>walk</travel_mode_main><travel_mode_sub_category/><Total_people_in_trip_party>1</Total_people_in_trip_party><Non_household_member_s_on_trip>0</Non_household_member_s_on_trip><Household_member_s_on_trip>1</Household_member_s_on_trip><group_bn3jd79><Vehicle_trip_with_2_Driver_or_passenger/><Vehicle_trip_Parking_location/><Parking_cost/></group_bn3jd79><Rail_payment_method/><Transit_fees_AUD/><Taxi_fees/></group_ee9rk21><__version__>vLX8oPb8xP2pLvxNBagygU</__version__><_version_>vwpCadzPYVsW9XEjxZzwnR</_version_><_version__001>vSuFDQ2DcmpRpKcRfSXReq</_version__001><_version__002>veMfY6eZ2W9qoGhmfzJX8F</_version__002><meta><instanceID>uuid:69bb39ae-6420-471a-afe0-3012e4de23c8</instanceID></meta></trip-end-survey_v5>"}]
[...]

PR #575, opened 2019-05-26
Cherry-pick mode and purpose survey changes to make them easier to review

shankari:
Original PR: https://github.com/e-mission/e-mission-phone/pull/572

Commands to generate this PR:
```
git checkout --track upstream/rk-unsw
git pull upstream rk-unsw
git checkout -b rk-unsw-survey rk-unsw
git cherry-pick a41de7a6a528a7f16e3b6f333eab9084f1e51f12..87499054f5ac5b3676fd940343f52c24e2d2c2e1
```

Testing done: None (assume that testing was done in original PR)


shankari:
Doh! Created PR to wrong branch.

shankari:
Let's see if we can just re-base

shankari:
Other commands:

```
git revert 5d3c689c62ac3b10f925c3e98e6118fad86f0125
git mv www/js/tripconfirm/trip-confirm-service.js www/js/tripconfirm/trip-confirm-services.js
```

+ a bunch of manual whitespace reverts

shankari:
One final command

```
git cherry-pick a41de7a6a528a7f16e3b6f333eab9084f1e51f12
```

It looks like the `git cherry-pick` excludes the first commit

shankari:
Testing done: see attached video

[test_survey_launch.mp4.gz](https://github.com/e-mission/e-mission-phone/files/3221019/test_survey_launch.mp4.gz)

shankari:
@atton16 this is my definition of a clean PR.
Note that there are no changes that are not relevant to the feature/function being implemented.
The change is still pretty large (21 files!) but you can see how the changes fit together.

Please use this as the model for future functionality PRs.
I welcome formatting changes as well, but in a separate PR, which contains NO functionality changes, ONLY formatting changes.

Also, note that comment with the testing done :)

www/index.html
```diff
@@ -62,7 +62,7 @@
     <script src="manual_lib/enketo/enketo-bundle.js"></script>
     <!-- Manually installed javascript for libraries that are not published through bower -->
     <!-- cordova script (this will be a 404 in the browser) -->
-    <script type="text/javascript" src="cordova.js"></script>
+    <script src="cordova.js"></script>
```
shankari: you don't need to revert this change in the dev branch. I have already done so in the deployment branch. Keep it around and enjoy the benefits of live reload.

www/index.html
```diff
@@ -95,7 +95,8 @@
     <script src="js/common/place-detail.js"></script>
     <script src="js/common/services.js"></script>
     <script src="js/survey/launch.js"></script>
-    <script src="js/survey/enketo_launch.js"></script>
+    <script src="js/survey/enketo-survey.js"></script>
```
shankari: I am not sure why you needed to rename `enketo_launch.js` to `enketo-survey.js` but it was a small file so I am not going to hold up the change for this.

www/js/app.js
```diff
@@ -63,6 +63,11 @@ angular.module('emission', ['ionic',
             Logger.displayError("Error reading or setting connection defaults", err);
         });
     });
+
+    $rootScope.$on('$stateChangeStart', function (event, toState, toStateParams, fromState, fromParams) {
```
shankari: do you still need this? In an earlier change, you had indicated that this was primarily for development - e.g. 

> That is only for development purpose. It appears that if I opened this page and the code got reloaded, the back button disappear. This help fix it by navigate back to diary page. It's not perfect but it works on the development environment and it should not interfere with the live deployment.

www/js/diary/list.js
```diff
@@ -150,7 +165,8 @@ angular.module('emission.main.diary.list',['ui-leaflet',
      * Embed 'mode' to the trip
      */
     $scope.populateModeFromTimeline = function (tripgj, modeList) {
-        var userMode = DiaryHelper.getUserInputForTrip(tripgj.data.properties, modeList);
+        if (!modeList) return;
```
shankari: why do you need this? the timeline code fills the data with a blank list if there are no modes

www/js/diary/list.js
```diff
@@ -171,7 +187,8 @@ angular.module('emission.main.diary.list',['ui-leaflet',
      * Embed 'purpose' to the trip
      */
     $scope.populatePurposeFromTimeline = function (tripgj, purposeList) {
-        var userPurpose = DiaryHelper.getUserInputForTrip(tripgj.data.properties, purposeList);
+        if (!purposeList) return;
```
shankari: Ditto.

www/js/diary/list.js
```diff
@@ -740,5 +771,11 @@ angular.module('emission.main.diary.list',['ui-leaflet',
           $rootScope.displayingIncident = false;
         }
       });
+
+      $rootScope.$on('$stateChangeStart', function (event, toState, toStateParams, fromState, fromParams) {
+        if (fromState.url === '/enketosurvey/:form_location/:opts' && toState.url === '/diary') {
+          readAndUpdateForDay(Timeline.data.currDay);
```
shankari: This seems super heavyweight. why not just modify the $scope variables like I do with the  mode and purpose changes? The `$scope` variables will be invalidated on the next update, but that will fill them in with the new values anyway.

www/js/survey/enketo-survey.js
```diff
@@ -0,0 +1,43 @@
+'use strict';
```
shankari: when you added this, did you use `rm` followed by `add` or `mv`? `mv` retains the git history. Again, this is not critical for the current file which was pretty simple to begin with, but is very critical if you begin refactoring other files that have accumulated history.

www/js/tripconfirm/trip-confirm-services.js
```diff
@@ -1,4 +1,4 @@
-angular.module('emission.tripconfirm.services', ['ionic', "emission.plugin.logger"])
```
shankari: why is this rename critical?

www/templates/diary/list.html
```diff
@@ -49,7 +49,7 @@
                 </div>
 
 
-                <div id="diary-card" ng-class="tripgj.listCardClass" ng-style="{height: '335px'}">
+              <div id="diary-card" ng-class="tripgj.listCardClass" ng-style="{height: '315px'}">
```
shankari: What is wrong with the previous size?

www/index.html
atton16: I still got this part in my code. I only do it to make the remote clean.

www/js/diary/list.js
atton16: I did try with `$scope` before I jump to this. There must be something with `$state.go` and the Back button on the nav bar. Apparently, the button label does not get updated if I use `$scope` unless I refresh the entire timeline.

www/js/diary/list.js
atton16: If modeList is not an array, `DiaryHelper.getUserInputForTrip` will throw an error. When I implement the survey into `list.js` with a clean data, it appprears that `modeList` is not an object anymore.

www/js/survey/enketo-survey.js
atton16: Not sure, I use vscode.

www/js/tripconfirm/trip-confirm-services.js
atton16: It just when we started to discussing refactoring so I think my mind started to refactor as well.

www/templates/diary/list.html
atton16: We get rid of the label on top and simply display button. So I adjust the height of the container to match.
Screen Shot 2019-05-25 at 21 25 22

www/js/diary/list.js
atton16: I which you could help figure out a good way to refresh the data without reloading the timeline.

www/index.html
shankari: I think it is better to keep these changes in the branch. There is a separate deployment branch where I pull the changes before pushing to the channel and I have removed this part on that branch. This way, if anybody else wants to start developing in the branch, they will have live reload on by default.

www/js/diary/list.js
shankari: hm, when I test it in an environment which has only mode data, I get the purpose and survey as empty lists. I bet this is because you were originally using the usercache directly - the UnifiedDataLoader is guaranteed to always return a list 

```
          var localResult = [];
          var localError = null;

          var remoteResult = [];
          var remoteError = null;
...
      var combinedList = list1.concat(list2);
```

www/js/survey/enketo-survey.js
shankari: Looks like in vscode you need to commit as soon as you rename and modify the file in a subsequent commit. If you do that, git will automatically detect the mv; otherwise it won't.
https://github.com/Microsoft/vscode/issues/6127

www/js/diary/list.js
shankari: can you tell me what you tried? In particular, did you use `$scope.$apply` around the scope updates?

www/js/survey/enketo-survey-service.js
```diff
@@ -0,0 +1,137 @@
...
+    .then(function(answers){
+      if (!_opts_populateLabels) return answers;
+
+      const xmlParser = new $window.DOMParser();
+      if (key === 'manual/confirm_survey') {
+        return answers.map(function(answer){
+          const xmlStr = answer.data.survey_result;
+          const xml = xmlParser.parseFromString(xmlStr, 'text/xml');
+          // Data injection
+          answer.travel_mode_main = _parseAnswerByTagName(xml, 'travel_mode_main');
```
shankari: Potential enhancement: now that you have pulled this out into a separate function, you could theoretically create a JSON file where users can specify the tags and potentially how to format them into the string. Then it is easier for other travel survey people to customize the displayed button without modifying code.

e.g.

```
{'extracted_tags': ['travel_mode_main', 'o_purpose_main', 'd_purpose_main'],
 'display': '<b>{{tripgj.userSurveyAnswer.extracted_tag[0]}}</b> from <b>{{tripgj.userSurveyAnswer.extracted_tag[1]}}</b> to <b>{{tripgj.userSurveyAnswer.extracted_tag[2]}}</b>'
}
```

This is definitely optional, depending on how easy you want to make it for future deployers to customize the UX and how comfortable you think they will be with code v/s editing a JSON file.

And even if you want to do it, I would suggest doing so as part of a separate PR. This one is complex enough already and it works.

The goal is to go from working -> working...

www/index.html
atton16: Oh okay, so the live channel will always remove this code.

www/js/app.js
atton16: This part is now mandatory. It helps to go back to the diary page when user save the survey.

www/js/diary/list.js
atton16: Here is what I did
```
                $scope.$apply(function() {
                    $scope.data.currDayTripWrappers.forEach(function(tripgj, _index, _array) {
                        $scope.populateSurveyAnswerFromTimeline(tripgj, $scope.data.unifiedConfirmsResults.surveyAnswers);
                    });
                });
```

atton16: This sometimes make the UI freeze and sometimes the router navigate back to `diary/list.js` but nothing happens.

www/index.html
shankari: Yup. I will ensure that the live channel will not have this code; you can remove it from yours.

shankari:
> Is your git-fu strong enough to pull from this PR, making the changes and generating a new PR? 

Are you able to generate a new PR?

> Please ensure that the new PR again is clean - e.g. has only the changes that are relevant to this feature.

To do this, you should turn off autoformatting in visual studio.
https://stackoverflow.com/questions/737222/turn-off-auto-formatting-in-visual-studio

atton16:
> > Is your git-fu strong enough to pull from this PR, making the changes and generating a new PR?
> 
> Are you able to generate a new PR?
> 
> > Please ensure that the new PR again is clean - e.g. has only the changes that are relevant to this feature.
> 
> To do this, you should turn off autoformatting in visual studio.
> https://stackoverflow.com/questions/737222/turn-off-auto-formatting-in-visual-studio

I normally use Sourcetree for git-related tasks. But I can make sure that auto formatting in vscode is off.

www/js/diary/list.js
shankari: I think this is because you are not actually modifying `$scope.data.unifiedConfirmsResults` when the user fills out the survey. See what I do with `mode_confirm` and `purpose_confirm`. I just set the `tripgj.userSurveyAnswer` directly.

atton16: If you had a look inside ` $scope.populateSurveyAnswerFromTimeline`, it try to set `tripgj.userSurveyAnswer` already.

shankari: That's true. But it does so from `$scope.data.unifiedConfirmsResults`. And `$scope.data.unifiedConfirmsResults.surveyAnswers` is populated *only* as part of `readAndUpdateForDay` - e.g. in

https://github.com/e-mission/e-mission-phone/pull/575/files#diff-9bf2100b27e218fd87648699604df6f9R922

So if you call `$scope.populateSurveyAnswerFromTimeline` on a newly filled out survey, the survey will not be in `$scope.data.unifiedConfirmsResults.surveyAnswers` and nothing will be populated.

atton16: I also did try with this

```js
                $scope.$apply(function() {
                    $scope.data.currDayTripWrappers.forEach(function(tripgj, _index, _array) {
                        $scope.populateSurveyAnswerFromTimeline(
                            tripgj,
                            EnketoSurvey.getAllSurveyAnswers("manual/confirm_survey", { populateLabels: true })
                        );
                    });
                });
```

but still no lucks.

shankari: This should work; I am happy to have a short call to debug why it doesn't, just like last time.
Although it is still not a perfect solution.

In particular, you read `EnketoSurvey.getAllSurveyAnswers` for each trip, although the results are not trip-specific, and that (since it uses the unified data loader), effectively makes both local and remote calls.

Why must you use `populateSurveyAnswerFromTimeline`? Why can't you set the `$scope` variables like I do with `mode_confirm`?

atton16: Ok, I think I have figured it out.

I forgot that `EnketoSurvey.getAllSurveyAnswers` returns a promise.

```javascript
                EnketoSurvey.getAllSurveyAnswers("manual/confirm_survey", { populateLabels: true }
                ).then(function(answers) {
                    $scope.$apply(function() {
                        $scope.data.currDayTripWrappers.forEach(function(tripgj, _index, _array) {
                            $scope.populateSurveyAnswerFromTimeline(
                                tripgj,
                                answers
                            );
                        });
                    });
                });
```

Does work.

atton16: So how should I submit the PR ? Or perhaps can you make a commit to `shankari:rk-unsw-survey` ?

shankari: @atton16 that is a lot better since you make a single call to `getAllSurveyAnswers` for every state change. But if people are filling out multiple surveys for a particular day (maybe at the end of the day), then you will still make multiple calls to `EnketoSurvey.getAllSurveyAnswers`.

I'd still like to know why you don't just modify the in-memory variable `tripgj.userSurveyAnswer` directly. That is the obvious UX change in a controller; the user makes a change and you reflect it immediately in a `$scope` variable while also saving remotely for persistence.

Why do you have to read the saved value from the database?

shankari: > So how should I submit the PR ? Or perhaps can you make a commit to shankari:rk-unsw-survey ?

I already asked you "Is your git-fu strong enough to pull from this PR, making the changes and generating a new PR?" What is the answer?

atton16: > > So how should I submit the PR ? Or perhaps can you make a commit to shankari:rk-unsw-survey ?
> 
> I already asked you "Is your git-fu strong enough to pull from this PR, making the changes and generating a new PR?" What is the answer?

So I can directly commit to `shankari:rk-unsw-survey` ?

> I'd still like to know why you don't just modify the in-memory variable `tripgj.userSurveyAnswer` directly. That is the obvious UX change in a controller; the user makes a change and you reflect it immediately in a `$scope` variable while also saving remotely for persistence.
> 
> Why do you have to read the saved value from the database?

As we directed out of the page, the only was to go back is to restore to previous state. So I cannot immediately knows which `tripgj` I wanted to get. The only thing that works is to refresh the page. If remote calls is what you are worrying then maybe we can go back to `BEMUserCache`.

shankari: > If remote calls is what you are worrying then maybe we can go back to BEMUserCache.

You cannot go back to `BEMUserCache` because it will not work for data that has already been pushed to the server - e.g. surveys from the previous day.

shankari: > As we directed out of the page, the only was to go back is to restore to previous state. So I cannot immediately knows which tripgj I wanted to get. 

Since you are already using `$rootScope` as part of the state navigation, you can also use it pass in the `tripgj` that is effectively part of the survey state. That is what we do in the detail view.

This also has the benefit that the survey options can be modified based on trip characteristics (e.g. speed? time of day?)

shankari: > So I can directly commit to shankari:rk-unsw-survey ?

No, you would pull from my branch into a new branch, make your changes and generate a new PR - e.g. https://stackoverflow.com/questions/9153598/how-do-i-fetch-a-branch-on-someone-elses-fork-on-github

shankari: @atton16 if it turns out that your git-fu is not enough to strong enough to figure out to do this, I am open to making the changes to my branch and merging. But would be good to have a patch that I can apply since the changes to use `tripgj` are more complex than this small snippet.

www/js/survey/enketo-survey-service.js
```diff
@@ -0,0 +1,137 @@
...
+    });
+  }
+
+  function displayForm() {
+    // Load survey with previous answer
+    if (__session &&
+        __session.data_key &&
+        __session.data_key === 'manual/confirm_survey'
+    ) {
+      return getAllSurveyAnswers(__session.data_key)
```
shankari: @atton16, if you would access the `tripgj` object directly in this controller, you would be able to load the form with the survey results directly instead of making another call to retrieve them

www/js/diary/list.js
shankari: > Since you are already using $rootScope as part of the state navigation, you can also use it pass in the tripgj that is effectively part of the survey state. 

One caveat with this is that you cannot call `$scope.$apply` from the survey screen since you are in a different scope. Couple of potential workarounds (neither of which are that great) involve:
1. passing in the scope variable, which a bunch of the ionic components (e.g. `$ionicModal`, `$ionicPopover`) also do anyway
1. storing the result in a different variable (e.g. `tripgj.tempSurveyAnswer` and setting `surveyAnswer` back in `list.js` within a `$scope.$apply`

PR #576, opened 2019-06-03
Update carbon calculations for GIS server (richer modes)

tms-martins:
Update the carbon calculations so that they use the richer modes in the GIS-based server ([ground_truth_matching](https://github.com/shankari/e-mission-server/tree/ground_truth_matching) branch).
- Aggregate (average) data won't work properly on local servers, due to an [outstanding issue](https://github.com/e-mission/e-mission-docs/issues/288).
- In `fillFootprintAggVals` it can happen that that the mode `ON_FOOT` has a NaN value; a workaround was provided for that, but the source of the problem should be investigated (perhaps it is due to different modes in app VS server?)


shankari:
> Aggregate (average) data won't work properly on local servers, due to an outstanding issue.

https://github.com/e-mission/e-mission-docs/issues/288 is resolved (you need to modify the CSP in `index.html`). I had left it open to track potential enhancements, but now I've opened https://github.com/e-mission/e-mission-docs/issues/408 instead.

shankari:
> In fillFootprintAggVals it can happen that that the mode ON_FOOT has a NaN value; a workaround was provided for that, but the source of the problem should be investigated (perhaps it is due to different modes in app VS server?)

Yes, the server supports both `ON_FOOT` and `WALKING` for sensed modes (the ones read from the phone API and smoothed) (`emission/core/wrapper/motionactivity.py`).

When the mode inference step is run successfully, both `ON_FOOT` and `WALKING` are collapsed to `WALKING` (`emission/core/wrapper/modeprediction.py`). You should assume that the mode is almost always `WALKING` with a fallback to `ON_FOOT` if it turns out that the mode inference step has been turned off or is not working for some reason.

www/index.html
```diff
@@ -106,6 +106,8 @@
     <script src="lib/angular-nvd3/dist/angular-nvd3.min.js"></script>
     <script src="lib/nz-tour/dist/nz-tour.min.js"></script>
     <link rel="stylesheet" href="lib/nvd3/build/nv.d3.css">
+    <!-- TIAGO: allow debug with weinre -->
+    <script src="http://192.168.1.238:8090/target/target-script-min.js#anonymous"></script>
```
shankari: Remove? Seems to be hardcoded to your IP.

www/js/metrics-factory.js
```diff
@@ -5,37 +5,61 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
...
+        result += footprint[mode] * mtokm(userMetrics[i].values);
+      }
+      else {
+        console.debug('WARNING: FootprintHelper.getFootprintFromMetrics() was requested for an unknown mode: ' + mode + " metrics JSON: " + JSON.stringify(userMetrics));
+      }
+    }
+    return result;
+  }
+  fh.getLowestFootprintForDistance = function(distance) {
+    var lowestFootprint = 9999;
```
shankari: Why `9999` instead of `Number.MAX_VALUE`?

www/js/metrics-factory.js
```diff
@@ -130,12 +154,6 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
         mets: 9.8
       }
     },
-    "IN_VEHICLE": {
```
shankari: I would retain `IN_VEHICLE` as a backup in case the mode inference step has been excluded from the pipeline or it is not running properly. There are multiple layers of backup/robustness in the e-mission code (e.g. the display in the list/detail views continues to support the `speedometer` mode for `IN_VEHICLE` trips, so would be good to be consistent.

www/js/metrics.js
```diff
@@ -410,6 +410,9 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
       delete clonedData.metric;
       clonedData.metric_list = [DURATION, MEDIAN_SPEED, COUNT, DISTANCE];
       clonedData.is_return_aggregate = true;
+      // TIAGO we should use our own server here, but for now it will return empty metrics, see:
```
shankari: I responded to this.

www/js/metrics.js
```diff
@@ -699,77 +702,91 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
-          }
-        }
+
+        $scope.carbonData.userCarbon    = FootprintHelper.readableFormat(FootprintHelper.getFootprintForMetrics(userCarbonData));
+        $scope.carbonData.optimalCarbon = FootprintHelper.readableFormat(FootprintHelper.getLowestFootprintForDistance(optimalDistance));
+        $scope.carbonData.worstCarbon   = FootprintHelper.readableFormat(FootprintHelper.getHighestFootprintForDistance(worstDistance));
+        lastWeekCarbonInt               = FootprintHelper.getFootprintForMetrics(userCarbonData);
+      }
+      else {
+        // TIAGO: -------------
```
shankari: nit: remove?

www/json/connectionConfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "connectUrl": "http://192.168.1.238:8080",
```
shankari: Remove from PR since it is related to your IP address.

www/templates/metrics/metrics-control.html
```diff
@@ -1,6 +1,6 @@
 <ion-modal-view style="background-color: #eeeeee">
 	<ion-header-bar>
-		<button side="left" class="button" ng-click="closeModal()">Cancle</button>
+		<button side="left" class="button" ng-click="closeModal()">Cancel</button>
```
shankari: Thank you so much for finally fixing this eyesore.

shankari:
> Yes, the server supports both ON_FOOT and WALKING for sensed modes (the ones read from the phone API and smoothed) (emission/core/wrapper/motionactivity.py).

So I realized that we would need to handle both `ON_FOOT` and `WALKING` correctly anyway for the calorie calculations, so I went looking for it, and I found https://github.com/e-mission/e-mission-phone/blob/master/www/js/metrics.js#L832

So it looks like we convert all `WALKING` to `ON_FOOT` anyway. Given that, I am not sure why this should be NaN. Since you are still connecting to https://e-mission.eecs.berkeley.edu for the aggregates, I can try to debug this if you file a new issue.

www/js/metrics-factory.js
```diff
@@ -5,37 +5,61 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
...
-    if (mode === "IN_VEHICLE") {
-      return [footprint.train * mtokm(distance), footprint.car * mtokm(distance)];
-    } else {
-      return footprint[mode] * mtokm(distance);
+
+  fh.readableFormat = function(v) {
+    return v > 9999? Math.round(v / 1000) + 'k kg CO₂' : Math.round(v) + ' kg CO₂';
+  }
+  fh.getFootprintForMetrics = function(userMetrics) {
+    var result = 0;
```
shankari: this is fairly straightforward, you just iterate over the various modes, look up their footprint, multiply with the distance and report the value. The main difference is that instead of returning a range, you return a single value.

www/js/metrics-factory.js
```diff
@@ -5,37 +5,61 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
...
+    return result;
+  }
+  fh.getLowestFootprintForDistance = function(distance) {
+    var lowestFootprint = 9999;
+    for (var mode in footprint) {
+      if (mode == 'ON_FOOT' || mode == 'BICYCLING') {
+        // these modes aren't considered when determining the lowest carbon footprint
+      }
+      else {
+        lowestFootprint = Math.min(lowestFootprint, footprint[mode]);
```
shankari: The main difference here is that instead of hardcoding that the lowest footprint is the train, you iterate over the modes and determine the lowest footprint from the data.

www/js/metrics-factory.js
```diff
@@ -5,37 +5,61 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
...
   }
-  fh.getFootprint = function(distance, mode) {
-    if (mode === "IN_VEHICLE") {
-      return readable(footprint.train * mtokm(distance)) + ' ~ ' + readable(footprint.car * mtokm(distance));
-    } else {
-      return readable(footprint[mode] * mtokm(distance));
+  fh.getHighestFootprintForDistance = function(distance) {
+    var highestFootprint = 0;
+    for (var mode in footprint) {
+      highestFootprint = Math.max(highestFootprint, footprint[mode]);
```
shankari: Ditto.

www/js/metrics.js
```diff
@@ -699,77 +702,91 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
-          //$scope.carbonData.userCarbon.push({key: userCarbonData[i].key, values: FootprintHelper.getFootprint(userCarbonData[i].values, userCarbonData[i].key)});
-          if (userCarbonData[i].key === "IN_VEHICLE") {
-            $scope.carbonData.userCarbon = FootprintHelper.getFootprint(userCarbonData[i].values, userCarbonData[i].key);
-            $scope.carbonData.optimalCarbon = FootprintHelper.getFootprint(optimalDistance, userCarbonData[i].key);
-            $scope.carbonData.worstCarbon = FootprintHelper.getFootprint(worstDistance, userCarbonData[i].key);
-            lastWeekCarbonInt = FootprintHelper.getFootprintRaw(userCarbonData[i].values, userCarbonData[i].key);
-          }
-        }
+
+        $scope.carbonData.userCarbon    = FootprintHelper.readableFormat(FootprintHelper.getFootprintForMetrics(userCarbonData));
```
shankari: Big difference here is that these are individual values and not arrays any more. So we can also make them readable right now.

www/js/metrics-factory.js
```diff
@@ -5,37 +5,61 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
...
   var mtokm = function(v) {
     return v / 1000;
   }
-  fh.getFootprintRaw = function(distance, mode) {
-    if (mode === "IN_VEHICLE") {
-      return [footprint.train * mtokm(distance), footprint.car * mtokm(distance)];
-    } else {
-      return footprint[mode] * mtokm(distance);
+
+  fh.readableFormat = function(v) {
```
shankari: Is this a replacement for `readable` that is exposed externally? If so, you can remove `readable`. I don't see any difference in functionality.

www/js/metrics.js
```diff
@@ -699,77 +702,91 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
-          if (aggrCarbonData[i].key === "IN_VEHICLE") {
-            $scope.carbonData.aggrVehicleRange = FootprintHelper.getFootprintRaw(aggrCarbonData[i].values, aggrCarbonData[i].key);
-            $scope.carbonData.aggrCarbon = FootprintHelper.getFootprint(aggrCarbonData[i].values, aggrCarbonData[i].key);
+          if (isNaN(aggrCarbonData[i].values)) {
+            console.debug("TIAGO: fillFootprintAggVals() WARNING: value is NaN for mode " + aggrCarbonData[i].key + ", changing to 0");
+            aggrCarbonData[i].values = 0;
           }
         }
+
+        $scope.carbonData.aggrVehicleRange = FootprintHelper.getFootprintForMetrics(aggrCarbonData);
```
shankari: Not your change, but I don't see this variable being used anywhere.
So you could remove it, since `aggrCarbon` seems to represent the same value anyway.

shankari: If not should probably rename to indicate that it is not a range any more.

www/js/metrics-factory.js
tms-martins: By the way shouldn't the comparison be ```v > 999``` (or even ```v >= 1000```)?

www/js/metrics-factory.js
tms-martins: Yes. So that if the data or dataset changes, the calculation still holds. Another option would be to pre-calculate and store this value only once (when it doesn't exist, e.g. the first time this screen is shown).

www/js/metrics.js
tms-martins: Removed :)

tms-martins:
> When the mode inference step is run successfully, both `ON_FOOT` and `WALKING` are collapsed to `WALKING` (`emission/core/wrapper/modeprediction.py`). You should assume that the mode is almost always `WALKING` with a fallback to `ON_FOOT` if it turns out that the mode inference step has been turned off or is not working for some reason.

Does this mean that in the app I will always use `WALKING`, including in the `var footprint` in `metrics-factory.js`; and that any label `ON_FOOT` should be treated as `WALKING` or even converted (in app) to 'WALKING'?

www/js/metrics.js
tms-martins: I will remove these and similar before the final PR (they have my name on them so I can find them quickly).

www/templates/metrics/metrics-control.html
tms-martins: No problem, just happened to spot it.

www/js/metrics-factory.js
tms-martins: Previously you had `IN_VEHICLE` as an average of "train" and "car". Would it make sense now to  average `TRAIN`, `CAR` and `BUS`, or would you have a different suggestion? 
e.g.
```
else if (mode == "IN_VEHICLE") {
    result += ((footprint[CAR] + footprint[BUS] + footprint[TRAIN]) / 3) * mtokm(userMetrics[i].values);
}
```

www/js/metrics-factory.js
```diff
@@ -5,37 +5,61 @@ angular.module('emission.main.metrics.factory', ['emission.plugin.kvstore'])
...
+  }
+  fh.getFootprintForMetrics = function(userMetrics) {
+    var result = 0;
+    for (var i in userMetrics) {
+      var mode = userMetrics[i].key;
+      if (mode in footprint) {
+        result += footprint[mode] * mtokm(userMetrics[i].values);
+      }
+      else if (mode == "IN_VEHICLE") {
+        result += ((footprint[CAR] + footprint[BUS] + footprint[TRAIN]) / 3) * mtokm(userMetrics[i].values);
```
tms-martins: Previously you had `IN_VEHICLE` as an average of "train" and "car". Would it make sense now to keep that option and average `TRAIN`, `CAR` and `BUS`, or would you have a different suggestion?

tms-martins:
> > When the mode inference step is run successfully, both `ON_FOOT` and `WALKING` are collapsed to `WALKING` (`emission/core/wrapper/modeprediction.py`). You should assume that the mode is almost always `WALKING` with a fallback to `ON_FOOT` if it turns out that the mode inference step has been turned off or is not working for some reason.
> 
> Does this mean that in the app I will always use `WALKING`, including in the [`var footprint` in `metrics-factory.js`](https://github.com/e-mission/e-mission-phone/blob/b0f31498099f4c1dd635779c66228e78d219fc48/www/js/metrics-factory.js#L10); and that any label `ON_FOOT` should be treated as `WALKING` or even converted (in app) to 'WALKING'?

@shankari need your input on this, I think it's the only outstanding issue in the PR - besides removing all my test code (e.g. `console.debug()`'s) of course 😉

www/js/metrics-factory.js
shankari: Hm, let me map out the current behavior and see if it makes sense:
100 kg -> 100 kg
998 kg -> 998 kg
9998 kg -> 9998 kg
9999 kg -> 9.9 k kg

Which does seem wrong.
Yeah I agree that this should be `> 999`.

Also maybe instead of 9.9 k kg, we should use 9.9 `mega gm`?
The standard SI units definitely don't have `kilo kilo`, 10^6 is mega.
On the other hand, `mega gm` looks unwieldy and `Mg` is not very well known.
UX thoughts?

www/js/metrics-factory.js
shankari: average is fine by me

www/js/metrics-factory.js
shankari: I think that maybe you can document that as a potential future optimization, but we can move on now. premature optimization is the root of all evil.

www/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://api.ionic.io/push/tokens emission: 'unsafe-eval'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com https://*.tile.stamen.com http://*.tile.stamen.com 'unsafe-inline'">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org http://192.168.1.238 https://api.ionic.io/push/tokens emission: 'unsafe-eval'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com https://*.tile.stamen.com http://*.tile.stamen.com 'unsafe-inline'">
```
shankari: also remove from PR :)

shankari:
> Does this mean that in the app I will always use WALKING, including in the var footprint in metrics-factory.js; and that any label ON_FOOT should be treated as WALKING or even converted (in app) to 'WALKING'?

We already do this conversion in the app, but in the reverse direction.
https://github.com/e-mission/e-mission-phone/blob/master/www/js/metrics.js#L832

Sorry about the lack of consistency :(

I can fix that or you can fix it as part of the PR.
At the end, we should use WALKING everywhere, and the app should convert any `ON_FOOT` to `WALKING`

www/js/metrics-factory.js
tms-martins: Agreed :)

www/js/metrics-factory.js
tms-martins: I'll change it to `> 999` for now. I can venture that big numbers should be displayed for rhetoric purposes, i.e. as the means to emphasise the amount/impact.

tms-martins:
> We already do this conversion in the app, but in the reverse direction.
> https://github.com/e-mission/e-mission-phone/blob/master/www/js/metrics.js#L832
> 
> At the end, we should use WALKING everywhere, and the app should convert any `ON_FOOT` to `WALKING`
> 
> LMK what you prefer.

When using the GIS-based server, it seems that `ON_FOOT` is only introduced there in `getDataFromMetrics()` since all the data I get from the server uses `WALKING`. So I would simply remove this conversion, if it doesn't break anything else (?).

It also begs the question: should I ever expect the mode `RUNNING`? If so, I have to either provide a carbon metric for this, or treat it as `WALKING` when calculating carbon footprints. 

Nevertheless, should I **always** check server data for old modes (i.e. `ON_FOOT`) and replace with new ones (i.e. `WALKING`) before proceeding?

www/js/metrics.js
```diff
@@ -692,77 +699,82 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
...
+
+        // TIAGO: please remove before PR, and turn into an issue?
+        // It can happen that the mode "ON_FOOT" has a NaN value.
+        // This may be due to the fact that the aggregated values are requested
+        // from the main e-mission server and not from our own.
         for (var i in aggrCarbonData) {
-          if (aggrCarbonData[i].key === "IN_VEHICLE") {
-            $scope.carbonData.aggrVehicleRange = FootprintHelper.getFootprintRaw(aggrCarbonData[i].values, aggrCarbonData[i].key);
-            $scope.carbonData.aggrCarbon = FootprintHelper.getFootprint(aggrCarbonData[i].values, aggrCarbonData[i].key);
+          if (isNaN(aggrCarbonData[i].values)) {
```
tms-martins: This has been investigated a bit more using the master phone branch and turned into an issue.
[https://github.com/e-mission/e-mission-docs/issues/422](https://github.com/e-mission/e-mission-docs/issues/422)

www/js/metrics.js
```diff
@@ -844,11 +844,11 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
                 // Here, we check if the string is all upper case by
                 // converting it to upper case and seeing if it is changed
                 if (field == field.toUpperCase()) {
-                    if (field === "WALKING" || field === "RUNNING") {
-                      fieldPhone = "ON_FOOT";
-                    } else {
+                    // if (field === "WALKING" || field === "RUNNING") {
+                    //   fieldPhone = "ON_FOOT";
```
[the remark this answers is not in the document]
shankari: Or I guess you are handing this in the footprint calculations (e.g. `getFootprintForMetrics`). Any reason for the change?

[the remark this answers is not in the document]
tms-martins: I knew I didn't want to convert WALKINg to ON_FOOT; but removing/commenting that code was also lazy, because then I can have RUNNING labels which aren't handled in `getFootprintForMetrics`. I think a separate branch should address the mode consistency (i.e. check for and convert certain modes) at some point. 
I will restore this line and avoid addressing the issue piecemeal before I break something :)

tms-martins:
I'm tidying up now, including removing the `connectionConfig.json`. 
It may come as a surprise at this point, but I'm not using the complete toolchain. I've been developing and testing using the DevApp on an Android device, and using Weinre to view the console and inspect UI elements. If you feel that this may open up the possibility of inconsistent behavior once the app is actually built for Android/iOS please let me know.

tms-martins:
> the github file changes list still shows the changes in that PR mixed with your changes.

Yes. I rebased my branch, and when I looked at the GitHub diff last time all looked well. I don't know exactly why this happens, although I've seen it before. I'm afraid that if I try to fix it now I end up making matters worse.

www/json/connectionConfig.json
tms-martins: Done.

tms-martins:
Tidied up. @shankari what is the best way to add testing videos? Something like G Drive or is there a Git Repo for those?

tms-martins:
> > the github file changes list still shows the changes in that PR mixed with your changes.
> 
> Yes. I rebased my branch, and when I looked at the GitHub diff last time all looked well. I don't know exactly why this happens, although I've seen it before. I'm afraid that if I try to fix it now I end up making matters worse.

I've read up that the PR-diff is made "against the common ancestor of the source branch and the target branch". Also that a quick fix is to use the `edit` button to temporarily change the base to something other than `master` and then back again, which forces Git to update the PR.

shankari:
> I'm afraid that if I try to fix it now I end up making matters worse.

No worries. I can just review off the compare since I don't expect to have a lot of review comments.

shankari:
> Tidied up. @shankari what is the best way to add testing videos? Something like G Drive or is there a Git Repo for those?

G Drive is fine. I typically make the video short and gzip it so that I can upload it to github directly (.gz extensions are supported).

www/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://api.ionic.io/push/tokens emission: 'unsafe-eval'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com https://*.tile.stamen.com http://*.tile.stamen.com 'unsafe-inline'">
```
shankari: can you retain the `https://e-mission.eecs.berkeley.edu` for now since we have that in `metrics.js` as well?

www/js/metrics.js
```diff
@@ -802,9 +802,9 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
                 // Here, we check if the string is all upper case by
                 // converting it to upper case and seeing if it is changed
                 if (field == field.toUpperCase()) {
-                    if (field === "WALKING" || field === "RUNNING") {
+                    // if (field === "WALKING" || field === "RUNNING") {
```
shankari: this seems actively wrong when you remove the `if`. It looks like you always set the field to `ON_FOOT` no matter what the original field was (e.g. `CAR` will become `ON_FOOT`). what am I missing here?

www/index.html
shankari: I just made this change myself since it is so simple.

www/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org emission: 'unsafe-eval'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com https://*.tile.stamen.com http://*.tile.stamen.com 'unsafe-inline'">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu emission: 'unsafe-eval'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' data: http://*.tile.openstreetmap.org http://tile.stamen.com https://*.tile.stamen.com http://*.tile.stamen.com 'unsafe-inline'">
```
tms-martins: Thank you! The more time I spend on this feature the more dumb mistakes I make :(

www/js/metrics.js
tms-martins: It **is** wrong 😮 I'm very sorry. My intention was to restore the original code... I'll correct that ASAP

tms-martins:
@shankari Testing video is [here](https://drive.google.com/file/d/1jcUhAVwrx1ymsd8nYOckk-de9lsECWWA/view?usp=sharing). I'm not sure if there were specific guidelines to follow; in case I'm missing something please let me know.

shankari:
Video looks good.

PR #580, opened 2019-06-09
Mode/Purpose Confirm Survey (Cherry-Pick)

atton16:
I have pulled from `shankari:rk-unsw-survey` and continue the work we have discussed here.


www/js/diary/list.js
```diff
@@ -774,7 +774,22 @@ angular.module('emission.main.diary.list',['ui-leaflet',
 
       $rootScope.$on('$stateChangeStart', function (event, toState, toStateParams, fromState, fromParams) {
         if (fromState.url === '/enketosurvey/:form_location/:opts' && toState.url === '/diary') {
-          readAndUpdateForDay(Timeline.data.currDay);
+            $rootScope.$on("$stateChangeStart", function (_event, toState, _toStateParams, fromState, _fromParams) {
+                if (fromState.url === "/enketosurvey/tripconfirm/:form_location/:opts" && toState.url === "/diary") {
+                    EnketoSurvey.getAllSurveyAnswers("manual/confirm_survey", { populateLabels: true }
```
shankari: @atton16 I had feedback on this approach in the previous PR
https://github.com/e-mission/e-mission-phone/pull/575#discussion_r289944826

> that is a lot better since you make a single call to getAllSurveyAnswers for every state change. But if people are filling out multiple surveys for a particular day (maybe at the end of the day), then you will still make multiple calls to EnketoSurvey.getAllSurveyAnswers.

to which you said

> As we directed out of the page, the only was to go back is to restore to previous state. So I cannot immediately knows which tripgj I wanted to get.

to which I said

> Since you are already using $rootScope as part of the state navigation, you can also use it pass in the tripgj that is effectively part of the survey state. That is what we (indirectly) do in the detail view. We pass in the tripID and the detail view looks it up from the timeline service. This is another option if you don't want to modify the $rootScope.
> This also has the benefit that the survey options can be modified based on trip characteristics (e.g. speed? time of day?)

shankari:
@atton16 thanks for generating a clean PR so that we can finish the discussion and I can merge!

www/js/diary/list.js
atton16: Like this https://github.com/e-mission/e-mission-phone/pull/580/commits/4b3ea980c6d48c06f1039b9f9e8c322308fbc332 ?

shankari: Yes. The structure of https://github.com/e-mission/e-mission-phone/commit/4b3ea980c6d48c06f1039b9f9e8c322308fbc332 looks fine. Does it work? Can you upload a screenshot/video?

atton16: Here is the testing video http://drive.attawit.me/s/NGJci5BSRm8T8g7

shankari:
@atton16 LGTM! Merging now...

shankari:

PR #581, opened 2019-06-16
post-trip-prompt launch survey

atton16:
I am trying to get the `post-trip-prompt` to launch the survey.

At this stage, I am not certain how should I get
1. `trip.data.properties.start_ts`
2. `trip.data.properties.start_ts`

perhaps you can help what is inside `notification.data` and `data` variable.


www/js/tripconfirm/post-trip-prompt.js
```diff
@@ -119,7 +119,20 @@ angular.module('emission.tripconfirm.posttrip.prompt', ['emission.plugin.logger'
   var displayCompletedTrip = function(notification, state, data) {
```
atton16: Please help clarify `notification.data` and `data`

www/js/tripconfirm/post-trip-prompt.js
```diff
@@ -119,7 +119,20 @@ angular.module('emission.tripconfirm.posttrip.prompt', ['emission.plugin.logger'
...
       Logger.log("About to display completed trip from Notification");
-      $state.go("root.main.tripconfirm", notification.data);
+      $state.go("root.main.enketosurvey", {
+        form_location: "json/trip-end-survey_v6.json",
+            opts: JSON.stringify({
+              session: {
+                data_key: 'manual/confirm_survey',
+                trip_properties: {
+                  // TODO: use the right variable
+                  start_ts: trip.data.properties.start_ts,
```
atton16: Not sure how to get to this variable, perhaps something inside `notification.data` or `data`.

shankari: `notification.data.start_ts`. You can see that we go to the `tripconfirm` screen using params = `notification.data`
https://github.com/e-mission/e-mission-phone/blob/master/www/js/tripconfirm/post-trip-prompt.js#L122
and then we get the start and end ts in the state from the `$stateParams` directly.

shankari: also, you may not want to use the current trip end code since it has a different screen for trip end mode/purpose versus diary mode/purpose. The trip end code shows a map and the mode/purpose choices while the diary just has popovers.

The previous incident reporting trip end code would basically redirect the user to the diary screen, which seems to be closer to what you want with your survey.
https://github.com/e-mission/e-mission-phone/blob/master/www/js/incident/post-trip-prompt.js#L133

and then in the diary if we are supposed to be dealing with an incident, we refresh. you can also got to the survey screen and everything else that you wrote should Just Work.

https://github.com/e-mission/e-mission-phone/blob/master/www/js/diary/list.js#L724

atton16:
Ok, so I have get rid of all popups and make the notification tigger navigate straight to the survey.

www/js/incident/post-trip-prompt.js
```diff
@@ -119,7 +119,20 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
       Logger.log("About to display completed trip from Notification");
-      $state.go("root.main.incident", notification.data);
+      $state.go("root.main.enketosurvey", {
+        form_location: "json/trip-end-survey_v6.json",
+            opts: JSON.stringify({
+              session: {
+                data_key: 'manual/confirm_survey',
+                trip_properties: {
+                  // TODO: use the right variable
+                  start_ts: notification.data.properties.start_ts,
```
shankari: @atton16 I thought you were now passing in the trip object directly instead of the start and end timestamps?

atton16: Ahh yes, I get it now why you are referring to `$rootScope.displayingIncident`. But even though I called `readAndUpdateForDay`, how am I supposed to know which trip the notification is referring to?

www/js/diary/list.js
```diff
@@ -766,10 +766,20 @@ angular.module('emission.main.diary.list',['ui-leaflet',
           if (angular.isDefined(Timeline.data.currDay)) {
               // page was already loaded, reload it automatically
               readAndUpdateForDay(Timeline.data.currDay);
+
+              const tripProp = notificationData.properties;
```
atton16: Not sure if this is what you are suggesting...

www/js/diary/list.js
```diff
@@ -766,10 +766,20 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
+              const trip = $scope.data.currDayTripWrappers.find(function(tripgj, index, array) {
+                return (
+                  tripgj.properties.start_ts === tripProp.start_ts &&
+                  tripgj.properties.end_ts === tripProp.end_ts
+                );
+              });
+              $scope.confirmSurvey(trip);
```
atton16: Not sure if this is what you are suggesting...

shankari: Yeah, that works. Since you know that this is a trip-end prompt, I think that a simpler and more performant option would be to just use the last trip (e.g. sth like `$scope.data.currDayTripWrappers[$scope.data.currDayTripWrappers.length - 1]`

shankari: Since Timeline is a service, you could also call `Timeline.updateForDay(day);` from the `post-trip-prompt.js` directly. But I think that after the survey you want to return to the diary, so this is probably the better option after all.

www/js/incident/post-trip-prompt.js
```diff
@@ -118,21 +118,8 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
 
   var displayCompletedTrip = function(notification, state, data) {
       $rootScope.displayingIncident = true;
+      $rootScope.notificationData = notification.data;
       Logger.log("About to display completed trip from Notification");
-      $state.go("root.main.enketosurvey", {
```
shankari: I don't think you need this when you already have the code in `www/js/tripconfirm/post-trip-prompt.js`

www/js/tripconfirm/post-trip-prompt.js
```diff
@@ -118,8 +118,8 @@ angular.module('emission.tripconfirm.posttrip.prompt', ['emission.plugin.logger'
 
   var displayCompletedTrip = function(notification, state, data) {
     $rootScope.displayingIncident = true;
+    $rootScope.notificationData = notification.data;
       Logger.log("About to display completed trip from Notification");
-      $state.go("root.main.tripconfirm", notification.data);
```
shankari: don't you need `$state.go` somewhere in here? Just setting the `notificationData` is not enough.

www/js/diary/list.js
atton16: > Yeah, that works. Since you know that this is a trip-end prompt, I think that a simpler and more performant option would be to just use the last trip (e.g. sth like `$scope.data.currDayTripWrappers[$scope.data.currDayTripWrappers.length - 1]`

Just to be cautious, sometimes users has multiple trips in a short period and got multiple notifications. This optimization will definitely cause the problem.

www/js/tripconfirm/post-trip-prompt.js
atton16: I've got `$state.go` in `diary/list.js` instead.

shankari: but that assumes that you are in the diary screen. Don't you need to `$state.go` to the diary first? What is your default start tab/state?

shankari: In particular, look at the diffs for `incident/post-trip-prompt.js` that you are trying to adapt. We have a `$state.go` which you have removed

atton16: Done.

shankari:
This looks good structurally; please upload a testing video and I will merge

atton16:
> This looks good structurally; please upload a testing video and I will merge

How should I test the notification? This is the one I have never done before. I am guessing using the real phone. But how can I side-load the app to the phone??

shankari:
https://github.com/e-mission/e-mission-docs/blob/b4f0278d30655c6a712a0e46933b4bc599ee2c4b/docs/tutorials/change_trip_end_prompt.md#testing

https://github.com/e-mission/e-mission-docs/blob/1fef61ac8c73c44d72dc37c18303e9a7781767e5/docs/overview/high_level_faq.md#how-to-get-the-app-onto-a-real-phone

atton16:
> Did you search the docs?
> 
> https://github.com/e-mission/e-mission-docs/blob/master/docs/tutorials/change_trip_end_prompt.md#testing
> 
> https://github.com/e-mission/e-mission-docs/blob/master/docs/overview/high_level_faq.md#how-to-get-the-app-onto-a-real-phone

I have been looking at the `on iOS: Debug -> Location -> Freeway Drive` but cannot find such control.

shankari:
Still shows up for me

Screen Shot 2019-06-22 at 12 07 24 AM

shankari:
BTW, I also found this page on how to install a .app onto a physical device
https://codeburst.io/latest-itunes-12-7-removed-the-apps-option-how-to-install-ipa-on-the-device-3c7d4a2bc788

haven't had a chance to try out the "Apple Configurator" option yet, but it looks promising.
If it works, maybe you can submit a PR?

atton16:
The new trip is not included in `$scope.data.currDayTripWrappers` even after `readAndUpdateForDay`. I also noticed that the trip is not displaying on the map.

Screen Shot 2019-06-22 at 14 19 06

shankari:
Are you on the correct day? The trips that you are showing appear to be from 2015. The latest trip will show up today.

atton16:
> Are you on the correct day? The trips that you are showing appear to be from 2015. The latest trip will show up today.

Right, but as I executed `readAndUpdateForDay(Timeline.data.currDay);` I need to change to the day of the trip of the selected notification. I am not sure if I should use `$rootScope.notificationData.start_ts` or `$rootScope.notificationData.end_ts`.

Says, user start a trip at `1st Jan 23:59` and end trip at `2nd Jan 00:02`. Which date would the trip be displayed on? `1st Jan` or `2nd Jan`.

atton16:
> BTW, I also found this page on how to install a .app onto a physical device
> https://codeburst.io/latest-itunes-12-7-removed-the-apps-option-how-to-install-ipa-on-the-device-3c7d4a2bc788
> 
> haven't had a chance to try out the "Apple Configurator" option yet, but it looks promising.
> If it works, maybe you can submit a doc PR?

Tried with Apple Configurator but got the app greyed out. I did it twice but same thing happen. Not sure if I have to do app signing or not.

Screen Shot 2019-06-22 at 14 51 45
IMG_2507

shankari:
> Tried with Apple Configurator but got the app greyed out. I did it twice but same thing happen. Not 
> sure if I have to do app signing or not.

The instructions I linked to didn't say anything about signing. Maybe post there and ask the author?

shankari:
> Says, user start a trip at 1st Jan 23:59 and end trip at 2nd Jan 00:02. Which date would the trip be displayed on? 1st Jan or 2nd Jan.

I am pretty sure 1st Jan. I think this is even in the unit tests but probably best to just test.

www/js/diary/list.js
```diff
@@ -277,14 +277,34 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
+                tripgj.data.properties.start_ts === $rootScope.notificationData.start_ts &&
+                tripgj.data.properties.end_ts === $rootScope.notificationData.end_ts
+              ) {
+                tripFromNotification = tripgj;
+              }
+            }
           });
           $ionicScrollDelegate.scrollTop(true);
+
+          if (tripFromNotification) {
```
atton16: It looks like that I have to display the confirm survey here because I have to wait for `Timeline.UPDATE_DONE` instead of displaying it right after `readAndUpdateForDay`.

However, as the trips are displayed correctly on the date the notification is selected but I cannot match any trip to the notification at all. Not sure what happens.

www/js/diary/list.js
```diff
@@ -763,23 +783,11 @@ angular.module('emission.main.diary.list',['ui-leaflet',
           $rootScope.barDetail = false;
           }
         if($rootScope.displayingIncident == true) {
-          if (angular.isDefined(Timeline.data.currDay)) {
+          const tripProp = $rootScope.notificationData;
```
atton16: Here we load the timeline based on the `start_ts` of notification data to get the diary page display the correct day.

shankari:
testing video?

atton16:
There is still a bug that make the system unable to match notificationData.start_ts to any of the trip stored in the system. Still not sure why this happen.

shankari:
I just pulled your changes to a local branch and it seems to work fine for me at least at first check

```
[Log] DEBUG:Unprocessed trip started at {"metadata":{"time_zone":"America/Los_Angeles","plugin":"none","write_ts":1561268142.048285,"platform":"ios","read_ts":0,"key":"statemachine/transition","type":"message"},"data":{"currState":"STATE_WAITING_FOR_TRIP_START","transition":"T_EXITED_GEOFENCE","ts":1561268142.048106},"_id":{"$oid":"5d0f0fddb058561569e20c83"},"user_id":{"$uuid":"bd457bc1b2c64689a605587a34d7d51a"}} (cordova.js, line 1732)
...
[Log] DEBUG:Deduped list = 1 (cordova.js, line 1732)
...
[Log] tripgj => – {data: Object, style: function, onEachFeature: function, …} display_end_time: "10:35 pm"display_start_time: "10:35 pm"...
[Log] $rootScope.notificationData => – {start_ts: 1561268146.657119, end_ts: 1561268146.657119, end_lat: 37.787359, …}...
[Log] tripFromNotification => – {data: Object, style: function, onEachFeature: function, …} display_end_time: "10:35 pm"display_start_time: "10:35 pm"...
```

And the survey pops up when I force the trip end from the developer zone.
But the results don't show up in the list page since this is a draft trip and you only show the mode and purpose selection for cleaned trips.

shankari:
ok so I just tried it again to take a video and it doesn't work now. It looks like it doesn't work when there are multiple trips for the day? Retesting to confirm...

shankari:
Could this be the problem? The notification data that I in the non-working case is

```
[Log] $rootScope.notificationData => (cordova.js, line 1732)
Object

end_lat: 37.33367998
end_lng: -122.05028306
end_ts: 1561269198.274788
start_lat: 37.33367998
start_lng: -122.05028306
start_ts: 1561269198.274788
Object Prototype
```

But on converting that using moment, I get 

```
> moment(1561269198.274788).format()
< "1970-01-18T17:41:09-08:00"
< "1970-01-18T17:41:09-08:00"
```

Is this some millisecond timestamp versus second timestamp issue?

shankari:
e-mission uses timestamps expressed in seconds, consistent with python.
I suspect javascript uses milliseconds and I bet we convert back and forth for display and storage.

atton16:
> Could this be the problem? The notification data that I in the non-working case is
> 
> ```
> [Log] $rootScope.notificationData => (cordova.js, line 1732)
> Object
> 
> end_lat: 37.33367998
> end_lng: -122.05028306
> end_ts: 1561269198.274788
> start_lat: 37.33367998
> start_lng: -122.05028306
> start_ts: 1561269198.274788
> Object Prototype
> ```
> 
> But on converting that using moment, I get
> 
> ```
> > moment(1561269198.274788).format()
> < "1970-01-18T17:41:09-08:00"
> < "1970-01-18T17:41:09-08:00"
> ```
> 
> Is this some millisecond timestamp versus second timestamp issue?

I did multiply `ts` by `1000` before performing `readAndUpdateForDay` though.

shankari:
So what have you tried to debug this? Have you put in breakpoints in the matching code? added additional logs?

shankari:
I added the following line 

```
diff --git a/www/js/diary/list.js b/www/js/diary/list.js
index 93babefd..e0cac6a7 100644
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -289,6 +289,11 @@ angular.module('emission.main.diary.list',['ui-leaflet',
             if($rootScope.displayingIncident == true && $rootScope.notificationData) {
               console.log('tripgj =>', tripgj);
               console.log('$rootScope.notificationData =>', $rootScope.notificationData);
+              console.log("Delta start_ts = "+
+                ($rootScope.notificationData.start_ts - tripgj.data.properties.start_ts)
+                +" and delta end_ts = "+
+                ($rootScope.notificationData.end_ts - tripgj.data.properties.end_ts));
+
               if (
                 tripgj.data.properties.start_ts === $rootScope.notificationData.start_ts &&
```

and got the output below. I think you should consider using the trip matching code in `trip-confirm-services.js` instead of assuming a perfect match.


```
[Log] tripgj => – {data: Object, style: function, onEachFeature: function, …} (cordova.js, line 1732)
{data: Object, style: function, onEachFeature: function, pointToLayer: function, start_place: Object, …}Object
[Log] $rootScope.notificationData => – {start_ts: 1561295775.317517, end_ts: 1561295775.317517, end_lat: 37.33054431, …} (cordova.js, line 1732)
{start_ts: 1561295775.317517, end_ts: 1561295775.317517, end_lat: 37.33054431, start_lat: 37.33054431, end_lng: -122.03058311, …}Object
[Log] Delta start_ts = 16.99403405189514 and delta end_ts = 0 (cordova.js, line 1732)
```

atton16:
> I added the following line
> 
> ```
> diff --git a/www/js/diary/list.js b/www/js/diary/list.js
> index 93babefd..e0cac6a7 100644
> --- a/www/js/diary/list.js
> +++ b/www/js/diary/list.js
> @@ -289,6 +289,11 @@ angular.module('emission.main.diary.list',['ui-leaflet',
>              if($rootScope.displayingIncident == true && $rootScope.notificationData) {
>                console.log('tripgj =>', tripgj);
>                console.log('$rootScope.notificationData =>', $rootScope.notificationData);
> +              console.log("Delta start_ts = "+
> +                ($rootScope.notificationData.start_ts - tripgj.data.properties.start_ts)
> +                +" and delta end_ts = "+
> +                ($rootScope.notificationData.end_ts - tripgj.data.properties.end_ts));
> +
>                if (
>                  tripgj.data.properties.start_ts === $rootScope.notificationData.start_ts &&
> ```
> 
> and got the output below. I think you should consider using the trip matching code in `trip-confirm-services.js` instead of assuming a perfect match.
> 
> ```
> [Log] tripgj => – {data: Object, style: function, onEachFeature: function, …} (cordova.js, line 1732)
> {data: Object, style: function, onEachFeature: function, pointToLayer: function, start_place: Object, …}Object
> [Log] $rootScope.notificationData => – {start_ts: 1561295775.317517, end_ts: 1561295775.317517, end_lat: 37.33054431, …} (cordova.js, line 1732)
> {start_ts: 1561295775.317517, end_ts: 1561295775.317517, end_lat: 37.33054431, start_lat: 37.33054431, end_lng: -122.03058311, …}Object
> [Log] Delta start_ts = 16.99403405189514 and delta end_ts = 0 (cordova.js, line 1732)
> ```

I tested with your delta-printing code and got an interesting output where the delta is in `negative` value which I presumed it is the trip we are looking for.

Screen Shot 2019-06-30 at 22 26 41

I also did try with the  trip matching code per your suggestion but it will not match any trip (output above (`trip => undefined`)).

Screen Shot 2019-06-30 at 22 29 01

www/js/diary/list.js
```diff
@@ -289,14 +289,25 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
                 tripFromNotification = tripgj;
               }
             }
           });
+          if($rootScope.displayingIncident == true && $rootScope.notificationData) {
+            const trip = ConfirmHelper.getUserInputForTrip({
+              start_ts: $rootScope.notificationData.start_ts*1000,
+              end_ts: $rootScope.notificationData.end_ts*1000,
+            }, $scope.data.currDayTripWrappers);
+            console.log('trip => ', trip);
```
atton16: This prints `undefined` even if I multiply with `1000`

www/js/diary/list.js
```diff
@@ -289,14 +289,25 @@ angular.module('emission.main.diary.list',['ui-leaflet',
             if($rootScope.displayingIncident == true && $rootScope.notificationData) {
               console.log('tripgj =>', tripgj);
               console.log('$rootScope.notificationData =>', $rootScope.notificationData);
+              console.log("Delta start_ts = "+
+                ($rootScope.notificationData.start_ts - tripgj.data.properties.start_ts)
+                +" and delta end_ts = "+
```
atton16: One of the output shows a `negative` delta which I suspect it is the trip we are looking for but I am not certain how should I refer to it. I am started to think that `start_ts` and `end_ts` from `notification` is somewhat not related to the trip but might be related to the time the notification triggered or the time I tap on the notification.

shankari:
> One of the output shows a negative delta which I suspect it is the trip we are looking for but I am not certain how should I refer to it. I am started to think that start_ts and end_ts from notification is somewhat not related to the trip but might be related to the time the notification triggered or the time I tap on the notification.

That is interesting. It should not be the time you tap on the notification since notificationdata is set into the notification object when it is created. If I remember the notify code correctly, the local notifications are generated from the generic transitions, so they should happen **after** the generic transitions are generated. Is this on android or iOS?

shankari:
> One of the output shows a negative delta which I suspect it is the trip we are looking for but I am not certain how should I refer to it.

Is this reproducible? Or did it happen only one time?

shankari:
So I looked through the native iOS code (android is similar) and the key is that when we generate the notification data, we do so by looking up corresponding location points. For the end point, we use the last point
https://github.com/e-mission/e-mission-transition-notify/blob/master/src/ios/BEMTransitionNotifier.m#L158

but for the start point, we look at the points around the start transition and potentially pick points before/after that
https://github.com/e-mission/e-mission-transition-notify/blob/master/src/ios/BEMTransitionNotifier.m#L210

this is how we can end up with a negative value.

shankari:
I don't currently remember why I pick the point just before the first start transition. And I don't see any comments in the PR that help me figure out why I did it either :)
https://github.com/e-mission/e-mission-transition-notify/pull/2

But until I investigate and fix that, we need to assume that untracked trips can have a negative delta *but* the max delta can be 5 minutes. I will change the `getUserInputForTrip` to match.

Let me fix the user input

atton16:
@shankari I have successfully migrated to use the new **Implement more complex user input matching** https://github.com/e-mission/e-mission-phone/pull/594.

shankari:
@atton16 so does this work now? If you attach the testing video, I can merge.

atton16:
Just update the latest trip file and also prepping the test video.

atton16:
trip-end-survey_v9+notification mov

www/js/diary/list.js
```diff
@@ -277,14 +273,44 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
           });
+          if($rootScope.displayingIncident == true && $rootScope.notificationData) {
+            const matchingTrips = $scope.data.currDayTripWrappers.filter(trip => {
+              const matchingInput = ConfirmHelper.getUserInputForTrip(trip, [{
+                data: {
+                  start_ts: $rootScope.notificationData.start_ts,
+                  end_ts: $rootScope.notificationData.end_ts,
+                },
+                metadata: {
+                  time_zone: 'Asia/Bangkok'
```
shankari: you need to use the current timezone instead of the hardcoded value.

www/js/diary/list.js
```diff
@@ -747,7 +773,9 @@ angular.module('emission.main.diary.list',['ui-leaflet',
     };
 
     $ionicPlatform.ready().then(function() {
-      readAndUpdateForDay(moment().startOf('day'));
+      // readAndUpdateForDay(moment().startOf('day'));
+      // DEBUG
+      readAndUpdateForDay(moment("2015-07-22").startOf("day"));
```
shankari: remove hardcoded `readAndUpdateForDay`

www/js/diary/list.js
```diff
@@ -277,14 +273,44 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
+
           $scope.data.currDayTripWrappers.forEach(function(tripgj, index, array) {
             $scope.populateModeFromTimeline(tripgj, $scope.data.unifiedConfirmsResults.modes);
             $scope.populatePurposeFromTimeline(tripgj, $scope.data.unifiedConfirmsResults.purposes);
             $scope.populateSurveyAnswerFromTimeline(tripgj, $scope.data.unifiedConfirmsResults.surveyAnswers);
             $scope.populateBasicClasses(tripgj);
             $scope.populateCommonInfo(tripgj);
           });
+          if($rootScope.displayingIncident == true && $rootScope.notificationData) {
+            const matchingTrips = $scope.data.currDayTripWrappers.filter(trip => {
```
shankari: arrow functions are not supported in iOS9 - they result in a WSOD. Please switch to explicit `function(trip)` unless you plan to publish your own app to the store and set the min iOS version to a higher value. the default emission app supports older versions of both android and iOS for maximum compatibility.

You can test this using the emulator as well.
https://webkit.org/blog/4054/es6-in-webkit/

www/js/incident/post-trip-prompt.js
```diff
@@ -212,19 +213,20 @@ angular.module('emission.incident.posttrip.prompt', ['emission.plugin.logger'])
...
-            } else {
-              Logger.log("Skipped incident reporting");
-            }
-          });
-        } else {
-          Logger.log("About to go to prompt page");
-          displayCompletedTrip(notification, state, data);
-        }
+        displayCompletedTrip(notification, state, data);
+        // if($ionicPlatform.is('ios')) {
```
shankari: this is the code that displays the prompt if you don't close the app when you end the trip. You saw this during my demo last week and said that you don't see it. This is why :)

In that case, there is no visible notification, it is delivered directly to the open app.

www/js/survey/enketo-survey.js
```diff
@@ -17,7 +17,11 @@ angular.module('emission.survey.enketo.launch', [
       $state.go("root.main.diary")
     });
   } else {
-    EnketoSurvey.init($stateParams.form_location, $stateParams.opts)
+    EnketoSurvey.init({
+      form_location: $stateParams.form_location,
+      opts: $stateParams.opts,
```
shankari: just to confirm, these `opts` are currently unused, right?

www/js/tripconfirm/post-trip-prompt.js
```diff
@@ -210,19 +211,20 @@ angular.module('emission.tripconfirm.posttrip.prompt', ['emission.plugin.logger'
...
-              } else {
-                Logger.log("Skipped confirmation reporting");
-              }
-            });
-        } else {
-          Logger.log("About to go to prompt page");
-          displayCompletedTrip(notification, state, data);
-        }
+        displayCompletedTrip(notification, state, data);
+        // if($ionicPlatform.is('ios')) {
```
shankari: see comment in `incident/post-trip-prompt.js`

shankari: after some real-world testing you can decide whether to re-enable this or not

www/js/incident/post-trip-prompt.js
atton16: I have modified this part along with

```js
  var displayCompletedTrip = function(notification, state, data) {
      $rootScope.displayingIncident = true;
      $rootScope.notificationData = notification.data;
      Logger.log("About to display completed trip from Notification");
      $state.go('root.main.diary');
  };
```

to direct notification to display the survey

www/js/survey/enketo-survey.js
atton16: `opts` is use to pass the `data_key` selector in case we have multiple surveys or data sources.

atton16:
```js
      $state.go("root.main.enketosurvey", {
      form_location: "json/trip-end-survey_v9.json",
          opts: JSON.stringify({
            session: {
              data_key: 'manual/confirm_survey',
            }
          }),
      });
```

www/js/incident/post-trip-prompt.js
shankari: I see. If the user has the app open when the trip ends, it can be a bit disorienting to suddenly have the app move to a different screen with no user input. But I agree that the odds of the user having the app UI open during a trip end is low.

www/js/diary/list.js
shankari: @atton16 are you going to resolve this? You can use the arrow library to get this information.

atton16: done

shankari:
Minor fixes LGTM.
Merging now...

shankari:
Deployed, should be live in ~10 mins

PR #583, opened 2019-06-16
changing color for control branch

sam11b:
Making UI change from greentrip branch also in control branch.


shankari:

PR #584, opened 2019-06-18
i18n e-mission-phone

DO3B:
As mentioned in https://github.com/e-mission/e-mission-docs/issues/354#issuecomment-500748094, my work on the i18n of the phone app using angular-translate. I may have forgotten some strings.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -11,7 +11,8 @@ angular.module('emission', ['ionic',
     'emission.controllers','emission.services', 'emission.plugin.logger',
     'emission.splash.customURLScheme', 'emission.splash.referral',
     'emission.splash.updatecheck',
-    'emission.intro', 'emission.main'])
+  'emission.intro', 'emission.main',
+  'pascalprecht.translate'])
 
 .run(function($ionicPlatform, $rootScope, $http, Logger,
     CustomURLScheme, ReferralHandler, UpdateCheck) {
@@ -70,7 +71,7 @@ angular.module('emission', ['ionic',
   console.log("Ending run");
 })
 
-.config(function($stateProvider, $urlRouterProvider) {
+.config(function($stateProvider, $urlRouterProvider, $translateProvider) {
   console.log("Starting config");
   // alert("config");
 
@@ -102,5 +103,23 @@ angular.module('emission', ['ionic',
   // alert("about to fall back to otherwise");
   // if none of the above states are matched, use this as the fallback
   $urlRouterProvider.otherwise('/splash');
+
+  // Allow the use of MessageForm interpolation for Gender and Plural.
+  $translateProvider.addInterpolation('$translateMessageFormatInterpolation');
+
+  // Define where we can find the .json and the fallback language
+  $translateProvider
+    .fallbackLanguage('en')
+    .registerAvailableLanguageKeys(['en', 'fr'], {
+      'en_*': 'en',
+      'fr_*': 'fr',
+      '*': 'en'
+    })
+    .determinePreferredLanguage()
+    .useStaticFilesLoader({
+      prefix: 'i18n/',
+      suffix: '.json'
+    });
+  
   console.log("Ending config");
 });
--- a/www/js/diary/current.js
+++ b/www/js/diary/current.js
@@ -86,18 +86,7 @@
 
   var startTimeFn = function (ts) {
     var date = new Date(ts * 1000);
-    if ($translate.use() != "en") {
-      return moment(date).format('HH:mm');
-    } else {
-      return moment(date).format('HH:mmA')
-    }
-    // var hours = date.getHours();
-    // var minutes = date.getMinutes();
-    // var amOrPm = hours < 12 ? 'AM' : 'PM';
-    // hours = hours % 12;
-    // hours = hours ? hours : 12;
-    // minutes = minutes < 10 ? '0'+ minutes : minutes;
-    // return hours + ':' + minutes + ' ' + amOrPm;
+    return moment(date).format("LT");
   };
 
   var getSpeed = function(curr_lglat, last_lglat, curr_ts, last_ts) {
--- a/www/js/diary/current.js
+++ b/www/js/diary/current.js
@@ -8,7 +8,7 @@
                                                 'emission.plugin.logger'])
 
 .controller('CurrMapCtrl', function($scope, Config, $state, $timeout, $ionicActionSheet,leafletData, 
-                                    Logger, $window, PostTripManualMarker, CommHelper, $http, KVStore, $ionicPlatform) {
+                                    Logger, $window, PostTripManualMarker, CommHelper, $http, KVStore, $ionicPlatform, $translate) {
     
   console.log("controller CurrMapCtrl called from current.js");
   var _map;
@@ -46,9 +46,9 @@
       hideLimitLabels: true,
       translate: function(value) {
           if (value === 1) {
-            return "Yesterday";
+            return $translate.instant('diary.current-yesterday');
           } else if (value === 7) {
-            return "Week ago";
+            return $translate.instant('diary.current-weekagos');
           }
         return "";
       }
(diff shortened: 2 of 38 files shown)


shankari:
@DO3B thank you so much for doing this! It is OK if you have forgotten some strings, we can always add them later as they are reported. The important thing is that the structure is in place.

shankari:
@DO3B in terms of process, I assume that as they add new features, developers will add new keys to the json file for their language. I will enforce this (no hardcoded strings) through code review :) 

But what happens until translatewiki finishes translating the text to the other languages? Will angular-translate use the string from the translation that does exist, or will it be blank?

DO3B:
@shankari If the json of the language does not have the keys, it will use the values of the [fallback language ](https://angular-translate.github.io/docs/#/guide/08_fallback-languages)(in our case, english).

shankari:
@DO3B great! I could have RTFM but I was lazy and decided to ask you instead :)

www/js/app.js
```diff
@@ -102,5 +103,23 @@ angular.module('emission', ['ionic',
...
   $urlRouterProvider.otherwise('/splash');
+
+  // Allow the use of MessageForm interpolation for Gender and Plural.
+  $translateProvider.addInterpolation('$translateMessageFormatInterpolation');
+
+  // Define where we can find the .json and the fallback language
+  $translateProvider
+    .fallbackLanguage('en')
+    .registerAvailableLanguageKeys(['en', 'fr'], {
+      'en_*': 'en',
```
shankari: so as we add translations from translatewiki, we would add the list here?
As a future enhancement, is it possible to determine this dynamically by iterating through the files in `www/i18n`?

www/js/diary/current.js
```diff
@@ -84,15 +84,20 @@
...
-    hours = hours % 12;
-    hours = hours ? hours : 12;
-    minutes = minutes < 10 ? '0'+ minutes : minutes;
-    return hours + ':' + minutes + ' ' + amOrPm;
+  var startTimeFn = function (ts) {
+    var date = new Date(ts * 1000);
+    if ($translate.use() != "en") {
+      return moment(date).format('HH:mm');
+    } else {
+      return moment(date).format('HH:mmA')
```
shankari: Thanks for doing this in the principled way. But I'm curious about why you use the am/pm format only when the current language is in english. Does it not work for French? Have you tried the locale aware formats (e.g. `LT`). I will not hold up the merge for this, but you might want to test and fix in a second round.

www/js/diary/list.js
```diff
@@ -42,6 +42,8 @@ angular.module('emission.main.diary.list',['ui-leaflet',
     // TODO: Convert the usercache calls into promises so that we don't have to
     // do this juggling
     Timeline.updateForDay(day);
+    // This will be used to show the date of datePicker in the user language.
+    $scope.currDay = moment(day).format('DD MMM YYYY');
```
shankari: again, I would suggest using `LL` or `ll`. The default string format in the US is `MMM DD YYYY` not `DD MMM YYYY`. As somebody who moved to the US from India, I had to get used to the change as well :)

www/js/diary/list.js
```diff
@@ -59,23 +61,23 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
-        hh: "%d h",
-        d:  "a day",
-        dd: "%d days",
-        M:  "a month",
-        MM: "%d months",
-        y:  "a year",
-        yy: "%d years"
-    }
-});
+//   moment.locale('en', {
```
shankari: Looks like this can be fixed by including the standard locales in `index.html`
https://momentjs.com/docs/#/i18n/loading-into-browser/

www/js/diary/services.js
```diff
@@ -166,12 +166,17 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
     return retVal;
   };
 
-  dh.getLocalTimeString = function(dt) {
-      var hr = ((dt.hour > 12))? dt.hour - 12 : dt.hour;
-      var post = ((dt.hour >= 12))? " pm" : " am";
-      var min = (dt.minute.toString().length == 1)? "0" + dt.minute.toString() : dt.minute.toString();
-      return hr + ":" + min + post;
+  dh.getLocalTimeString = function (dt) {
```
shankari: Ditto

www/js/intro.js
```diff
@@ -19,7 +19,27 @@ angular.module('emission.intro', ['emission.splash.startprefs',
...
 .controller('IntroCtrl', function($scope, $state, $ionicSlideBoxDelegate,
-    $ionicPopup, $ionicHistory, ionicToast, $timeout, CommHelper, StartPrefs) {
+    $ionicPopup, $ionicHistory, ionicToast, $timeout, CommHelper, StartPrefs, $translate, $cordovaFile) {
+  
+  $scope.getConsentFile = function () {
+    var lang = $translate.use();
+    $scope.consentFile = "templates/intro/consent.html";
+    if (lang != 'en') {
+      var url = "www/i18n/intro/consent-" + lang + ".html";
+      $cordovaFile.checkFile(cordova.file.applicationDirectory, url).then( function(result){
```
shankari: Does `cordova.file.applicationDirectory` work on both android and iOS? I generally use `$http.get` to read files from the `www` directory (e.g. `$http.get(`templates/intro/consent.html`) which does work crossplatform, and you would use the http status to determine whether the file existed or not.

www/js/tripconfirm/trip-confirm-services.js
```diff
@@ -27,24 +27,45 @@ angular.module('emission.tripconfirm.services', ['ionic', "emission.plugin.logge
...
             return Promise.resolve(ch.modeOptions);
         }
     }
 
     ch.getPurposeOptions = function() {
         if (!angular.isDefined(ch.purposeOptions)) {
-            return loadAndPopulateOptions("json/trip_confirm_options.json")
+            var lang = $translate.use();
+            if (lang != "en") {
+                return loadAndPopulateOptionsLocales("i18n/trip_confirm_options-fr.json")
```
shankari: nit: use `lang` instead of hardcoded `fr`

www/js/tripconfirm/trip-confirm-services.js
```diff
@@ -27,24 +27,45 @@ angular.module('emission.tripconfirm.services', ['ionic', "emission.plugin.logge
             });
     }
     
+    var loadAndPopulateOptionsLocales = function (filename) {
```
shankari: Instead of adding this new function, can't you just expand `loadAndPopulateOptions` to have the following logic.

- if lang is not 'en', try to read language specific json
- if that fails, fall back to english json
- if that also fails, fall back to english json sample

DO3B: As you suggested, I changed loadAndPopulate: https://github.com/e-mission/e-mission-phone/pull/584/commits/5b947033d941f187ae6757678ccc37d92bafa1a7.

I used `lang` instead of `filename` as parameters since we were repeating it twice (for Modes and Purposes).

www/js/tripconfirm/trip-confirm-services.js
DO3B: Oups, 5b94703 fixed it !

www/js/diary/current.js
DO3B: AM/PM does not work in French as most of the french speaking people think that AM meaning is "après-midi" (afternoon). But the good news is that locale aware formats seem to work ! So I will trust `moment` now (https://github.com/e-mission/e-mission-phone/pull/584/commits/f8db039d97f3d403f7a1dbd16174c5f0c4423fee). Thank you for your suggestion.

www/js/diary/services.js
DO3B: https://github.com/e-mission/e-mission-phone/pull/584#discussion_r295166532

www/js/intro.js
DO3B: I tried it in both OS and it worked but I can use `$http.get` if you prefer.

www/js/app.js
DO3B: As we determined the language by accessing browser properties, It seemed necessary to "filter" the locales. For instance, a browser will return `en_US` or `en_UK` which could be seen as the same language and will try to read `en_US.json` or `en_UK.json`. When I register the languages keys, I avoid this behaviour. [Language Negotiation](https://angular-translate.github.io/docs/#/guide/09_language-negotiation)

www/js/diary/list.js
DO3B: I am doing it [there](https://github.com/fabmob/e-mission-phone-fabmob/blob/f8db039d97f3d403f7a1dbd16174c5f0c4423fee/www/js/main.js#L154). The main idea is that if there is no spanish (for instance) it will not use the spanish locales but the locales of the fallback language instead.

www/js/tripconfirm/trip-confirm-services.js
```diff
@@ -12,19 +12,28 @@ angular.module('emission.tripconfirm.services', ['ionic', "emission.plugin.logge
...
+            .catch(function (err) {
+                console.log("error "+JSON.stringify(err)+" while reading confirm options in your language, reverting to english options");
+                return loadAndPopulateOptions("en");
+            });
+        }
+
+        return $http.get("json/trip_confirm_options.json")
+        .then(fillInOptions)
+        .catch(function(err) {
+           // prompt here since we don't have a fallback
```
shankari: nit: we don't prompt here since we do have a fallback

www/js/intro.js
shankari: It's fine. If I do try to remove the cordova file plugin later, I can change this to `$http.get`

www/js/diary/list.js
shankari: ah, I see. I was looking at the moment documentation, 
https://momentjs.com/docs/#/i18n/loading-into-browser/
and I thought that you needed to include individual locale files - e.g.

```
<script src="locale/fr.js" charset="UTF-8"></script>
<script src="locale/pt.js" charset="UTF-8"></script>
```

But I see that you are using option B - e.g. have the combined minified locale file instead

```
     <script src="lib/moment/min/locales.min.js"></script> 
```

so that's all good.

PR #585, opened 2019-06-29
Cistup bangalore

deepalics0044:


www/json/trip_confirm_options.json.sample
```diff
@@ -1,25 +1,41 @@
 {
+
   "modeOptions" : [
-       {"text":"Walk", "value":"walk"},
```
shankari: @deepalics0044 this is a sample file (`trip_confirm_options.json.sample`). In general, instead of editing the sample file directly (which can cause merge conflicts later), you want to copy it into (`trip_confirm_options.json`). Can you make that change?

www/templates/intro/instruction.html
```diff
@@ -0,0 +1,19 @@
+<ion-content class="has-footer">
```
shankari: where is this used? I don't see any changes to the slidebox to display it...

www/templates/main.html
```diff
@@ -26,10 +28,12 @@
     <ion-nav-view name="main-metrics"></ion-nav-view>
   </ion-tab>
 
-  <!-- Heatmap Tab -->
-  <ion-tab title="{{'main-heatmap.title' | translate}}" icon="ion-fireball" href="#/root/main/heatmap">
+  <!-- Heatmap Tab 
+  <ion-tab title="Heatmap" icon="ion-fireball" href="#/root/main/heatmap">
```
shankari: You are only commenting out the heatmap and game tabs here, but are retaining the dashboard. Is this what you want?

deepalics0044: Yeah

PR #586, opened 2019-06-30
Cistup bangalore

deepalics0044:


www/json/trip_confirm_options.json.sample
```diff
@@ -1,25 +1,41 @@
 {
+
   "modeOptions" : [
-       {"text":"Walk", "value":"walk"},
```
shankari: To clarify, you need to modify `trip_confirm_options.json` and leave `trip_confirm_options.json.sample` unchanged to avoid merge conflicts.

shankari:
Squashing before merge to keep the git blame clean.

shankari:

shankari:
had to resolve some minor merge conflicts, but all looks good now.

deepalics0044:
Thank you :)

PR #588, opened 2019-06-30
User Profile Survey

atton16:
In this PR,

we are trying to put use survey for user profile purpose.

So we are,

1. Putting 'My Profile' button in survey tab
1. Insert user profile survey during the on-boarding process


atton16:
@shankari I am trying to bind the survey to both `main-dairy` tab and `main-control` tab as the survey controller is associated to these views. This is minor but it will help with the back button going back properly. The only way I can think of is to separate into two state, i.e. `root.main.enketosurvey` and `root.main.enketosurvey2`. This theoretically should work but the underlying `library` of the survey some how got initiated twice and messed up both `confirm survey` and `user profile survey`. Now I have to choice but keep using single `root.main.enketosurvey`. Do you have any suggestion??

shankari:
> Do you have any suggestion??

An option is to use a popover instead of a state. 
https://ionicframework.com/docs/v1/api/service/$ionicPopover/
which can define a fairly complex file to be displayed in a separate screen without defining a new state.

Sometimes, I have had issues with popovers not working as expected, mainly for the map display (e.g. https://github.com/shankari/e-mission-phone/commit/15cf5e163ceba313b12c4e1490641ba2ed184151), which is why I went with a state to begin with, but it is worth a try.

shankari:
> Do you have any suggestion??

If you do stick with states, I don't think you should duplicate the states. If I understand the design of angular states, a state represents one set of functionality, not how you navigated to that functionality. The whole point is to separate the functionality from the navigation.

If the rest of the functionality is the same, you should be able to have a single state and use the `$stateParams` to figure out where you came to the state from and how you should navigate back.

shankari:
This claims that you can have one states updated two views, although again, I am not sure that is what you want.
http://txt.fliglio.com/2013/05/angularjs-state-management-with-ui-router/

shankari:
At a high level, you may not want to use the same structure for profiles as for trip confirmations. Not that there are multiple trip confirmations for each user, but there is only one profile.

So for profiles, we support a simple CRUD model. The related wrappers are `CommHelper.registerUser` and `CommHelper.updateUser` and `CommHelper.getUser`. You can just store the profile and retrieve it directly without any start/end timestamps and without any matching required.

www/js/control/general-settings.js
```diff
@@ -399,6 +399,23 @@ angular.module('emission.main.control',['emission.services',
...
         });
     };
+    $scope.editUserProfile = function() {
+        CommHelper.getUser().then(function(profile){
+            const uuid = profile.user_id["$uuid"];
+            $state.go("root.main.enketosurvey", {
+                form_location: "json/user-profile_v1.json",
+                opts: JSON.stringify({
+                    session: {
+                        data_key: "manual/user_profile_survey",
```
shankari: e.g. you could just populate the profile directly from here.

atton16: We will be using user profile as external data for confirm survey in the yet-to-come future. So that's why I am implementing the service this way.

atton16: But yes, I think it could work from either in the service or controller.

shankari: For using the profile as external data, presumably you will want to retrieve the profile in the background on app startup and cache it locally instead of making remote calls while taking the survey. Again, since the profile is not trip specific and there is only one profile per app installation, it is much easier to work with.

atton16: We are allowing the profile to be changed on-demand. So getting the most recent snapshot of the profile prior to launching the confirm survey is a must. I am told that the answer from the profile will be dynamically displayed on the confirm survey's multiple choice question(s)

shankari: But you only allow users to modify the profile from the app, right? So you can cache the modified profile after any modifications occur as well.

atton16: So perhaps moving back to BEMUserCache right?

shankari: I was thinking more in terms of an in-memory cached value. Since you are already persisting this on the remote server and you can retrieve it on demand from the server, it seems like an in-memory variable would be the most efficient and easiest to use.

atton16:
It seems that I cannot have two different surveys and switching between each other in the app.
Once the app is started, only one survey works. I am stuck at this for a long time now.

I have tried:
1. separate view and controller
2. separate the service
3. use modal instead to `$state.go`

but they all exhibit the same behaviour.

I am not sure if there is something to do with the underlying library or not.

Here are the test videos:

Test 1: Load user profile survey then trip confirm survey [multiple-survey-1 mov](https://user-images.githubusercontent.com/5247367/61036131-a78d9e00-a3f2-11e9-8da7-6e9dfac94585.gif)
multiple-survey-1 mov

Test 2: Load trip confirm survey then user profile survey [multiple-survey-2 mov](https://user-images.githubusercontent.com/5247367/61036191-cc821100-a3f2-11e9-99ba-52589aa36413.gif)
multiple-survey-2 mov

shankari:
@atton16 are the testing videos above with the state or with the modal? with the state, it may be because states are cached in ionic, so if you may need to reset the state display by hooking into `ionic.enter` or `ionic.afterEnter` (https://ionicframework.com/docs/v1/api/directive/ionNavView/, scroll to **Caching**)

But with the modal, there is no state, so it should just work. Let's work through this during our chat today.

atton16:
It works now by launching the new instance of modal every time we need the survey. Now there are only two things left and one bug to fix.

TODO:
1. Cleanup the code
2. Display the survey when user login for the first time

Bug:
1. Scrolling, we can only scroll a bit, if the page is too long then scrolling will stuck at some point.

www/js/survey/enketo-survey.js
```diff
@@ -53,6 +53,11 @@ angular.module('emission.survey.enketo.launch', [
     });
   }
 
+  function resetView() {
+    EnketoSurvey.getState().form.resetView();
+    $('.enketo-plugin article > form').remove();
```
shankari: why do you have to remove the form here? Why not just reset the view? I actually don't even think you need to reset the form if you are using a modal that you remove and re-add every time since that removes the entire modal including the form automatically.

www/templates/survey/enketo-survey-modal.html
```diff
@@ -3,7 +3,7 @@
         <h1 class="title">Survey</h1>
         <button
             class="button button-clear button-positive"
-            ng-click="surveyModal.hide()"
+            ng-click="surveyModalHide()"
```
shankari: `surveyModal.hide()` should work if you are passing in the correct `$scope`

www/js/control/general-settings.js
```diff
@@ -401,25 +401,28 @@ angular.module('emission.main.control',['emission.services',
...
-        EnketoSurveyLaunch.initProfileSurvey();
-        $scope.surveyModal.show();
+        $ionicModal.fromTemplateUrl('templates/survey/enketo-survey-modal.html', {
+          scope: $scope
+        }).then(function (modal) {
+          $scope.surveyModal = modal;
+          $scope.surveyValidateForm = EnketoSurveyLaunch.validateForm;
+          $scope.surveyModalHide = function() {
+            EnketoSurveyLaunch.resetView();
+            $scope.surveyModal.hide();
```
shankari: nit: I don't think you need to hide if you are going to remove anyway

www/js/control/general-settings.js
```diff
@@ -399,6 +400,31 @@ angular.module('emission.main.control',['emission.services',
...
+            $scope.surveyModal.remove();
+            $scope.surveyModal = null;
+          }
+          EnketoSurveyLaunch.initProfileSurvey();
+          $scope.surveyModal.show();
+        });
+    };
+
+    $scope.$on("USERPROFILE_SUBMIT", function(_event, _args) {
+      $scope.surveyModal.hide();
```
shankari: ditto

www/js/diary/list.js
```diff
@@ -763,29 +799,19 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
-          $rootScope.confirmSurveyAnswer
-        ) {
-          $scope.$apply(function() {
-            $rootScope.confirmSurveyTrip.userSurveyAnswer = $rootScope.confirmSurveyAnswer;
-            $rootScope.confirmSurveyTrip = undefined;
-            $rootScope.confirmSurveyAnswer = undefined;
-          });
-        }
+      $scope.$on("CONFIRMSURVEY_SUBMIT", function(_event, _args) {
+        $scope.surveyModal.hide();
```
shankari: ditto

www/js/intro.js
```diff
@@ -83,10 +86,23 @@ angular.module('emission.intro', ['emission.splash.startprefs',
     window.cordova.plugins.BEMJWTAuth.signIn().then(function(userEmail) {
       // ionicToast.show(message, position, stick, time);
       // $scope.next();
-      ionicToast.show(userEmail, 'middle', false, 2500);
+      $scope.userEmail = userEmail;
       CommHelper.registerUser(function(successResult) {
-        $scope.finish();
+          const uuid = successResult.uuid;
+          return EnketoSurvey.getAllSurveyAnswers("manual/user_profile_survey"
```
shankari: Recall that you can use CRUD operations directly for the profile instead of adding an entry into the timestream.

www/js/survey/enketo-survey-service.js
```diff
@@ -49,8 +55,22 @@ angular.module('emission.enketo-survey.service', [
     return loadErrors;
   }
 
+  function getUserProfile(user_properties, answers) {
```
shankari: ditto

shankari:

shankari:
wrt scrolling, I think I already sent a link to 
https://stackoverflow.com/a/37231467
there is another answer specifically for modals
https://stackoverflow.com/a/27638851/4040267

The ionic documentation on scrolling is 
https://ionicframework.com/docs/v1/api/directive/ionScroll/
and
https://ionicframework.com/docs/v1/api/directive/ionContent/

It looks like `ionScroll` turns on the browser scrolling, so you want to use `ionContent` in general.
But if `ionContent` does not work, then you use `ionScroll` but **without** `ionContent`. There are also considerations around specifying the height of the scroll view etc.

So basically I would try with some combination of `ionScroll`, `ionContent` taking into account the considerations documented (e.g. `If there is any dynamic content inside the ion-content, be sure to call .resize() with $ionicScrollDelegate after the content has been added.`), which may be relevant here since we add the questions to the form dynamically.

shankari:
Also, FYI, the modal documentation has a link to a sample codepen
https://ionicframework.com/docs/v1/api/service/$ionicModal/

If I edit the codepen, duplicate the email input multiple times and then open it in the responsive design view, the scrolling works fine. So it definitely works, you just need to experiment with the various options for getting it to work with the form.

codepen_content_scroll_works mov

asiripanich:
> Bunch of comments for the cleanup. But I am merging this now so that @asiripanich could start testing the functionality on his phone.

This PR broke the app.

shankari:
Argh! I didn't test it because I know @atton16 had just tested it on his phone as we were online. It's probably a missing file of some sort. Let me take a look now.

shankari:
@asiripanich can you test again? Since @atton16 restored the `text/javascript` in 
https://github.com/e-mission/e-mission-phone/pull/588/commits/ad230b517724a61018c8e3a943c5b7d5fcfa543c we also pulled it to the channel, which breaks the app.

I have now removed it **in the channel only**. @atton16 don't change it any more; removing and restoring will restore it in the channel as well :)

asiripanich:
> @asiripanich can you test again? Since @atton16 restored the `text/javascript` in
> [ad230b5](https://github.com/e-mission/e-mission-phone/commit/ad230b517724a61018c8e3a943c5b7d5fcfa543c) we also pulled it to the channel, which breaks the app.
> 
> I have now removed it **in the channel only**. @atton16 don't change it any more; removing and restoring will restore it in the channel as well :)

Still not working for me after the fix. Btw, I removed and reinstalled the app.

atton16:
> It works now by launching the new instance of modal every time we need the survey. Now there are only two things left and one bug to fix.
> 
> TODO:
> 
> 1. Cleanup the code
> 2. Display the survey when user login for the first time
> 
> Bug:
> 
> 1. Scrolling, we can only scroll a bit, if the page is too long then scrolling will stuck at some point.

Let me finish point 2 in todo and we should be good to go.

shankari:
I was just about to upload a video saying that I am not getting a WSOD on android (here it is)
works_on_android mov

shankari:
@atton16 when you are fixing this, you might also want to cherry-pick
https://github.com/e-mission/e-mission-phone/pull/596
which has a couple of minor fixes for race conditions with the trip-end prompt.
These may not affect you, since you don't go to a separate screen, only the diary.
However, if you can leverage testing from other groups, why not?!

PR #591, opened 2019-07-02
Unify the implementations of `getDataFromMetrics`

shankari:
and handle `WALKING` and `RUNNING` correctly in both cases.

This fixes https://github.com/e-mission/e-mission-docs/issues/422

Concretely:
- since we can have `WALKING`, `RUNNING` and `ON_FOOT` show up in the metrics
  and we want all of them to be merged to `ON_FOOT`, we maintain a running count of
  foot-like values
- once we have iterated over all the fields, we add an entry for `ON_FOOT` with
  the running count
- unify the previously duplicated `getDataFromMetrics` and
  `getAvgDataFromMetrics` functions by passing in the function that extracts
  the value from the metric. Note that this bug had been fixed (naively) in one of the
  functions in https://github.com/e-mission/e-mission-phone/commit/7f38a53786f462e473bb0adb560570c50cacb286 but the bug persisted in the other. $%$^$$#$# copy-paste.
- change call sites to pass in the correct extractor method

Testing done:
- retrieved data from June 16 2015 and confirmed that we did not see any NaN values

### DIFF
--- a/www/js/metrics.js
+++ b/www/js/metrics.js
@@ -769,10 +769,10 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
    };
 
     $scope.showCharts = function(agg_metrics) {
-      $scope.data.count = getDataFromMetrics(agg_metrics.count);
-      $scope.data.distance = getDataFromMetrics(agg_metrics.distance);
-      $scope.data.duration = getDataFromMetrics(agg_metrics.duration);
-      $scope.data.speed = getDataFromMetrics(agg_metrics.speed);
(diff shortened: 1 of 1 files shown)


shankari:
Screen Shot 2019-07-01 at 6 03 04 PM

shankari:
@tms-martins would you like to review this? Also, please let me know whether I should merge this before or after https://github.com/e-mission/e-mission-phone/pull/576. If I merge it first, you may need to rebase again.

tms-martins:
> after #576
@shankari after, please :)
I'll have a quick look at it in the meantime

tms-martins:
Looks good! 👍

PR #597, opened 2019-07-08
Select carbon dataset via settings menu

tms-martins:
The user can select which carbon-footprint dataset is used in carbon footprint calculations, via the settings menu. A service was created to handle storing of the user's choice and provide the data for the selected dataset. Datasets are hard-coded and include the US dataset (previously in the app and used by default) as well as Germany, France, Austria, Sweden, Norway and Switzerland; and an EU dataset based on the values from these countries. Sources: Tremod and HBEFA.


www/js/control/general-settings.js
```diff
@@ -429,6 +432,19 @@ angular.module('emission.main.control',['emission.services',
             }
         }
     }
+    $scope.changeCarbonDataset = function() {
+        $ionicActionSheet.show({
+          buttons: CarbonDatasetHelper.getCarbonDatasetOptions(),
+          titleText: "Select a dataset for carbon footprint calculations",
+          cancelText: "Cancel",
```
tms-martins: `titleText` and `cancelText` should be translated.

www/js/control/general-settings.js
```diff
@@ -51,6 +52,8 @@ angular.module('emission.main.control',['emission.services',
       ionicDatePicker.openDatePicker(datepickerObject);
     };
 
+    $scope.carbonDatasetString = $translate.instant('carbon-dataset') + ": " + CarbonDatasetHelper.getCurrentCarbonDatasetName();
```
tms-martins: Instead of using `CarbonDatasetHelper.getCurrentCarbonDatasetName()` would it be sensible to retrieve a translation for these too, based on locale/code (e.g. "US", "EU", DE")? 
This would imply that adding more datasets would require also adding more strings in the localization/translation file(s).

www/js/control/general-settings.js
shankari: So you are going to put these into the i18n file, right?

www/js/control/general-settings.js
```diff
@@ -429,6 +432,19 @@ angular.module('emission.main.control',['emission.services',
...
     }
+    $scope.changeCarbonDataset = function() {
+        $ionicActionSheet.show({
+          buttons: CarbonDatasetHelper.getCarbonDatasetOptions(),
+          titleText: "Select a dataset for carbon footprint calculations",
+          cancelText: "Cancel",
+          buttonClicked: function(index, button) {
+            console.log("changeCarbonDataset(): chose locale " + button.value);
+            CarbonDatasetHelper.setCurrentCarbonDatasetLocale(button.value);
+            $scope.carbonDatasetString = "Carbon Dataset: " + CarbonDatasetHelper.getCurrentCarbonDatasetName();
```
shankari: `$translate.instant('carbon-dataset')` here too?

www/js/services.js
```diff
@@ -428,6 +429,160 @@ angular.module('emission.services', ['emission.plugin.logger'])
...
+    }
+  };
+
+  var defaultCarbonDatasetCode = 'US';
+  var currentCarbonDatasetCode;
+
+  this.initialize = function() {
+    console.debug("Tiago: CarbonDatasetHelper.initialize()");
+    currentCarbonDatasetCode = defaultCarbonDatasetCode;
+    var value = KVStore.getDirect(CARBON_DATASET_KEY);
```
shankari: Not sure `getDirect` is the right call here. We used to use local storage quite a lot, but then it turns out that iOS can delete local storage arbitrarily. So I now use `getDirect` only for state that won't annoy the user if cleared (e.g. is the walkthrough complete).

See https://github.com/e-mission/e-mission-phone/issues/443#issuecomment-423066729 and https://github.com/e-mission/e-mission-phone/pull/470 for more details on why I had to change the storage.

www/js/services.js
```diff
@@ -428,6 +429,160 @@ angular.module('emission.services', ['emission.plugin.logger'])
...
+    console.debug("Tiago: CarbonDatasetHelper.initialize() obtained value from storage [" + value + "]");
```
shankari: You may want to turn some of these to Logger.log messages just in case we run into complaints. Those are logged to the app logs, which can be emailed and checked if unexpected heisenbugs occur.

www/js/control/general-settings.js
shankari: I would assume so. I presume that `Kohle Datei: Germany` is going to look really weird. A workaround that would make it possible to add datasets without translations is to use the locale code directly (e.g. US/EU/DE/FR, etc). I think most people know country codes well enough to figure that out.

Thoughts? @PatGendre @jf87 @pgranato any thoughts from your side?

www/js/control/general-settings.js
tms-martins: Yes :)

www/js/control/general-settings.js
tms-martins: Another option is 
```
if translate(countryCode) is defined
  use translation
else
  use countryCode
```

www/js/control/general-settings.js
tms-martins: Yes :)

www/js/services.js
tms-martins: Thank you, I will.

www/js/services.js
tms-martins: Should I be using `storage.set()` and `storage.get()` instead?

www/js/services.js
tms-martins: I'll try to keep the ones which are useful and less repetitive

www/js/control/general-settings.js
tms-martins: I just tried requesting a translation string which doesn't exist:
```
var foo = $translate.instant('value-which-doesnt-exist');
``` 
instead of getting an `undefined` it apparently just broke (and the settings UI has all the Angular properties between curly braces). Is this intended and/or am I doing something wrong?

For now I will simply use country codes.

shankari: @DO3B do you know what happens if we try to request a translation string that doesn't exist?

www/js/services.js
shankari: Yes.

www/js/control/general-settings.js
DO3B: It should just print the key because it didn't find any values. 

`$translate.instant('value-which-doesnt-exist'`) will return `value-which-doesnt-exist`

shankari:
I think that the main steps left are `getDirect` -> `get` and testing. I will merge as soon as those are done - I will not wait for @DO3B's reply since it is more of a long-term consideration.

www/js/services.js
tms-martins: I can't seem to solve this in an elegant way (maybe limited knowledge of JS). Having the CarbonDatasetHelper initialize on-the-fly with a promise inside leads to the function returning with the default dataset value (US) before the promise is solved and the correct dataset code loaded from KVStore.
Would you have any suggestions on where/when to read the KVStore value beforehand? Or does it seem that I am doing something wrong, or overlooking a simple solution perhaps?

www/js/services.js
```diff
@@ -428,6 +429,162 @@ angular.module('emission.services', ['emission.plugin.logger'])
...
+      }
+    }
+  };
+
+  var defaultCarbonDatasetCode = 'US';
+  var currentCarbonDatasetCode;
+
+  this.initialize = function() {
+    currentCarbonDatasetCode = defaultCarbonDatasetCode;
+    var value = KVStore.getDirect(CARBON_DATASET_KEY);
```
shankari:
```suggestion
     return KVStore.get(CARBON_DATASET_KEY).then(function(value) {
          // stuff to do with the value. This will run asynchronously,
          // but it doesn't look like you rely on this for any reason
          // if you do, then you can just `then` on the returned promise to execute the dependencies **after** this is loaded.
         Logger.log("CarbonDatasetHelper.initialize() obtained value from storage [" + value + "]");
       if (!value) {
          value = defaultCarbonDatasetCode;
          Logger.log("CarbonDatasetHelper.initialize() no value in storage, using [" + value + "] instead");
            }
            this.setCurrentCarbonDatasetLocale(value); 
        });
```

www/js/services.js
shankari: Added a suggestion below

www/js/services.js
shankari: Does this work?

shankari: Actually, wait. You do use this to return values from `getCurrentCarbonDatasetName`. Let me add the suggestions there too...

www/js/services.js
```diff
@@ -428,6 +429,162 @@ angular.module('emission.services', ['emission.plugin.logger'])
...
+    return carbonDatasets[currentCarbonDatasetCode].regionName;
+  };
+
+  this.getCurrentCarbonDatasetCode = function () {
+    if (!currentCarbonDatasetCode) this.initialize();
+    return currentCarbonDatasetCode;
+  };
+
+  this.getCurrentCarbonDatasetFootprint = function () {
+    if (!currentCarbonDatasetCode) this.initialize();
```
shankari:
```suggestion
    return new Promise(resolve, reject) {
        if (!currentCarbonDatasetCode) this.initialize().then(resolve(carbonDatasets....);
        else resolve(carbonDataset...
    };

```

tms-martins:
> I am not sure that your lazily loaded get/set structure works well with promises

I agree. And if I return promises from `getCurrentCarbonDatasetFootprint` and `getCurrentCarbonDatasetCode` the code/logoc will feel needlessly complicated. 

I don't want to use the lazy loading, I would prefer to ensure that the service is initialized once and only once when the app starts. Even so, I have the issue that:

1. I need to initialize the carbon dataset service by reading a user preference from KVStore
2. Reading from KVStore is asynchronous
3. By the time the metrics view is displayed, the service hasn't been initialized yet (metrics will perform calculations based on the default data)

This seems like an issue that would be generalized to other apps/views, so by now there should be a way to solve it - e.g. a synchronous app initialization step where user preferences are loaded. Would this be the case somehow?

...or should I initialize the service with the view, returning a promise and then refreshing once the promise is resolved?

tms-martins:
> initialize the service with the view, returning a promise and then refreshing once the promise is resolved

That is the approach I am trying now. The service is initialized in the `metrics` controller; this assumes that the metrics view is the app UI entry point.

shankari:
> This seems like an issue that would be generalized to other apps/views, so by now there should be a way to solve it - e.g. a synchronous app initialization step where user preferences are loaded. Would this be the case somehow?

There is no synchronous app inititialization step, unfortunately, since javascript is designed from the bottom up to be asynchronous so that the browser can start parsing and displaying the HTML template and then you fill in the values asynchnronously when the results come in.

e.g. https://stackoverflow.com/questions/21819858/how-to-wrap-async-function-calls-into-a-sync-function-in-node-js-or-javascript

> Asynchronous function using callback/promise is the DNA of JavasSript and Node.js. 

That post also suggests some options for working around async calls, notably fiber or the deasync module, but I haven't tried them out.

shankari:
> ...or should I initialize the service with the view, returning a promise and then refreshing once the promise is resolved?

You typically don't pass in the view to the service, instead you call the service functions, which return promises, and in the view, you refresh once the promise is resolved.

> That is the approach I am trying now. The service is initialized in the metrics controller; this assumes that the metrics view is the app UI entry point.

And that is all fine; it just assumes that `getCurrentCarbonDatasetFootprint` is synchronous. But it is actually asynchronous. So instead of 

```
     var footprint = CarbonDatasetHelper.getCurrentCarbonDatasetFootprint();
```

in the metrics controller, you need to do something like

```
CarbonDatasetHelper.getCurrentCarbonDatasetFootprint().then(function(value) {
    footprint = value;
    ....
}
```

You could even have the initial footprint be the default, and then have it be overwritten once the async values come in, - e.g.

```
var footprint = // default footprint value; will be set synchronously
CarbonDatasetHelper.getCurrentCarbonDatasetFootprint().then(function(value) {
    // overwritten when async call comes in
    footprint = value;
    ....
}
```

Remember to use `$scope.$apply` to ensure that angular sees the modification.

tms-martins:
@shankari Thanks for the feedback! 

> You typically don't pass in the view to the service, instead you call the service functions

Yes... I explained myself wrong 😅 I wrote "should I initialize the service with the view" but I meant something like "should I initialize the service from within the controller code, when the controller is initialized"

shankari:
@tms-martins do you have an ETA for this change? Note that @DO3B is adding support for tram and subway as well (#601), so you could add them to the footprint map as well...

tms-martins:
> do you have an ETA for this change?

In a couple of hours :) I managed to set some time aside for this.

www/js/metrics.js
```diff
@@ -38,6 +38,10 @@ angular.module('emission.main.metrics',['nvd3',
     };
 
     $ionicPlatform.ready(function() {
+        CarbonDatasetHelper.loadCarbonDatasetLocale().then(function(result) {
+          // refresh the page so that all metrics are re-calculated with the correct footprint dataset
+          $scope.doRefresh();
```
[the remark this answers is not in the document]
tms-martins: > you can alternatively move this to the other `$ionicPlatform.ready` and call `getData()` only after this data is loaded.

seems to be the most straightforward solution

tms-martins:
@shankari I moved `getData()` up and merged with the latest master. That worked out better than rebasing (like last time, which made a mess of everything).

shankari:
@DO3B I am not sure @tms-martins is going to upload a testing video in the near future since his contract is complete. The change looks pretty clean and I don't anticipate any issues.

Would you mind a lot if I just merged the change since I know that @PatGendre is waiting for this functionality? If you find any issues, we can ask @tms-martins to fix it, but I think that is unlikely and that way we don't have to wait for another week to merge it.

shankari:
Note that you will also need to add the support for the expanded modes to the new carbon footprint maps as well. Fortunately, @tms-martins has linked to the sources (Tremod, HBEFA)

DO3B:
I don't mind at all, I will work on it Friday.

tms-martins:
> LGTM! @tms-martins Can you add a short testing video so I can ensure that it minimally works?

I can do that later this evening (Berlin time). 

Also thank you @DO3B for working on the new modes. Please message me if you have any questions. As of now I cannot contribute much more to the project.

shankari:
@tms-martins don't even bother - I already merged the change and @DO3B will test it when he integrates with the PR for the other modes. Can you instead figure out the correct values for the new modes from TREMOD and add them as a comment in #601? The TREMOD report is in German and so it is hard for us to figure it out ourselves.

shankari:
> As of now I cannot contribute much more to the project.

Completely understand, and I appreciate your finishing up your pending changes even after your contract is done.

PR #598, opened 2019-07-11
Add italian support and missing translations

DO3B:
While working on the fabmob app, I noticed some missings translations. Moreover, I have removed the HTML chars and the HTML from some of the keys to use the **escape** strategy so no more warning from `angular-translate`. I have also took the liberty to add the "it" key to `angular-translate` available language key since we have now an italian translation :)

### DIFF
--- a/www/js/metrics.js
+++ b/www/js/metrics.js
@@ -97,7 +97,8 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
       showFilter: false,
       showVis: true,
       showResult: true,
-      current: "Last week",
+      current: true,
+      currentString: $translate.instant('metrics.last-week'),
       showChart: false,
       showSummary: true,
       showMe: true,
@@ -193,10 +194,10 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
       return $scope.userData.weightUnit == 0? "unit-toggle-right-active hvcenter" : "unit-toggle-right hvcenter";
     }
     $scope.currentQueryForCalorie = function() {
-      return $scope.uictrl.current == "Last week"? "user-calorie-percentage" : "user-calorie-no-percentage";
+      return $scope.uictrl.current ? "user-calorie-percentage" : "user-calorie-no-percentage";
     }
     $scope.currentQueryForCarbon = function() {
-      return $scope.uictrl.current == "Last week"? "user-carbon-percentage" : "user-carbon-no-percentage";
+      return $scope.uictrl.current ? "user-carbon-percentage" : "user-carbon-no-percentage";
     }
     $scope.showRange = function() {
       if ($scope.uictrl.showFilter) {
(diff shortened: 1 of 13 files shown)


www/js/services.js
```diff
@@ -158,6 +158,27 @@ angular.module('emission.services', ['emission.plugin.logger'])
           window.cordova.plugins.BEMServerComm.getUserPersonalData("/pipeline/get_complete_ts", resolve, reject);
       });
     };
+  
+    this.getEmailConfig = function () {
```
shankari: Hm. In general, the functions in `CommHelper` are for remote calls only. That's why it is called `CommHelper` for the communication. I do see that you want a common email configuration function though.

This is more work, but would you consider creating a new service for emailing (e.g. something like `www/js/control/emailService.js`) that would include `getEmailConfig`, `emailCache` and `emailLogs`.

I haven't looked at in detail, but I think that `emailCache` and `emailLogs` are pretty similar in implementation, so for extra credit, you could also refactor them into a single implementation that takes the database to email as a parameter. But I will not insist on this for the merge

www/js/metrics.js
```diff
@@ -193,10 +194,10 @@ angular.module('emission.main.metrics',['nvd3', 'emission.services', 'ionic-date
       return $scope.userData.weightUnit == 0? "unit-toggle-right-active hvcenter" : "unit-toggle-right hvcenter";
     }
     $scope.currentQueryForCalorie = function() {
-      return $scope.uictrl.current == "Last week"? "user-calorie-percentage" : "user-calorie-no-percentage";
+      return $scope.uictrl.current ? "user-calorie-percentage" : "user-calorie-no-percentage";
```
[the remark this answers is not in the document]
DO3B: Yup, to avoid to compare two `$translate.instant('metrics.<>')`, I decided to use a boolean instead.

www/js/services.js
DO3B: I added `emailService.js`, and merged `emailCache` and `emailLogs` into `sendEmail(boolean isLog)` since it was just the enclosed file which changed between the two functions. If you prefer, we can still use the database to email as a parameter.

shankari: @DO3B I do prefer to use the database to email as the parameter for two reasons:
1. It reduces the dependency between the standard service (`EmailHelper`) and the places it is called from
1. It makes it easier to reuse `EmailHelper` if we want to email out other files or databases

I see that in addition to the database, the log message is slightly different as well, which might be why you wanted to use `isLog`. But I think that you can change the log message to include the filename instead and it will still be meaningful.

DO3B: So, it's done (https://github.com/e-mission/e-mission-phone/pull/598/commits/2ff0d81b71a7ab54a9688c3713e0076faed5f363). I already removed the misleading comment in the other commit (https://github.com/e-mission/e-mission-phone/pull/598/commits/cd1c0263df3fb680d6df3a0f3f6b45629f05cce8).

PR #600, opened 2019-08-06
Fix today button by upgrading to most recent version of the datepicker

shankari:
Unfortunately, there is no matching release and no matching dist file.
Therefore, we remove it from bower, check in the manually compiled dist file
and switch index.html to use the manually specified value

Testing done: see animated gif below
today_works_with_new_changes mov


shankari:
This fixes https://github.com/e-mission/e-mission-docs/issues/432
@asiripanich @DO3B would you like to review this fairly simple change?

PR #601, opened 2019-08-12
Add tram and subway

DO3B:
This PR contains the UI changes linked to https://github.com/e-mission/e-mission-server/pull/709. I didn't find the footprint of both subway and tram (or at least I don't know where to find them).


shankari:
@DO3B at least for the US 
https://github.com/e-mission/e-mission-server/blob/master/front/server/carbon_calc_details.html#L80

shankari:
BART is a subway, Caltrain is a commuter train (S-Bahn?) and the other two are tram lines, IIRC.
You can also check their OSM tags to confirm and map back to OSM.

shankari:
I just checked this on OSM:
- Green Line (https://www.openstreetmap.org/way/25653645) is light_rail
- BART (https://www.openstreetmap.org/relation/2851613) is subway
- Muni (https://www.openstreetmap.org/way/160305461) is tram
- Caltrain (https://www.openstreetmap.org/relation/9606325) is rail

So shouldn't TRAIN be Caltrain (74) instead of 92?
Also, are you going to support light_rail as separate from tram, or is that not very common in Europe?

shankari:
Europe also has light rail 
https://wiki.openstreetmap.org/wiki/Tag:railway%3Dlight_rail

PatGendre:
@shankari @DO3B hello from Bangkok,
Given the OSM examples, I suggest to consider light rail to be grouped with metro

shankari:
@PatGendre I don't think there is an OSM tag for metro, only subway. And subway typically implies underground. From the description on the wiki, I think that they are using light rail for a subway-like line but aboveground?

e.g.

> Compared to light rail, a subway is predominantly underground, and, while overground, has a separate right of way nearly without level crossings.
> constructed stops rather than simple bus-like stops at the road level.

shankari:
I don't see an icon in fontawesome for light rail though so you would probably just use tram or subway but with a different color

PatGendre:
@Shankari yes I think you are right in OSM, subway is for underground, so we we can consider bith subway and light rail as metro in e-mission.

shankari:
@PatGendre given the current code structure, I think it would be harder for @DO3B to merge two other train modes into metro than to just support a new mode. But I don't think either will be too hard. I'll leave this up to your judgement...

DO3B:
When reading the description of the light_rail in the [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dlight_rail). I think it is better to differentiate it from the subway.

DO3B:
I added the light rail. For now, it is just a blue subway icon (I changed manually the data to test it on the phone, we don't have light rail in Marseille). 

|Light Rail|
|---|
|Screenshot_1565772442|

PatGendre:
@DO3B ok fine! Thanks :-)

shankari:
I have merged the server-side changes, and @tms-martins changes which have resulted in a merge conflict in the footprint calculation. Once you resolve that, and include the correct carbon footprint values for the EU for the new modes, I can merge this 😃

shankari:
@tms-martins can you please add the tram, subway and light_rail CO2 values to this PR as a comment for @DO3B to incorporate?

I found the TREMOND website, https://www.ifeu.de/en/methods/models/tremod/ but the reports are in pretty dense German and @DO3B is French :smiley so I don't know if he can figure out the values directly.

DO3B:
Well, I tried to look at the report but after few hours I can conclude that I don't speak German at all . I used the US values (for now) to fix the merge conflict. This might be my last contribution because my internship is over and my classes resume.

tms-martins:
@shankari @DO3B The Tremod table I used was actually [this one](https://www.umweltbundesamt.de/themen/verkehr-laerm/emissionsdaten#verkehrsmittelvergleich_personenverkehr). I think it's fine to use the US values there right now. Later today or early tomorrow I will have a look and see if I can find more accurate values in Tremod (I think tram, subway and light rail are grouped there) and make a separate PR for those, if that's ok?

shankari:
@DO3B @tms-martins all sounds good. I am merging this now. @tms-martins thanks again for being conscientious to finish up 😃

PR #602, opened 2019-08-17
Updated carbon footprint values for LIGHT_RAIL and TRAM

tms-martins:
... based on Tremod stats (which group tram, urban train and subway, so the value is the same for both). Also removed duplicate TRAIN values.


DO3B:
Thank you @tms-martins ! The duplicated values was a refactoring fail, that should be "SUBWAY" instead.

tms-martins:
Ah, ok ;)
I'll fix that; it's the same value.

PR #604, opened 2019-09-01
fix: disable notifications

atton16:
Disable post-trip notifications.

ref issue: https://github.com/asiripanich/e-mission-phone/issues/16


shankari:
If that's what you want, that's what you get...

PR #610, opened 2019-10-27
patch(survey): speedup server calls

atton16:


www/js/diary/services.js
```diff
@@ -991,12 +991,24 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
       var tripsFromServerPromise = timeline.updateFromServer(day);
       var isProcessingCompletePromise = timeline.isProcessingComplete(day);
 
+      // Also mode/purpose and survey answers
+      var tq = $window.cordova.plugins.BEMUserCache.getAllTimeQuery();
+      var modesPromise = UnifiedDataLoader.getUnifiedMessagesForInterval('manual/mode_confirm', tq);
+      var purposesPromise = UnifiedDataLoader.getUnifiedMessagesForInterval('manual/purpose_confirm', tq);
+      var surveyAnswersPromise = EnketoSurvey.getAllSurveyAnswers("manual/confirm_survey", { populateLabels: true });
```
shankari: you don't need the `surveyAnswersPromise` here since the main branch does not use the survey answers. It also doesn't have the `EneketoSurvey` file, so this won't even load. Have you even tested this on master?

PR #613, opened 2019-12-31
Rk unsw

atton16:
Features:
1. Add timestamp to survey for both profile survey and confirm survey (trip survey)
   - This helps us track user more efficiently via our dashboard
2. Remove unused settings that might make our users confuse:
   1. Dark theme
   2. Carbon dataset
   3. Developer zone
3. Add `.nvmrc` to lock node version to `9.4.0`


shankari:
This looks pretty straightforward, and I assume you have tested it with the devapp.

shankari:

PR #614, opened 2020-01-01
Rk unsw

atton16:
Features
1. Fix invalid date displayed for processed trips

Apparently, the date stored in `tripgj.data.properties.start_local_dt` and `tripgj.data.properties.end_local_dt` use a `1-12` month value while `moment` needs `0-11` month value. The test was done on December which translated to value `12`. Hence moment throw `Invalid date`. However, this problem does not persist for `draft trip` so we skip the fix for those trips.

Before:
Screen Shot 2563-01-01 at 21 00 50

After:
Screen Shot 2563-01-01 at 21 00 25

2. Limit date picker to the day the user register (we use timestamp from the oldest user profile survey record)

Example: I answered on January 1st, 2020
Screen Shot 2563-01-01 at 21 24 36


atton16:
Wrong branch

PR #620, opened 2020-03-11
Rciti1

atton16:


shankari:
@asiripanich @atton16 I just deployed this build to the rciti1 channel. It should be live in ~ 10 mins.
Could you test it using
https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=rciti1&clear_usercache=true&clear_local_storage=true

If it works correctly, then I will follow the same procedure to set up the other channels as well so we can make sure the setup stuff is under control. Please let me know ASAP.

asiripanich:
I only see a blank screen after I applied the UI update.

shankari:
@asiripanich I debugged this further. This is because the `index.html` has the line

```
<script type="text/javascript" src="cordova.js"></script>
```

That is required for the live-reload in the devapp, but it will break emTripLog. I had already ensured that the master for the deploy code included it properly.
https://github.com/e-mission/e-mission-phone/blob/deploy-upload-master/www/index.html#L64

```
<script src="cordova.js"></script>
```

However, @atton16 had edited this line to remove (https://github.com/e-mission/e-mission-phone/commit/6040e1887cffd1dd65d7b2b2fdf307e783cc7b84) and reinsert (https://github.com/e-mission/e-mission-phone/commit/ad230b517724a61018c8e3a943c5b7d5fcfa543c) the `type="text/javascript" ` part, so the final merged version had it inserted.

@atton16 since I assume you are not going to be testing live-reload from the rciti1...rciti4 branches, can you remove `type="text/javascript"` from them? Then any further pulls won't reintroduce it...

atton16:
Fixed!

asiripanich:
@shankari I'm still seeing the white screen. @atton16 mentioned that it worked on the devapp.

shankari:
I merged and pushed the change, but forgot to launch the build on the ionic dashboard. This is what happens if you try to work on things after midnight. Try again now.

> @atton16 mentioned that it worked on the devapp.

This doesn't help because the cause of the WSOD is the one piece of code that needs to be different between devapp and emTripLog.

asiripanich:
;) @shankari thank you. It is working now.

shankari:
great, so I will deploy the other channels as well so that at least that part of the setup is complete.

shankari:
Just deployed to the rciti2, rciti3 and rciti4 channels. Please test that the setup worked in ~ 10 mins. After this, you should be able to test only on one of the channels, or on the `rkunsw` channel :)

PR #624, opened 2020-03-11
Rk unsw End Survey

atton16:
Hi,

We are trying to implement the End Survey button with a custom post-survey-save logic.

Basically what we want is to
1. Force Sync data
2. Logout

I have borrowed the `forceSync` and `returnToIntro` functions from `general-settings.js` for force sync and logout, respectively.

During the test, it seems that `ControlSyncHelper.forceSync()` does not executed. Hence the survey modal dismissed and do nothing.

I am not sure what we have to do to get force sync work. I have tried with the force sync button on the settings screen. The original force sync button does not seem to work as well.

Perhaps, you might be able to give us some advise.


www/js/survey/enketo-survey.js
```diff
@@ -206,6 +207,7 @@ angular.module('emission.survey.enketo.launch', [
         console.log("Added " + ClientStats.getStatKeys().BUTTON_FORCE_SYNC + " event");
       });
     return ControlSyncHelper.forceSync().then(function () {
+      console.log('forceSync start');
```
atton16: This line does not get executed.

shankari: Unfortunately, that is not a valid assumption.

The sync code was originally designed to be only automatically triggered so that the data would sync periodically. I later added an interface to force the sync, but that just triggers the sync process. So `forceSync` triggers the sync, but you can't rely on knowing when the sync is complete.

There were some prior issues in which I debugged this in detail. Did you check through prior issues/code history to validate your assumption?

atton16: > Unfortunately, that is not a valid assumption. There were some prior issues in which I debugged this in detail. Did you check through prior issues/code history to validate your assumption?
> 
> IIRC, the sync code was originally designed to be only automatically triggered so that the data would sync periodically. I later added an interface to force the sync, but that just triggers the sync process. So `forceSync` triggers the sync, but you can't rely on knowing when the sync is complete.

How does this any different than the force sync button in the settings.js page though? Or does this mean devapp and emTripLog behave differently for the sync process?

asiripanich: @shankari so speaking on the high-level, what is your suggestion if we want to make sure the end survey get sent right away?

shankari: It is not any different from the `forceSync` button in the `settings.js` page. The sync is launched and completes. It just doesn't trigger the `then` callback.

@asiripanich note that we only sync values before the current end of trip. You want to use the `endTripAndForceSync` method to push the end survey.  The sync will still happen, but you won't be able to display a popup when the sync is complete since the callback is not triggered. Please see the history around this code for more details.

shankari:
@atton16 incidentally, I think that copying functions around is a pretty terrible idea wrt maintenance. Specially if you are going to argue that you built features that others can re-use later, you need to think in terms of modularity and long-term maintenance, not just what works right now.

asiripanich:
@shankari I agree it should be done properly for long term maintainability. However, as you know we need to deploy this ASAP hence Attawit are working hard to make this possible and some shortcuts had to be made. We will come back and refactor this for the master branch if things go well. Hope you understand.

shankari:
Closing this since it is superceded by https://github.com/e-mission/e-mission-phone/pull/629

shankari:
@atton16 @asiripanich please reopen if you disagree

PR #625, opened 2020-03-12
Update index.html

atton16:
fix(Deploy): fix white screen


shankari:
deployed, should be live in ~10 mins.
@asiripanich, can you check this out again? If it still doesn't work, I will debug tomorrow am my time.

PR #636, opened 2020-03-30
Upgrade the config.xml and plugins

shankari:
- upgrade to API 28
- turn off google analytics
- upgrade to the latest version of the plugins


shankari:
At this point, the android build works

```
$ cordova build android
Copying locales: Languages found -> fr,it
Copying locales: I found fr, I will now copy the files.
...
BUILD SUCCESSFUL in 18s
45 actionable tasks: 43 executed, 2 up-to-date
Built the following apk(s):
	/Users/shankari/e-mission/e-mission-phone/platforms/android/build/outputs/apk/debug/android-debug.apk
```

shankari:
Will return to this to fix the iOS build tonight.

shankari:
Apple will not accept apps using anything other than WKWebView. According to the ionic docs, this requires `cordova-ios 5.1.0` or above.

> Using Cordova? Update to cordova-ios 5.1.0 and above, install the WKWebView plugin, and update each Cordova plugin to the latest version.

And yes, that is in fact required
https://cordova.apache.org/announcements/2019/11/25/cordova-ios-release-5.1.0.html
includes 
> The ability to disable UIWebView during the compile-time.

which is needed because of
https://github.com/apache/cordova-ios/issues/661

There are apparently several limitations with WKWebView.
https://cordova.apache.org/news/2018/08/01/future-cordova-ios-webview.html

Of the limitations they highlight, 

>    Cookies don't persist. This is a WebKit bug, but someone has created a plugin for a workaround. See CB-12074

We don't use cookies

>     Can't delete cookies. This is/was a WebKit bug (2015), we need to test for the iOS 11/12. See CB-11297

We don't use cookies

>    Can't execute JavaScript code in the background. There are several issues related to this. See CB-12815

Most of our functionality is already native code. But we do 

>    XmlHttpRequests don't work, because of Cross-Origin Resource Sharing issue (CORS). There is a workaround plugin created by Oracle (UPL licensed, which is Apache-2.0 compatible). See CB-10143

We use XHR for the aggregate calls, need to fix them.

>    Migration of localStorage from UIWebView. There is a migration plugin available. See CB-11974

Fortunately, we moved away from local storage to the database a long time ago. Should probably remove alert about data missing from local storage.

>     iframes will not be supported any longer (they are now CORS restricted in WKWebView), and may be partially or completely broken. This may lead to incompatibilities with the same code in other Cordova platforms.

No iframes

>    Known issues with WKWebView on iOS pre-11 which will be deprecated and dropped in a future Cordova release

???

shankari:
Given all the stuff that broke when we last tried to upgrade to cordova-ios@5, and the newer things that will break, I will migrate to the new system in the new app (https://github.com/covid19database/phone-app), which has fewer dependencies. 

Then, I can migrate the changes back here.

shankari:
It turns out that I needed to upgrade not just the iOS version but also the android version in order to support kotlin. I will migrate all the changes over, but for now, I will merge these changes.

PR #637, opened 2020-04-06
Emission contribution

tarekeg:
- Fixing undefined Time in Diary 
- Fixing streets display in diary
- Changing map tiles and attribution to oms due to loading problem on android

### DIFF
--- a/www/js/services.js
+++ b/www/js/services.js
@@ -564,9 +564,9 @@ angular.module('emission.services', ['emission.plugin.logger',
 
     config.getMapTiles = function() {
       return {
-          tileLayer: 'http://tile.stamen.com/terrain/{z}/{x}/{y}.png',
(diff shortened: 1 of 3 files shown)


www/js/services.js
```diff
@@ -564,9 +564,9 @@ angular.module('emission.services', ['emission.plugin.logger',
 
     config.getMapTiles = function() {
       return {
-          tileLayer: 'http://tile.stamen.com/terrain/{z}/{x}/{y}.png',
+          tileLayer: 'http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
```
shankari, line 567: Minor correction: with the changes that require https on android 9+, we should use the https version of this.
https://medium.com/@son.rommer/fix-cleartext-traffic-error-in-android-9-pie-2f4e9e2235e6

PR #638, opened 2020-04-18
Import the setup scripts and the CI workflows for Github Actions

shankari:


shankari:
This failure is expected since we haven't migrated over all the plugins yet. We will first migrate over the other changes and then fix the CI in a separate branch.

PR #642, opened 2020-04-20
patch: update survey version

atton16:


shankari:
Channels for `rciti1`, `rciti2`, `rciti3` and `rciti4` updated.

PR #643, opened 2020-04-20
Rciti2

atton16:


www/js/intro.js
```diff
@@ -109,7 +109,11 @@ angular.module('emission.intro', ['emission.splash.startprefs',
       $scope.userEmail = userEmail;
       CommHelper.registerUser(function(successResult) {
           const uuid = successResult.uuid;
+<<<<<<< HEAD
```
shankari, line 112: This has a merge conflict.

atton16: I thought I have resolve that before creating PR. However, I just did now.

PR #651, opened 2020-04-23
Rciti2 - test deploy

asiripanich:
- Several updates from master 
- added a second on boarding screen to explain and guide the user to setup the location permission for emTripLog of their phone 
- Updated survey forms
- Justified text in the consent page
- Cleaned up older versions of the surveys


www/i18n/en.json
```diff
@@ -245,17 +245,28 @@
         }
     },
 
+    "intro": {
+        "permissions": {
+            "locationPermExplanation-android-lt-6": "you accepted the permission during installation. You don't need to do anything now.",
+            "locationPermExplanation-android-gte-6": "please select 'allow'",
+            "locationPermExplanation-ios-lt-13": "please select 'Always allow'. This allows us to protect you even when you are not actively using the app",
```
shankari, line 252: Didn't you want to change "protect" here?

www/js/survey/enketo-survey.js
```diff
@@ -100,7 +100,7 @@ angular.module('emission.survey.enketo.launch', [
 
   function initConfirmSurvey(opts) {
     return initSurvey({
-      form_location: 'json/20200420-01-trip-end-survey.json',
+      form_location: 'json/trip-end-survey.json',
```
shankari, line 103: Is this what you want? I thought you wanted the most recent version of the survey...

www/i18n/en.json
asiripanich: 'protect' is actually not a bad idea for now.. :)

www/js/survey/enketo-survey.js
asiripanich: That's the most recent version. I removed all older forms.

shankari:
I'm approving this for now since the changes are relatively minor, but in the future, to make it easier to review, it would be good to split to merge from master (no review needed) and the changes that you made.

asiripanich:
> approving this for now since the changes are relatively minor, but in the future, to make it easier to review, it would be good to split to merge from master (no review needed) and the changes that you made.

Sorry for the mess. :( Will do better next time.

shankari:
no worries, as you probably saw, I deployed the channel already

PR #646, opened 2020-04-21
Rciti1 - update survey titles and consent

atton16:

PR #652, opened 2020-04-25
fix: white screen after consent page

asiripanich:
Just a note, you forgot to mention `www/templates/intro/sensor_explanation.html` file https://github.com/e-mission/e-mission-phone/pull/646#issuecomment-617267333. That was why we got a white screen after the consent page.


shankari:
You're right! I forgot that I had checked in the sensor_explanation file as a separate commit.

asiripanich:
deploy in 10mins? :)

shankari:
deployed, should be live now.

PR #653, opened 2020-04-26
Rciti2 - use inapp browser to launch the user profile survey

asiripanich:
Avoiding significant changes required to make this work as discussed in https://github.com/e-mission/e-mission-docs/issues/525.

We use socketio for checking the status of profile submission.
This is the workaround because postMessage API is not available in the the current inapp browser version.

screencast


shankari:
Cool workaround! FYI, as part of the upgrade to the latest `cordova-ios`, we had to switch to WKWebView, which has restrictions on file loads. So the cordova workaround is to run a local browser and local from `http://localhost`. 

I wonder if the enketo stuff will work better in that case. But it is definitely consistent with this approach.

I did notice, in the video you have uploaded, that there is a Loading error. What's up with that? I am happy to merge the changes right now but and fix any issues later.

asiripanich:
Deploy in 10mins? :)

asiripanich:
@atton16 can u explain why that enketo error occur in the devapp?

atton16:
> @atton16 can u explain why that enketo error occur in the devapp?
> 
> It happens every time on the devapp, but works fine on emTripLog from what I can tell.

The only reason I can think of is for some reason JavaScript is disabled in in-app browser on the devapp. Oddly, this does not occur in emTripLog.

atton16:
> Cool workaround! FYI, as part of the upgrade to the latest `cordova-ios`, we had to switch to WKWebView, which has restrictions on file loads. So the cordova workaround is to run a local browser and local from `http://localhost`.
> 
> I wonder if the enketo stuff will work better in that case. But it is definitely consistent with this approach.
> 
> I did notice, in the video you have uploaded, that there is a Loading error. What's up with that? I am happy to merge the changes right now and fix any issues later.

if `http://localhost` works, then we need to find a solution to self-host the socket server from inside the app. If we could make everything self-contained then we could achieve a more reliable behaviour.

asiripanich:
@atton16 Umm I have just tried this on emTripLog but got stuck on the google login page after I answered the user profile survey.

shankari:
> if http://localhost works, then we need to find a solution to self-host the socket server from inside the app. If we could make everything self-contained then we could achieve a more reliable behaviour.

http://localhost works on the currently in-progress upgrade to the latest `cordova-ios` which is not going to be done today and which needs to be deployed to the app store after passing app store review once it is actually done.

This was more of a musing for future reference, not an immediate fix for you.

shankari:
I took a quick look at this since you are under a time crunch. Although I need to really get back to the workshop now. I also saw the same behavior as @asiripanich 

The error appears to be related to 

```
04-26 14:08:16.760  6378  6378 D SystemWebChromeClient: file:///data/user/0/edu.berkeley.eecs.embase/app_aa811daf-4a66-47d4-b004-2d2b77226dcd/manual_lib/socket.io-2.3.0/socket.io.js: Line 7 : Refused to connect to 'https://emission-socket.byamarin.com/socket.io/?EIO=3&transport=polling&t=N6uKRpt' because it violates the following Content Security Policy directive: "default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://api.ionic.io/push/tokens emission: 'unsafe-eval'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.
```

@atton16 did you fix the `index.html`?

shankari:
also, if your goal is to just close the inapp browser when the survey is done, you can use the redirect URL trick that `survey.js` uses. Basically redirect to a known URL at the end of the survey and add a listener to `loadURL` that closes the browser when we see the final redirect URL

atton16:
> I took a quick look at this since you are under a time crunch. Although I need to really get back to the workshop now. I also saw the same behavior as @asiripanich
> 
> The error appears to be related to
> 
> ```
> 04-26 14:08:16.760  6378  6378 D SystemWebChromeClient: file:///data/user/0/edu.berkeley.eecs.embase/app_aa811daf-4a66-47d4-b004-2d2b77226dcd/manual_lib/socket.io-2.3.0/socket.io.js: Line 7 : Refused to connect to 'https://emission-socket.byamarin.com/socket.io/?EIO=3&transport=polling&t=N6uKRpt' because it violates the following Content Security Policy directive: "default-src 'self' data: gap: https://ssl.gstatic.com http://nominatim.openstreetmap.org https://e-mission.eecs.berkeley.edu https://api.ionic.io/push/tokens emission: 'unsafe-eval'". Note that 'connect-src' was not explicitly set, so 'default-src' is used as a fallback.
> ```
> 
> @atton16 did you fix the `index.html`?

Ahh, I did not know that there is a CORS on mobile app.

atton16:
> also, if your goal is to just close the inapp browser when the survey is done, you can use the redirect URL trick that `survey.js` uses. Basically redirect to a known URL at the end of the survey and add a listener to `loadURL` that closes the browser when we see the final redirect URL

This sounds much better than our socket.io method. I cannot find the file though. I cannot find the usage of `loadURL` method in any of the file as well...

shankari:
> Ahh, I did not know that there is a CORS on mobile app.

How did it work on your testing?

> This sounds much better than our socket.io method. I cannot find the file though. I cannot find the usage of loadURL method in any of the file as well...

https://github.com/e-mission/e-mission-phone/search?q=survey.js&unscoped_q=survey.js

atton16:
> > also, if your goal is to just close the inapp browser when the survey is done, you can use the redirect URL trick that `survey.js` uses. Basically redirect to a known URL at the end of the survey and add a listener to `loadURL` that closes the browser when we see the final redirect URL
> 
> This sounds much better than our socket.io method. I cannot find the file though. I cannot find the usage of `loadURL` method in any of the file as well...

I think I found the snippet that you just describe in `loadstop` event here: https://github.com/e-mission/e-mission-phone/blob/5fe208d8623fcb1b933e167e83bed3cebf17134f/www/js/goals.js#L164

shankari:
that is for the game, I gave you a link to the actual survey file above. And while you can copy that snippet over to your code, you can also use the survey launch file, which is intended to be used as a library by other code.

atton16:
> that is for the game, I gave you a link to the actual survey file above. And while you can copy that snippet over to your code, you can also use the survey launch file, which is intended to be used as a library by other code.
> 
> I believe that is what the interscity branch uses
> (https://github.com/e-mission/e-mission-phone/blob/interscity/www/js/survey/launch.js)
> 
> If the behavior in
> https://e-mission.eecs.berkeley.edu/#/client_setup?new_client=interscity&clear_usercache=true&clear_local_storage=true
> is fine, then you can use the library and ensure that you don't have last minute regressions.

We need to use the `$scope` inside `intro.js`. So I decided not to use the `StartSurvey.launch` function. Otherwise, we gonna have to create the new method inside the `StartSurvey` service anyways.

PR #654, opened 2020-04-27
patch(User Profile): Use loadstart event instead of socketio for surv…

asiripanich:
…ey submit detection

Fixes the hiccups in #653.


asiripanich:
@shankari can you please have a look at this pr for us? Thanks

PR #655, opened 2020-04-28
Merge rciti2 to rciti1

asiripanich:


shankari:
Don't you also need a merge to `rciti4`?

asiripanich:
we are only using 3 channels for now. :)

PR #672, opened 2020-05-11
Add explanations for relaxing background restrictions on samsung and …

shankari:
…huawei

@asiripanich I don't have a way to test this because I don't have either a Huawei or a Samsung.

The Huawei fix is documented here https://stackoverflow.com/questions/31638986/protected-apps-setting-on-huawei-phones-and-how-to-handle-it and was also reported by @PatGendre 
The Samsung fix was reported by you.

With these changes, we can cover the two worst offenders in https://dontkillmyapp.com/ but it is not a comprehensive list.

Thoughts?

### DIFF
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -253,6 +253,10 @@
             "locationPermExplanation-ios-gte-13": "please open emTripLog in Settings and change the location permission to 'Always'. This allows us to protect you even when you are not actively using the app. After a few days, you will be asked whether you want to give the app background permission. Please continue to say 'Always Allow' whenever that comes up."
         }
     },
+    "allow_background": {
+        "samsung": "Settings → Power mode → High performance",
+        "huawei": "Settings → Battery manager → Protected apps → Enable protection for emTripLog"
+    },
     "consent":{
         "button-accept": "I accept",
         "button-decline": "I refuse"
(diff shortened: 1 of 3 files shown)


shankari:
Testing done: doesn't break anything. I am unable to test further right now since I don't have access to either Huawei or Samsung phones.

asiripanich:
Looks good to me. Please merge. ;)

PatGendre:
Hi FYI currently I don't have a Huawei phone anymore for testing. anyway your approach seems very sound.

shankari:
@asiripanich Please report the results of field-testing this approach in your study. Was it confusing? Was it clear?

I can then cherry-pick this change (potentially with some modifications) into master.

asiripanich:
I will need to ask the impacted user for feedback. 

By the way, are you sure that all samsung phones have the same battery management feature?

shankari:
I don't. I don't have any samsung phones. We could change the text from "Your phone restricts app operation in the background." -> "Some Samsung phones restrict app operation in the background."

I would also check with the other samsung user to confirm that everything is actually working. Maybe it's not working and they are too busy to report it.

PatGendre:
@shankari that's great that you could get the device manufacturer programmatically :-)   
For Huawei the text explanations is fine, AFAIK!   
I have no idea how many small phone vendors restrict background apps, though, but now we have a solid way of adding new ones to the list, thanks.

PR #679, opened 2020-05-13
Rciti1 - patch(tripEndSurvey): update destination and mode categories

asiripanich:
The last version was too simplified and missed many important categories such as change mode, delivery, bicycle, motorcycle, etc.


shankari:
Deployed, should be live by now. Good luck with the survey - hope I was correct to deploy given that the survey has already started!

PR #685, opened 2020-05-19
patch(TripEndSurvey): add 'trip_not_valid' option

asiripanich:
Users have reported that emTripLog detected trips that were not valid. For example, walking in your own house shouldn't be detected as a trip. I can confirm this behaviour as well and I do not live in a palace just a small apartment, with no room to run around.


shankari:
I am happy to merge this, but have some more questions because there is code to fix this in the post-processing pipeline, which apparently needs some tweaking.
The code that currently deals with this is the squishing code at:
https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/intake/cleaning/clean_and_resample.py#L927

As you can see from the implementation of `_is_squished_untracked`, 
https://github.com/e-mission/e-mission-server/blob/master/emission/analysis/intake/cleaning/clean_and_resample.py#L1172
in the clean and resample stage, we squish together any trips whose start and end are less than 100m apart.

So even if you see spurious green trips, they should be squished by the time they turn white. Can you confirm that this you can see **white** trips that are within your apartment as well?

If so, can you check the spurious trips to see why this is not happening? You can also extract the timeline for the user, run the pipeline and check the log files.

asiripanich:
Let me see how up-to-date my server code is.

But here are what I’m have been seeing.

image

image

asiripanich:
Btw, is there anyone working on the trip segmentation feature?

shankari:
I can't see the maps related to many of the invalid trips, so I don't know how valid (or otherwise) they are. But the 11:57pm trip should have been squished. The full length was only 100m, so the distance between the start and end should have been under 100m and it should have triggered the squished code.

I would suggest pulling that user's timeline for the day and checking the logs to see why it wasn't squished.

> I think we are using the gis branch but we do not see good results, also this is a common complaint we get from our users.

Do not see good results on which metric? Trip segmentation/section segmentation/mode inference? The GIS pipeline is generally the better of the two pipelines for mode inference.

Note also that the GIS branch uses GIS information from open street map (OSM) to distinguish between motorized modes. If OSM data is lacking in your area, then motorized mode detection will not work. Again, the logs should tell you which GIS queries were performed, their results and the rules that applied for the determination of the mode.

> is there anyone working on the trip segmentation feature?

Not that I am aware of. Trip segmentation is much more stable than section segmentation and mode inference so I am not aware of anybody prioritizing it.

asiripanich:
> I can't see the maps related to many of the invalid trips, so I don't know how valid (or otherwise) they are. But the 11:57pm trip should have been squished. The full length was only 106m, so the distance between the start and end should have been under 100m and it should have triggered the squished code.
> 
> 
> 
> I would suggest pulling that user's timeline for the day and checking the logs to see why it wasn't squished.
> 
> 
> 
> > I think we are using the gis branch but we do not see good results, also this is a common complaint we get from our users.
> 
> 
> 
> Do not see good results on which metric? Trip segmentation/section segmentation/mode inference? The GIS pipeline is generally the better of the two pipelines for mode inference.
> 
> 
> 
> Note also that the GIS branch uses GIS information from open street map (OSM) to distinguish between motorized modes. If OSM data is lacking in your area, then motorized mode detection will not work. Again, the logs should tell you which GIS queries were performed, their results and the rules that applied for the determination of the mode.
> 
> 
> 
> > is there anyone working on the trip segmentation feature?
> 
> 
> 
> Not that I am aware of. Trip segmentation is much more stable than section segmentation and mode inference so I am not aware of anybody prioritizing it.

Should you change that to as the crow flies distance?

shankari:
> Should you change that to the crow flies distance?

Hm, I thought it was the crow flies distance - e.g. 

```
    next_distance = ecc.calDistance(cleaned_untracked.data.start_loc.coordinates,
                       next_unsquished_trip.data.start_loc.coordinates)
    logging.debug("_is_squished_untracked: distance to next clean start (%s) = %s" %
                  (next_unsquished_trip.get_id(), next_distance))
```

but actually, that is only for squishing untracked time. Untracked time is if you have turned off tracking or have the phone turned off. If you didn't move a lot in that untracked time, we squish it.

For skipping spurious trips, I do look at the full trip length. 

```
    trip_distance = sum([section.data.distance for section in list(section_map.values())]) + \
                    sum([stop.data.distance for stop in list(stop_map.values())])
    filtered_trip_data.distance = trip_distance
    filtered_trip_entry["data"] = filtered_trip_data
    if filtered_trip_data.distance < 100:
        logging.info("Skipped single point trip %s (%s -> %s) of length %s" %
                     (trip.get_id(), trip.data.start_fmt_time,
                      trip.data.end_fmt_time, filtered_trip_data.distance))
        return None
```

You could tweak that check to meet your needs.

Concrete suggestions:
- use crow flies distance, but then what about home -> home tours?
- you can run a quick search over "trip not valid" entries and bump up the distance from 100m to something that will cover 90%, etc of the entries
- you can add in a time check as well. I could have sworn that I had a time check (~ 10 minutes) but it is not in the code.

Basically, you have to come up with a rule for spurious trips that is better than the one that is there now, driven by the data in your region.

> I have many trips where I made a quick stop (e.g. to grab coffee) but the starting and end points are at the same location. 

Not sure what you mean by this. Can you give a more concrete example?

> So eventually, if we want to make emission a perfect app for travel surveys, we should give the ability for users to correct their trips. Let see if RCITI can contribute on this feature.

That sounds like a great idea. If you are going to contribute at this level though, I would suggest joining the PMC. Would be great to coordinate on this with the other community members...

asiripanich:
> Not sure what you mean by this. Can you give a more concrete example?

See the first trip. Instead of breaking it into two trips, it detected as one, where the start and end points were both at my apartment. So the stop I made at the coffee shop was not recorded.

UNADJUSTEDNONRAW_thumb_143a

> That sounds like a great idea. If you are going to contribute at this level though, I would suggest joining the PMC. Would be great to coordinate on this with the other community members... It is not a lot of commitment - I send out an email maybe once a month or so.

Unfortunately that decision is not up to me, as you may ready know.

shankari:
> See the first trip. Instead of breaking it into two trips, it detected as one, where the start and end points were both at my apartment. So the stop I made at the coffee shop was not recorded.

Since the current trip segmentation is a dwell-time algorithm, you can tweak the dwell time to be shorter in your case. But then you may start getting breaks when you stop at red lights.

You are right that a truly correct solution will allow trip editing (both splitting and joining). There's a branch out there that has an initial version of that code, but it might be easier for you to just start from scratch.

asiripanich:
> That sounds like a great idea. If you are going to contribute at this level though, I would suggest joining the PMC. Would be great to coordinate on this with the other community members... It is not a lot of commitment - I send out an email maybe once a month or so.

dump question.. what is the PMC? :D I think I misunderstood that.

shankari:
PMC stands for Project Management Committee. It's a bunch of e-mission deployers that share updates and help steer the direction and the roadmap of the underlying platform. It is not very time consuming - I send out updates ~ 1x/month. I think you would be a great addition, and would welcome your joining it.

asiripanich:
I see. I would be glad to join if you think that my involvement will be good for the community.

shankari:
Woo! I think your involvement would be great. I will add you when I send out the next update.

asiripanich:
Looking forward to it!

PR #690, opened 2020-05-21
Rciti4

atton16:
- Display "Trip not valid" when user reported it.
- Fix label does not update after user answered the survey.
- Add EnketoSurvey.getAllLocalSurveyAnswers function to support the fix mentioned above.

Demo (GIF):
Untitled2

### DIFF
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -15,6 +15,7 @@ angular.module('emission.main.diary.list',['ui-leaflet',
                                       'emission.tripconfirm.services',
                                       'emission.services',
                                       'emission.survey.enketo.launch',
+                                      'emission.enketo-survey.service',
                                       'ng-walkthrough', 'nzTour', 'emission.plugin.kvstore',
     'emission.plugin.logger'
   ])
@@ -25,13 +26,19 @@ angular.module('emission.main.diary.list',['ui-leaflet',
                                     $ionicActionSheet,
                                     ionicDatePicker,
                                     leafletData, Timeline, CommonGraph, DiaryHelper,
-    Config, PostTripManualMarker, ConfirmHelper, nzTour, KVStore, Logger, UnifiedDataLoader, $ionicPopover, $ionicModal, EnketoSurveyLaunch, $translate) {
+    Config, PostTripManualMarker, ConfirmHelper, nzTour, KVStore, Logger, UnifiedDataLoader, $ionicPopover, $ionicModal, EnketoSurveyLaunch, EnketoSurvey, $translate) {
   console.log("controller DiaryListCtrl called");
     var MODE_CONFIRM_KEY = "manual/mode_confirm";
     var PURPOSE_CONFIRM_KEY = "manual/purpose_confirm";
 
     $scope.confirmSurvey = function(trip) {
-      EnketoSurveyLaunch.launch($scope, 'ConfirmSurvey', { trip: trip });
+      EnketoSurveyLaunch.launch($scope, 'ConfirmSurvey', { trip: trip }).then(function() {
+        return EnketoSurvey.getAllLocalSurveyAnswers("manual/confirm_survey", { populateLabels: true });
+      }).then(function(surveyAnswers) {
+        $scope.$apply(function() {
+          $scope.populateSurveyAnswerFromTimeline(trip, surveyAnswers);
+        });
+      });
     }
 
   // Add option
--- a/www/js/diary/list.js
+++ b/www/js/diary/list.js
@@ -33,10 +33,8 @@ angular.module('emission.main.diary.list',['ui-leaflet',
 
     $scope.confirmSurvey = function(trip) {
       EnketoSurveyLaunch.launch($scope, 'ConfirmSurvey', { trip: trip }).then(function() {
-        return EnketoSurvey.getAllLocalSurveyAnswers("manual/confirm_survey", { populateLabels: true });
-      }).then(function(surveyAnswers) {
         $scope.$apply(function() {
-          $scope.populateSurveyAnswerFromTimeline(trip, surveyAnswers);
+          $scope.updateCustomSurveyLabelForTrip(trip);
         });
       });
     }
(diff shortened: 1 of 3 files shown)


atton16:
@asiripanich

asiripanich:
@atton16 This one sparks joy.

www/js/survey/enketo-survey-service.js
```diff
@@ -174,6 +188,7 @@ angular.module('emission.enketo-survey.service', [
     displayForm: displayForm,
     validateForm: validateForm,
     getAllSurveyAnswers: getAllSurveyAnswers,
+    getAllLocalSurveyAnswers: getAllLocalSurveyAnswers,
```
shankari, line 191: Not sure why you need `getAllLocalSurveyAnswers`. `getAllSurveyAnswers` uses the UnifiedDataLoader, which loads data from both local and remote.

asiripanich: I guess he doesn't want the loading message to come up every time a trip end survey has been filled?

shankari: but why does he need to retrieve anything in the first place? when the survey is complete, the javascript gets back the survey response object, which it saves as a `manual/confirm_survey` object. Why can't you just use that object directly?

I thought that is what was happening before. Thanks for the image of the fix, but I don't understand the problem to begin with 😦

atton16: There was a report that the button label on the trip does not refresh after the survey is completed.
Normally we reference to the answer with `tripgj.userSurveyanswer`. But this time the rendering of the button is fairly complex so we introduce `tripgj.customSurveyLabel` (see: https://github.com/asiripanich/e-mission-phone/blob/6394a824201bd25e3abc6476e4bd8fee4060d50d/www/js/diary/list.js#L220-L258).

After we move to `tripgj.customSurveyLabel`, the button label stop to automatically refresh because we need to update it manually. So we have to find the right time to do that.

I recalled that I have implemented `EnketoSurvey.launch` to return a promise. So I decided to put the refresh code in there.

I audited the `EnketoSurvey.launch` and it looks like it does not return the answer. It simply return the modal state which is not usable for us.

We wanted to get the latest answer, we have `EnketoSurvey.getAllSurveyAnswers`. But this function makes a call to the server which is fairly expensive.

So I implement `EnketoSurvey.getAllLocalSurveyAnswers` to aid this concern.

www/js/survey/enketo-survey-service.js
```diff
@@ -103,22 +103,36 @@ angular.module('emission.enketo-survey.service', [
     return answer;
   }
 
+  function _processSurveyAnswers(answers, key, opt_populateLabels = false) {
```
shankari, line 106: why do you need this new function? It seems like it is just code restructuring...

www/js/survey/enketo-survey-service.js
shankari: > After we move to tripgj.customSurveyLabel, the button label stop to automatically refresh because we need to update the variable manually. So we have to find the right time to do that.

I don't buy this. Both `userSurveyAnswer` and `customSurveyLabel` are variables. `customSurveyLabel` is completely dependent on `userSurveyAnswer`. In the old `$scope.populateSurveyAnswerFromTimeline`, you set `tripgj.userSurveyAnswer`, now you are setting `tripgj.customSurveyLabel`. There is no reason why `tripgj.customSurveyLabel` should behave differently.

Although `EnketoSurvey.getAllLocalSurveyAnswers` is a local call, it is still a database call + refresh. Even with local data, too many calls can seriously affect performance as we saw in
https://github.com/e-mission/e-mission-phone/pull/553#issuecomment-482313867

shankari: In the javascript, `tripgj.userSurveyAnswer` is only accessed in `populateSurveyAnswerFromTimeline`. I really don't see how translating `tripgj.userSurveyAnswer` -> `tripgj.customSurveyLabel` should cause anything to break. 

At worst, you can just choose to not introduce `tripgj.customSurveyLabel` and get it to work that way.

atton16: I didn't buy this at first as well. But some how `tripgj.customSurveyLabel` break the UI update. That's why I implement this fix.

www/js/survey/enketo-survey-service.js
atton16: It is shared between `getAllSurveyAnswers` and `getAllLocalSurveyAnswers`.

www/js/survey/enketo-survey-service.js
asiripanich: @shankari if you see the earlier PR of this branch, the trip label didn't work properly as @atton16 already explained in this PR. If you don't agree with his approach, could you give a direction for us to fix this issue that doesn't a huge wait time like changing between dates?

asiripanich: Apparently, @atton16 can't figure out why `customSurveyLabel` doesn't work after we updated to the new survey so he implemented a fix, which he already tested.

shankari: > that doesn't a huge wait time like changing between dates?

I am trying to avoid even the call to the local database, so definitely not proposing a huge wait time

shankari: > I didn't buy this at first as well. But some how tripgj.customSurveyLabel break the UI update. That's why I implement this fix.

As a naive approach, can you go back to filling out `tripgj.userSurveyAnswer`? Yes, the logic to do so might be more complex, but it won't require a database call. If you want to keep the code clean, you can create a local variable (`userSurveyAnswerOld`) that you set at the beginning of the function and use to populate `tripgj.userSurveyAnswer`.

shankari: Alternatively, put breakpoints in the code, step through, and see how `tripgj.customSurveyLabel` is modified.

www/js/survey/enketo-survey-service.js
shankari: Yeah I figured that out. But it would have been helpful to state that upfront instead of making me try and understand what you were doing...

www/js/survey/enketo-survey-service.js
shankari: basically, I think you need to convert "X doesn't work" to "X doesn't work because of Y and there is no workaround for Y" before you decide to go to the database.

asiripanich:
Hi Shankari

Could you please accept this pr? I want to test this on my phone and if it doesnt unusable then we will fix it. But we are under a time crunch as you know.. we have an on going study that requires this feature. I know it is not ideal but we will come back to clean this up later.

asiripanich:
I guess it is fixed now.. 😬 @atton16 thanks

atton16:
Apparently Angular automatically detect string changes inside the object. When the answer gets updated, `EnketoSurvey` replace the reference to `tripgj.userSurveyAnswer` with the new answer. The older code works because angular perform change detection automatically.
This is fundamentally different when the UI is depend on the other object that get updated outside `EnketoSurvey`. In our case, we simply update the string with the newly replaced `trip.userSurveyAnswer` right after the survey is closed. We force UI refresh using `$scope.$apply`.

shankari:
if you make the changes, I'll stay up late tonight and get this merged!

asiripanich:
Thanks for the review.

@atton16 wont see this for at least another 5 hours..

atton16:
As `userSurveyAnswer` and `customSurveyLabel` are two distinct variables, we need to get these two synced up in two scenarios:
1. When we load all trips and survey answer is loaded.
2. When survey answer gets updated.

With your suggestion to transform the string in to object of string, we still need to cover the second scenario. The only thing we probably will get rid of it the need to call `$scope.$apply`.

With that said, I can try it out later today. But I don’t see the need to patch this up further.

www/js/diary/list.js
```diff
@@ -25,13 +26,17 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
+    Config, PostTripManualMarker, ConfirmHelper, nzTour, KVStore, Logger, UnifiedDataLoader, $ionicPopover, $ionicModal, EnketoSurveyLaunch, EnketoSurvey, $translate) {
   console.log("controller DiaryListCtrl called");
     var MODE_CONFIRM_KEY = "manual/mode_confirm";
     var PURPOSE_CONFIRM_KEY = "manual/purpose_confirm";
 
     $scope.confirmSurvey = function(trip) {
-      EnketoSurveyLaunch.launch($scope, 'ConfirmSurvey', { trip: trip });
+      EnketoSurveyLaunch.launch($scope, 'ConfirmSurvey', { trip: trip }).then(function() {
+        $scope.$apply(function() {
+          $scope.updateCustomSurveyLabelForTrip(trip);
```
shankari, line 37: The code to update the survey label as part of the survey close is added here, as part of this PR. Before this, the only time the displayed label was updated was as part of `populateSurveyAnswerFromTimeline`

atton16:
It is because this line here:

https://github.com/e-mission/e-mission-phone/blob/152c86a5bdd569a8766483cb2fbd92d1edaab76d/www/js/survey/enketo-survey.js#L151

shankari:
Ah that is much clearer. Adding this description to the comment would have made the review process smoother. Also, as I said earlier, are you going to remove `EnketoSurvey.getAllLocalSurveyAnswers`, and by extension, `_processSurveyAnswers`?

shankari:
once you do that, I am happy to merge.

shankari:
I am taking @asiripanich at his word that this will be cleaned up later :)

As part of this later cleanup, please set both `userSurveyAnswer` and `customSurveyLabel` in the same place (either in `validateForm` or in `confirmSurvey`).

Having them set in different places is confusing and is likely to lead to errors in the future where code maintainers modify one place, don't realize that they need to modify the other and run into errors.

I am not going to insist on that change now in the middle of an ongoing study. But this should definitely be fixed post-study.

asiripanich:
Hi Shankari 

Which part of the codes do you want us to remove? If you could point that out for me I will clean that up so eveyone can enjoy their weekend. :)

shankari:
It's 11pm here, so would appreciate a quick patch so I can merge before I go to bed.

asiripanich:
I'm not sure if @atton16 is available right now..

shankari:
From https://github.com/e-mission/e-mission-phone/pull/690#issuecomment-633183064, you need to `EnketoSurvey.getAllLocalSurveyAnswers` and `_processSurveyAnswers` and go back to the inline implementation

asiripanich:
sorry.. where is the inline implementation?

asiripanich:
these right ?

Screen Shot 2020-05-24 at 4 08 59 pm

www/js/survey/enketo-survey-service.js
```diff
@@ -103,23 +103,19 @@ angular.module('emission.enketo-survey.service', [
...
       if (!_opts_populateLabels) return answers;
-
       const xmlParser = new $window.DOMParser();
       if (key === 'manual/confirm_survey') {
         return answers.map(function(answer){
           return populateLabels(answer, xmlParser);
         });
-      }
-
-      return answers;
```
shankari, line 121: @asiripanich you need to fix this too. Basically, there should be no changes in `enketo-survey-service.js`.

shankari:
why did you revert that change? now `EnketoSurvey.getAllLocalSurveyAnswers` and `_processSurveyAnswers` are back. you need to remove them correctly so there are no actual changes to `enketo-survey-service.js` visible in the diffs for this PR.
https://github.com/e-mission/e-mission-phone/pull/690/files

asiripanich:
Good?

shankari:
yes, this is fine. Please do not forget to clean up https://github.com/e-mission/e-mission-phone/pull/690#issuecomment-633184314 after the study is done.

shankari:
@atton16 did you test after reverting that file? Just checking...

atton16:
flawlessly.

shankari:

asiripanich:
Thanks Shankari for staying up late to help us with this :)

atton16:
> I am taking @asiripanich at his word that this will be cleaned up later :)
> 
> As part of this later cleanup, please set both `userSurveyAnswer` and `customSurveyLabel` in the same place (either in `validateForm` or in `confirmSurvey`).
> 
> Having them set in different places is confusing and is likely to lead to errors in the future where code maintainers modify one place, don't realize that they need to modify the other and run into errors.
> 
> I am not going to insist on that change now in the middle of an ongoing study. But this should definitely be fixed post-study.

Just wanted to make sure that you want me to put all label logic into the `service`?

shankari:
hm, now that I think about this further, it might actually be more modular to separate the core storage (into `userAnswers`) from the postprocessing for the specific screen (into `confirmSurvey`).

But in that case, there needs to be better usage documentation on how to add the post-processing callbacks.

Right now, there is no indication of what a future user needs to do wrt using the saved data from the survey.

I would be fine with clear documentation on the survey, including code snippet examples - e.g. add a `how_to_XXX` document to https://github.com/e-mission/e-mission-docs/tree/master/docs/dev/front

PR #707, opened 2020-08-06
Upgrade to API 29

shankari:
Most of the changes were in the plugin, which we upgraded from version `1.4.4` to `1.4.5`.
The main change in here was to change the intro to specify the choice for android 10+ ("Always allow")

Other minor fixes:
- Fix the release script
  - Bump up the minSDK number
  - Fix the location of the production keystore
  - Fix the version of the buildtools used in the build
  - Change apk file names a bit since we build only one apk right now

### DIFF
--- a/bin/sign_and_align_keys.sh
+++ b/bin/sign_and_align_keys.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+PROJECT=$1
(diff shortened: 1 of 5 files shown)


shankari:
This fixes https://github.com/e-mission/e-mission-docs/issues/557

PR #714, opened 2020-08-24
survey code from rciti1 branch merged with leading master

robfitzgerald:
this PR attempts to take the survey from the rciti1 branch and merge it with the current state of the master branch. as a result, we should have an example of how to wire in a simple HTTP survey page, a (desirably) limited set of navigation tabs, as well as have the current features of the e mission phone app, including the "trip purpose" and "trip mode" button selections on the diary page.


www/js/tripconfirm/post-trip-prompt.js
```diff
@@ -220,6 +220,23 @@ angular.module('emission.tripconfirm.posttrip.prompt', ['emission.plugin.logger'
...
+        //         displayCompletedTrip(notification, state, data);
+        //       } else {
+        //         Logger.log("Skipped confirmation reporting");
+        //       }
+        //     });
+        // } else {
+        //   Logger.log("About to go to prompt page");
+        //   displayCompletedTrip(notification, state, data);
+        // }
+=======
```
shankari, line 239: Unresolved merge conflict.

www/manual_lib/enketo/css/plain-print.css
```diff
@@ -0,0 +1,332 @@
+/** CORE **/
```
shankari, line 1: We are not using the enketo embedded forms at this point, so none of the enketo `www/manual_lib` stuff is required.

www/i18n/en.json
```diff
@@ -214,7 +214,7 @@
         "tour-sectionPct-content": "% of time spent in each mode for this trip"
     },
 
-    "list-explainDraft-alert": "This trip has not yet been analysed. If it stays in this state, please ask your sysadmin to check what is wrong.",
+    "list-explainDraft-alert": "This trip has not yet been analysed. Once it is analysed you will be able to add your trip details. Normally, all draft trips get analysed at 10.00am and 8.30pm everyday.",
```
shankari, line 217: This is true of the UNSW survey, not necessarily of our system.

www/js/app.js
```diff
@@ -47,9 +47,6 @@ angular.module('emission', ['ionic',
     // Configure the connection settings
     Logger.log("about to get connection config");
     $http.get("json/connectionConfig.json").then(function(connectionConfig) {
-        if(connectionConfig.data.length == 0) {
```
shankari, line 50: This needs to be retained for emTripLog

www/templates/diary/list.html
```diff
@@ -89,16 +89,38 @@
...
+                    ng-click="confirmSurvey(tripgj)"
+                    >
+                    <b>{{tripgj.customSurveyLabel}}</b>
+                  </button>
+                  <button
+                    ng-if="!tripgj.customSurveyLabel"
+                    class="button btn-trip-confirm btn-trip-confirm-white"
+                    ng-click="confirmSurvey(tripgj)"
+                    >
+                    Add Trip Details
```
shankari, line 105: This looks like it still has the enketo survey logic - the label is `"Add Trip Details"` instead of `{{ .purpose }}`

shankari (changes requested):
@robfitzgerald I don't think you want to merge the entire `rciti` branch here. In particular, as we discussed, we want to combine the onboarding survey of rciti with the list component from master.

And honestly the code to add the onboarding survey is fairly minimal. So I would not merge from any of the rciti branches now unless you wanted to use any of the more advanced survey functionality.

shankari:
Closing this since it is more trouble than it is worth to fix it.
Let's just start with a new PR.

PR #717, opened 2020-09-01
Improve the sensor explanation to not be COVID-focused

shankari:


shankari:
Changes already cherry-picked

PR #719, opened 2020-09-22
Couple of small fixes to the login screen

shankari:
- Ensure that we don't accept null and empty logins
- Ensure that we always store the current channel so that every study doesn't
  have to modify it
- Ensure that if there is a login error, we don't proceed any further until the
  user resolves it

### DIFF
--- a/www/js/intro.js
+++ b/www/js/intro.js
@@ -1,6 +1,7 @@
 'use strict';
 
 angular.module('emission.intro', ['emission.splash.startprefs',
+                                  'emission.splash.updatecheck',
                                   'ionic-toast'])
 
 .config(function($stateProvider) {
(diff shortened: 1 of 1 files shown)


shankari:
Testing done:
- Generated null and empty logins; got an error


| null |  empty |
|--------|-------|
| Screenshot_1600801691 | Screenshot_1600801701 |

- Checked the profile on the server

```
 {'_id': ObjectId('5f6a47036b07dd3c2be9ab2d'),
  'user_id': UUID('d9c223c1-3e9c-424a-84ed-6e0caa67d04d'),
  'mpg_array': [32.044384997811726],
  'source': 'Shankari',
  'update_ts': datetime.datetime(2020, 9, 22, 12, 2, 17, 802000),
  'client': None},
```

```
{'_id': ObjectId('5f6a47036b07dd3c2be9ab2b'),
 'user_email': 'testme',
 'uuid': UUID('d9c223c1-3e9c-424a-84ed-6e0caa67d04d'),
 'update_ts': datetime.datetime(2020, 9, 22, 12, 2, 17, 800000)}
```

PR #734, opened 2021-02-05
Heavy lifting for supporting the option to upload the phone log

shankari:
Instead of emailing it.
This is a partial fix for https://github.com/e-mission/e-mission-docs/issues/609

Has only been tested on android, needs to be tested on iOS as well

Main changes:
- add a new upload service
- figure out how to read the phone logs into memory from the app ecosystem
- figure out how to format them as phone data
- include a popup to gather additional information
- combine info from popup along with phone logs and upload to a test site

Change the control screen to upload instead of emailing logs

There are other improvements we can make, but let's get this checked in first

### DIFF
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -60,6 +60,7 @@
         "check-ui-updates": "Check for UI updates",
         "download-json-dump": "Download json dump",
         "email-log": "Email log",
+        "upload-log": "Upload log",
         "user-data": "User data",
         "erase-data": "Erase data",
         "dev-zone": "Developer zone",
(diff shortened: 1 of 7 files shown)


shankari:
Testing done:
- deployed a server based on https://github.com/e-mission/em-bug-upload to production
- uploaded a file from this UI
- confirmed that the file is decodable

```
$ python bin/csv_export_add_date.py /tmp/2021-02-16_Test_with__America_Los_Angeles.loggerDB
exporting csv to /tmp/2021-02-16_Test_with__America_Los_Angeles.loggerDB.csv
adding dates to the dataframe
exporting csv with date to /tmp/2021-02-16_Test_with__America_Los_Angeles.loggerDB.withdate.log
```

shankari:
This is now a full fix for https://github.com/e-mission/e-mission-docs/issues/609
And potentially other use cases?!

I am going to merge this now. Let's see if people want to use it. We can remove it if not.

PR #740, opened 2021-02-18
Switch to es5-compatible version of bottleneck

shankari:
So it works on older versions of android and iOS
Not sure how necessary this backward-compat is, but the test phones are set up
to not upgrade, so I ran into this with them.
We should decide when to move beyond es5 to es2017 at least :)
es5 = 2009
es2017 = 2017

without this fix, we get a syntax error and white screen on older android
(force no update for webview) or iOS 12


shankari:
Testing done:
- old version of emulator had WSOD without fix
- old version of emulator showed the output with the fix

PR #742, opened 2021-02-20
Pull from master

shankari:


shankari:
Testing done:
- Launched app
- Clicked through screens
- Confirmed that replaced_mode showed up properly in both the diary screen and the labels screen

PR #745, opened 2021-02-27
Move the privacy policy link up

shankari:
So the text is not hidden on larger phones. See screenshots for a visual example.
The new location is close to the tracking on/off button, so it makes sense conceptually as well

### DIFF
--- a/www/js/control/general-settings.js
+++ b/www/js/control/general-settings.js
@@ -65,13 +65,19 @@ angular.module('emission.main.control',['emission.services',
     };
 
     $scope.viewPrivacyPolicy = function($event) {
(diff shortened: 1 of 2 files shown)


shankari:
| iPhone 12 Pro Max (only) |  Text is hidden and invisible | Works after moving to the top | Even on small screens |
|--------------|---------------|-------------|------------|
| Simulator Screen Shot - iPhone 12 Pro Max - 2021-02-27 at 09 12 01 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-02-27 at 09 12 23 | Simulator Screen Shot - iPhone 12 Pro Max - 2021-02-27 at 09 13 03 | Simulator Screen Shot - iPod touch (7th generation) - 2021-02-27 at 09 14 42 |

PR #747, opened 2021-03-10
Update the visualization code to match the new spec format

shankari:
The start_loc, end_loc and route_coords are now arrays instead of objects.
We need to find the correct valid object from the array and return it


shankari:
Related server PR
https://github.com/MobilityNet/mobilitynet-analysis-scripts/pull/52/commits/bb91f3cf49172409808b679f232bf09c7176a6e6

Spec was changed in MobilityNet/mobilitynet-analysis-scripts#48, fixing MobilityNet/mobilitynet.github.io#11

PR #750, opened 2021-04-04
Add support for using the staging environment for STAGE_* tokens

shankari:
- Store the full connection config in the app
- if the token starts with STAGE_*, set the connection config URL to stage.canbikeco.org
- change the connection settings and proceed only after the change is complete


shankari:
Testing done:
- logged in with `STAGE_*` tokens, confirmed that the host is stage.canbikeco.org

| android | iOS          |
|---------|----------|
| Screenshot_1617578845 | simulator_screenshot_B8A24BA5-7618-480C-A85E-38CB6BABD456 |

PR #753, opened 2021-04-05
Ensure that we set the values into the rootSpec even when we are usin…

shankari:
…g the saved value


shankari:
Without this, we cannot switch from a non-stage token to a stage token because the `$rootSpec.connectionConfig` variable is not set.

PR #754, opened 2021-04-15
Switch the trip confirm screen to also go to the diary directly

shankari:
Avoids race conditions and confusion; highlights the diary
We may want to merge this to master based on feedback from the CEO project


shankari:
This fixes https://github.com/e-mission/e-mission-docs/issues/575

Testing done:
- applied this change
- started a trip
- ended the trip
- the diary screen loaded

PR #755, opened 2021-04-26
App-only implementation of the bike library

shankari:
Next week, I will finish the server implementation and hook them up together


shankari:
Additional testing done:
Generated QR codes for NorthTower1, SouthTower1 and WestTower1 and scanned them with a physical phone. Everything works.

PR #762, opened 2021-06-28
Add UI features and algorithm to support label inference

GabrielKS:
+ Phone now receives and processes label inference data structure from the server to determine what inferences to show the user
+ Makes labels red if they have no value, yellow if they are inferred, and green if they are user input
+ Adds a "verify" button that turns yellow labels green

### DIFF
--- a/www/templates/diary/infinite_scroll_list.html
+++ b/www/templates/diary/infinite_scroll_list.html
@@ -77,9 +77,14 @@
                             {{trip.userInput[input.name].text}}
                         </button>
                     </div>
-                    <div  ng-if="!trip.userInput[input.name]" class="input-confirm-container">
-                        <button ng-click ="openPopover($event, trip, input.name)" class="button btn-input-confirm btn-input-confirm-white" translate>
-                        {{input.choosetext}}
+                    <div ng-if="!trip.userInput[input.name] && trip.finalInference[input.name]" class="input-confirm-container">
+                        <button ng-click ="openPopover($event, trip, input.name)" class="button btn-input-confirm btn-input-confirm-yellow" translate>
+                            {{trip.finalInference[input.name].text}}
+                        </button>
+                    </div>
+                    <div ng-if="!trip.userInput[input.name] && !trip.finalInference[input.name]" class="input-confirm-container">
+                        <button ng-click ="openPopover($event, trip, input.name)" class="button btn-input-confirm btn-input-confirm-red" translate>
+                            {{input.choosetext}}
                         </button>
                     </div>
                 </div>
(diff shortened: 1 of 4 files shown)


shankari:
@GabrielKS can you add some screenshots or animated gifs here to make it easier to review without pulling the changes and running them?

www/templates/diary/infinite_scroll_list.html
```diff
@@ -48,43 +48,52 @@ <h3 ng-if="!infScrollControl.reachedEnd">{{'diary.scroll-to-load-more' | transla
...
+                            <i class="icon ion-ios-location" style="font-size: 16px; left: 0; color: #33e0bb;"></i>
+                                    {{trip.start_display_name}}
+                            </div>
+                            <div ng-class="listLocationClass" id="no-border" href="#" style="background-color: transparent; font-size: 0.8em; padding-top: 5px; padding-bottom: 5px; padding-left: 30px; margin-top: 0; margin-bottom: 0;">
+                            <i class="icon ion-ios-location" style="font-size: 16px; left: 0; color: #ff5251;"></i>
+                                    {{trip.end_display_name}}
+                            </div>
+                        </div>
+                        <div ng-click="verifyTrip($event, trip)" class="col-10 diary-checkmark-container center-vert center-horiz">
+                            <i ng-class="trip.verifiability" class="ion-checkmark-round"></i>
```
shankari, line 69: FYI, please continue using properties instead of functions here. Using functions to retrieve the `ng-class`, although it works, plays havoc with the caching and the performance really suffers.

shankari (approved):
Looks pretty straightforward; I will merge after you upload some screenshots/animated gifs

GabrielKS:
Below is a GIF that showcases all of the changes implemented. The inference data is made up for instructional purposes.
Showcase GIF

Here's what is happening for each trip:
* Trip 1: No inference possible. Same behavior as before these changes, except blank labels are red instead of gray.
* Trip 2: Both labels can be inferred. Verify button marks labels as user-confirmed, turning them green.
* Trip 3: Only mode can be inferred; verify button confirms it.
* Trip 4: Displayed inferences are incorrect. User inputs mode, at which point the system can correctly infer purpose.
* Trip 5: Only mode can be inferred with enough confidence to display it; once user confirms that, confidence in the purpose inference increases to the point that it can be displayed.
* Trip 6: Only mode can be inferred with enough confidence to display it, but it is incorrect; user enters purpose, at which point the system can correctly infer mode.

shankari:
Please submit the PR to fix the confirmation check for Trip 1 after I merge this.

PR #763, opened 2021-07-02
Fix verifiability bug

GabrielKS:
Previously, trip.verifiability would not get updated when certain label
changes were made, leading to an improperly styled checkmark.


shankari (approved):
I am not sure I fully understand this, but the change seems simple enough.

PR #765, opened 2021-07-16
Add "To Label" screen

GabrielKS:
This PR implements the `To Label` screen as described in the [UI proposal](https://github.com/e-mission/e-mission-docs/files/6743130/UI.Draft.2.pdf). Though the proposal is bit out of date and could use a new draft, it describes the purpose of the `To Label` screen well. Currently I have only implemented the `"all"` and `"none"` expectation types on the server side, but given the information the phone receives from the server it constructs `To Label` exactly as proposed.

Smaller changes:
 * I changed the styling of the filter buttons at the top of the `Label` pane so they would look more like the tabs we have implemented elsewhere. I'd like to change their positioning as well (maybe center them and right-align the refresh button), but that's less of a priority.
 * I made the only filter options `To Label` and `All` by default. The others are still in the codebase if we want to add them back.

Demo GIF:
Showcase GIF
This is a new fake dataset that has all the trips marked as expected to be labeled except the one that takes place from 1:15 PM to 1:22 PM. The video shows:
 1. The presence of the 1:15 PM to 1:22 PM trip on the `All` tab and its absence on the `To Label` tab
 2. How confirming a trip removes it from the `To Label` tab but not the `All` tab (note: in the demo, I confirm all trips with the `verify` button, but trips can just as easily be confirmed by manually selecting labels)
 3. How a user might confirm all trips in `To Label` (assuming all inferred labels happen to be correct); now `To Label` is empty and the only trip in `All` with yellow labels is the trip the user was not expected to label


shankari:
@GabrielKS I wonder if it would be worthwhile to have three buttons at the top:
- To Label
- All Unlabeled
- All

If we are in a relaxed mode, but somebody still wants to verify all their trips, we should make it easy for them, no?
I am happy to defer this until after we get initial feedback, but just wanted to document the idea.

I also think the buttons (specially "All") should be bigger. While it is easy enough to click on them in the emulator, I'm concerned that the button may be too small for fat fingers on a real phone

www/js/diary/infinite_scroll_filters.js
```diff
@@ -17,7 +17,7 @@ angular.module('emission.main.diary.infscrollfilters',[
     var unlabeledCheck = function(t) {
        return ConfirmHelper.INPUTS
            .map((inputType, index) => !angular.isDefined(t.userInput[inputType]))
-           .reduce((acc, val) => acc && val, true);
+           .reduce((acc, val) => acc || val, false);
```
shankari, line 20: I assume this is to ensure that users have all labels filled out before we consider the trip as labeled. I think I had it the other way around because I didn't anticipate that users would leave some labels blank. But we did see this case in Ft. Collins, so this is a good fix.

www/js/diary/infinite_scroll_list.js
```diff
@@ -294,6 +302,12 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
         $scope.editingTrip = angular.undefined;
     }
 
+    $scope.updateTripProperties = function(trip) {
+      $scope.inferFinalLabels(trip);
+      $scope.updateVerifiability(trip);
+      $scope.recomputeDisplayTrips();
```
shankari, line 308: This is the part that removes the trips from the list once they are labeled correctly, right? I think this should be fine since it doesn't make any remote calls, but we should keep an eye on the performance.

shankari (changes requested):
Again, this looks fine code-wise; I have some minor comments about formatting and requests for additional documentation. If @corinne-hcr is ready with her changes over the weekend (currently unsure whether this will happen), I will merge anyway so we can start end-to-end testing, and you can address the changes as part of the notification PR.

shankari:
@GabrielKS High level question for testing: when we merge this to the CanBikeCO branch, would it be possible to enable this functionality only for "stage" users, similar to (similar to https://github.com/e-mission/e-mission-phone/blob/ceo_ebike_project/www/templates/main.html#L27). We want the equivalent of preprocessor directives :)

If the code is too distributed, I can try to push this to a new ionic channel, but then I need to get all the beta testers to switch to the new channel. And I will need to push all future changes to the staging channel before production. I guess that is actually a Good Thing, but I'm not sure that the beta testers want to test minor changes to the privacy policy, etc.

Anyway, let's first see if it is trivial to put in the equivalent of pre-processor directives to enable the new code only on the staging environment. If not, let's consider the option of using a channel.

www/js/diary/infinite_scroll_list.js
```diff
@@ -45,11 +45,13 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
   // reset all filters
   $scope.filterInputs = [
-    InfScrollFilters.UNLABELED
+    InfScrollFilters.TO_LABEL
   ];
   $scope.filterInputs.forEach((f) => {
     f.state = false;
   });
+  $scope.filterInputs[0].state = true;
+  $scope.allTrips = false;
```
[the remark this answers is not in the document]
GabrielKS: Line 53 sets the first filter tab to be active by default; line 48 specifies that that's `To Label`. Line 54 declares a variable keeping track of whether the `All` tab is active so we can highlight it or not just like all the others. I need a separate variable because `All` is not one of the `filterInputs`. I could use a function that checks whether all of the `state`s in `filterInputs` are `false`, but you advised that I should use properties instead of functions for CSS classes. I can add documentation for this.

www/templates/diary/infinite_scroll_list.html
```diff
@@ -3,16 +3,19 @@
...
-          <button class="button button-icon" class="col-5 button" ng-click="refresh()">
-            <i class="icon ion-refresh"></i>
-          </button>
-       </div>
+            <button ng-repeat-start="selF in filterInputs" ng-if="$first" ng-click="select(selF)" class="{{selF.width}} button labelfilter first" ng-class="{on:selF.state}" style="text-align: center;font-size: 14px;font-weight: 600;" translate>
+                {{selF.text}}
+                </button>  
+                <button ng-repeat-end ng-if="!$first" ng-click="select(selF)" class="{{selF.width}} button labelfilter middle" ng-class="{on:selF.state}" style="text-align: center;font-size: 14px;font-weight: 600;" translate>
+                {{selF.text}}
+                </button>
```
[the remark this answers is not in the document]
GabrielKS: This part is actually unrelated to `infinite_scroll_list.js:53` — it's just to get the rounded left corners on the first element. I am realizing I can probably do this more elegantly with the CSS `:first` selector; I'll look into that.

www/js/diary/infinite_scroll_filters.js
```diff
@@ -29,16 +29,27 @@ angular.module('emission.main.diary.infscrollfilters',[
...
 
+    var toLabelCheck = function(trip) {
+        console.log(trip.expectation.to_label)
+        return trip.expectation.to_label && unlabeledCheck(trip);
+    }
+
     sf.UNLABELED = {
         text: $translate.instant(".unlabeled"),
-        width: "col-80",
+        width: "col-30",
```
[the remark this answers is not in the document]
GabrielKS: Okay. I'd like to change the positioning and width of the tabs more drastically: ideally, the refresh button would be aligned right, the tabs would be centered, each tab's width would be the same as the others (we shouldn't be setting widths based on the length of the word as we seemed to do before this PR unless we want to include widths in a localization file), and the widths would automatically be as large as possible without the tabs overlapping the refresh button. I tried for some time to make progress on this but wasn't really getting anywhere, so I left it for later. For now I can hardcode a larger value again.

www/js/diary/infinite_scroll_filters.js
GabrielKS: Yep!

www/js/diary/infinite_scroll_list.js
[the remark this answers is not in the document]
shankari: yes, please. I figured out most of that from the code, but making life easier for others helps with maintainability. Even just a note saying that this is related to highlighting the all tab (turning it green) would be helpful because it will help others know what to expect in the code and find the parts that match their expectations.

But again, you can defer this to a later commit.

www/js/diary/infinite_scroll_list.js
GabrielKS: `inferFinalLabels` recomputes what inferences we should be showing based on what the user just entered. `updateVerifiability` recomputes whether the verify checkbox should be clickable. `recomputeDisplayTrips` removes trips from `To Label` (or `Unlabeled`, etc.) once they are labeled. There is some inefficiency here as `recomputeDisplayTrips` goes through all trips, whereas we only need to check the trip that was just changed; that would be the optimization if it is necessary.

GabrielKS:
> @GabrielKS I wonder if it would be worthwhile to have three buttons at the top:
> 
> * To Label
> * All Unlabeled
> * All
> 
> If we are in a relaxed mode, but somebody still wants to verify all their trips, we should make it easy for them, no?
> I am happy to defer this until after we get initial feedback, but just wanted to document the idea.
> 
> I also think the buttons (specially "All") should be bigger. While it is easy enough to click on them in the emulator, I'm concerned that the button may be too small for fat fingers on a real phone

Sure, that makes sense, I can add the `Unlabeled` tab back. I discuss button size and positioning above.

GabrielKS:
> @GabrielKS High level question for testing: when we merge this to the CanBikeCO branch, would it be possible to enable this functionality only for "stage" users, similar to (similar to https://github.com/e-mission/e-mission-phone/blob/ceo_ebike_project/www/templates/main.html#L27). We want the equivalent of preprocessor directives :)
> 
> If the code is too distributed, I can try to push this to a new ionic channel, but then I need to get all the beta testers to switch to the new channel. And I will need to push all future changes to the staging channel before production. I guess that is actually a Good Thing, but I'm not sure that the beta testers want to test minor changes to the privacy policy, etc.
> 
> Anyway, let's first see if it is trivial to put in the equivalent of a simple pre-processor directive (a set of if statements) to enable the new code only on the staging environment. Note that we don't need to exclude the new functions, just avoid invoking them. If not, let's consider the option of using a channel.
> 
> Note that this applies not just to this PR but also to the previous PR with the colored buttons. The non-staging UI should be unchanged; only the staging UI should change.

I will investigate this today.

www/js/diary/infinite_scroll_list.js
shankari: Right, modifying an individual trip is fine, I was just worried about `recomputeDisplayTrips`. But we don't need to optimize that just yet.

shankari:
Once you change the button sizes and positioning, I will merge this to master. Can you then pull these changes to the  CanBikeCO branch, explore the `if(stage)` for a few hours and let me know if it is feasible?

If it does work, you could submit a PR to the CanBikeCO that includes these changes (from the pull) and the `if(stage)` changes.

GabrielKS:
New Label screen header, with `Unlabeled` added back in and the buttons resized and repositioned:
New Label screen header

shankari:
Build failure is due to 

```
Build-tool 31.0.0 is missing DX at /Users/runner/Library/Android/sdk/build-tools/31.0.0/dx
```

which seems to be related to the version of build tools used

https://stackoverflow.com/questions/68269060/ionic-failure-build-failed-with-an-exception 

Going ahead and merging this PR since the failure is not related to these changes.

shankari:
Filed https://github.com/actions/virtual-environments/issues/3757 to track the issue, since the error is

```
> Installed Build Tools revision 31.0.0 is corrupted. Remove and install again using the SDK Manager.
```

and the virtual environment changed from `20210712.4` to `20210718.2`

PR #766, opened 2021-07-20
Incorporate new Label UI conditionally

GabrielKS:
This is a draft — don't merge yet!

This PR incorporates the new Label screen UI (red/yellow/green labels, inferences, verify button, To Label view, redesigned header, etc.) — but only for `"stage"` users (and dependent on a manual override). However, this conditional functionality is still a little buggy. In particular, the header (new or old) does not load properly unless one accesses the Label screen, accesses some other screen, and goes back to the Label screen. I plan to investigate this soon. It might be easier for someone with more Angular, Ionic, etc. experience than me to see how best to fix this — I'm sure it's just an issue of the header rendering before the data has been properly populated.


GabrielKS:
Also, I have found that the URL we need to check is often `undefined` if one uses the hot reload functionality, whereas it generally has a proper value if the app is restarted after saving changes. I haven't had time to look into why this is the case in any detail.

shankari (commented):
I was surprised to see a bunch of changes that updated plugins in this PR. Then I found that I had unpushed changes - I had pulled from master at

```
commit b08b4ea37a344b64196bb8f820647e4b6055ff03 (upstream/master)
Merge: 1811495f 755ad143
Author: shankari <shankari@eecs.berkeley.edu>
Date:   Sat Jun 5 17:01:39 2021 -0700
```

but have not pushed that (and subsequent changes) to `ceo_ebike_project`.

If I push the changes now, will you be able to deal with the merge conflict?

shankari:
Closing this because it is superceded by https://github.com/e-mission/e-mission-phone/pull/767

PR #768, opened 2021-08-03
Update Cordova-Android to 10.0.1 to prevent build issues with Android build-tools 31

WilliamChan4:
Update Cordova-Android to 10.0.1 to prevent build issues with Android build-tools 31

Replace phonegap-plugin-push with cordova-plugin-push to remove dependency on cordova-support-google-services causing build error with GradlePluginGoogleServicesEnabled. 

https://github.com/phonegap/phonegap-plugin-push/pull/2873
https://github.com/havesource/cordova-plugin-push/pull/8/files

### DIFF
--- a/package.cordovabuild.json
+++ b/package.cordovabuild.json
@@ -8,19 +8,18 @@
     "url": "git+https://github.com/e-mission/e-mission-phone.git"
   },
   "devDependencies": {
-    "cordova": "10.0.0",
+    "@havesource/cordova-plugin-push": "^2.0.0",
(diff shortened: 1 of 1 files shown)


shankari:
@WilliamChan4 alas, this looks like it didn't actually work.
I'm going to start with downgrading the tools on GHA so we can get this to work, and then we need more testing with the new cordova version before we upgrade.

shankari:
Closing this since I am not going to work on this right now.
Workaround committed in #779

PR #769, opened 2021-08-04
Reverse Label screen scrolling direction

GabrielKS:
The Label screen now loads with the most recent trip at the bottom of the screen and loads older trips as the user scrolls upwards. The start and end time labels have also been reversed to correspond to this layout, place names load from the most recent trip back in time, and a border was added to the trip cards per a user request.

Demo GIF:
output
Note that not all the place names have loaded. The GIF was recorded at this stage on purpose to show how they load bottom to top. This GIF was made using a real human's data with their informed consent.

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -82,7 +82,8 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
             $scope.inferFinalLabels(trip);
             $scope.updateVerifiability(trip);
         });
-        ctList.forEach(function(trip, index) {
+        // Fill places on a reversed copy of the list so we fill from the bottom up
+        ctList.slice().reverse().forEach(function(trip, index) {
             fillPlacesForTripAsync(trip);
         });
         $scope.data.allTrips = ctList.concat($scope.data.allTrips);
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -55,9 +55,11 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
   $scope.allTrips = false;
   const ONE_WEEK = 7 * 24 * 60 * 60; // seconds
 
-  $scope.infScrollControl = {};
+  $scope.infScrollControl = {threshold: 75, scrollLock: true, fromBottom: -1, callback: undefined};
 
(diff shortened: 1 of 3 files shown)


shankari:
Before I merge this, can you confirm that it doesn't work to just putting `<ion-infinite-scroll` to the top of the list instead of the bottom (e.g. right below this line 
https://github.com/e-mission/e-mission-phone/pull/769/files#diff-56fd9d875d4df695032e9c2a8fb9c9e2f61abb295b8c49984643e2a53ad575a1L34) doesn't work?

I am sure this works, but from a maintainability perspective it might be better to use the standard ionic component rather than a home-grown solution. I don't want you to spend too much time on this (ionic v1 is essentially EOL, so its maintainability isn't too great either), but I just wanted to do my due diligence.

GabrielKS:
Oh, I tried that — among many other things. I have convinced myself that there is not a way to do this using the standard Ionic component; I would love to be proven wrong.

For the record: [here](https://ionicframework.com/docs/api/infinite-scroll) is the reference that describes how our desired behavior can be achieved with the `ion-infinite-scroll` element's `position` property in recent Ionic versions, and [here](https://ionicframework.com/docs/v1/api/directive/ionInfiniteScroll/) is the corresponding reference for v1, which doesn't mention a `position` property and only discusses triggering at the bottom of the page. Not that I didn't try `position` anyway….

www/js/diary/infinite_scroll_list.js
```diff
@@ -137,6 +141,31 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
+    if ($scope.infScrollControl.reachedEnd) {
+      console.log("reached end");
+      return;
+    }
+    if ($ionicScrollDelegate.getScrollPosition().top < $scope.infScrollControl.threshold) {
+      $scope.infScrollControl.callback = function() {
+        // This whole "infinite scroll upwards" implementation is quite hacky, but after hours of work on it, it's the only way I could approximate the desired behavior.
+        $ionicScrollDelegate.scrollBottom();
+        const clientHeight = $ionicScrollDelegate.getScrollView().__clientHeight;
+        $ionicScrollDelegate.scrollBy(0, -$scope.infScrollControl.fromBottom+clientHeight);
```
shankari, line 163: I can't believe you actually implemented this! I would have given up after the nth hour of fiddly work.

shankari (approved):

PR #773, opened 2021-08-07
wrap native calls in intro.js in a ready() statement

shankari:
Well, first wrap them into a function, and then call the function from ready()
Without this, live-reload on the intro wouldn't work
With this fix, live-reload on the intro works (see video in PR)


shankari:
Live reload works, including hitting breakpoints in the code.

live-reload demo small mov

PR #756, opened 2021-05-14
Use random token authentication

asiripanich:
@shankari hopefully this works. :)

PR #774, opened 2021-08-09
patch(UserId): display userid without prompt auth secret

WilliamChan4:
Fix to ensure the "secret" isn't displayed on the profile page concatenated with the user id token. 

Originally by atton16.

### DIFF
--- a/www/js/control/general-settings.js
+++ b/www/js/control/general-settings.js
@@ -7,6 +7,7 @@ angular.module('emission.main.control',['emission.services',
                                         'emission.main.control.tnotify',
                                         'ionic-datepicker',
                                         'ionic-datepicker.provider',
+                                        'emission.splash.secretcheck',
                                         'emission.splash.startprefs',
                                         'emission.splash.updatecheck',
                                         'emission.main.metrics.factory',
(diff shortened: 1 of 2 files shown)


WilliamChan4:
@shankari A pull request for Amarin's study on Thursday.

asiripanich:
@shankari could you please help us review this and the other pending PR. 🙏

shankari:
I don't know why you need the secret part of the code - that was a requirement from NREL legal because they did not want users outside NREL to use the app. But since you don't have that restriction, you only need the part where we autogenerate the token.

shankari:
Just to clarify, when you first merged the random token authentication, I suggested that you only merge the random token part, and you added the secret check because I think you pulled the most recent version from NREL branch and not the commit related to this
https://github.com/e-mission/e-mission-phone/pull/756

So @WilliamChan4 probably better to remove the secret check as in https://github.com/e-mission/e-mission-phone/pull/756#issuecomment-841985351

and all the secret check dependencies because you don't need them.

WilliamChan4:
As per discussion, the secret check is no longer necessary for rCITI. Also, the user sign up survey is no longer required. The only change from master is the automatic generation of an "Email" (called a "Token") and the corresponding user interface for signing up.

PR #775, opened 2021-08-09
Fix tiles not loading at closer zoom levels on Android

WilliamChan4:
This pull request cherry picks some patches applied to Master to change 
* http://nominatim.openstreetmap.com to https://nominatim.openstreetmap.com (HTTPS) to prevent a CORS issue causing 404 Not Founds. 
* The street map tile provider to Open Street Map from Stamin since the latter has missing tiles at closer zoom levels leading to blank maps at some zoom levels (not applicable for all regions but visible on the outskirts of Sydney)

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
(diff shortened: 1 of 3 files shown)


WilliamChan4:
@shankari A pull request for Amarin's study on Thursday.

shankari:
There is already a commit to switch back
https://github.com/e-mission/e-mission-phone/commit/46cf4efb5ab52a5eaefa43d786e31122cde9db09

I would suggest cherry-picking that change instead of committing a new change so that you don't have merge conflicts while pulling from master in the future.

shankari:
Going to work off master instead.

PR #777, opened 2021-08-10
Fix a dead link.

asiripanich:


shankari (approved):
Thanks for making this change!

PR #782, opened 2021-08-13
Update intro screens

asiripanich:
Closes https://github.com/rciti/e-mission-phone/issues/41.

Please see the commit messages below.

### DIFF
--- a/www/css/intro.css
+++ b/www/css/intro.css
@@ -31,11 +31,19 @@
   padding-left: 8%;
   font-size: 18px;
 }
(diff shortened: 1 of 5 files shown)


shankari (approved):
LGTM

shankari:
You theoretically don't need to change the login screen; you can use the prompted auth on the phone. But it works both ways, and you have tested this way, so let's go with it.

shankari:
should be live in ~ 10 mins

PR #785, opened 2021-08-16
Switch the infinite scroll from auto-load on scroll to a button

shankari:
The auto-load on scroll feature, so common to messaging and chat apps is actually fairly complicated to implement correctly in the case in which we display a filtered list. In particular:
- the additional scrolling required to maintain "place" in the list triggers multiple "scroll" calls even when the user is not doing anything, which leads to multiple calls to the server.
- if new entries are not added to the list after the server call, then the condition that triggered the server call (top < threshold) persists, leading to multiple calls again.

See https://github.com/e-mission/e-mission-docs/issues/662

This is an inherent limitation of using auto-load on scroll for a filtered list and also happens with the standard ionic component, https://github.com/e-mission/e-mission-docs/issues/662#issuecomment-899312254

So instead of butting my head against the wall getting auto-load to work, I switched to manual load, similar to the diary screen. https://github.com/e-mission/e-mission-phone/pull/769

As a bonus, this also helps with https://github.com/e-mission/e-mission-docs/issues/659
It indicates how many of the trips are displayed in **this filter**, should help to clarify questions around "missing trips" and allow users to explore other tabs as necessary.

Testing done:
    - loaded data for a user with a single trip
    - labeled that trip so the user had no trips
    - loaded data for a user with multiple trips

In all cases, used the button to scroll all the way to the oldest trip.
Did not encounter any issues.

+ bonus fix to scroll to the bottom of the screen when switching filters, which is one of the fit-and-finish issues in https://github.com/e-mission/e-mission-docs/issues/662

### DIFF
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -160,6 +160,7 @@
         "no-trips-found": "No trips found",
         "for-current-filter": "for current filter. Show All to remove filters",
         "scroll-to-load-more": "Scroll to load more",
+        "filter-display-status": "Displaying {{displayTripsLength}} of {{allTripsLength}} trips upto {{currentEnd | date}}",
         "choose-mode": "Mode 📝 ",
         "choose-replaced-mode": "Replaces 📝",
         "choose-purpose": "Purpose 📝",
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -160,7 +160,7 @@
         "no-trips-found": "No trips found",
         "for-current-filter": "for current filter. Show All to remove filters",
         "scroll-to-load-more": "Scroll to load more",
-        "filter-display-status": "Displaying {{displayTripsLength}} of {{allTripsLength}} trips upto {{currentEnd | date}}",
+        "filter-display-status": "Displaying {{displayTripsLength}} of {{allTripsLength}} trips up to {{currentEnd | date}}",
         "choose-mode": "Mode 📝 ",
         "choose-replaced-mode": "Replaces 📝",
         "choose-purpose": "Purpose 📝",
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -210,6 +210,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
     });
     $scope.allTrips = false;
     $scope.recomputeDisplayTrips();
+    $ionicScrollDelegate.scrollBottom();
     ClientStats.addReading(ClientStats.getStatKeys().LABEL_TAB_SWITCH, {"source": prev, "dest": $scope.getActiveFilters()});
   }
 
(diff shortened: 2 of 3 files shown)


shankari:
@GabrielKS I am done with the slow loading screen changes, would you be able to review them? Regardless of your review, I plan to deploy in ~30 mins so the changes are in place before the decision meeting.

www/i18n/en.json
```diff
@@ -160,6 +160,7 @@
         "no-trips-found": "No trips found",
         "for-current-filter": "for current filter. Show All to remove filters",
         "scroll-to-load-more": "Scroll to load more",
+        "filter-display-status": "Displaying {{displayTripsLength}} of {{allTripsLength}} trips upto {{currentEnd | date}}",
```
GabrielKS, line 163: "up to"

shankari:
new_load_screen mp4

www/js/diary/infinite_scroll_list.js
```diff
@@ -200,6 +210,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
     });
     $scope.allTrips = false;
     $scope.recomputeDisplayTrips();
+    $ionicScrollDelegate.scrollBottom();
```
GabrielKS, line 213: Ah, so we scroll to the bottom every time we switch tabs. Maybe a comment explaining why?

GabrielKS (changes requested):
It's kind of hard to see in the GIF — when we are able to load more trips, it looks like the load button is gray with a white square. Is this intuitive? Maybe the standard download icon (downward arrow pointing at horizontal line) would work better. Also, what does this button look like when we've loaded all trips? Does it look grayed out?

shankari:
I apologize for the quality; my initial attempt was too large to upload.
These screenshots may show it better.

|  Gray download icon | Turning to check mark when all trips have been retrieved |
|-------------------|-----------------|
| Screenshot_1629138539 | Screenshot_1629138555 |

shankari:
@GabrielKS added some screenshots to answer your questions + fixed minor comments

GabrielKS:
Ah, okay. Download icon looks good. Maybe green on gray would make it clearer that it's clickable? Other than that, looks good.

shankari:
Confirmed that this works on iOS as well. The spacing is broken on the iPod Touch, but other parts of the spacing are broken too, so I'm not going to fix it now. I also tried green on gray, but it looks a bit ugly IMHO, so will leave it off for now. We can change it at the same time as the start and end colors.

| iPhone 12 | iPod Touch | iPhone 12 with green on gray |
|------|------|-----|
| Simulator Screen Shot - iPhone 12 - 2021-08-16 at 11 56 55 | Simulator Screen Shot - iPod touch (7th generation) - 2021-08-16 at 11 58 53 | Simulator Screen Shot - iPhone 12 - 2021-08-16 at 12 00 05 |

PR #786, opened 2021-08-18
Display the start date **and end date** for the processed trips

shankari:
So users can figure out whether the trips are delayed or just in the other
filters.

Testing done:
- Loaded data, confirmed that the start and the end were displayed


shankari:
Not going to wait for review of this trivial change.

PR #787, opened 2021-08-18
Delay the deletion of entries from the To Label screen

shankari:
to give users a chance to correct the trip.

The implementation is fairly straightforward.
- as a trip is modified, set its `waiting for modification` flag to true
- ensure that trips that are waiting for modification are not filtered out
- create a timeout to set the flag to false and recompute
- cancel any existing timeouts to avoid recomputations

Testing done:
#### Set all three labels for a trip
- Saw the logs:

Open first popover, create new timeout promise, nothing to cancel

```
DEBUG:in openPopover, setting draftInput = {"start_ts":1628366684.0019567,"end_ts":1628367280.652}
DEBUG:in storeInput, after setting input.value = walk, draftInput = {"start_ts":1628366684.0019567,"end_ts":1628367280.652,"label":"walk"}
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: creating new timeout
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: cancelling existing timeout undefined
```

Open second popover, create new timeout promise, cancel previous

```
DEBUG:in openPopover, setting draftInput = {"start_ts":1628366684.0019567,"end_ts":1628367280.652}
DEBUG:in storeInput, after setting input.value = home, draftInput = {"start_ts":1628366684.0019567,"end_ts":1628367280.652,"label":"home"}
ionic.bundle.js:3063 [Violation] 'touchend' handler took 150ms
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: creating new timeout
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: cancelling existing timeout [object Object]
ionic.bundle.js:5301 [Intervention] Ignored attempt to cancel a touchstart event with cancelable=false, for example because scrolling is in progress and cannot be interrupted.
self.touchStart @ ionic.bundle.js:5301
```

Open third popover, create new timeout promise, cancel previous

```
DEBUG:in openPopover, setting draftInput = {"start_ts":1628366684.0019567,"end_ts":1628367280.652}
DEBUG:in storeInput, after setting input.value = bike, draftInput = {"start_ts":1628366684.0019567,"end_ts":1628367280.652,"label":"bike"}
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: creating new timeout
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: cancelling existing timeout [object Object]
```

After a significant time, recompute

```
DevTools failed to load SourceMap: Could not load content for http://10.0.2.2:3000/socket.io/socket.io.js.map: Connection error: net::ERR_CONNECTION_TIMED_OUT
DEBUG:trip starting at 2021-08-07T13:04:44.001957-07:00: executing recompute
```

#### Set two labels for a trip and then the third

Set first two labels

```
DEBUG:in openPopover, setting draftInput = {"start_ts":1628294651.152,"end_ts":1628297929.551}
DEBUG:in storeInput, after setting input.value = walk, draftInput = {"start_ts":1628294651.152,"end_ts":1628297929.551,"label":"walk"}
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: creating new timeout
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: cancelling existing timeout undefined
DEBUG:in openPopover, setting draftInput = {"start_ts":1628294651.152,"end_ts":1628297929.551}
DEBUG:in storeInput, after setting input.value = work, draftInput = {"start_ts":1628294651.152,"end_ts":1628297929.551,"label":"work"}
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: creating new timeout
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: cancelling existing timeout [object Object]
```

Recompute is run, but the trip is not filtered because only two labels are filled out

```
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: executing recompute
```

Third label is filled out

```
DEBUG:in openPopover, setting draftInput = {"start_ts":1628294651.152,"end_ts":1628297929.551}
DEBUG:in storeInput, after setting input.value = no_travel, draftInput = {"start_ts":1628294651.152,"end_ts":1628297929.551,"label":"no_travel"}
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: creating new timeout
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: cancelling existing timeout undefined
```

Final recompute is called and trip is filtered

```
DEBUG:trip starting at 2021-08-06T17:04:11.152000-07:00: executing recompute
```

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -23,6 +23,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
                                     $ionicScrollDelegate, $ionicPopup, ClientStats,
                                     $ionicLoading,
                                     $ionicActionSheet,
+                                    $timeout,
                                     ionicDatePicker,
                                     leafletData, Timeline, CommonGraph, DiaryHelper,
                                     InfScrollFilters,
@@ -236,7 +237,8 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
                 Logger.displayError("multiple filters not supported!", undefined);
             } else {
                 // console.log("Trip n before: "+$scope.data.displayTrips.length);
-                $scope.data.displayTrips = $scope.data.allTrips.filter(f.filter);
+                $scope.data.displayTrips = $scope.data.allTrips.filter(
+                    t => InfScrollFilters.waitingForMod(t) || f.filter(t));
                 // console.log("Trip n after:  "+$scope.data.displayTrips.length);
                 alreadyFiltered = true;
             }
@@ -356,10 +358,36 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
         $scope.editingTrip = angular.undefined;
     }
 
+    /*
+     * Embody the logic for delayed update:
+     * the recompute logic already keeps trips that are waitingForModification
+     * even if they would be filtered otherwise.
+     * so here:
+     * - set the trip as waiting for potential modifications
+     * - create a one minute timeout that will remove the wait and recompute
+     * - clear the existing timeout (if any)
+     */
+    $scope.updateVisibilityAfterDelay = function(trip) {
+      // We have just edited this trip, and are now waiting to see if the user
+      // is going to modify it further
+      trip.waitingForMod = true;
+      let currTimeoutPromise = trip.timeoutPromise;
+      let ONE_MINUTE = 60 * 1000;
+      Logger.log("trip starting at "+trip.start_fmt_time+": creating new timeout");
+      trip.timeoutPromise = $timeout(function() {
+        Logger.log("trip starting at "+trip.start_fmt_time+": executing recompute");
+        trip.waitingForMod = false;
+        trip.timeoutPromise = undefined;
+        $scope.recomputeDisplayTrips();
+      }, ONE_MINUTE);
+      Logger.log("trip starting at "+trip.start_fmt_time+": cancelling existing timeout "+currTimeoutPromise);
+      $timeout.cancel(currTimeoutPromise);
+    }
+
     $scope.updateTripProperties = function(trip) {
       $scope.inferFinalLabels(trip);
       $scope.updateVerifiability(trip);
-      $scope.recomputeDisplayTrips();
+      $scope.updateVisibilityAfterDelay(trip);
     }
 
     /**
--- a/www/js/diary/infinite_scroll_filters.js
+++ b/www/js/diary/infinite_scroll_filters.js
@@ -39,6 +39,10 @@ angular.module('emission.main.diary.infscrollfilters',[
         }
     }
 
+    sf.waitingForMod = function(t) {
+        return t.waitingForMod == true;
+    }
+
     sf.UNLABELED = {
         key: "unlabeled",
         text: $translate.instant(".unlabeled"),


shankari:
Tried to "animate" the delete by turning on 

```
diff --git a/www/templates/diary/infinite_scroll_list.html b/www/templates/diary/infinite_scroll_list.html
index 7aa8521e..969cdfbc 100644
--- a/www/templates/diary/infinite_scroll_list.html
+++ b/www/templates/diary/infinite_scroll_list.html
@@ -49,7 +49,7 @@
             <div class="start-time-tag-inf-scroll">{{trip.display_start_time}}</div>
             <div  style="padding-left: 19%;">

-                        <ion-item id="diary-item" style="background-color: transparent;" class="list-item">
+                        <ion-item id="diary-item" style="background-color: transparent;" class="list-item item-remove-animate">
                 <div id="diary-card" ng-class="trip.listCardClass" ng-style="{height: '150px'}">
                     <div class="row">
                         <div ng-click="showDetail($event, trip)" class="col-90 center-vert" ng-class="listTextClass" style="font-size: 14px; padding-left: 30px; margin-bottom: 0; justify-self: flex-start;" translate=".date-distance-in-time" translate-value-date="{{ trip.display_date }}" translate-value-distance="{{ trip.display_distance }}" translate-value-time="{{ trip.display_time }}"></div>
```

Consistent with https://github.com/ionic-team/ionic-v1/blob/eefba1331bc9206767c651912db71e9ed34850d5/test/html/tabs-starter.html#L58
but it didn't appear to work.

shankari:
delayed_filter mov

shankari:
@GabrielKS if you have time, can you review this? I'm going to merge and deploy it anyway in ~ 20 mins so I can have some real testing done before we push out to participants.

www/js/diary/infinite_scroll_list.js
```diff
@@ -236,7 +237,8 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
                 Logger.displayError("multiple filters not supported!", undefined);
             } else {
                 // console.log("Trip n before: "+$scope.data.displayTrips.length);
-                $scope.data.displayTrips = $scope.data.allTrips.filter(f.filter);
+                $scope.data.displayTrips = $scope.data.allTrips.filter(
```
GabrielKS, line 240: This is quite an edge case, but: what if the user goes to All Unlabeled, labels a trip that is not in To Label (triggering the delayed disappearance) and then switches to To Label? We want the trip to disappear immediately, given that it should not be on To Label at all. Will this keep it around?

GabrielKS (changes requested):

www/js/diary/infinite_scroll_list.js
shankari: Good catch! This will keep it around.

I am not sure I agree with "we want the trip to disappear immediately" though. They did label it very very recently and maybe they still want to change it. In other words, the simple rule for labeling would be "we will give you some time to fix your changes" and showing it in "To Label" will keep that rule.

If that is too confusing, the fix is fairly straightforward - any time we change the selection, we reset the `timeoutPromise` and the `trip.waitingForMod` for all trips.

GabrielKS: I think it will be confusing to break the promise of what the filters do in that way (even though we are already breaking that promise by having labeled trips stick around). Ideal would be to have the trip disappear immediately if you switch to To Label, but then if you switch back to All Unlabeled have it still be there for the rest of the minute. That sounds more complicated to implement though.

shankari: That is way too complicated to implement.

But going back to the promise of what the filters do, my argument is that with this change, we are fundamentally altering the promise of what those filters do (i.e. "even though we are already breaking that promise by having labeled trips stick around").

The new promise is: the "to label" screen will have the trips that you need to label, and any trips that you have labeled recently, giving you a chance to correct them.

This implementation is completely consistent with that promise. I would argue that it is actually *more* confusing to have the behavior be different if the labeling happens directly from the "To Label" screen versus the "All Trips" -> "To Label" screen

shankari:
I'm going to go ahead and merge this now so I can get some testing on the basic new functionality today morning. If we decide to handle that corner case differently, we can introduce it as a bugfix, potentially even after the main deployment, since it does not change the fundamental UX.

PR #788, opened 2021-08-18
Updates

asiripanich:
See the commit messages below.


shankari (approved):
You may want to use icons for `pick_drop_person` versus `pick_drop_item`, otherwise the text is to long to be visible.

But I'm sure you can tweak that as necessary

shankari:
should be deployed by now

asiripanich:
> You

Thanks! :) Emojis to the rescue! #789

PR #789, opened 2021-08-18
Shorten trip confirm options with emojis 🙉

asiripanich:


shankari (approved):

shankari:
Deployed, should be live now

PR #790, opened 2021-08-18
Change the delete delay to 30 sec + add day of the week for clarity

shankari:


shankari:
Not going to wait for review for trivial change.

PR #791, opened 2021-08-18
Walkthrough for new label UI

GabrielKS:
Adds a walkthrough button and tour for the new label UI. Planning to make this play the first time a user views the new UI; haven't done that yet.

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -613,13 +613,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
       },
       steps: [{
         target: '.ion-view-background',
-        content: $translate.instant('new_label_tour.0'),
-        before: function() {
-          var d = $q.defer();
-          $ionicScrollDelegate.scrollTop(true);
-          d.resolve();
-          return d.promise
-        }
+        content: $translate.instant('new_label_tour.0')
       },
       {
         target: '.labelfilter',
@@ -635,7 +629,13 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
       },
       {
         target: '.control-icon-button',
-        content: $translate.instant('new_label_tour.4')
+        content: $translate.instant('new_label_tour.4'),
+        before: function() {
+          var d = $q.defer();
+          $ionicScrollDelegate.scrollTop(true);
+          d.resolve();
+          return d.promise
+        }
       },
       {
         target: '.input-confirm-row',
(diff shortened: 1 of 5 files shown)


www/templates/diary/infinite_scroll_list.html
```diff
@@ -2,12 +2,15 @@
     <ion-nav-bar class="bar-stable">
     </ion-nav-bar>
     <ion-nav-buttons side="left" id="toget" class="row buttons">
-        <button ng-repeat="selF in filterInputs" ng-click="select(selF)" class="col col-50 button labelfilter" ng-class="{on:selF.state}" style="text-align: center;font-size: 14px;font-weight: 600;" translate>
-            {{selF.text}}
-        </button>
-        <button ng-click="resetSelection()" class="col col-50 button labelfilter last" ng-class="{on:allTrips}" style="text-align: center;font-size: 14px;font-weight: 600;" translate>
-            {{'.show-all'}}
-        </button>
```
shankari, line 10: The advantage of using entries such as `col-50` is that they (at least theoretically) auto-resize on different size screens. Can't you use a lower level of `col` - e.g. `col-33` etc instead of hardcoding values?

shankari: There are lots of options!
https://ionicframework.com/docs/v1/components/#grid-explicit

shankari (changes requested):

www/js/diary/infinite_scroll_list.js
```diff
@@ -611,13 +611,71 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
       },
-      steps: []
+      steps: [{
+        target: '.ion-view-background',
+        content: $translate.instant('new_label_tour.0'),
+        before: function() {
+          var d = $q.defer();
+          $ionicScrollDelegate.scrollTop(true);
+          d.resolve();
+          return d.promise
```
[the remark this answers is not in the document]
GabrielKS: Sure, I just saw `$q` on [the tour library reference](http://nozzle.github.io/nzTour/) and assumed that was the standard way to do it.

www/templates/diary/infinite_scroll_list.html
GabrielKS: My intent wasn't to hardcode values, it was to have the buttons automatically resize to fit the text within them, which they do now. Is that objectionable?

GabrielKS: Before I first got to this code, there was a `width` parameter stored in each `infinite_scroll_filter`. I changed it to use the `col` values because that _was_ hardcoded (and not accommodating of multiple languages), but my new solution seems even less hardcoded.

www/js/diary/infinite_scroll_list.js
```diff
@@ -611,27 +611,89 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
+        content: $translate.instant('new_label_tour.3')
+      },
+      {
+        target: '.control-icon-button',
+        content: $translate.instant('new_label_tour.4'),
+        before: function() {
+          var d = $q.defer();
+          $ionicScrollDelegate.scrollTop(true);
+          d.resolve();
+          return d.promise
```
shankari, line 637: I think this is the replacement

```suggestion
          return new Promise(function(resolve, reject)  {
             $ionicScrollDelegate.scrollTop(true);
             resolve();
          });
```

shankari (approved):

PR #795, opened 2021-08-20
Fix the new Label UI walkthrough bug

GabrielKS:
Fix for https://github.com/e-mission/e-mission-docs/issues/669.

We have not yet tested this fix as much as we can; do not merge.

PR #796, opened 2021-09-22
Cherry-pick the walkthrough changes to master

shankari:
This is a replacement for https://github.com/e-mission/e-mission-phone/pull/795
which got messed up with a bunch of unrelated commits.

The walkthrough now works on both physical android and physical iOS, even after hitting the refresh button.
It highlights the entire list instead of individual list items since our current walkthrough doesn't work when there are multiple items with the same class or id.

nzTour is deprecated, so we should move to a new walkthrough mechanism anyway. We can take a look at what that mechanism supports and tighten up the walkthrough.

The build seems to be failing, but that should not be related to our changes since they are UI only.

### DIFF
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -180,7 +180,6 @@ div.labelfilterlist {
 }
 
 .labelfilter:first-of-type {
-  margin-left:5%;
   border-top-left-radius: 5px;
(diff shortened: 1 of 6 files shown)


shankari:
@GabrielKS "worse quality but not broken" fix

PR #797, opened 2021-09-23
Call ForceSync on app resume

shankari:
This is a UI-only hack to experiment with syncing on every app start/resume.
In general, our assumption was that people will open the app only sporadically,
and we want to push data more frequently than that. So we do so in a background
process.

However, due to the various restrictions on background operations in android,
sometimes the background process is not launched. We can ask the user to "force
sync" but that is not a viable solution as well.

We push a UI-only update that force-syncs on every app launch to see the extent
of this problem and confirm that this fixes it.

Once we know that this works, we should move this to native code, potentially
by pushing data at the end of a trip as we do on iOS. We may also want to
introduce silent push notifications on android as well.

Testing done:
- Started app
- Paused and resumed app

Saw message "app has started/resumed, forcing sync on android" in the console
in both cases.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -11,11 +11,12 @@ angular.module('emission', ['ionic',
     'emission.controllers','emission.services', 'emission.plugin.logger',
     'emission.splash.customURLScheme', 'emission.splash.referral',
     'emission.splash.updatecheck', 'emission.services.email',
+    'emission.main.control.sync',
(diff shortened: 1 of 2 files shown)


ericafenyo:
> ... if you could submit a PR to push the data at the end of the trip on android (IIRC, we already do this on iOS), I would be happy to merge it.

Sure, I will submit it once it is ready. Thanks :)

PR #798, opened 2021-09-24
Rciti

atton16:
- Fast forward rciti branch to the most recent master commit
- Migrate Enketo Survey as well as simplify and re-implement the feature to be easier to use
- Use Enketo Survey for Trip confirm
- Hide label and heatmap tabs


www/js/diary/services.js
```diff
@@ -406,6 +413,12 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
   }
 
   dh.getUserInputForTrip = function(tripgj, nextTripgj, userInputList) {
+    // If there is only one item in the list, return it.
+    // This make it compatible when fake list is given (for Enketo Survey).
+    // As well as optimize the performance.
+    if (userInputList.length === 1) {
```
atton16, line 419: Need this inorder for https://github.com/e-mission/e-mission-phone/pull/798/commits/fb338f12df472f002f4b591bc315d9cd0023fef7#diff-b254f964755cdc55e579741bb46be006886e57fa24e7302e413556ab39ccb014R492 to work.

atton16:
In this EnketoSurvey implementation, we use `json/enketoSurveyConfig.json` like so:

```json
{
  "TripConfirmSurvey": {
    "formPath": "json/trip-end-survey-multiple-select.json",
    "version": 1.2,
    "compatibleWith": 1
  }
}
```

We distinguish survey by their key name in the config file.

To launch the survey, simple call `EnketoSurveyLaunch.launch($scope, 'TripConfirmSurvey')`.

The launch method also support options (`opts`) which allows to launch the survey that is linked to `trip` or user (via `uuid`) described here: https://github.com/e-mission/e-mission-phone/commit/fb338f12df472f002f4b591bc315d9cd0023fef7#diff-cb2163d07f53ca7cd4d0b2a3940106b728d96a370f1158e50f7c06ec69b5efa6R79.

To resolve the button label for displaying, one must define the function that takes `xmlDoc` as input and return `string` in `enketo-survey-answer.js`. See example: https://github.com/e-mission/e-mission-phone/commit/fb338f12df472f002f4b591bc315d9cd0023fef7#diff-08e4bb957a2ad1e61811540af7799aca93cb431031e13cb412970f7ed3612fadR47.

Note that the label is resolved at the answer time. So it will also be saved to the database along with other data. This helps us simplify logic on `www/js/diary/list.js` file here https://github.com/e-mission/e-mission-phone/commit/fb338f12df472f002f4b591bc315d9cd0023fef7#diff-b254f964755cdc55e579741bb46be006886e57fa24e7302e413556ab39ccb014R175.

asiripanich:
@shankari could you please review this PR? :)

shankari:
@atton16 can you confirm that I only need to review changes after https://github.com/e-mission/e-mission-phone/pull/798/commits/610fdab4c3ffc452e988c1efc99bbfa46902ddf4?

The rest of it is just a straight pull from master, right?

atton16:
> @atton16 can you confirm that I only need to review changes after [610fdab](https://github.com/e-mission/e-mission-phone/commit/610fdab4c3ffc452e988c1efc99bbfa46902ddf4)?
> 
> The rest of it is just a straight pull from master, right?

Yes, you only need to review after that commit.

shankari:
@atton16 are you done with your changes? BTW, the current refactoring is at:
https://github.com/e-mission/e-mission-phone/pull/799

As part of https://github.com/e-mission/e-mission-docs/issues/674 and long-term maintainability, I plan to think through how to create a similar directive for the enketo survey. So it would be good if this is as close to your final version as possible.

asiripanich:
> 
> 
> @atton16 are you done with your changes? BTW, the current refactoring is at:
> #799
> 
> As part of [e-mission/e-mission-docs#674](https://github.com/e-mission/e-mission-docs/issues/674) and long-term maintainability, I plan to think through how to create a similar directive for the enketo survey. So it would be good if this is as close to your final version as possible.

Yes, the PR is ready to be reviewed.

www/templates/intro/consent-text.html
```diff
@@ -1,17 +1,117 @@
...
+</div>
+<h3>Inclusion criteria</h3>
+<div style="text-align: justify;">
+Before you decide to participate, we need to ensure that you meet the inclusion criteria of this research study. You can participate if the following criteria apply to you:
+</div>
+<ul>
+  <li>You are eighteen years old or older.</li>
+  <li>You reside in a metropolitan area of Australia.</li>
+  <li>You own a smartphone running Android (compatibility varies with device) or Apple iOS 9 (or newer).</li>
+  <li>You have a Google/Gmail account.*</li>
```
shankari: Do you need this given the switch to randomized token generation?

www/js/intro.js
```diff
@@ -134,27 +134,80 @@ angular.module('emission.intro', ['emission.splash.startprefs',
...
+              const iab = $window.cordova.InAppBrowser.open(url, '_blank');
+              const listener = function(event) {
+                console.log("started loading, event = " + JSON.stringify(event));
+                if (event.url === returnUrl || event.url === 'https://ee.kobotoolbox.org/thanks') {
+                  iab.removeEventListener('loadstart', listener);
+                  iab.close();
+                  ionicToast.show(userEmail, 'middle', false, 2500);
+                  $scope.finish();
+                }
+              };
```
shankari: This is to auto-close the survey after completion, right?
would be good to put this into `surveylaunch.js`, since:
- others might also want to use it
- it is more likely to survive changes to the onboarding

In general, I would recommend against putting substantial functionality in `intro.js` since that tends to be heavily customized.

www/json/trip_confirm_options.json
```diff
@@ -0,0 +1,30 @@
+{
+  "MODE": [
```
shankari, line 2: are you using the trip confirm functionality? If not, I would omit this - it just causes additional confusion and potential future merge conflicts. There is already a `trip_confirm_options.json.sample` to provide an overview.

www/templates/diary/list.html
```diff
@@ -86,12 +86,16 @@
                     </div>
                 </div>
              </div>
-             <div class="row" style="padding-left: 5px;padding-right: 5px;">
+             <div 
+                ng-if="isAnalyzed(tripgj)"
```
shankari, line 90: let's avoid calling functions from the HTML. It significantly reduces performance because of the way in which angular . Create a `$scope.isAnalyzed` variable and use it instead

See https://github.com/e-mission/e-mission-phone/commit/30561b94d8691e37fc3e50b8488f5da82f6dd815 for the change which removed function based conditionals.

www/index.html
```diff
@@ -39,7 +46,7 @@
     <script src="lib/moment/min/locales.min.js"></script>
     <script src="lib/moment/min/moment-with-locales.min.js"></script>
     <script src="lib/moment-timezone/builds/moment-timezone-with-data.min.js"></script>
-    <script src="lib/ngCordova/dist/ng-cordova.js"></script>
+    <!-- <script src="lib/ngCordova/dist/ng-cordova.js"></script> -->
```
shankari, line 49: Note to self: we removed the ngCordova library as part of 
https://github.com/e-mission/e-mission-phone/commit/61513d92e2223aabef929c072b05110b518c0462#diff-dedb6c21fdaac9cba4da4afde881ffe5c80f80896274dbf7971d312fc6f5207d
since it is now deprecated.

www/js/diary/list.js
```diff
@@ -170,11 +171,19 @@ angular.module('emission.main.diary.list',['ui-leaflet',
...
-            var userInputEntry = $scope.inputParams[inputType].value2entry[userInput.data.label];
-            if (!angular.isDefined(userInputEntry)) {
-              userInputEntry = ConfirmHelper.getFakeEntry(userInput.data.label);
-              $scope.inputParams[inputType].options.push(userInputEntry);
-              $scope.inputParams[inputType].value2entry[userInput.data.label] = userInputEntry;
+            var userInputEntry;
+            switch(inputType) {
+              case 'SURVEY':
+                userInputEntry = {text: userInput.data.label};
+                break;
```
shankari, line 178: As I said earlier, we should really create directives/modules for this functionality to make it easier to re-use without editing code.

www/js/diary/list.js
```diff
@@ -364,6 +373,10 @@ angular.module('emission.main.diary.list',['ui-leaflet',
     // $scope.expandEarlierOrLater = DiaryHelper.expandEarlierOrLater;
     // $scope.increaseRestElementsTranslate3d = DiaryHelper.increaseRestElementsTranslate3d;
 
+    // rciti start
+    $scope.isAnalyzed = DiaryHelper.isAnalyzed;
+    // rciti end
+
```
shankari, line 379: Ditto. Also, we should store the result of `isAnalysed` as a variable instead of calling a function every time.

www/js/diary/services.js
```diff
@@ -1114,18 +1127,25 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
...
-      // var surveyAnswersPromise = EnketoSurvey.getAllSurveyAnswers("manual/confirm_survey", { populateLabels: true });
 
       // Deal with all the trip retrieval
       Promise.all([tripsFromServerPromise, isProcessingCompletePromise].concat(manualPromises))
         .then(function([processedTripList, completeStatus, ...manualResults]) {
         console.log("Promise.all() finished successfully with length "
           +processedTripList.length+" completeStatus = "+completeStatus);
         var mrString = 'with ' + manualResults.map(function(item, index) {
-            return ' ${mr.length} ${ConfirmHelper.INPUTS[index]}';
+            return ` ${mr.length} ${ConfirmHelper.INPUTS[index]}`;
```
shankari, line 1148: thanks for fixing this annoying bug that I haven't had time to fix before.

www/js/diary/services.js
shankari: I see. But why do you need to call

```
$scope.populateInputFromTimeline(tripgj, undefined, inputType, [{data: result}])
```

from `openPopover` in the first place? 

`populateInputFromTimeline`, and by extension, `getUserInputForTrip` are used for matching the input to the trip. You don't really need that because you know the match already; it is the new trip.
I don't call `populateInputFromTimeline` for the trip confirm inputs

www/js/survey/enketo-survey-service.js
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  /**
+   * _saveData save survey answer data
+   * @returns {Promise<EnketoAnswerData>} answer data
+   */
+  function _saveData() {
+    /** @type {DOMParser} xmlParser */
+    const xmlParser = new $window.DOMParser();
+    const xmlResponse = _state.form.getDataStr();
```
shankari, line 137: nit: Do you plan to store the JSON response as well? Of course, totally fine to do that as a separate enhancement.

www/js/survey/enketo-survey-service.js
```diff
@@ -0,0 +1,213 @@
...
+    const xmlDoc = xmlParser.parseFromString(xmlResponse, 'text/xml');
+
+    const data = {
+      label: EnketoSurveyAnswer.resolveLabel(_state.name, xmlDoc),
```
shankari, line 141: I see this reference to "Note that the label is resolved at the answer time."

shankari (changes requested):
Overall LGTM. The Enketo code looks pretty good now! 

I have a couple of questions and nits and one real change.
Again, this really needs another round of cleanup and modularization before merging to master, but I am pretty much ready to merge to `rciti` once the comments here have been resolved so you can be unblocked for your specific project.

I haven't reviewed it `www/js/survey/*.js` in detail for correctness since I assume that it is primarily a copyover from the previous implementation.

www/js/survey/enketo-survey-service.js
shankari: Having it in JSON will help with the deployer dashboard since it can be automagically displayed just like the various `confirm` fields in the trip table.

www/js/diary/list.js
shankari: In fact, there is already an `isDraft` that you can use instead of creating a new function/variable.

```
tripgj.isDraft = $scope.isDraft(tripgj);
```

atton16: Does the opposite of `isDraft` equal to **analyzed**? If so, i think it is safe to use it.

www/js/diary/services.js
atton16: Inside `openPopover`, we call `EnketoSurveyLaunch.launch` to launch the survey. Then when we receive the answer, it comes in the `.then()` clause of the promise. Only that time that we can get the answer and update the label for that specific trip.

www/templates/intro/consent-text.html
atton16: @asiripanich

www/json/trip_confirm_options.json
atton16: I just put the file to avoid 404 error when debugging the phone.

www/js/intro.js
atton16: I can put this bits into `surveylaunch.js`.

www/js/survey/enketo-survey-service.js
```diff
@@ -105,25 +105,26 @@ angular.module('emission.enketo-survey.service', [
...
-    if (_state.opts.uuid) {
-      answers = answers.filter(answer => answer.data.user_uuid === _state.opts.uuid);
-      answer = answers.length ? answers[0] : null;
-    }
-    return answer ? answer.data.xmlResponse : null;
+      if (_state.opts.trip) {
+        answer = DiaryHelper.getUserInputForTrip(_state.opts.trip, undefined, answers);
+        if (answer) {
+          return answer.data.xmlResponse;
+        }
```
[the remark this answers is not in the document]
atton16: When the survey is launched, user can launch by adding `opts.trip`, `opts.uuid` or without any options at all.
If they use `opts.trip`, then it tries to restore answer based on the trip `start_ts` and `end_ts`.
If they use `opts.uuid`, then it tries to restore answer based on `data.uuid`.
If none `opts` is given, we fall back to line `126`.

www/templates/diary/list.html
atton16: Patched in https://github.com/e-mission/e-mission-phone/pull/798/commits/b57742a68746273b62a4b3bbe91b76d209d365c4.

www/js/survey/enketo-survey-service.js
atton16: We need to store it as  XML doc to make it compatible with Enketo library when we restore the answer. But storing JSON along with XML would be nice as you suggested. I will to this one later.

www/js/survey/enketo-survey-service.js
atton16: I changed the flow to make it easier to get the answer when we call `populateInputFromTimeline. Thus avoid further processing.

www/json/trip_confirm_options.json
shankari: you can ignore the 404 error; there is a fallback to read the sample file in that case and everything still works

www/js/diary/list.js
shankari: The implementation is not identical, but I think they should have the same effect. Feel free to look at the `isDraft` implementation!

www/index.html
shankari: Right, that is why it is a note **to self**

www/js/survey/enketo-survey-service.js
shankari: again, this is a note to self 😄

www/js/survey/enketo-survey-service.js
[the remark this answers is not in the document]
shankari: I understand that.

However, the code here is

```
        if (answer) {
          return answer.data.xmlResponse;
        }
```

The code at line 126 is 

```
       return answer ? answer.data.xmlResponse : null;
```

They both seem to have the same functionality, so I was just wondering why you used two different code fragments. It is not a huge issue, (which is why it is labeled "nit"), but I like to follow standard patterns if possible.

www/js/diary/list.js
```diff
@@ -199,6 +199,9 @@ angular.module('emission.main.diary.list',['ui-leaflet',
         tripgj.display_time = $scope.getFormattedTimeRange(tripgj.data.properties.start_ts,
                                 tripgj.data.properties.end_ts);
         tripgj.isDraft = $scope.isDraft(tripgj);
+        // rciti start
+        tripgj.isAnalyzed = $scope.isAnalyzed(tripgj);
+        // rciti end
```
shankari, line 204: I would prefer using `isDraft` (if possible) to avoid rciti specific code to the maximum extent

www/js/intro.js
shankari: great, but please make sure it is configurable/modular. Not everybody will use Enketo (the CEO project is using Google Forms for now), and not every enketo user will want to auto-close surveys.

shankari: LMK the ETA for this change. I'd like to figure out whether I should wait for it or not.

atton16: Should be done in a few hours.

www/js/diary/services.js
shankari: I understand how promises work. What I don't understand is why you need to call `populateInputFromTimeline`. For example, `populateInputFromTimeline` calls, in its first line:

```
        var userInput = DiaryHelper.getUserInputForTrip(tripgj, nextTripgj, inputList);
```

This iterates through all the user inputs and tries to match the trip. But in your case, you already know which trip this is the input for. So why take that additional performance hit instead of just setting the value into the trip directly?

e.g. something like

```
tripgj.userInput[inputType] = result;
```

or 

```
tripgj.userInput[inputType] = {data: result};
```

or something similar

As I said earlier, that's what I do with the current user labels (see `$scope.save` for example).

atton16:
Ready for another round of review.

www/js/diary/list.js
atton16: migrated to `isDraft`.

www/js/survey/enketo-survey-service.js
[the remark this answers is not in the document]
atton16: If the answer is found by checking with `trip`, I just wanted to return right away and not checking for `uuid` any further. I could not return `null` if answer is not found when checking against `trip` since we can further check on `uuid`. That is why two bits are slightly different.

www/templates/intro/consent-text.html
atton16: Updated in this .merge commit https://github.com/e-mission/e-mission-phone/pull/798/commits/b1b8bfdca11393bd4ac91a2cdb128993a310cc41

www/json/trip_confirm_options.json
atton16: Ok but how can I un-commit it?

www/js/intro.js
atton16: Done

www/js/diary/services.js
atton16: hmm, haven't think of that before. Thank you for pointing out. Let me try...

atton16: I only thought that putting all ui update in the same place would be nice.

atton16: Done https://github.com/e-mission/e-mission-phone/pull/798/commits/c5ead07d23f4e18f33bcd2bfdea7943f9ddcf534

www/json/trip_confirm_options.json
shankari: `git rm --cached www/json/trip_confirm_options.json`
and then `git commit`

www/js/survey/enketo-survey-service.js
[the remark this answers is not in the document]
shankari: I think we are speaking at cross-purposes here, but this is a nit, so I am going to let it go.

shankari (changes requested):
I think this is pretty much ready now.

Three issues to resolve before this merge:
- Remove the trip_confirm_options.json file
- Confirm that removal of survey functionality was intentional + add testing results on google forms
- Fix the intro after feedback from @asiripanich 

Two long-term issues that will be handled after this merge
(you may want to check out the restructuring PR first)
- store the result as JSON so it is easier to display generically in the dashboard
- create an <enketo-survey> directive (similar to the restructuring) that removes all the `if/else` survey code from `list.js`.
- Store JSON

www/js/survey/launch.js
```diff
@@ -161,31 +161,52 @@ angular.module('emission.survey.launch', ['emission.services',
...
-        if (returnUrl) {
-          // NOTE: We manually append with & to make it compatible when we embed uuid as path param
-          // i.e. http://pe.byamarin.com/${uuid}&return_url=https://www.google.com
-          // However, if uuid is appended normally `return_url` will correctly be appended with & anyways.
-          modifiedURL = `${modifiedURL}&return_url=${encodeURIComponent(returnUrl)}`;
-        }
+        const urlObj = new URL(url);
+        urlObj.searchParams.append(opts.returnURLSearchParam, encodeURIComponent(opts.returnURL));
+        urlObj.searchParams.append(opts.uuidSearchParam, uuid);
+        const modifiedURL = urlObj.toString();
```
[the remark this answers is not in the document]
atton16: We do not need it anymore. So I simplify the code. However we need the search param to be there always even with the blank value to make pattern matching works on our end. And the code looks nicer with out if condition.

The survey launch fine, screenshot below:
Screen Shot 2564-09-29 at 09 05 15

But you have to change the function calling to:

```javascript
        SurveyLaunch.startSurveyPrefilled(
            'https://docs.google.com/forms/d/e/1FAIpQLSeMqtl_rHJkg-lZ-9xnQQHrJyBYOQl83zpKmwXfgOUFVOOwLg/viewform', {
                uuidSearchParam: 'entry.779873496',
        });
```

www/json/trip_confirm_options.json
atton16: I thought there is a magic command that you can scoop the file out of commit history. Anyways, done.

www/templates/intro/consent-text.html
```diff
@@ -1,125 +1,94 @@
...
-</div>
+
 <ul>
-  <li>You are eighteen years old or older.</li>
-  <li>You reside in a metropolitan area of Australia.</li>
-  <li>You own a smartphone running Android (compatibility varies with device) or Apple iOS 9 (or newer).</li>
-  <li>You have a Google/Gmail account.*</li>
+<li>- You are eighteen years old or older.</li>
+<li>- You reside in a metropolitan area of Australia.</li>
+<li>- You own a smartphone running Android (compatibility varies with device) or Apple iOS 9 (or newer).</b></li>
```
atton16: The consent is fixed in merge commit here.

shankari (approved):

shankari:
Deployed, should be live in ~ 20 mins.

asiripanich:
@shankari, i just emailed you a new connectionConfig file. Could you please update that before deploying it? Thanks.

shankari:
already did update before deploying

PR #806, opened 2022-01-15
Handle the case where no trips have been labeled for two weeks

shankari:
This is likely fairly common for those people who have given up on labeling, so
it is worth addressing. The problem is that in that case,

```
lastWeek.footprint.low = 0
thisWeek.footprint.low = 0

thisWeek.footprint.low/lastWeek.footprint.low = NaN
```

So we end up with a blank "low" on the diff display, which looks weird.

To fix, we add a pre-processing step to the directive, which converts NaN -> 0
and display the % if it is >= 0

This then leads to the marginally less weird case in which we don't show the
arrow and we don't show the OR

To fix that as well, we remove the `ng-if` for the OR (not sure why it was ever added)
and add an = for the case in which the metric is 0

### DIFF
--- a/www/js/metrics.js
+++ b/www/js/metrics.js
@@ -1390,6 +1390,11 @@ angular.module('emission.main.metrics',['nvd3',
         scope: {
             change: "="
         },
+        link: function(scope) {
+            if (isNaN(scope.change.low)) scope.change.low = 0;
+            if (isNaN(scope.change.high)) scope.change.high = 0;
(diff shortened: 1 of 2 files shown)


shankari:
| Before changes | zero replace and `>= 0` fix | Final version |
|------------------|------------------|-----------------|
| Screenshot_1642250891 | Screenshot_1642251584 | Screenshot_1642254790 |

PR #808, opened 2022-01-21
Fit and finish fixes to the dashboard screen

shankari:
Based on program admin feedback.
- Add in the "air" mode
- Add "My" to all headings
- Change "worst" to "CO2 saved vs. worst"
- Add a meaningful equivalent of the CO2 saved


shankari:
Note that the meaningful equivalent is off by a bit since we used the rounded up value (67) to check against the EPA website.

| Renamed everything as "My" |  Changed worst to saved | Meaningful equivalent check|
|-----------------|---------------|------------------|
| Screenshot_1642742951 | Screenshot_1642742958  | EPA Comparison |

shankari:
After flipping the saved values correctly. The saved charges are also now closer to the manual check against the EPA values.

PR #812, opened 2022-02-10
Create the iOS version of the status screen

shankari:
- Replace placeholders with appropriate text
- Add `scope.setupIOSLocChecks` and `scope.setupIOSFitnessChecks`
- For notifications, we re-use the android settings
- For background restrictions, we disable them since they don't exist (or at
  least, are not user controllable) on iOS

### DIFF
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -321,7 +321,7 @@
                 "description": {
                     "android-lt-9": "Location services should be enabled and set to High Accuracy. This allows us to accurately record the trajectory of the travel",
                     "android-gte-9": "Location services should be enabled. This allows us to access location data and generate the trip log",
-                    "ios": "Placeholder"
+                    "ios": "Location services should be enabled. This allows us to access location data and generate the trip log"
                 }
             },
             "locperms": {
@@ -331,11 +331,12 @@
                     "android-6-9": "Please select 'allow'",
                     "android-10": "Please select 'Allow all the time'",
                     "android-gte-11": "On the app settings page, choose the 'Location' permission and set it to 'Allow all the time'",
-                    "ios": "Placeholder"
+                    "ios-lt-13": "Please select 'Always allow'",
+                    "ios-gte-13": "On the app settings page, please select 'Always' and 'Precise' and return here to continue"
                 }
             },
             "overall-fitness-name-android": "Physical activity",
-            "overall-fitness-name-ios": "Motion & Fitness",
+            "overall-fitness-name-ios": "Motion and Fitness",
             "fitnessperms": {
                 "name": "Fitness Permission",
                 "description": {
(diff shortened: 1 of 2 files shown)


shankari:
|  Screen top | Screen bottom |
| -------------|----------------|
| Simulator Screen Shot - iPhone 13 - 2022-02-10 at 15 56 04 | Simulator Screen Shot - iPhone 13 - 2022-02-10 at 15 56 11 |

shankari:
on testing, the iOS version of the dashboard is missing the footprint. It works on android. The issue seems to be that the date range is 3 lines - it is 2 lines on android and works fine there. It also used to work on iOS through the UI update but it doesn't today. When I launched the production app, I briefly saw a correctly formatted dashboard, which was then displayed incorrectly.

| Android | iOS |
|------|-------|
| Screenshot_1645574268 | Simulator Screen Shot - iPhone 8 - 2022-02-22 at 15 55 46 |

shankari:
Digging a little deeper:
- on android, the width is 206px for the date range and 120px for the "change dates" button
- on iOS, the width is 195 for the `ng-if` div and 105 px for the `metric-second-half` div, it is 195px for the change dates parent and 120px for the actual button

Why is there such a large margin?


```
<div ng-if="uictrl.showSummary" style="width:50%; float:right; padding-top: 10px;" class="">
      <div class="metric-second-half">
      <div class="ng-binding"> Feb 16, 2022 ➡️  Feb 23, 2022 </div>
      </div>
    </div>
```

Ah it is because the `metrics-second-half` class has a width of 105px 

```
/*Change later*/
.metric-second-half {
  width: 105px;
  margin: auto;
}
```
Apparently that hardcoded width is large enough on android but not on iOS?

shankari:
Setting an explicit % width (90%) for the range seems to fix it, but let's see why we had a hardcoded value to begin with.

- It was changed from a % (38%) to a hardcoded value in https://github.com/e-mission/e-mission-phone/commit/75255b4200530dc14ef3f7a6a0b88d1da962df48#diff-f0b16f371e059afcc066174f37fb73072e9d99b5a5f08a348f5def05855ca4c6, mixed in with a bunch of other changes.

- Is it used anywhere else or can we just change it? Only used in this location, for both halves of that row (the date range and the "change date" button.

```
      <div class="metric-second-half">
      <div> {{ selectCtrl.fromDateTimestamp.format('ll') }} ➡️  {{ selectCtrl.toDateTimestamp.format('ll') }} </div>
      </div>
    </div>
    <div ng-if="uictrl.showChart" style="width:50%;float:right; padding-top: 10px">
      <div class="metric-second-half">
```

Ok, let's go ahead and change it.

shankari:
Changing it to 90% fixed it on iOS without breaking on android.

|  iOS | Android |
|-------|------|
| Simulator Screen Shot - iPhone 12 Pro - 2022-02-22 at 21 30 01 | Screenshot_1645594188 |

shankari:
Making some final wording changes since we don't have any fallbacks.

shankari:
This finally closes https://github.com/e-mission/e-mission-docs/issues/680

PR #814, opened 2022-03-23
Remove most of the devDependencies from leaflet

shankari:
Since we don't build it now and we are never going to build it


shankari:
Should remove the dependabot alert as well.

PR #827, opened 2022-05-17
View QR Code Row

aGuttman:
Added a row to the profile screen to view qr code. Pressing the icon creates a popup where qr code will go.

### DIFF
--- a/www/js/control/general-settings.js
+++ b/www/js/control/general-settings.js
@@ -82,9 +82,10 @@ angular.module('emission.main.control',['emission.services',
     $scope.viewPrivacyPolicy = function($event) {
         // button -> list element -> scroll
         // const targetEl = $event.currentTarget.parentElement.parentElement;
-        if ($scope.ppp) {
-            $scope.ppp.show($event);
-        } else {
+
+        // if ($scope.ppp) {
+        //     $scope.ppp.show($event);
+        // } else {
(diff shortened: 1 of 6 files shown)


www/js/control/general-settings.js
```diff
@@ -82,17 +82,28 @@ angular.module('emission.main.control',['emission.services',
...
                     $scope.ppp = p;
                     $scope.ppp.show($event);
                 });
             }).catch((err) => Logger.displayError("Error while displaying privacy policy", err));
-        }
+        // }
+    }
+
+    $scope.viewQRCode = function($event) {
+        i18nUtils.geti18nFileName("templates/", "control/qrc", ".html").then((qrcFileName) => {
```
shankari, line 100: you don't need the QR code template to be internationalized.

www/templates/control/qrc.html
```diff
@@ -0,0 +1,12 @@
+<script src="/node_modules/qrcode-generator/js/qrcode.js"></script>
+<script src="/node_modules/qrcode-generator/js/qrcode_UTF8.js"></script>
+<script src="/node_modules/angular-qrcode/angular-qrcode.js"></script>
```
shankari, line 3: You need to have this module in `package.json`
Also, the code in `/node_modules` is typically not available to the app. Are you use this works?

shankari (changes requested):
Please indicate if this actually works.

PR #831, opened 2022-05-26
Revert unsw specific changes

shankari:


shankari:
All UNSW changes seem to have been reverted.
The demographic survey works.
The trip labels have some minor issues but we can resolve them later.
https://github.com/e-mission/e-mission-docs/issues/727#issuecomment-1138942682

Squash merging this so that we can resolve the trip label issues and finally merge.

PR #833, opened 2022-05-27
Switch to an autogenerated token with a manual backup

shankari:
This allows users to:
- quickly login without typing anything
- switch phones and handle resets
- login without providing PII

It also means that if people didn't store their token, their data is lost.
Maybe we should explore people opting in to associate that token with an email address, but that is too much work at this time.

Future work:
- Support logging in by scanning a QR code
- Support logging in by scanning a program-provided QR code (remove autogenerated option)
- Remove the native auth methods since this is the only option we will support


shankari:
The android build has been failing consistently with 

```
Running command: pod install --verbose
  Cloning into 'trunk'...

  fatal: unable to access 'https://cdn.cocoapods.org/': The requested URL returned error: 403

Failed to install 'cordova-plugin-em-datacollection': Error: pod: Command failed with exit code 1 Error output:
Cloning into 'trunk'...
  fatal: unable to access 'https://cdn.cocoapods.org/': The requested URL returned error: 403
    at ChildProcess.whenDone (/Users/runner/work/e-mission-phone/e-mission-phone/node_modules/cordova-common/src/superspawn.js:136:25)
    at ChildProcess.emit (events.js:400:28)
    at maybeClose (internal/child_process.js:1058:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:293:5)
```

This is apparently a known cocoapods issue:
https://github.com/CocoaPods/CocoaPods/issues/11355

and should not affect the android build anyway

And this is a UI only change which should not affect the UI build at all.

So if the CDN issue is not resolved by tonight, I will merge anyway.

PR #840, opened 2022-06-07
Added Share QR code button

sebastianbarry:
DISPLAY QR CODE

bower.json
- Added the dependency for angular-qrcode

index.html
- Added the scripts for angular-qrcode

general-settings.js
- Added monospaced.qrcode into the angular.Module so that it works in general-settings.js
- Added viewQRCode which calls the qrc.html file to display the QR code

main-control.html
- Added button in Profile menu to View the QR code (named My
Bikecode)

Created qrc.html
- Translates from general-settings (i was using en.json)
- Displays the QR code and uses the "download" attribute which adds an element in the background called "a" which generates the base64 image data for the QR code (we are not using this currently, we are using the canvas which is translated into base64. Using this auto-generated attribute would be ideal if we can jquery it correctly)

SHARE QR CODE

en.json
- Added EN text for QR code based text lines

general-settings.js
- Added prepopulateQRMessage which contains the text for the sharing feature
- Added shareQR which..
    - Converts the displayed canvas QR code into base64 and saves that into prepopulateQRMessage so that the sharing feature can share the QR code
    - Shares using prepopulateQRMessage

qrc.html
- Displays a ionic popover that displays the My Bikecode and added a "share" button which runs shareQR()


www/i18n/en.json
```diff
@@ -101,7 +102,11 @@
         "consented-ok": "OK",
         "share-message": "Join me in making transportation greener and healthier \nDownload the emission app:",
         "share-subject": "Emission - UC Berkeley Research Project",
-        "share-url": "https://bic2cal.eecs.berkeley.edu/#download"
+        "share-url": "https://bic2cal.eecs.berkeley.edu/#download",
+        "qrcode": "My Bikecode",
```
shankari, line 106: Why is this "Bikecode"? OpenPATH tracks all our travel, not just bike trips.

shankari (changes requested):
Can you also add some screenshots of what this looks like?
I don't see any reason why the privacy code should open the QR code - can you also share a video of that happening so I can verify it?

If we can verify that it is reproducible, let's add additional logging and see what's going on under the hood.

Does this work only on iOS or also on android?

sebastianbarry:
I also added a commit to this branch which replaces "My Bikecode" with "My OPcode"

sebastianbarry:
I see, there is an interesting glitch happening:
When I boot up the app first time, if I click the privacy policy button it correctly brings up the privacy policy popover
If I then click on the QR code button AFTER clicking on the privacy policy, then any time after when I click the privacy policy button again, it starts opening the QR button.

---------------------------------------------------

I walked through the debugger and here is the behavior I'm seeing:
When I click the Privacy policy button, the `viewPrivacyPolicy()` function is called 
Screen Shot 2022-06-08 at 12 35 09 PM

Then I step through and when stepping into the line highlighted in the above image, the code reaches this point which is the last point before the incorrect popover is displayed:
Screen Shot 2022-06-08 at 12 36 48 PM

And then once it goes to the next line the QR code popover is displayed instead of the Privacy policy popover:
Screen Shot 2022-06-08 at 12 34 46 PM

www/js/control/general-settings.js
```diff
@@ -617,4 +629,26 @@ angular.module('emission.main.control',['emission.services',
         });
     }
 
+    var prepopulateQRMessage = {
+        message: $translate.instant('general-settings.qrcode-share-message'), // not supported on some apps (Facebook, Instagram)
+        subject: $translate.instant('general-settings.qrcode-share-subject') // fi. for email
+    }
```
shankari, line 635: Instead of having pre-defined messages here, can we put in the token, or even better, a URL encoding the token? 
Let's start with text.

That would allow people to copy/paste the value or click on a link instead of using a QR code. What if they have only a phone and no computer? They can't use their phone to scan a QR code displayed via email on the same phone.

shankari (commented):
@sebastianbarry this code looks good. Since it looks like you have a better sense of what you are doing now, do you want to try to send the token as text as well? Spend a few hours on it and see if you can get it done; give up if not 😄 

I think that will be useful to get a sense of how long it will take you to make changes now that you are a bit familiar with angular and with the codebase. and in turn be helpful to think through how ambitious we should be.

www/js/control/general-settings.js
sebastianbarry: I went ahead and added a `url:` field to the `prepopulateQRMessage` var. I tested out copying and it produces correctly. My emulator can't test texting or emailing, but I believe it will work since it follows the exact same structure as the `share()` function.

sebastianbarry:
Screen Shot 2022-06-09 at 1 24 39 PM

Here is what the copy/paste looks like

shankari (approved):
LGTM!
I will merge as soon as the tests finish running. Let's discuss adding the scanning at today's 1:1

shankari:
@sebastianbarry all the automated checks failed! i know what the problem is, but see if you can figure it out.
Sorry I missed that detail, but that's why we have the automated tests 😄

sebastianbarry:
> @sebastianbarry all the automated checks failed! i know what the problem is, but see if you can figure it out. Sorry I missed that detail, but that's why we have the automated tests 😄

I think I fixed it in my last commit, let's see if that fixed it 🤞

shankari:
@sebastianbarry whoops! Looks like I didn't actually merge this yesterday afternoon because I got caught up in meetings. Not sure if you have created your new branch for scanning the QR code yet, but if you have, you should pull from `upstream/master` again.

PR #841, opened 2022-06-10
Adding scan OPcode button to login page

sebastianbarry:
index.html
- Added scripts needed to run scan QR code

intro.js
- Added button to "enter existing token" popover called "Scan OPcode" which has some test scan-qrcode code thrown in there. This code is not intended to be final, I just want to make sure it works first
- Right now, my emulator does not have a camera so this fails saying "Scanning failed: unable to detect video capture device"
- Added a $scope.scanQRCode function but it also does not have any meaninful code in it. This is how I think we should do it (by having the scanQRCode function and then calling it from the popover button), but I was struggling to get the function to call. I don't know how to type it (is it just "$scope.scanQRCode();" and it starts working?)

### DIFF
--- a/www/js/intro.js
+++ b/www/js/intro.js
@@ -148,6 +148,29 @@ angular.module('emission.intro', ['emission.splash.startprefs',
             onTap: function(e) {
               return null;
             }
+          },{
+            text: '<b>Scan OPcode</b>',
+            type: 'button-positive',
+            onTap: function(e) {
+              cordova.plugins.barcodeScanner.scan(
+                function (result) {
+                // the code from here is copied and pasted from https://www.sitepoint.com/scanning-qr-code-cordova/
+                // this is not how the feature is intended to be developed, this is just to test it out to see if scanning a qr code works
+                  if(!result.cancelled)
+                  {
+                    alert("Barcode type is: " + result.format);
+                    alert("Decoded text is: " + result.text);
+                  }
+                  else
+                  {
+                    alert("You have cancelled scan");
+                  }
+                },
+                function (error) {
+                    alert("Scanning failed: " + error);
+                }
+              );
+            }
           }
         ]
     });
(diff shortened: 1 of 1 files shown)


www/index.html
```diff
@@ -67,6 +67,10 @@
     <script src="lib/angularLocalStorage/dist/angularLocalStorage.min.js"></script>
     <script src="lib/angularjs-slider/dist/rzslider.min.js"></script>
     <script src="manual_lib/enketo/enketo-bundle.js"></script>
+
+    <script src="https://code.jquery.com/jquery-2.1.4.min.js"></script>
+    <script src="https://code.jquery.com/mobile/1.4.5/jquery.mobile-1.4.5.min.js"></script>
+    <script type="text/javascript" src="cordova.js"></script>
```
shankari, line 73: how do you know that these are the files needed for this to work? We already have jquery included in the index.html In general for hybrid apps, you don't want to use remote locations for code, since then the app won't work if the network is not available.

shankari (changes requested):
> Added button to "enter existing token" popover called "Scan OPcode" which has some test scan-qrcode code thrown in there. This code is not intended to be final, I just want to make sure it works first

Makes sense

> Right now, my emulator does not have a camera so this fails saying "Scanning failed: unable to detect video capture device"

Ah yes, it would be much easier to test this on a physical device. I generally use android to test on a physical device since it is easier but I think you can also use your iPhone even if you don't have a developer account. You would have to build the em-dev-app using the standard cordova instructions, open it as an existing project in xcode and run it on your phone.

Not sure if you are ready to get setup for that 😄 

> Added a $scope.scanQRCode function but it also does not have any meaninful code in it. This is how I think we should do it (by having the scanQRCode function and then calling it from the popover button), but I was struggling to get the function to call. I don't know how to type it (is it just "$scope.scanQRCode();" and it starts working?)

Pretty much. I would actually first scan the QR code using a button directly on the main screen instead of the "enter token" popover since it is easier. functions in javascript are first class objects and you can pass them around just like variables (read up on functional programming).

sebastianbarry:
I amended my previous commit to hopefully resolve those 2 changes you recommended @shankari 

>AMEND
>- I reverted back the changes to comment out the intro slides - my mistake
>- I also reverted the changes to this
>
>> index.html
>> - Added scripts needed to run scan QR code
>
>Removed the scripts from there, they are unnecessary

PR #849, opened 2022-06-28
"Upload Logs" UI fixes

sebastianbarry:
uploadService.js
- Added functionality on 'Cancel' button so that the button closes when you click it, instead of continuing to upload the logs
- Added didCancel variable which keeps track of whether the cancel button was clicked, or not, to either execute the rest of the function, or to not
- Switched the $ionicPopup.alert for the uploading popup to an $ionicPopup.show so that we could change the button text from "OK" to "Cancel"
- Added an ionic spinner in the Uploading popup
- Added the button type 'button-cancel' to make the cancel button display as red

style.css
- Created button type 'button-cancel' which makes buttons red with white text

### DIFF
--- a/www/js/control/uploadService.js
+++ b/www/js/control/uploadService.js
@@ -103,26 +103,26 @@ angular.module('emission.services.upload', ['emission.plugin.logger'])
                     +' placeholder="{{ \'upload-service.please-fill-in-what-is-wrong \' | translate}}">',
                 scope: newScope,
                 buttons: [
-                  { 
+                { 
                     text: 'Cancel',
                     onTap: function(e) {
                         didCancel = true;
                         detailsPopup.close();
                     }
-                  },
-                  {
+                },
+                {
                     text: '<b>Upload</b>',
                     type: 'button-positive',
                     onTap: function(e) {
                         didCancel = false;
                         if (!newScope.data.reason) {
-                        //don't allow the user to close unless he enters wifi password
+                            //don't allow the user to close unless he enters wifi password
                             e.preventDefault();
                         } else {
                             return newScope.data.reason;
                         }
                     }
-                  }
+                }
                 ]
             });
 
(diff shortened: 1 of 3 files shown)


www/js/control/uploadService.js
```diff
@@ -94,46 +94,62 @@ angular.module('emission.services.upload', ['emission.plugin.logger'])
...
+                        if (!newScope.data.reason) {
                         //don't allow the user to close unless he enters wifi password
-                        e.preventDefault();
-                      } else {
-                        return newScope.data.reason;
-                      }
+                            e.preventDefault();
+                        } else {
+                            return newScope.data.reason;
+                        }
```
shankari, line 123: nit: Can you remove the whitespace changes so that we don't get incorrect history/blame for the commits?

www/js/control/uploadService.js
```diff
@@ -94,46 +94,62 @@ angular.module('emission.services.upload', ['emission.plugin.logger'])
...
                 const params = {
                     reason: reason,
                     tz: Intl.DateTimeFormat().resolvedOptions().timeZone
                 }
                 uploadConfig.forEach((url) => {
-                    const progressPopup = $ionicPopup.alert({
+                    const progressPopup = $ionicPopup.show({
                         title: $translate.instant("upload-service.upload-database",
-                            {db: database}),
+                            { db: database }),
```
shankari, line 144: ditto: wrt whitespace changes. You may need to change your vscode settings since it looks like it autoformats code.

shankari (changes requested):
Only very small changes around removing unneeded whitespace.

sebastianbarry:
I adjusted the whitespace, but I don't think the changes on my commit quite reflect it, so please open it and take a look and let me know.

Additionally, I also set my VScode tab size be 2 spaces as opposed to 4.
But for this file, uploadConfig.js, it looks like it was already being formatted as 4 space tab, so I kept my adjustments to be 4 space tab. Let me know if you'd like me to reformat the entire document to 2 space tab uniform.

PR #850, opened 2022-06-30
"Upload Logs" UI fixes (REDO)

sebastianbarry:
Redoing this commit because last time there were some whitespace/blame changes that I did not intend. This time they should be fixed

uploadService.js
- Added functionality on 'Cancel' button so that the button closes when you click it, instead of continuing to upload the logs
- Added didCancel variable which keeps track of whether the cancel button was clicked, or not, to either execute the rest of the function, or to not
- Switched the $ionicPopup.alert for the uploading popup to an $ionicPopup.show so that we could change the button text from "OK" to "Cancel"
- Added an ionic spinner in the Uploading popup
- Added the button type 'button-cancel' to make the cancel button display as red

style.css
- Created button type 'button-cancel' which makes buttons red with white text

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -51,6 +51,11 @@ body.platform-ios {
   height: 40px;
 
 }
+.button.ng-binding.button-cancel {
+  background-color: #d02001;
+  height: 40px;
+  color: #ffffff
+}
 .selected_date_full.ng-binding {
   color: #01D0A7;
 }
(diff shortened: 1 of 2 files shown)


sebastianbarry:
I deleted the old branch of this issue https://github.com/e-mission/e-mission-phone/pull/849
There were some weird whitespace changes and things I could not figure out how to undo, so I just recreated my branch and made sure there were no formatting changes to the existing code

shankari (approved):
LGTM! @sebastianbarry can you also upload some screenshots/video of the new behavior for the record?

PR #853, opened 2022-07-07
New InfiniteScrollTripItem Directive, Issue #740

tyleryandt18:
Pulled out functionality from the InfiniteDiaryListCtrl into the new module, 'infscrolltripitem'. This new scope includes the surveyOpt, itemHt, and showDetail function for the new directive. This relates to issue #740, Unifying label and diary views.


tyleryandt18:
Added the new diary item directive. 

Also, noticed a formatting issue in the map detail screen. The text of the time of the trip and the trip destination does not properly align.
Screen Shot 2022-07-08 at 1 10 33 PM

PR #854, opened 2022-07-10
Adding "Save OPcode" prompt after login during setup

sebastianbarry:
intro.js
- Added the monospaced.qrcode to the libraries
- Added a saveLogin function
- Checks if token is empty (as is the case if the user selects the "Save your OPcode" button) and automatically fills in the token with the random token
- Otherwise, it uses the passed in token
- The token that gets displayed as an OPcode and shared is the $scope.currentToken, because when useing the $ionicPopup.show() function, the template requires a $scope variable in the HTML injection
- Added the shareQR function, which is a variation of the shareQR function seen in www/js/control/general-settings.js. It allows for sharing the QR image together with the OPcode text
- This should be later added as a directive, so that the code is not copied/pasted between these 2 files (which it is now), but for now the functionality works
- Added a currentToken scope variable which keeps track of the currentToken, which will be displayed on the saveToken popup
- Added an "alreadySaved" scope variable to know if the user has already saved their OPcode, so to not bug them twice about it
- Made sure that after scanning or entering in an OPcode manually, it prompts to save
- If a user saves their random token before logging in with their random token, it will not prompt them after logging in. Otherwise, it will.

login.html
- Added the "Save your OPcode" button underneath the randomly generated OPcode token during the setup/login screen

general-settings.js
- Changed the share message to be just the OPcode image and the OPcode text (copyable)
- This is because Shankari and I agreed that it should only save the QR image and the OPcode text, any additional text is confusing and excessive

### DIFF
--- a/www/js/control/general-settings.js
+++ b/www/js/control/general-settings.js
@@ -634,20 +634,11 @@ angular.module('emission.main.control',['emission.services',
         });
     }
 
-    var prepopulateQRMessage = {
-        message: $translate.instant('general-settings.qrcode-share-message'), // not supported on some apps (Facebook, Instagram)
-        subject: $translate.instant('general-settings.qrcode-share-subject') // fi. for email
-    }
-
     $scope.shareQR = function() {
-        //const c = $(".qrcode"); // selects the canvas element containing the QR code
-        //const cbase64 = c[0].toDataURL(); // converts the canvas element into base64 data
-        //prepopulateQRMessage.files = [cbase64]; // adds the base64 data into our share message
-
+        var prepopulateQRMessage = {};  
         const c = document.getElementsByClassName('qrcode-link');
         const cbase64 = c[0].getAttribute('href');
         prepopulateQRMessage.files = [cbase64];
-
         prepopulateQRMessage.url = $scope.settings.auth.email;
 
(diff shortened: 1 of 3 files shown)


shankari (approved):
LGTM!

@sebastianbarry can you add some screenshots/videos of what this looks like before I merge?

shankari:
also, the changes seem like they will work on both `​master`​ and ​`master_for_platform`​ since they are not related to the different `ng-if` in the HTML. So I am going to merge into master first and then pull down to `master_for_platform`.

shankari:
Ah, rebasing to `master` also includes all the config fixes, so not rebasing after all.
Might want to cherry-pick to master later, or just handle the config better in master.

sebastianbarry:
Video showcasing the save button + no show save prompt after clicking the save button

https://user-images.githubusercontent.com/61334340/178170662-fd4b0089-4638-4942-9c61-6d07187a7ec4.mov

Video showcasing Show save prompt after not clicking the save button

https://user-images.githubusercontent.com/61334340/178170665-10401659-c254-4dcd-bf91-13177c272fdc.mov

Video showcasing Show save prompt after inputting token

https://user-images.githubusercontent.com/61334340/178170667-2da1b905-5fed-409e-8e8c-c01b3813a86b.mov

PR #856, opened 2022-07-10
Make the back button and the jump nav at the bottom of the enketo sur…

shankari:
…vey configurable

And turn both on for the demographic survey
While keeping them both turned off for the trip level survey (the default)

### DIFF
--- a/www/js/survey/enketo/enketo-demographics.js
+++ b/www/js/survey/enketo/enketo-demographics.js
@@ -44,7 +44,8 @@ angular.module('emission.survey.enketo.demographics',
   $scope.openPopover = function ($event) {
     return EnketoDemographicsService.loadPriorDemographicSurvey().then((lastSurvey) => {
         return EnketoSurveyLaunch
-          .launch($scope, 'UserProfileSurvey', { prev_demographic_survey: lastSurvey })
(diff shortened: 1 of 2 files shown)


shankari:
| `showBackButton` = false/undefined and `showFormFooterJumpNav` = false/undefined | `showBackButton` = true and `showFormFooterJumpNav` = false/undefined | `showBackButton` = true and `showFormFooterJumpNav` = true
|--------|-------|-----|
| Simulator Screen Shot - iPhone 13 Pro - 2022-07-10 at 16 27 52 | Simulator Screen Shot - iPhone 13 Pro - 2022-07-10 at 16 26 15 |  Simulator Screen Shot - iPhone 13 Pro - 2022-07-10 at 16 29 14 |

shankari:
@sebastianbarry can you review this? Will also give you a sense of how to change this based on the next round of feedback.

sebastianbarry:
This looks great! I would go with 

`showBackButton` = true and 
`showFormFooterJumpNav` = false/undefined

shankari:
I'm going to leave both in place now because the back button looks a bit better visually in that case - it is below "Powered by Enketo" instead of on the same line. You can always change it back in the next release based on feedback for whether it is confusing.

PR #863, opened 2022-07-12
Internationalization/Spanish added to popups

sebastianbarry:
intro.js
- Added $translate to the text displays for the "save token popup" and to the "login existing token popup"

en.json
- Added a "login" section with...
- enter-existing-token
- button-accept
- button-decline
- save-your-opcode
- continue


shankari:
@sebastianbarry can you remove the spanish translations from this repo? they should be in the translate repo.
We don't want them to be in two places
Having all of them in one place makes it easier to hand the translations off to specialized translators.

shankari:
@sebastianbarry and do make sure to check them into the translate repo so I can pull them

shankari:
@sebastianbarry ah nvm, the translations are in https://github.com/e-mission/e-mission-translate/pull/15/files
you can just remove the other files from here; probably faster to do via command line than one by one from the github UI

shankari (approved):

PR #867, opened 2022-07-21
Turning "Save Opcode" into an Angular Module (Directive/Service)

sebastianbarry:
Heads up - this is a huge commit containing the adding of the new Opcode angular module, implementing it into the login/setup screen (+ cleanup of the old Opcode save code), and implementing it into the profile/settings screen (+ cleanup of the old Opcode save code)

save-opcode-item.js
- Created new directive <opcodeshare token="{{token}}" alwayssave="true/false"></opcodeshare> and references save-opcode-item.html
- Created controller which has show and share functions. These functions are just $scope references of the other .service functions
- Created .service with show and share functions, as well as a reset function so that if the user logs in multiple times, it will reset the alreadySaved status

index.html
- Added the libraries for emission.save-opcode-item, and the Opcode object in the controller
- Added an Opcode.reset right after the program generates a randomToken
- Replaced the "alreadySaved" variables to "alwayssaveopcode" to work with the new format (checking if we need to always save, not if we already saved) and the word opcode lets the developer know that this is an opcode-related variable being held in the scope
- Replaced the "alreadySaved" check at the end of login() to just go straight to saveLogin
- Removed currentToken and alreadySaved variables which are no longer needed
- Removed the savePopup chunk of code from saveLogin(), and replacing it with Opcode.show(...);
- Removed ShareQR (both shareQR and the savePopup are now in the new angular module/directive/service)

login.html
- Replaced the save button code with the <opcodeshare> directive

save-opcode-item.hmtl
- Added this file
- This file contains the code that was previously in login.html

index.html
- Added save-opcode-item script

CLEANUP OF OLD OPCODE STUFF

qrc.hmtl
- Deleted this file. The HTML was old, and has been replaced with the directive

en.json
- Deleted the "qrcode" related text that was being used in qrc.html

general-settings.js
- Added save-opcode-item to the libraries
- Added Opcode to the controllers
- Removed the code that displays the qrc.html code in viewQRCode, and replaced it with viewOpcode, that calls the Opcode.show service

main-control.hmtl
- Replaced deleted viewQRCode() with new viewOpcode()


sebastianbarry:
We have moved away from Angular directives in favor of React hooks, closing

PR #868, opened 2022-07-28
Added Survey XLS file / Updated Demographics Survey

sebastianbarry:
Demo-survey-v1.xlsx
- Added this file so that it lives in the repository, and can be imported to KoBotoolbox by other devs (right now, KoBotoolbox can't import XML, only XLS files, despite having an export to both option)
- Added spanish translations to the KoBotoolbox XLS version

Demo-survey-v1.xml
- Added "What is your race/ethnicity?" question
- Added "Do you consider yourself transgender?" question
- Added answer to "What is your gender?" -> "Non-binary/genderqueer/gender non-conforming"
- Added answer to "Which one below describes you best?" -> "None of the above"
- Added answer to "Which best describes your current job?" -> "Other"
- Added skip rules to "Job related questions" if user indicates they don't work

Demo-survey-v1.json
- Export of Demo-survey-v1.xml


sebastianbarry:
@shankari 
My test of the survey worked:
Simulator Screen Shot - iPhone 11 - 2022-07-28 at 11 55 08

### However,
**It looks like the "Back" button is missing.** I checked your PR for how you did this (https://github.com/e-mission/e-mission-phone/pull/856/files) But it looks like you edited the files...
- www/js/survey/enketo/enketo-demographics.js
- www/js/survey/enketo/launch.js

Which are already edited in _my push_:

```
  $scope.openPopover = function ($event) {
    return EnketoDemographicsService.loadPriorDemographicSurvey().then((lastSurvey) => {
        return EnketoSurveyLaunch
          .launch($scope, 'UserProfileSurvey', { prev_demographic_survey: lastSurvey,
                showBackButton: true, showFormFooterJumpNav: true  })
          .then(result => {
            console.log("demographic survey result ", result);
          });
    });
  };
```
**etc.**

shankari:
> So, I'm not sure why the "Back" button isn't displaying on my updated version of the survey

@sebastianbarry the first page will never have the back button since there is nothing to go back to 😄

shankari:
@sebastianbarry please do confirm that the back button shows up on the second page

sebastianbarry:
Thank you for pointing this out! Disregard my last message, the back button is displaying: 
Simulator Screen Shot - iPhone 11 - 2022-07-28 at 14 48 14

sebastianbarry:
I tested it and found 2 issues:
One of the Spanish translations was not being displayed, so I fixed that here:
Simulator Screen Shot - iPhone 11 - 2022-07-28 at 15 02 04
I also found that the Jobs section was displaying as a whole list, instead of one page per question, so I fixed that here too:
Simulator Screen Shot - iPhone 11 - 2022-07-28 at 14 59 58
-------
Now, after testing finally in English and Spanish, I do not see any further issues. This PR should be good to close
------

shankari:
I didn't actually review the files since they are all autogenerated. Thank you for taking the initiative to get the XLS file and to merge all the manual changes into it so we can make changes more easily going forward.

shankari:
@sebastianbarry 

Couple of checks before I merge this:

> Added skip rules to "Job related questions" if user indicates they don't work

what if the user says "Prefer not to say"?

> Added answer to "Which best describes your current job?" -> "Other"

If they choose "other" can they type out a value? Otherwise, what is the difference between "Other" and "Prefer not to say"?

sebastianbarry:
>| Added skip rules to "Job related questions" if user indicates they don't work
>what if the user says "Prefer not to say"?

The skip rule only skips the "Job Related Questions" section  if the user chooses "No", so if a user chooses "Prefer not to say", then they will still receive the job related questions. Right now it functions this way, so that if someone does work but did not want to indicate that they did in this particular question, it would still ask them the questions. This is also so that, if someone chose to select "prefer not to say" for all questions, they would still fall under the general audience and thus be given these questions. The more questions we ask, the better (unless they indicate specifically that they do not work, in which case we don't ask them job related questions). @shankari **What are your thoughts on this reasoning?**


>| Added answer to "Which best describes your current job?" -> "Other"
>If they choose "other" can they type out a value? Otherwise, what is the difference between "Other" and "Prefer not to say"?

Right now, there is no difference between "Other" and "Prefer not to say". Since users are going through the survey and they see "prefer not to say" as an answer to every question, they may overlook that as an option on this question. _There was user feedback for this question that "there should be an 'other' option", despite there already being a prefer not to say option_.
However I agree, this "Other" should be followed with a question that asks something along the lines of "Since you chose Other, How would you describe your job?". I will put this as a change to make in  Demographics Survey Changes 2 #771

shankari:
I would actually skip the job related questions if they say "prefer not to say". Because they already told us that they don't want to say anything about jobs, not even whether they are currently employed or not. So no need to waste their time making them go through multiple pages of job-related questions.

> I will put this as a change to make in Demographics Survey Changes 2 https://github.com/e-mission/e-mission-phone/pull/771

Sounds good

sebastianbarry:
> @sebastianbarry I would actually skip the job related questions if they say "prefer not to say". Because they already told us that they don't want to say anything about jobs, not even whether they are currently employed or not. So no need to waste their time making them go through multiple pages of job-related questions.

I agree with this! I'll put it as a change in Demographics Survey Changes 2 #771 as well

shankari:
Cherry-picked this to master

```
$ git cherry-pick b20776c3bcb04bb4e7a2ce19b20ceb431b10971c
[master 7fd34714] Added Survey XLS file / Updated Demographics Survey
 Author: Sebastian Barry <sebbarry@comcast.net>
 Date: Thu Jul 28 12:01:23 2022 -0600
 5 files changed, 2314 insertions(+), 1968 deletions(-)
 rewrite survey-resources/data-json/demo-survey-v1.json (96%)
 create mode 100644 survey-resources/data-xls/demo-survey-v1.xlsx
 rewrite www/json/demo-survey-v1.json (96%)

$ git cherry-pick 82c997cb32b6ca23c31a75f2f78ce643951d1d45
[master 381066f6] Small testing fixes
 Author: Sebastian Barry <sebbarry@comcast.net>
 Date: Thu Jul 28 15:01:46 2022 -0600
 5 files changed, 40 insertions(+), 40 deletions(-)
 rewrite survey-resources/data-json/demo-survey-v1.json (73%)
 rewrite survey-resources/data-xls/demo-survey-v1.xlsx (75%)
 rewrite www/json/demo-survey-v1.json (73%)
```

shankari:
Cherry-picked commits:
- https://github.com/e-mission/e-mission-phone/commit/7fd34714f1aac31cf723113449207e10a4a626dd
- https://github.com/e-mission/e-mission-phone/commit/381066f6a72cb65ffa991306055b8452576df9f9

PR #871, opened 2022-08-11
Unified label screen

tyleryandt18:
This PR should include the commit history from SavoireFaire Linux. I cloned the Unified-Diary-Screen branch, pulled the savoire Dev branch, manually resolved the merge conflicts, and then lastly pulled my Unified-Diary-Screen branch changes onto it.


shankari:
I don't think this is the right approach - it pulls over a bunch of changes that we don't want (notably to the config.xml etc).
And of course, it has some merge conflicts as well.
I think we only want to pull the changes to the `www/js/diary` and `www/templates/diary` files.

@lgharib what do you think about this change (unifying the diary and label screens)? Would you be able to pull these changes when you next update the app?

PR #872, opened 2022-08-12
Demographics survey changes 2

sebastianbarry:
- removed question "How did you usually get to your primary job last week? If you used more than one mode of transportation, please select the one used for most of the distance"
- removed question "What is your typical access/egress mode to/from the transit stop?"
- added skip logic to "Job related questions" section to only not skip if user selects "Yes" to "Are you a paid worker?". If they choose "no" or "prefer not to say", this section is skipped
- added question "Please describe your primary job" if user selects "Other" to "Which best describes your current job?"
- renamed survey to "OpenPATH Demographics Survey", before it was "Location History Data based Travel Pattern Prediction Survey"
- named the new demo-survey file to "demo-survey-v2" and reflected that in enketoSurveyConfig.json.sample
- added spanish translation to new questions


shankari (approved):
Don't have an easy way to test, but the summary is pretty good 😄

PR #873, opened 2022-08-12
Add support for e-car to give people a better sense of their carbon f…

shankari:
…ootprint

Also duplicated at: https://github.com/e-mission/e-mission-docs/issues/688#issuecomment-1212776169

Since we are NREL, we want to also support the difference in energy intensity between car vs. e-car.

Estimates for the average e-car:
- From Andrew Kotz: 3 - 4 miles/kWH = 333 - 250 WH/mile
- From Andy's Leaf: 244 WH/mile
- From my Leaf (4.2 miles/kWH average, probably because I am in the SF Bay Area): 238 WH/mile

Let's go with 250 WH/mile (0.25 kWH/VMT) since most of the numbers seem to be clustered around there, and 250 is a nice round number 😄

So we will add entries for:
- e_car_drove_alone: 250 WH/mile (versus e-bike at 22 WH/mile)
- e_car_shared_ride:  125 WH/mile (versus e-bike at 22 WH/mile)

We will continue to use 1166 lb per MWH, consistent with egrid estimate for CO at the time of the CEO mini-pilot
https://www.epa.gov/egrid/power-profiler#/RMPA

1,000,000 WH = 1166 lb
250 WH = (250 * 1166) / 1000000 = 0.2915 lb
125 WH = (125 * 1166) / 1000000 = 0.1458 lb

- e_car_drove_alone: 0.2915 lb/mile
- e_car_shared_ride:  0.1458 lb/mile

We want the values in kg/PkmT

0.2915 lb = 1 mile
0.1322 kg = 1.609 km
(0.1322 / 1.609) = 0.08216 kg/PkmT (versus 0.00728 for e-bike, so the magnitude seems right)

0.1458 lb = 1 mile
0.0661 kg = 1.609 meters
(0.0661 / 1.609) = 0.04108 kg/PkmT (half of drove_alone) so the magnitude seems right

- e_car_drove_alone: 0.08216 kg/meter
- e_car_shared_ride:  0.04108 kg/meter

Testing done:
- Labeled a bunch of trips as e-car
- Ensured that the carbon footprint dropped significantly

### DIFF
--- a/www/json/trip_confirm_options.json.sample
+++ b/www/json/trip_confirm_options.json.sample
@@ -6,8 +6,10 @@
             "ALL": {"range": [0, -1], "mets": 4.9}
        }, "co2PerMeter": 0.00728},
        {"text":"Scooter share","value":"scootershare", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.00894},
-       {"text":"Drove Alone","value":"drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.22031},
-       {"text":"Shared Ride","value":"shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.11015},
+       {"text":"Gas Car Drove Alone","value":"drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.22031},
+       {"text":"Gas Car Shared Ride","value":"shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.11015},
+       {"text":"E-Car Drove Alone","value":"e_car_drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.08216},
+       {"text":"E-Car Shared Ride","value":"e_car_shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.04108},
        {"text":"Taxi/Uber/Lyft","value":"taxi", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.30741},
        {"text":"Bus","value":"bus", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.20727},
        {"text":"Train","value":"train", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.12256},


shankari:
@sebastianbarry here you go

shankari:
Testing done:

|  Selecting e-car | works |
|--------|--------|
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-12 at 14 22 51 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-12 at 14 23 07 | 

| carbon footprint of driving alone | drops |
|----------|-------------|
| Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-12 at 14 50 19 | Simulator Screen Shot - iPhone 12 Pro Max - 2022-08-12 at 15 26 52 |

sebastianbarry:
LGTM, the only thing I may want to point out: 

```
{"text":"Gas Car Drove Alone","value":"drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.22031},
{"text":"Gas Car Shared Ride","value":"shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.11015},
{"text":"E-Car Drove Alone","value":"e_car_drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.08216},
{"text":"E-Car Shared Ride","value":"e_car_shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.04108},
```

For `Gas Car Drove Alone` and `Gas Car Shared Ride` we have `drove_alone` and `shared_ride, while for
`E-Car Drove Alone` and `E-Car Shared Ride` we have `e_car_drove_alone` and `e_car_shared_ride`

For clarity, since for e-car we specify in the `value` field `e_car_----`, we may want to for gas car specify in the `value` field `gas_car_-----`

www/json/trip_confirm_options.json.sample
```diff
@@ -6,8 +6,10 @@
             "ALL": {"range": [0, -1], "mets": 4.9}
        }, "co2PerMeter": 0.00728},
        {"text":"Scooter share","value":"scootershare", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.00894},
-       {"text":"Drove Alone","value":"drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.22031},
-       {"text":"Shared Ride","value":"shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.11015},
+       {"text":"Gas Car Drove Alone","value":"drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.22031},
+       {"text":"Gas Car Shared Ride","value":"shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.11015},
```
sebastianbarry, line 10:
```suggestion
       {"text":"Gas Car Drove Alone","value":"gas_car_drove_alone", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.22031},
       {"text":"Gas Car Shared Ride","value":"gas_car_shared_ride", "met_equivalent": "IN_VEHICLE", "co2PerMeter": 0.11015},
```

sebastianbarry:
Do we have an issue to attach this to? If not, I can create one and put it in this week's sprint on the [OpenPATH weekly sprints Project](https://github.com/orgs/e-mission/projects/1/views/1?layout=board)

shankari:
@sebastianbarry this was such a small change that I didn't create an issue for it
But feel free to create one if you have time!

www/json/trip_confirm_options.json.sample
shankari: If this were a brand-new system, this would make sense. However, we are updating an existing system.
If we change this to `gas_car_drove_alone` and the user tried to display a trip from last week which was labeled as `drove_alone`, then it would not find any match.

Keeping it as `drove_alone` gives us backwards compatibility.

We can think about how to deal with backwards-compat programmatically. For example, we can write a script to change all `drove_alone` to `gas_car_drove_alone` on the server. It's a little bit tricky because it is tricky to guarantee that the server and the phone are in sync since phone apps have to go through the app store review process.

There are ways around all of this (hardcoded, temporary hack that handles both modes on the phone, scheduling release at a particular date/time), but not sure I want to tackle that right now.

shankari: This update will not make things worse, since we always had `drove_alone` strings. We can make the change later as part of a separate release. File an issue, and maybe you can tackle it as a "small" change.

PR #874, opened 2022-08-13
Show the "replaced mode" only for e-bikes in a program

shankari:
Show the "replaced mode" only for e-bikes in a program

Based on feedback from program admins, and from NREL modelers who are using the
data, we don't really need the replaced mode for non-program interventions.
This also reduces user confusion ("what is the replaced mode?" is an FAQ)
and user burden.

Caveats:
- Currently, this is hardcoded to an e-bike intervention, pending
  implementation of https://github.com/e-mission/e-mission-docs/issues/755
- EDIT: ~This currently does not work for initial load; will be fixed in the next commit~ (fixed in https://github.com/e-mission/e-mission-phone/pull/874/commits/aba9dfb242fcb8d04f33030a9162d9c454472d1d)

Changes:
- change the inputDetails initialization to create a baseInputDetails in the
  case of programs. inputDetails will have all three options (MODE, PURPOSE,
  REPLACED_MODE), baseInputDetails will have only the two base options
- read them in the directive init as `fullInputDetails` (always exists) and
  `baseInputDetails` (only for programs)
- change the directive HTML to display `displayInputDetails`
- initialize `displayInputDetails` to `baseInputDetails` if it exists (program case)
    or `fullInputDetails` if it does not (study case)
- when the user chooses a value, we switch the `displayInputDetails` to
  `fullInputDetails` or `baseInputDetails` depending on the type of install
    (program vs. study) and the current mode.

Testing done:
- Select e-bike, changes to 3 labels
- Change purpose, stays at 3 labels
- Select walking, changes to 2 labels
- Change purpose, stays at 2 labels

### DIFF
--- a/www/js/survey/multilabel/multi-label-ui.js
+++ b/www/js/survey/multilabel/multi-label-ui.js
@@ -28,7 +28,7 @@ angular.module('emission.survey.multilabel.buttons',
   };
 })
 .controller("MultiLabelCtrl", function($scope, $element, $attrs,
-    ConfirmHelper, $ionicPopover, $ionicPlatform, $window, ClientStats, DynamicConfig, MultiLabelService) {
+    ConfirmHelper, $ionicPopover, $ionicPlatform, $window, ClientStats, DynamicConfig, MultiLabelService, Logger) {
   console.log("Created multilabel directive controller, waiting for init");
 
   var findViewElement = function() {
@@ -165,6 +165,28 @@ angular.module('emission.survey.multilabel.buttons',
       }
   };
 
+  var expandInputsIfNecessary = function(inputType, inputValue) {
+    if (inputType == "MODE") {
+        if (inputValue == "e-bike") {
+            if ($scope.displayInputDetails != $scope.fullInputDetails) {
+                Logger.log("Found e-bike mode in a program, displaying full details");
+                $scope.displayInputDetails = $scope.fullInputDetails;
+            } else {
+                Logger.log("Found e-bike mode in a study, already displaying full details", $scope.displayInputDetails == $scope.fullInputDetails);
+            }
+        } else {
+            if ($scope.baseInputDetails) {
+                Logger.log("Found non e-bike mode in a program, displaying base details");
+                $scope.displayInputDetails = $scope.baseInputDetails;
+            } else {
+                Logger.log("Found non e-bike mode in a study, nothing to change");
+            }
+        }
+    } else {
+        Logger.log("Non-mode change, ignoring");
+    }
+  }
+
   $scope.choose = function (inputType) {
     ClientStats.addReading(ClientStats.getStatKeys().SELECT_LABEL, {
       "userInput":  angular.toJson($scope.editingTrip.userInput),
@@ -180,6 +202,9 @@ angular.module('emission.survey.multilabel.buttons',
       isOther = true
       ConfirmHelper.checkOtherOption(inputType, checkOtherOptionOnTap, $scope);
     }
+    // special check for programs
+    // TODO: make this part of the dynamic config
+    expandInputsIfNecessary(inputType, $scope.selected[inputType].value);
     closePopover(inputType);
   };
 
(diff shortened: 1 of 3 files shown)


shankari:
@sebastianbarry here's the other change. Please ignore the comments before [52150ad](https://github.com/e-mission/e-mission-phone/pull/874/commits/52150ad4eb62fea8681eacb97cc5d61d871ec21a) - they are cherry-picked changes from master. As you can see from the file list, they are only fixing the commit history - they don't represent any code changes.

shankari:
Video of testing done:

https://user-images.githubusercontent.com/2423263/184472481-c7496a05-dee4-460e-b075-034e8501f89c.mov

shankari:
@sebastianbarry LMK if you have any cleanup comments and I will address them next week.

www/js/survey/multilabel/multi-label-ui.js
```diff
@@ -165,6 +165,29 @@ angular.module('emission.survey.multilabel.buttons',
       }
   };
 
+  var expandInputsIfNecessary = function(inputType, inputValue) {
+    console.log("Experimenting with expanding inputs for type "+inputType+" and value "+inputValue);
```
sebastianbarry, line 169: Is this console log still needed? It looks like it may be related to your debugging while implementing the change

www/js/survey/multilabel/multi-label-ui.js
```diff
@@ -180,6 +203,9 @@ angular.module('emission.survey.multilabel.buttons',
       isOther = true
       ConfirmHelper.checkOtherOption(inputType, checkOtherOptionOnTap, $scope);
     }
+    // special check for programs
+    // TODO: make this part of the dynamic config
+    expandInputsIfNecessary(inputType, $scope.selected[inputType].value);
```
sebastianbarry, line 208: would it be more clean to include the `    if (inputType == "MODE") {
        if (inputValue == "e-bike") {` statements within this block of code? as creating a new function call that is required to be called after every trip object may lead to security vulnerabilites, and unnecessary use of the stack. I'm interested as to why you chose to do it this way (having a function that gets called everytime, but it only affects the desired output if the internal if statements are met).

sebastianbarry: also, I am not sure what it means to "make it part of the dynamic config", if you could help me understand :)

sebastianbarry:
Otherwise this looks really good!! I like that you split the `baseInputDetails` and the `fullInputDetails`, that makes a lot of sense!

www/js/survey/multilabel/multi-label-ui.js
shankari: It is still useful to see what the values are, so that we can debug the behavior.
Note that this code was substantially refactored in subsequent PRs:
https://github.com/e-mission/e-mission-phone/pull/876
and
https://github.com/e-mission/e-mission-phone/pull/878

shankari:
@sebastianbarry please note that this was substantially refactored in two subsequent PRs:
https://github.com/e-mission/e-mission-phone/pull/876
and
https://github.com/e-mission/e-mission-phone/pull/878

that you might want to review as well to see how code is modified as necessary until it finally looks really really good 😄

www/js/survey/multilabel/multi-label-ui.js
shankari: wrt

> "may lead to security vulnerabilites"

I am not aware of any security vulnerability engendered by making a call to another function in the same file. Can you clarify?

wrt

> "unnecessary use of the stack"

it is absolutely true that there is additional overhead caused by a function call instead of inline code. However, with modern compilers, the penalty is vanishingly small. And pulling into out into a separate function helps with modularity and maintainability.

Concretely, it makes it easier to call from multiple locations (as in https://github.com/e-mission/e-mission-phone/pull/874/commits/aba9dfb242fcb8d04f33030a9162d9c454472d1d) without copy-pasting.

And when we moved the code around (as we did in the subsequent PRs), it was much easier to move this function than to potentially pull out inline code with unexpected side effects.

PR #876, opened 2022-08-13
Support filtering on base/expanded labels as well 

shankari:
The previous change worked fine (c0876131a550943a1a832838f7f541eab0297025,
aba9dfb242fcb8d04f33030a9162d9c454472d1d), and I even merged it before
realizing that it didn't work for filters.

For both the verify check code and the 'to label' filters, we used to iterate
through all the possible user inputs. For programs, that includes replaced mode.
But we don't show the replaced mode for non e-bike trips, so it is never filled
out, and we always show the trip as "To Label".

Fixing this required significant restructuring, or maybe finally bowing to
reality, since the computations for the filtering run outside the directive
controller. The only shared data structure is the trip, so we need to put the
desired INPUTs into the trip. And since some of the computation happens before
the trip is assigned to a directive, we need to assign the appropriate inputs
on load in the service. Note that we can potentially change this later as part
of the diary/label unification when we stop using `collection-repeat` and start
using `ng-repeat`.

Changes:
- move the `expandInputsIfNecessary` code to the service
- invoke it before computing verifiability while populating trip inputs
- change `expandInputsIfNecessary` to move the relevant input storage into the trip
    - trip INPUTS
    - trip inputDetails
- this allows us to remove the `$watch` on the trip

- remove the code to add the name to the input details based on the key by
  hardcoding during creation instead
- this allows us to remove all references to input details in the directive
  controller scope since it is now in the trip

- change the filters (`updateVerifiability` and `unlabeledCheck`) to use trip
  labels instead of static labels
- change the template to read input details from the trip

Testing done:
- Changed label to/from e-bike
- Confimed that it moved to/from the "To label" and "All Trips" screens

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -222,6 +222,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
   }
 
   $scope.recomputeDisplayTrips = function() {
+    console.log("recomputing display trips now");
     let alreadyFiltered = false;
     $scope.filterInputs.forEach((f) => {
         if (f.state == true) {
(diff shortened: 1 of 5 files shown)


shankari:
https://user-images.githubusercontent.com/2423263/184505212-f000cd80-a086-45dd-bb90-d142d0159481.mov

PR #877, opened 2022-08-13
Add e-car to the replaced mode as well

shankari:
This is a follow-on to 2cbcb8198e45ef6d426a3a3003d92fe2b065d262
but for the replaced mode as well
since the replaced mode is only in the `master_for_platform` branch


shankari:
@sebastianbarry should have fixed the replaced mode as well!

PR #879, opened 2022-08-23
Add display app version

sebastianbarry:
Created a section at the bottom of the Profile screen that displays the current app version

general-settings.js
- Added to the settings scope, clientAppVer which calls the ClientStats model to use the get app version function

style.css
- Created a new div class to display the app version

en.json
- Added an english translation for "App Version"

main-control.html
- Added div to display of the app version at the bottom
- Added app version translation into the HTML
- Used the new css style for the version number

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -707,6 +707,15 @@ background-image: none !important; }
   font-size:20px;
 }
 
+.control-version-number{
+  text-align: center;
+  float: right;
+  height: 100%;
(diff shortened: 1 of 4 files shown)


sebastianbarry:
style.css
- Changed text for version number from black to gray

shankari (approved):
LGTM!

Thanks for the detailed explanation of change and the screenshots.
Makes it much easier to review!

PR #884, opened 2022-09-27
Added very basic date picker button

sebastianbarry:
infinite_scroll_list.js
- Copied over functions from list.js, functions that relate to the date picker
- Sets current day as the current time in UTC within the getDataFromServer function

infinite_scroll_list.html
- Added an ng-click for the pickDay() function when the user clicks on the date in the diary i.e. "----- Tue 27 Sep 22 -----"
- Added a feedback for changing the date, for debugging purposes

https://user-images.githubusercontent.com/61334340/192608871-b1d09a91-2593-4d97-9422-c48b9aadbf30.mov

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -973,36 +973,6 @@ button.button.back-button.buttons.button-clear.header-item {
         top: -15px;
         box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23);
 }
-
-.start-time-tag-inf-scroll {
-        box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23);
-        text-align: center;
-        background-color: #80D0FF;
-        color: black;
-        width: 16%;
-        height: 18px;
-        border-radius: 8px;
-        font-size: 11px;
-        position: absolute;
-        left: 1%;
-        top: 15px;
-        line-height: 16px;
-
-}
-.stop-time-tag-inf-scroll {
-        box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23);
-        text-align: center;
-        background-color: #0088ce;
-        color: white;
-        width: 16%;
-        height: 18px;
-        border-radius: 8px;
-        font-size: 11px;
-        position: absolute;
-        left: 1%;
-        bottom: 15px;
-        line-height: 16px;
-}
 .stop-time-tag-lower { /* this is used in both lists (infinite_scroll_list.js & list.js) around the differentCommon function */
         box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23);
         text-align: center;
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -952,6 +952,21 @@ button.button.back-button.buttons.button-clear.header-item {
         bottom: -15px;
         box-shadow: 0 3px 6px rgba(0,0,0,0.16), 0 3px 6px rgba(0,0,0,0.23);
 }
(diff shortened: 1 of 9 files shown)


sebastianbarry:
As you can see it is not quite polished or finished by any means. Here's what I want to change about it:
- How I want it to work:
  -  I would for it to work like this: the user clicks on the date header for that day of trips (in this case they would click on the header that says " ------ Wed 22 Jul 15 ------ "); then, the datepicker popup comes up. They select the day and once they select a day the popup immediately closes and "skips" down the `infinite-scroll list` of trips to that day they chose (or the closest trip to that day in case there are no trips listed for that day)
- Problems with it right now:
  - Most noticeably, when you select a date nothing happens. I want it to skip/scroll/swipe down the infinite list to the date you picked
  - When you click on a date in the datepicker popup, you have to click set, and then close separately. I only want it to both set and close when the user clicks on the date (or at least close automatically when the user clicks `set`)
  - Whenever the datepicker popup opens up after selecting a date, it does set the currDay, however if you open the datepicker popup again it starts back again at the current date, instead of the last date you chose (or the closest date to the date you chose that contains trips)
  - Whenever the datepicker popup opens the first time, it starts with the current day, and not the day of the most recent trip in the list

@shankari I have a question: What container is holding the list of trips in this screen? I see there is `tripgj` which I believe is the object type for these trips, but what is the name of the list or container that is holding these?

sebastianbarry:
I was able to get the date picker to update the visible list of trips:

Testing:

https://user-images.githubusercontent.com/61334340/193933017-cd93367e-57e2-4c97-a227-b6ee458017d0.mov

There are some temporary changes in this commit, so this is not intended to be final product - I just pushed this commit out so you can see it @shankari 

1. I temporarily have it set to 11 days ago when we boot up - I have to change this to be the timestamp of the last logged trip
2. I temporarily have a big button which looks bad, to be the date picker button - I have to design and program the button to look better and fit in well to the UI as it currently is so it is familiar to users while also adding this new functionality

sebastianbarry:
I fixed the following issues:
- ~~When you choose a date and reopen the datepicker, it still always defaults to today, and not the currently selected day~~
- ~~When you choose a day, you still have to press "set"~~
- ~~When you set the day, data doesn't automatically update - you have to press the refresh button to get the data to populate~~
- I also fixed the issue that, the new variable I was using to populate the list of trips based on the date picker would not allow the original functionalities of the "to-label/all unlabeled/ all trips" buttons from working. Now, if you start to use the date picker the method it uses to populate is switched to work with the date picker; Otherwise it uses the original method to generate the trips (the date picker also works with the "all unlabeled/all trips" buttons too, I tested it)

Testing:

https://user-images.githubusercontent.com/61334340/194933148-a19e6735-511f-43af-8d89-9352f77de18a.mov


https://user-images.githubusercontent.com/61334340/194933163-49a2cae0-b48e-4f37-94dc-f591cfbf3923.mov

Issues:
- The fix I just implemented with the to-label/all unlabeled/all trips now looks more like a band-aid fix. Is this okay? @shankari Please take a quick look at my code in my last commit. I want it to implement nicely with the existing code but I also don't want it to take too long or change up the existing structure too much - **do you think by the way I wrote it that it fits this description**?
- If you load up the app for the first time, the selected datepicker input date is TODAY, however the loaded data will be for the most recent trip (with the test data, July 22 2016). This also translates over that - if you select the week of July 25 as your date and then scroll down to july 22 and click on that date picker button, it will still have July 25 highlighted, despite you just clicking on text that said "july 22". This isn't a huge problem, and may be overlooked, but it may be important to try to at least try to quickly fix

shankari:
This should be handled when we finish unifying diary and label screens.
let's hold off on this for now

sebastianbarry:
Date picker has been implemented in label screen, this PR is not longer valid

PR #887, opened 2022-09-29
Fixed diary screen checkmark box display issue

sebastianbarry:
For some reason the `class="control-icon-button"` was pushed to line 31, but not this line 30 so the display was incorrect
- changed class from "diary-button" (the incorrect style) to "control-icon-button" (the correct style)

I confirmed this is the fix for this issue by doing the following:
_Since I did not have enough unlabeled trips on my emulator (using the `test_july_22` test data) to trigger the `ng-if="!infScrollControl.reachedEnd"`, I set the `infScrollControl.reachedEnd` to always be `false` for my testing_

Here is how it was displaying before this change:
Screen Shot 2022-09-29 at 9 45 18 AM

Here is how it displays after this change:
Screen Shot 2022-09-29 at 9 49 22 AM


sebastianbarry:
@shankari This should fix the problem we were seeing with the weird button style displaying when `infScrollControl.reachedEnd = false`

PR #888, opened 2022-09-30
Fixing display issues for diary styles

sebastianbarry:
style.css
- Added placeholder styles to match the older versions of the stop-time-tag and start-time-tag which were changed in the unification process

diary_list_item.html
- Used the new temporary styles for start/stop time tag
- Changed the height size of the map to be smaller, to accompany the vertically stacked buttons for mode and purpose

### DIFF
--- a/www/templates/diary/diary_list_item.html
+++ b/www/templates/diary/diary_list_item.html
@@ -1,5 +1,5 @@
 <div>
-    <div class="start-time-tag">{{tripgj.display_start_time}}</div>
+    <div class="start-time-tag-diarytoremove">{{tripgj.display_start_time}}</div>
     <div  style="padding-left: 19%;">
 
     <ion-item id="diary-item" style="background-color: transparent;" class="list-item">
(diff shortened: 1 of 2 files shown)


sebastianbarry:
Testing:

Before:
image

After:
Screen Shot 2022-09-30 at 11 49 35 AM
Simulator Screen Shot - iPhone 11 - 2022-09-30 at 11 32 42

www/templates/diary/diary_list_item.html
```diff
@@ -31,7 +31,7 @@
         </div>
 
         <div class="row diary-arrow-container" ng-click="toDetail(tripgj.data.id)">
-            <leaflet class="col-90" geojson="tripgj" id="map_{{tripgj.data.id}}" defaults="defaults" height="150px"></leaflet>
+            <leaflet class="col-90" geojson="tripgj" id="map_{{tripgj.data.id}}" defaults="defaults" height="100px"></leaflet>
```
sebastianbarry, line 34: This change directly changes the leaflet element; could this be the reason that the maps are now loading with `ionic://...` and not `https://...`?

shankari: Correct me if I am wrong, but it looks like that only changes the element height. Not sure why it would change the map load prefix.

shankari (approved):

PR #889, opened 2022-10-03
Ensure that the map works properly on the iOS diary view as well

shankari:
The map tile location is configured in the `emission.services` module, `Config` service. Whenever we want to display a map, we read the tile location from the config and merge it with the existing map settings.

Before the refactor in 086dd85904f9be677ef2dd547f21d41119b36580, we displayed the list maps directly from `DiaryListCtrl`, so we configured it there.

As part of that refactor, we display the list map from the newly created `diaryListItem` directive, defined in `www/js/diary/diary_list_item.js`. However, the map tile configuration code was not moved in there.

This commit moves it in.

Testing done:
- without this change, the app tried to load maps from `ionic:://`, which would fail
- with this change, the app will try to load maps from `https://`, which succeeds The maps are now visible in the diary screen on iOS.

### DIFF
--- a/www/js/diary/diary_list_item.js
+++ b/www/js/diary/diary_list_item.js
@@ -4,7 +4,8 @@
 
 angular.module('emission.main.diary.diarylistitem', [
     'ui-leaflet',
-    'emission.main.diary.list'
+    'emission.main.diary.list',
+    'emission.services',
 ])
 
 .directive("diaryListItem", function(){
@@ -21,7 +22,7 @@ angular.module('emission.main.diary.diarylistitem', [
 .controller("DiaryListItemCtrl", function(
                                         $scope,
                                         SurveyOptions,
-                                        $state
+                                        $state, Config
 ){
     $scope.surveyOpt = SurveyOptions.MULTILABEL;
     const DEFAULT_ITEM_HT = 335;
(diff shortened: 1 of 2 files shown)


shankari:
@sebastianbarry please review this fix to https://github.com/e-mission/e-mission-phone/pull/888#issuecomment-1263862455

Testing done:

#### Before

Screen Shot 2022-10-03 at 1 40 25 AM


#### After

Screen Shot 2022-10-03 at 1 45 21 AM

sebastianbarry:
This looks good, I reviewed the code and I see what you did.

It looks like when Tyler made his changes to the Unified Diary Screen, he moved this code which you moved
```javascript
    angular.extend($scope, {
        defaults: {
            zoomControl: false,
            dragging: false,
            zoomAnimation: true,
            touchZoom: false,
            scrollWheelZoom: false,
            doubleClickZoom: false,
            boxZoom: false,
        }
    });

    angular.extend($scope.defaults, Config.getMapTiles())
});
```
To his directive, but never did the same for the diary screen (because he was anticipating this not being used)

One question I have @shankari is how did you find out that this was the issue? Did you look through the commits to see what Tyler did to display the maps, or did you just have a general idea of how it works and then went from there?

shankari:
Yup, the original plan was to replace the label by the diary, so @tyleryandt18 probably didn't put too much work into making sure that diary was still working well.

wrt debugging, it was pretty clear that the issue was with the base URL for the map. I vaguely remembered configuring the map URL in some consistent place, so I searched for "openstreetmap" in the codebase and found it in `Config`. I saw that it was configured in list.js as I had remembered, but not in the new directive. I then checked the label directive and confirmed that it was there. At that point, I was pretty sure I know what was going on and the fix was pretty trivial.

shankari:
wrt cleanup note also that if this was not a temporary fix, I should really remove the `Config` service dependency from `list.js` since we don't use it there any more. But since we will remove `list.js` in the future, I made the minimal change for now.

PR #892, opened 2022-10-12
fixed diary street name overlap

sebastianbarry:
main.diary.css
- added overflow styles to the diary-street css style which allows the ellipses ending to populate
- added a two-lines style that can be used in conjunction with diary-street to make it ellipses after 2 lines instead of just 1

trip_list_item.html
- added diary-street two-lines as the styles for both diary-street lines
- added a 100% width to the col-100 div so that it doesn't try to ellipses past the end of the div (which it was doing before without this 100% width)

### DIFF
--- a/www/templates/diary/trip_list_item.html
+++ b/www/templates/diary/trip_list_item.html
@@ -62,15 +62,15 @@
 
                     <!-- Row, containing Detail Screen button and Trip Origin/Destination -->
                     <div class="row" style="padding: 4px">
-                        <div class="col-100">
+                        <div class="col-100" style="width: 100%;">
                             <!-- Origin for Trip -->
-                            <div class="diary-street" ng-click="showDetail($event, trip)">
+                            <div class="diary-street two-lines" ng-click="showDetail($event, trip)">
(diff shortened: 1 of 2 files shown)


sebastianbarry:
Testing: (i used an iphone 7 so that the text displays bigger and forces the lines to be at least 3 lines long)

Simulator Screen Shot - iPhone 7 - 2022-10-12 at 13 10 44

www/templates/diary/trip_list_item.html
```diff
@@ -62,15 +62,15 @@
 
                     <!-- Row, containing Detail Screen button and Trip Origin/Destination -->
                     <div class="row" style="padding: 4px">
-                        <div class="col-100">
+                        <div class="col-100" style="width: 100%;">
                             <!-- Origin for Trip -->
-                            <div class="diary-street" ng-click="showDetail($event, trip)">
+                            <div class="diary-street two-lines" ng-click="showDetail($event, trip)">
```
shankari, line 67: I like the fact that you put this into a class instead of bulking up the style attribute!

shankari (approved):

PR #893, opened 2022-10-13
Fixed purpose dropdown text not being readable

sebastianbarry:
main.diary.css
- created a new shadow CSS element called two-lines which allows the purpose button to be 2 lines long and then ellipse if it extends past that

trip_confirm_options.json.sample
- Added spaces in front of any "/", this way the word isn't cut off in the purpose selection when it goes past 2 lines (for example, "Entertainment/Socia(next line)l")

mode-popover.html
- Added the use of the new CSS override two-lines

purpose-popover.html
- Added the use of the new CSS override two-lines


sebastianbarry:
Testing:


https://user-images.githubusercontent.com/61334340/195677976-e7973e16-6fa6-4d63-b7c5-b2acf70210f4.mov

www/json/trip_confirm_options.json.sample
```diff
@@ -41,12 +41,12 @@
        {"text":"Transit transfer", "value":"transit_transfer"},
        {"text":"Shopping","value":"shopping"},
        {"text":"Meal","value":"meal"},
-       {"text":"Pick-up/Drop off 👤","value":"pick_drop_person"},
-       {"text":"Pick-up/Drop off 📦","value":"pick_drop_item"},
-       {"text":"Personal/Medical","value":"personal_med"},
+       {"text":"Pick-up/ Drop off Person 👤","value":"pick_drop_person"},
+       {"text":"Pick-up/ Drop off Item 📦 and asdbvas absdva awebva","value":"pick_drop_item"},
```
sebastianbarry, line 45:
```suggestion
       {"text":"Pick-up/ Drop off Item 📦","value":"pick_drop_item"},
```

sebastianbarry: Missed this testing, removing it

sebastianbarry:
@shankari 
Here's my fix for the purpose popup, I also applied it to the mode popup since they work the same way, and it looks good for both

The other solution if you didn't like this as much, would be to increase the width of the popup, however I didn't like how it looked as much:
Simulator Screen Shot - iPhone 7 - 2022-10-13 at 12 34 33

shankari:
I am fine with this.

shankari:
What does it look like after one of these long strings is selected? Is the text still visible in the button or is it cut off?

sebastianbarry:
> What does it look like after one of these long strings is selected? Is the text still visible in the button or is it cut off?

When one of the long strings is selected (such as Pick-up/ Drop off Item 📦), the text gets cut off in the button. It will display like this in the button:
image

Should I fix this in this PR as well?

shankari:
> When one of the long strings is selected (such as Pick-up/ Drop off Item 📦), the text gets cut off in the button. It will display like this in the button:

Please fix this as well if possible. Might be hard to do if there is a lot of hardcoding of trip sizes.

shankari:
Also, we used the emojis (e.g. 📦) to fit within the previous popup width. If the width is not a consideration any more, 📦 can be replaced by the text "item"

sebastianbarry:
> Please fix this as well if possible. Might be hard to do if there is a lot of hardcoding of trip sizes.

> Also, we used the emojis (e.g. 📦) to fit within the previous popup width. If the width is not a consideration any more, 📦 can be replaced by the text "item"

I pushed out a change to fix the buttons, as well as remove the emojis

Testing:

Simulator Screen Shot - iPhone 11 - 2022-10-17 at 12 48 54
Simulator Screen Shot - iPhone 11 - 2022-10-17 at 12 48 57
Simulator Screen Shot - iPhone 11 - 2022-10-17 at 12 49 03

www/json/trip_confirm_options.json.sample
```diff
@@ -41,12 +41,12 @@
        {"text":"Transit transfer", "value":"transit_transfer"},
        {"text":"Shopping","value":"shopping"},
        {"text":"Meal","value":"meal"},
-       {"text":"Pick-up/Drop off 👤","value":"pick_drop_person"},
-       {"text":"Pick-up/Drop off 📦","value":"pick_drop_item"},
-       {"text":"Personal/Medical","value":"personal_med"},
+       {"text":"Pick-up/ Drop off Person","value":"pick_drop_person"},
+       {"text":"Pick-up/ Drop off Item","value":"pick_drop_item"},
```
shankari, line 45: @sebastianbarry you also need to fix the Spanish translations for this before we push to production

www/json/trip_confirm_options.json.sample
```diff
@@ -41,12 +41,12 @@
...
+       {"text":"Personal/ Medical","value":"personal_med"},
        {"text":"Access Recreation","value":"access_recreation"},
-       {"text":"Recreation/Exercise","value":"exercise"},
-       {"text":"Entertainment/Social","value":"entertainment"},
+       {"text":"Recreation/ Exercise","value":"exercise"},
+       {"text":"Entertainment/ Social","value":"entertainment"},
```
shankari, line 49: Ditto

shankari (approved):

shankari:
Merged into master first through cherry-picked commits in https://github.com/e-mission/e-mission-phone/pull/898
and then into master through https://github.com/e-mission/e-mission-phone/pull/901

Note that changes still need to be made in the Spanish translation as well.

www/json/trip_confirm_options.json.sample
sebastianbarry: I addressed this in my recent PR for e-mission-translate: https://github.com/e-mission/e-mission-translate/pull/19

www/json/trip_confirm_options.json.sample
sebastianbarry: I addressed this in my recent PR for e-mission-translate: https://github.com/e-mission/e-mission-translate/pull/19

PR #894, opened 2022-10-14
Fixed Label date header displaying wrong date if particular trips are labeled

sebastianbarry:
Before, when you had multiple days of trips in your label screen and labeled trips on the days on the inside but not on the outside (for example, I have trips from Mon, Tue, Wed and I only label the trips on Tue), the Tue trips would disappear but the Wed trips would have the label date header of "Tue", despite being Wed trips

This was because our logic would [for one single object] only check what today is, and the next day. If the trip's next day was different than this trip's day, then it would also display (after the trip card) the header for the next day, and then start displaying the adjacent unlabeled trips. This caused the wrong date header to display over a day of trips (for example, the Tue header would show up for Wed's day of trips)

The fix was to, instead of checking the next day in this object, check the previous day in the next object. That way, if you label trips for the previous day, it will still be different than the previously labeled trip, and the correct date will show


shankari:
Closing as a dup of https://github.com/e-mission/e-mission-phone/pull/895

PR #895, opened 2022-10-14
Fixed date separator displaying wrong date

sebastianbarry:
infinite_scroll_list.html
- Removed the trip.nextTrip.display_date div
- Changed the ng-show="$index==0" to an ng-if, so that I can concatenate the next change
- Added an OR to also display the date for a trip if the previous day's date is different than the current day's date

infinite_scroll_list.js
- Added an attribute to all trips called "prevTrip", which sets to undefined by default
- Got rid of the nextTrip attribute which is only set within that particular ctList.forEach((trip, tIndex) loop
- Replaced the nextTrip attribute set with a check: if it is the first trip in our list, it sets the prevTrip.display_date to its own display_date. Otherwise (all other trips) it sets the prevTrip object to the previous object in the list

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -109,7 +109,11 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
         Logger.log("Received batch of size "+ctList.length);
         ctList.forEach($scope.populateBasicClasses);
         ctList.forEach((trip, tIndex) => {
-            trip.nextTrip = ctList[tIndex+1];
+            if (tIndex == 0) {
+              trip.prevTrip = {display_date: trip.display_date};
+            } else {
+              trip.prevTrip = ctList[tIndex-1];
+            }
             $scope.labelPopulateFactory.populateInputsAndInferences(trip, $scope.data.manualResultMap);
         });
         // Fill places on a reversed copy of the list so we fill from the bottom up
@@ -320,6 +324,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
         // Pre-populate start and end names with &nbsp; so they take up the same amount of vertical space in the UI before they are populated with real data
         tripgj.start_display_name = "\xa0";
         tripgj.end_display_name = "\xa0";
+        tripgj.prevTrip = undefined;
     }
 
     const fillPlacesForTripAsync = function(tripgj) {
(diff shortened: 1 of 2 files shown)


sebastianbarry:
Before, when you had multiple days of trips in your label screen and labeled trips on the days on the inside but not on the outside (for example, I have trips from Mon, Tue, Wed and I only label the trips on Tue), the Tue trips would disappear but the Wed trips would have the label date header of "Tue", despite being Wed trips

This was because our logic would [for one single object] only check what today is, and the next day. If the trip's next day was different than this trip's day, then it would also display (after the trip card) the header for the next day, and then start displaying the adjacent unlabeled trips. This caused the wrong date header to display over a day of trips (for example, the Tue header would show up for Wed's day of trips)

The fix was to, instead of checking the next day in this object, check the previous day in the next object. That way, if you label trips for the previous day, it will still be different than the previously labeled trip, and the correct date will show

sebastianbarry:
Testing:
Part 1:

https://user-images.githubusercontent.com/61334340/195911997-3e0dffb0-006b-45da-8f76-c591cccced73.mov

Part 2:

https://user-images.githubusercontent.com/61334340/195912034-da0c3b6c-0b09-421f-9488-a67f5600ae24.mov

www/js/diary/infinite_scroll_list.js
```diff
@@ -109,7 +109,11 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
         Logger.log("Received batch of size "+ctList.length);
         ctList.forEach($scope.populateBasicClasses);
         ctList.forEach((trip, tIndex) => {
-            trip.nextTrip = ctList[tIndex+1];
+            if (tIndex == 0) {
+              trip.prevTrip = {display_date: trip.display_date};
+            } else {
+              trip.prevTrip = ctList[tIndex-1];
+            }
```
shankari, line 116: I can see adding `prevTrip`, but I am not sure why you want to remove `nextTrip`.
`nextTrip` is set for a reason - it is used in the user input matching code.

shankari: Given the amount of time that this change has been pending, I will just fix this as part of the merge.

shankari (approved):

shankari:
Merged into master first through cherry-picked commits in https://github.com/e-mission/e-mission-phone/pull/897
and then into master through https://github.com/e-mission/e-mission-phone/pull/901

Note that I have commented out part of this fix since it fails when there is data for only one day
https://github.com/e-mission/e-mission-phone/pull/899

And https://github.com/e-mission/e-mission-docs/issues/823 is tracking the pending change to fix that.

PR #896, opened 2022-10-21
Display the trip directly instead of displaying a copy generated afte…

shankari:
…r we get the trajectories

This ensures that we display at least part of the trip information even while waiting for the trajectory data

This is a test to see if we can fix
https://github.com/e-mission/e-mission-docs/issues/819


shankari:
Testing done:
Loading seems a bit better 


https://user-images.githubusercontent.com/2423263/197105014-d877a26c-e4a1-4526-99b3-9ffb790791ac.mov

PR #902, opened 2022-10-26
Multiple fixes to improve performance of the new label screen

shankari:
This is a snapshot before we start the restructuring with web workers.


shankari:
mapLimiter investigation:
https://github.com/e-mission/e-mission-docs/issues/819#issuecomment-1290771857
https://github.com/e-mission/e-mission-docs/issues/819#issuecomment-1290899187

PR #905, opened 2022-10-31
Iconized the start/stop in label screen

sebastianbarry:
main.diary.css
- Changed leaflet-div-icon-start/stop to have border-radius 100% instead of 6px, so that it displays as a circle instead of a rounded rectangle
- Added styles for the ICON portion of the leaflet div ionicon (the other styles are for the background circle of the icon)
- leaflet-div-ionicon-stop has margin styles applied so the weirdly-shaped ion-flag icon is more centered in the cirlce

services.js
- Changed the html for startIcon and stopIcon. Added an outer div which contains the css style, and an inner ionicon, start has a start icon, stop has a stop flag icon

trip_last_item.html
diary_list_item.html
infinite_scroll_detail.html
- Because we iconized the start/stop, i changed these icons to match and be consistent with the iconization

### DIFF
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -97,11 +97,26 @@ a.item-content {
 }
 .leaflet-div-icon-start {
   background-color: #80D0FF;
-  border-radius: 6px;
+  border-radius: 100%;
 }
 .leaflet-div-icon-stop {
   background-color: #0088ce;
-  border-radius: 6px;
+  border-radius: 100%;
+}
+.leaflet-div-ionicon {
+  background: transparent; 
+  color: white;
+  font-size: 100%;
+  justify-content: center; 
+  display: flex;
+}
+.leaflet-div-ionicon-start {
+  color: white;
+}
+.leaflet-div-ionicon-stop {
+  color: white;
+  margin-top: 1px;
+  margin-left: 3px;
 }
 .inner-icon {
   background-color: white;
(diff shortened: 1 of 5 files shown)


sebastianbarry:
Testing:

Simulator Screen Shot - iPhone 11 - 2022-10-31 at 10 45 45
Simulator Screen Shot - iPhone 11 - 2022-10-31 at 10 55 51

shankari:
Not sure that the location icon is the most intuitive for the start.
How about start and stop icons?
Something this for start: Screen Shot 2022-10-31 at 11 13 19 AM
and 
something like this for stop: image

sebastianbarry:
> Not sure that the location icon is the most intuitive for the start. How about start and stop icons? Something this for start: Screen Shot 2022-10-31 at 11 13 19 AM and something like this for stop: 

Good suggestions!

I had based my decision for the icon to use off of this image:

But I like that idea of using "start" and "stop" symbols to denote the start and stop of the trip. The only qualm I have with these logos is that they 1. resemble typically clickable buttons (which these are not) and 2. they are typically images used for audio/video play/stop instead of a location-based trip start/stop. Take for example google maps which looks like this:
image

or airplanes which look like this
image

I am okay with changing these to the icons you suggested, but perhaps we can have a discussion about what icons we should choose?

sebastianbarry:
Here is the cherry picked PR for merging to Master: https://github.com/e-mission/e-mission-phone/pull/907

shankari:
superceded by #907

PR #906, opened 2022-10-31
Removing prevTrip, displaying trip date on trip item

sebastianbarry:
infinite_scroll_list.js
- Removing prevTrip, because this was causing issues in the regression https://github.com/e-mission/e-mission-docs/issues/823 and we will replace it with another way to display the date to come

infinite_scroll_list.html
- Commenting out the hr-lines style that displays the trip date above the list of trips

trip_list_item.html
- Displaying the date from within the trip item
- Removed the previous styling that was unnecessary for this div since it previously didn't have any text inside it


sebastianbarry:
Testing:
Simulator Screen Shot - iPhone 11 - 2022-10-31 at 13 12 06

shankari (approved):

PR #907, opened 2022-10-31
Iconized the start/stop in label screen

sebastianbarry:
main.diary.css
- Changed leaflet-div-icon-start/stop to have border-radius 100% instead of 6px, so that it displays as a circle instead of a rounded rectangle
- Added styles for the ICON portion of the leaflet div ionicon (the other styles are for the background circle of the icon)
- leaflet-div-ionicon-stop has margin styles applied so the weirdly-shaped ion-flag icon is more centered in the cirlce

services.js
- Changed the html for startIcon and stopIcon. Added an outer div which contains the css style, and an inner ionicon, start has a start icon, stop has a stop flag icon

trip_last_item.html
diary_list_item.html
infinite_scroll_detail.html
- Because we iconized the start/stop, i changed these icons to match and be consistent with the iconization

### DIFF
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -97,11 +97,26 @@ a.item-content {
 }
 .leaflet-div-icon-start {
   background-color: #80D0FF;
-  border-radius: 6px;
+  border-radius: 100%;
 }
 .leaflet-div-icon-stop {
   background-color: #0088ce;
-  border-radius: 6px;
+  border-radius: 100%;
+}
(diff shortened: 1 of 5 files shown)


sebastianbarry:
Testing: 
Simulator Screen Shot - iPhone 11 - 2022-10-31 at 13 20 30
Simulator Screen Shot - iPhone 11 - 2022-10-31 at 13 20 35

shankari (approved):

PR #908, opened 2022-11-03
Remove loading overlay while loading trajectories

shankari:
Since we have already downloaded the trips and will be downloading the trajectories in the background


shankari:
Testing done:
- loaded the diary, only saw the initial loading screen while pulling trips
- added logs to the ionic bundle, saw

```
[Log] DEBUG:Turning on the ionic loading overlay in setupInfScroll
[Log] IONIC LOADING – {template: "Reading from server..."}
[Log] DEBUG:Turning on the ionic loading overlay in setupInfScroll
[Log] IONIC LOADING – {template: "Reading from server..."}
[Log] IONIC LOADING – {template: "Reading from server..."}
```

PR #911, opened 2022-11-16
Initial implementation of item height changes when the replaced mode …

shankari:
…is displayed

Changes by @sebastianbarry

- main.diary.css - Changed the way the map displays, so that it is variable and changes with the height of the trip item - Changed the way the diary card displays as min-content, dynamic, instead of a static hard-coded pixel value

- infinite_scroll_list.js - Added a function that returns a hard-coded pixel height value for the trip item in the template's collection-repeat

- infinite_scroll_list.html - Added the dynamic collection-repeat "item-height" attribute that is calling the function we made in the controller

- trip_list_item.html - Added some styling to the elements so that they display properly with the dynamic height now

Changes by @shankari

- Find the closest ancestor scroll element if present
- Cache it for future use
- Call scroll-resize on it

Testing done:
https://github.com/e-mission/e-mission-docs/issues/831#issuecomment-1316411689

Pending testing:
- test in all locations where we use the linked survey - at least infinite scroll, diary, diary detail and infinite scroll detail - also include others found through a recursive search

Pending improvement:
- only trigger the resize if the original value or the new value is e-bike

Testing done:
Connected via the inspector and verified that the item height changed when the mode was changed to/from e-bike. The distance between items also remained the same, unlike with the prior implementation where the card size changed, but the item size did not, so the gap between items narrowed.

### DIFF
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -47,7 +47,7 @@
   and (min-device-width: 320px) 
   and (max-device-width: 568px)
   and (-webkit-min-device-pixel-ratio: 2) {
-    .angular-leaflet-map { height: 150px; width: 245px; }
+    .angular-leaflet-map { position: absolute; height: 82.8%; width: 46.8%; }
 }
 
 /* ----------- iPhone 6 ----------- */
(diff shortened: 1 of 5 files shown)


shankari:
Superceded by #912

PR #913, opened 2022-11-21
feat: a time-use UI

asiripanich:


shankari:
@sebastianbarry can you pull this and start seeing what it already does? We can then have a meeting with @asiripanich on what their goals for the project are.

asiripanich:
> @sebastianbarry can you pull this and start seeing what it already does? We can then have a meeting with @asiripanich on what their goals for the project are.

Sounds good to me. Happy to meet again this week too.

sebastianbarry:
Update for today: I see that the time-use is already created as an angular module. I made some changes to try to get it to work, but I'm just making progress getting everything (modules) connected and loaded

--------

Here is how the app looked as is when I copied the branch:

| | | |
:---:|:---:|:---:
| Simulator Screen Shot - iPhone 11 - 2022-12-07 at 10 32 40 | Simulator Screen Shot - iPhone 11 - 2022-12-07 at 10 32 36 | Simulator Screen Shot - iPhone 11 - 2022-12-07 at 10 32 42 |
| Simulator Screen Shot - iPhone 11 - 2022-12-07 at 10 33 12 | Simulator Screen Shot - iPhone 11 - 2022-12-07 at 10 34 21 | Simulator Screen Shot - iPhone 11 - 2022-12-07 at 10 34 30 |

-------

I added a time-use tab to the `main.html`, so that I can test out the template and get an idea of what the time-use tab will look like:
```html
  <!-- Time Use Tab -->
  <ion-tab title="{{'main-time-use.tab' | translate}}" icon="ion-map" href="#/root/main/time_use">
    <ion-nav-view name="time-use"></ion-nav-view>
  </ion-tab>
```

I also loaded the files into `index.html`, and here is where I left off...

Screen Shot 2022-12-07 at 12 48 54 PM

shankari:
> I see that the time-use is already created as an angular module

It is created as a module, but only in the javascript. the HTML does not use modules aka directives. I think that the HTML is just a copy-paste of the old list view **and should be changed to use directives** after you get it to work in the first place.

wrt the error you are getting: I think I changed the module names as part of a big refactoring into directives. Check out the enketo directives at the locations that I listed above - the new module names should be pretty similar...

sebastianbarry:
I was able to all the enketo modules loaded properly without any errors. The next problem I'm facing is the time-use page is not loading:

Screen Shot 2022-12-09 at 9 34 35 AM

Surprisingly there are no errors, so I am not sure exactly why the page isn't loading, or how to debug this.

-------

I figured the problem exists in one of 2 places:
1. In the way the `state` for time-use is written/loaded
2. There is some error in the `template` or `controller` for time-use that is not writing an error message but is preventing the page from loading

I first tried making the state load the diary template and controller, and I tried to copy the diary.js state file almost exactly:

image

Then I tried completely wiping out the controller and templates, making them as barebones as I could, but that didn't fix it either:

image

-----

What's more surprising, is neither of my attempts resulted in there being error messages in the console:

image

This is what it looked like for both

shankari:
@sebastianbarry can you push your changes so far (to the point where there are no errors but nothing is loading)?

sebastianbarry:
I pushed my changes which can be seen here in this commit: https://github.com/e-mission/e-mission-phone/commit/e6077f89203401251afb15e9d3798998f3dde7e5

sebastianbarry:
I got it working!

I have opened up my pull request for this same topic, from my branch (not sure how to contribute to this PR)

https://github.com/e-mission/e-mission-phone/pull/915

shankari:
Superceded by #915

PR #914, opened 2022-11-29
Adding description for error 403

sebastianbarry:
If the error text is exactly as "During server call, error The operation couldn’t be completed. (com.google.HTTPStatus error 403.)", then the error is prepended with "Errr: OPcode not found."

This was made specifically for the error checking in the following files:
- services.js (reading confirmed trips)
- metrics.js (Error loading user data)
- storedevicesettings.js (Error in updating profile to store device settings)

### DIFF
--- a/www/js/services.js
+++ b/www/js/services.js
@@ -3,7 +3,7 @@
 angular.module('emission.services', ['emission.plugin.logger',
                                      'emission.plugin.kvstore'])
 
-.service('CommHelper', function($rootScope) {
+.service('CommHelper', function($rootScope, Logger) {
     var getConnectURL = function(successCallback, errorCallback) {
         window.cordova.plugins.BEMConnectionSettings.getSettings(
             function(settings) {
@@ -15,16 +15,37 @@ angular.module('emission.services', ['emission.plugin.logger',
             window.cordova.plugins.BEMServerComm.getUserPersonalData("/profile/create", successCallback, errorCallback);
     };
 
+    var processErrorMessages = function(errorMsg) {
+      var titleMsg = "Error"
+      if (errorMsg.includes("403")) {
+        titleMsg = "Error: OPcode not found";
+        errorMsg = "Error: OPcode not found. " + errorMsg;
+      }
+      console.error(errorMsg);
+      Logger.displayError(titleMsg, errorMsg);
+      return errorMsg;
+    }
+
     this.updateUser = function(updateDoc) {
-        return new Promise(function(resolve, reject) {
+        var updateUserPromise = new Promise(function(resolve, reject) {
             window.cordova.plugins.BEMServerComm.postUserPersonalData("/profile/update", "update_doc", updateDoc, resolve, reject);
         });
+        updateUserPromise.catch(error => {
+          error = "While updating user, " + error;
+          error = processErrorMessages(error);
+        });
+        return updateUserPromise;
     };
 
     this.getUser = function() {
-        return new Promise(function(resolve, reject) {
+        var getUserPromise = new Promise(function(resolve, reject) {
             window.cordova.plugins.BEMServerComm.getUserPersonalData("/profile/get", resolve, reject);
         });
+        getUserPromise.catch(error => {
+          error = "While getting user, " + error;
+          error = processErrorMessages(error);
+        });
+        return getUserPromise;
     };
 
     this.putOne = function(key, data) {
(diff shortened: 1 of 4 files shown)


sebastianbarry:
Removed the previous changes to the commit, and only made changes to services.js

www/js/services.js
```diff
@@ -15,15 +15,35 @@ angular.module('emission.services', ['emission.plugin.logger',
             window.cordova.plugins.BEMServerComm.getUserPersonalData("/profile/create", successCallback, errorCallback);
     };
 
+    var processErrorMessages = function(errorMsg) {
+      var titleMsg = "Error"
+      if (errorMsg.includes("403")) {
+        titleMsg = "Error: OPcode not found";
```
shankari, line 21: why do you need the titleMsg? Doesn't seem to be used anywhere.

www/js/services.js
```diff
@@ -15,15 +15,35 @@ angular.module('emission.services', ['emission.plugin.logger',
...
+      }
+      return errorMsg;
+    }
+
     this.updateUser = function(updateDoc) {
         return new Promise(function(resolve, reject) {
             window.cordova.plugins.BEMServerComm.postUserPersonalData("/profile/update", "update_doc", updateDoc, resolve, reject);
+        })
+        .catch(error => {
+          error = "While updating user, " + error;
```
shankari, line 33: Do you need the "While updating user" etc comments? The popup already has a description of when we got the error, right?

shankari (changes requested):
Bunch of minor changes.

shankari:
@sebastianbarry can you finish addressing my comments on this before you leave?

www/js/services.js
```diff
@@ -3,7 +3,7 @@
 angular.module('emission.services', ['emission.plugin.logger',
                                      'emission.plugin.kvstore'])
 
-.service('CommHelper', function($rootScope) {
+.service('CommHelper', function($rootScope, Logger) {
```
sebastianbarry, line 6:
```suggestion
.service('CommHelper', function($rootScope) {
```

Logger is no longer needed here

www/js/services.js
sebastianbarry, line 21:
```suggestion
      if (errorMsg.includes("403")) {
```

www/js/services.js
sebastianbarry: To keep it standard with all the other function error catching blocks, I think this might be helpful.

More error message info is better than less.

I will test one case of the error 403 real quick (not necessarily the `updatingUser` function) to see how it displays with the error message information and then make a decision with whether it's helpful or not

www/js/services.js
```diff
@@ -58,8 +88,13 @@ angular.module('emission.services', ['emission.plugin.logger',
      * Other fields can be added easily and the server can be modified at the same time.
      */
     this.habiticaRegister = function(regConfig) {
-        return new Promise(function(resolve, reject){
+      return new Promise(function(resolve, reject){
```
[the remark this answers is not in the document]
sebastianbarry: I think this specific catch block may not be necessary, as I can't think of a scenario where you would get to the `habiticaRegister` function with an error 403, _without already having triggered an error 403 before that in some capacity_

In other words, I will remove any changes to this function as it: 
> creates unnecessary clutter

www/js/services.js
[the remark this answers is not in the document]
sebastianbarry: Removed in https://github.com/e-mission/e-mission-phone/pull/914/commits/d1390cdb1ebfd370cec46a564c369c749c11ed11

www/js/services.js
```diff
@@ -15,15 +15,35 @@ angular.module('emission.services', ['emission.plugin.logger',
             window.cordova.plugins.BEMServerComm.getUserPersonalData("/profile/create", successCallback, errorCallback);
     };
 
+    var processErrorMessages = function(errorMsg) {
+      var titleMsg = "Error"
+      if (errorMsg.includes("403")) {
+        titleMsg = "Error: OPcode not found";
+        errorMsg = "Error: OPcode not found. " + errorMsg;
+        console.error("Error 403 found. " + errorMsg);
```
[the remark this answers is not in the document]
sebastianbarry: I think a better wording would be `Error: OPcode does not exist on the server`, what do you think @shankari ?

www/js/services.js
sebastianbarry: Removed in https://github.com/e-mission/e-mission-phone/pull/914/commits/a590bf6409a61d4f0a3cb999079cce9302c4a99b

sebastianbarry:
Once I finish my testing and post my first "final" draft for this change, I will recreate the PR to the `may_2023_point_release` branch

@shankari on a higher level, what do you think of this solution (in terms of using the `processErrorMessages` function which checks to see if they error message string contains `"403"` and altering the error message before displaying it?

shankari:
> @shankari on a higher level, what do you think of this solution (in terms of using the processErrorMessages function which checks to see if they error message string contains "403" and altering the error message before displaying it?

@sebastianbarry I would have preferred a change in the native code. But given that the only person who can currently make native code changes is me, this is fine from a javascript-oriented solution.

sebastianbarry:
This PR has been superseded by https://github.com/e-mission/e-mission-phone/pull/960

PR #916, opened 2022-12-19
consolidate filter buttons into dropdown <select>

JGreenlee:
the corresponding JS functions select() and resetSelection() have also been consolidated into updateSelection()

preview on iOS:
Screen Shot 2022-12-19 at 5 56 21 PM

android:
Screen Shot 2022-12-19 at 5 58 46 PM

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -54,6 +54,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
     f.state = false;
   });
   $scope.filterInputs[0].state = true;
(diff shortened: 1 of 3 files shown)


shankari (approved):
Looks good to me at a high level

shankari:
@JGreenlee this seems to generate merge conflicts with the changes we made to the label screen for the trip/place features. I'm going to remove these changes by force-pushing.

Can you generate a new PR against the new master after merging + testing?
Or can @sebastianbarry take that over?

shankari:
Argh! I can't force push because these were committed too long ago; will just have to test and hope that the recomputation works. If it doesn't, it is still not a showstopper!

PR #919, opened 2023-02-10
Places survey Part 2

sebastianbarry:


JGreenlee:
This video shows the behavior we should expect with the trip confirm button:

https://user-images.githubusercontent.com/15843932/218181145-12248a3e-33dc-4f3c-97f9-6aacffba4100.mp4


- Previous entries can be edited and the button label will update
- New entries can be added and the label shows
- All entries persist after refresh

shankari:
@JGreenlee we should actually revisit the metrics-mappings overall. If the survey is not the multilabel but is a fully configurable survey, I am not sure that we will have the correct energy intensity mappings for the dashboard. We might want to disable the dashboard completely in that case, or use only the sensed values.

JGreenlee:
@shankari Is looking into that a task you want me to prioritize? Or just something to keep in mind?

shankari:
Just keep it in mind for now. We should check with @asiripanich for some feedback on what to do since they are the primary users of the enketo survey instead of multilabel

JGreenlee:
To get place-level additions working, I have slightly revised the input matching functions (on [composite-trip-prep](https://github.com/sebastianbarry/e-mission-phone/tree/composite-trip-prep), to merge here soon). I want to make sure I document this in case it comes up later.

Previously, we only looked at trips to match inputs. The algorithm also looked ahead at the `nextTrip` for checking end timestamps.
Now, inputs can be matched to an entry that is either a trip or a place. Instead of `nextTrip` we will look ahead at the `nextEntry`, which is not necessarily a trip. If the current entry is a trip, `nextEntry` is its end place. If current is a place, `nextEntry` is the following trip.
This means that the end/exit of an entry will be the same as the start/enter of its `nextEntry`.

This necessitates a slight change in the matching algorithm -- instead of the addition start being **less than or equal** to the trip/place end/exit, it _must_ be **less**.
That is, an addition cannot start at the end/exit of a trip/place and be matched to it. The addition must start _before_.

The same changes will be mirrored on the server-side matching functions.

shankari:
@sebastianbarry @JGreenlee Is it possible to split this PR into two:
1. That only has the trip addition fixe (I think this is everything until https://github.com/e-mission/e-mission-phone/pull/919/commits/8c1cd86693a54a298f236c4e827f57fc7d6b0a97). This will let me push that out to TestFlight that I can push it out to TestFlight?
2. The composite trip stuff should be in a separate PR since we are still working on it and I should review the changes

Would be great to have that done ASAP so that I can finish the first review and push out to TestFlight.

JGreenlee:
In that case, the 1st PR (for trip fixes) should also include 1ae141778ec568342060e349f63bfa9cbeaa1c90 - but nothing after that

shankari:
@JGreenlee I don't think that changing `nextTrip` to `nextObject` will work correctly. We do really want to check the overlap with the next trip for the trip matching case.

I know that it appears to work now but that's because you are only working with processed trips, not draft trips.
If you check the blame for the change, you can see the reason in the commit message
https://github.com/e-mission/e-mission-server/commit/986a5340f3c38cb3e3c9d2f9c3c7cc287efab14c

I bet this doesn't pass the unit tests on the server

JGreenlee:
created "Trip addition fixes" #926

shankari:
So to test with draft trips, you want to first load data for day1, and then run the pipeline as you have been doing so far, and then load data for day 2 where day 2 is > day 1 and *not* run the pipeline. Trips from day 2 will show up as draft in the diary.

draft trips are:
(1) trips that are on the phone and haven't been pushed up *and*
(2) trips that are pushed up but not yet processed

This would be (2)

JGreenlee:
@shankari I have been able to run the unit tests locally and I am slowly progressing on revising the input matching.

I noticed that one of the test conditions is a trip with start 5 and end 6. The user input has start 6 and end 7.
Why should we expect this to match? The user input did not even start until the trip was over.

If inputs are going to be matched to places, then wouldn't this user input belong on the trip's *end place*, not the trip itself?

Is it safe to say that we will no longer match cases like this?

shankari:
@JGreenlee I'm looking at `tests.storageTests.TestTripQueries` and I don't see any trip matching for start 5 and end 6.
I searched for `= 5` and I only find `testQuerySectionsForTrip` and `testQueryStopsForTrip`, neither of which have anything to do with the user input.

Can you clarify which test you are seeing this in?

JGreenlee:
@shankari Considering `testUserInputForTripOneInput()`:

We call `create_fake_trip()` (if you trace this back, it ultimately returns a trip with start_ts of '5' and end_ts of '6')

Then,
```
new_mc["start_ts"] = new_trip.data.start_ts + 1
        new_mc["end_ts"] = new_trip.data.end_ts + 1
```
I believe this is how we get 6 and 7 for the user input start and end

When I debug the input matching function (`validUserInputForTimelimeEntry()`), this is what I'm faced with: trip start = 5, trip end = 6, userinput start = 6, userinput end = 7.

The test expects a match; if I return False in this case, the test fails

And the other test, I believe, is `testUserInputForTripTwoInputs()`. same scenario

sebastianbarry:
I noticed a bug with https://github.com/e-mission/e-mission-phone/pull/919/commits/f989d9e31f6e847b2602ba92a7cde280a9dd2ae2 on this line:
```javascript
$scope.getPrefillTimes = () => {
    const begin = $scope.timelineEntry.start_fmt_time || $scope.timelineEntry.enter_fmt_time;
    const stop = $scope.timelineEntry.end_fmt_time || $scope.timelineEntry.exit_fmt_time;

    const momentBegin = moment.parseZone(begin);
    let momentStop = moment.parseZone(stop);
    // stop could be undefined because the last place will not have an exit time
    if (!stop) {
      // if begin is the same day as today, we will use the current time for stop
      // else, stop will be begin + 1 hour
      if (moment(begin).isSame(moment(), 'day')) {
        momentStop = moment.parseZone();
      } else {
        momentStop = moment(momentBegin).add(1, 'hour');
      }
    }

    return {
      "Date": momentBegin.format('YYYY-MM-DD'),
      "Start_time": momentBegin.format('HH:mm:ss.SSSZ'),
      "End_time": momentStop.format('HH:mm:ss.SSSZ')
    }
  }
```

When the user clicks on either the places addition or trip addition button, this code runs to prefill the time bounds and date. But since `momentStop` doesn't get filled in `if (!stop) {...`, it is an undefined object and cannot use `.format`

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -42,23 +30,37 @@ angular.module('emission.survey.enketo.add-note-button',
...
+  // return a dictionary of fields we want to prefill, using start/enter and end/exit times
   $scope.getPrefillTimes = () => {
-    const timeBounds = $scope.timeBounds();
-    const begin = timeBounds.start_fmt_time || timeBounds.enter_fmt_time;
-    const stop = timeBounds.end_fmt_time || timeBounds.exit_fmt_time;
+    const begin = $scope.timelineEntry.start_fmt_time || $scope.timelineEntry.enter_fmt_time;
+    const stop = $scope.timelineEntry.end_fmt_time || $scope.timelineEntry.exit_fmt_time;
+
+    const momentBegin = moment.parseZone(begin);
+    let momentStop;
```
sebastianbarry, line 48:
```suggestion
    let momentStop = moment.parseZone(stop);
```

In implementation this causes this result: 
image

Where all the trips and place additions work properly, except the last place addition which prefills the stop_time as 1 hour after the start time. Does this sound correct?

JGreenlee: Please see the comment in the code right below this

In real-life usage, the last place will represent the place you're currently at. We will prefill with the current time

However, for dev testing we just load old test trips, and the current time can seem inappropriate. We could end up with an end time that is earlier than the start time, which would be invalid

So, I implemented a quick check for whether the start time is the same as today. If not, we just use start time + 1 hour.

sebastianbarry:
## TESTING:
All testing uses the following configuration
https://github.com/sebastianbarry/e-mission-phone/tree/places-survey
https://github.com/sebastianbarry/e-mission-server/tree/add-composite-trip

### Test 1:
iPhone 11 emulator
Program: `{name: "Devapp Program (Tripconfirm survey + Timeuse Survey)", joinurl: "emission://join_study?label=dev-emulator-program-tripconfirm&source=github" }`

| Trial | Notes |
:---:|:---:
| Trial 1: Load in jul-25-2016 data, label both trips and add timeuse to both trips | I am not able to open the timeuse survey button without having this code change https://github.com/e-mission/e-mission-phone/pull/919#discussion_r1133308042 . After having that change, the survey opens up just fine and i am able to add timeuse notes into my app which worked fine. I then moved on to trial 2 | 
| Trial 2: Load in jul-27-2016 data | After loading in the jul-27 data, I ran the pipeline and then pressed the "force trips + sync" button, and a significant error occurred. When I tried to load in data from the server the app becomes unresponsive, and upon app reload the app remains unresponsive at this point: Screen Shot 2023-03-14 at 11 58 00 PM |
| Trial 3: Load in jul-29-2016 data | I was not able to test this due to the above issue |
| Trial 4: Run the pipeline multiple times, to simulate multiple hours/days going by | I was not able to test this due to the above issue |


### Test 2:
Android Pixel (Google Playstore API) emulator
Program: `{name: "Devapp Program (Tripconfirm survey + Timeuse Survey)", joinurl: "emission://join_study?label=dev-emulator-program-tripconfirm&source=github" }`

| Trial | Notes |
:---:|:---:
| Trial 1: Load in jul-25-2016 data, label both trips and add timeuse to both trips | Same issue is happening here, here is a screenshot of it: Screen Shot 2023-03-15 at 12 19 39 AM | 
| Trial 2: Load in jul-27-2016 data | After loading in the jul-27 data, I ran the pipeline and INSTEAD of pressing the "force trips + sync" button, the same significant error occurred again. When I tried to load in data from the server the app becomes unresponsive, and upon app reload the app remains unresponsive at the same point |
| Trial 3: Load in jul-29-2016 data | I was not able to test this due to the above issue |
| Trial 4: Run the pipeline multiple times, to simulate multiple hours/days going by | I was not able to test this due to the above issue |


### Test 3:

Let's address the previous issue before moving onto more testing...

shankari:
@sebastianbarry can you please resolve the merge conflicts by pulling from `refactor_survey_expansion` before testing?

sebastianbarry:
> can you please resolve the merge conflicts by pulling from `refactor_survey_expansion` before testing?

After resolving the merge conflicts and pulling the new change, the same error as described above is still occurring: "After loading in the jul-27 data, I ran the pipeline and then pressed the "force trips + sync" button, and a significant error occurred. When I tried to load in data from the server the app becomes unresponsive, and upon app reload the app remains unresponsive at this point"

www/js/survey/enketo/enketo-notes-list.js
```diff
@@ -33,7 +33,8 @@ angular.module('emission.survey.enketo.notes-list',
     }
 
     $scope.setDisplayTime = function(entry) {
-      const timezone = $scope.timezone;
+      const timezone = $scope.timelineEntry.start_local_dt?.timezone
```
sebastianbarry, line 36: To confirm: We are doing this because if we were to use `$scope.timezone`, the display time would be in the device's local time rather than the local time to the TimelineEntry (i.e. if the TimelineEntry was initially tracked in Hawaii but the user is now in Colorado, the display time will still be in Hawaii time)?

Is `start_local_dt` the best option for determining time zone? I believe this is good because in testing with Hawaii TimelineEntries I haven't seen any issues, but I just want to get input on this.

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -42,23 +30,37 @@ angular.module('emission.survey.enketo.add-note-button',
 
   const updateLabel = () => {
     const localeCode = $translate.use();
-    if ($scope.notesConfig?.['filled-in-label'] && trip.tripAddition?.length > 0) {
+    if ($scope.notesConfig?.['filled-in-label'] && timelineEntry.additions?.length > 0) {
```
sebastianbarry, line 33: Question for @shankari, should it be standard for every program that if they are using notes-addition buttons, they MUST require a `'not-filled-in-label'` and they only need to require a `'filled-in-label'` if they want to?

Is it safe to assume this as the determining factor for whether we will display `'filled-in-label'`?

www/js/survey/enketo/enketo-add-note-button.js
sebastianbarry: I see the comment and the reasoning makes sense.

However in testing I still do need to have the above recommended change, as the add-notes-button will not work without `let momentStop = moment.parseZone(stop);`, since the return statement uses `momentStop.format('HH:mm:ss.SSSZ')` which will not be able to function if `momentStop` is not set to a value.

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -147,44 +149,43 @@ angular.module('emission.survey.enketo.add-note-button',
...
-  enbs.populateManualInputs = function (trip, nextTrip, inputType, inputList) {
+  enbs.populateManualInputs = function (timelineEntry, nextTimelineEntry, inputType, inputList) {
       // Check unprocessed labels first since they are more recent
-      const unprocessedLabelEntry = InputMatcher.getAdditionsForTrip(trip, inputList);
+      const unprocessedLabelEntry = InputMatcher.getAdditionsForTimelineEntry(timelineEntry, inputList);
       var userInputEntry = unprocessedLabelEntry;
-      if (userInputEntry.length == 0) {
-          userInputEntry = trip.trip_addition;
+      if (!angular.isDefined(userInputEntry)) {
+          userInputEntry = timelineEntry.timelineEntry_addition[enbs.inputType2retKey(inputType)];
```
sebastianbarry, line 175: Where is the property `timelineEntry_addition` coming from? I do not see any other references to this object when searching the e-mission-phone:places-survey code. Is it coming from the server?

www/js/diary/services.js
```diff
@@ -769,38 +774,27 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
...
-            features: features,
-            properties: { 
-              start_ts: trip.start_ts,
-              end_ts: trip.end_ts
-            }
-          }
-          return trip_gj;
-        }).catch((err) => {
-          Logger.displayError("while filling details", err);
-        });
```
sebastianbarry, line 803: Is there a reason why we removed this catch?

www/js/diary/services.js
```diff
@@ -769,38 +774,27 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
...
-        return Promise.resolve(undefined);
+        return undefined;
       }
-      Logger.log("About to pull location data for range "
-        + moment.unix(trip.start_ts).toString() + " -> " 
-        + moment.unix(trip.end_ts).toString());
-
-        const fillPromises = [
-            CommHelper.getRawEntries(["analysis/recreated_location"], trip.start_ts, trip.end_ts, "data.ts", 100)
-        ];
```
sebastianbarry, line 782: To confirm: We are not calling `CommHelper.getRawEntries` from `confirmedTrip2Geojson` anymore because the returned object from the server already contains a truncated version of its rawEntries, correct?

www/js/diary/services.js
```diff
@@ -432,21 +432,26 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
...
             return ctList.phone_data.map((ct) => {
-                const retVal = ct.data;
-                retVal.id = ct._id["$oid"];
-                return retVal;
+              ct.data._id = ct._id["$oid"];
+              const cp = ct.data.confirmed_place;
+              if (cp) {
+                cp.data._id = cp._id["$oid"];
+                delete ct.data.confirmed_place;
+                ct.data.confirmed_place = cp.data;
```
sebastianbarry, line 452: I am not familiar with the `delete` keyword, but how come we are deleting `ct.data.confirmed_place` first, and then immediately using it again? Won't it be deleted?

www/js/diary/infinite_scroll_list.js
```diff
@@ -143,22 +154,34 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
         });
-        // Fill places on a reversed copy of the list so we fill from the bottom up
+        console.time("fillTrajectories")
+        // Fill place names and trajectories on a reversed copy of the list so we fill from the bottom up
         ctList.slice().reverse().forEach(function(trip, index) {
             fillPlacesForTripAsync(trip);
-            fillTrajectoriesForTripAsync(trip);
+            fillTrajectoriesForTrip(trip);
         });
+        console.timeEnd("fillTrajectories")
```
sebastianbarry, line 181: Are we intending on leaving the `console.time` for the final product?

www/js/diary/infinite_scroll_list.js
```diff
@@ -267,30 +290,40 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
+    
+    $scope.data.displayTimelineEntries = []
+    $scope.data.displayTrips.forEach((t) => {
+      const place = t.confirmed_place;
+      const {confirmed_place, ...trip} = t;
+      $scope.data.displayTimelineEntries.push(trip);
+      if ($scope.showPlaces && place) {
+        $scope.data.displayTimelineEntries.push(place);
+      }
+    });
```
sebastianbarry, line 326: Putting in this comment for reference: we may want to investigate this for the issue where "after labeling trips with enketo survey and then adding trip/place addition notes -> then going on another trip -> then running the pipeline -> app becomes hung up and unresponsive"

www/js/diary/infinite_scroll_list.js
```diff
@@ -386,31 +424,33 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
                 tripgj.end_display_name = endName;
+                if (tripgj.confirmed_place) {
+                  tripgj.confirmed_place.display_name = endName;
+                }
             });
         });
     }
 
-    const fillTrajectoriesForTripAsync = function(trip) {
-        mapLimiter.schedule(() => Timeline.confirmedTrip2Geojson(trip)).then((tripgj) => {
```
sebastianbarry, line 394: We are still instantiating `mapLimiter` for this component, do we not need to use it during `fillTrajectoriesForTrip` anymore? If so then we may consider removing mapLimiter from above

sebastianbarry (commented):
First round of comments for the PR. There is such good thought put into this PR, most of the code looked amazing. While there is a lot of code changes purely centered around the object rename of `trips` to `TimelineEntries`, I added comments on some code segments I think can easily be removed from debugging, some questions for myself about what the code does, some though-provoking questions as to why we chose to do things a certian way, among other things.

At this point, I still think the main 2 issues are the following:
1. with places-survey as is right now, the notes-additions button doesn't work without changing `let momentStop;` to have some initial value
2. The app hangup when additional trips are loaded after labeling already

www/js/diary/infinite_scroll_list.js
```diff
@@ -76,16 +76,15 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
   }
 
-  $scope.getTripHeight = function(trip) {
-    let height = trip.INPUTS?.[2] ? 438 : 384;
-    if (trip.tripAddition) {
-      height += 40 * trip.tripAddition.length;
+  $scope.getCardHeight = function(entry) {
+    let height = 0;
+    if (entry.start_ts) { // entry is a trip
+      height = entry.INPUTS?.[1] ? 380 : 326;
```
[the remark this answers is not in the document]
JGreenlee: LGTM

www/js/diary/infinite_scroll_list.js
```diff
@@ -76,16 +76,15 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
...
+    } else if (entry.enter_ts) { // entry is a place
+      height = 106;
     }
-    if (trip.placeAddition) {
-      height += 40 * trip.placeAddition.length;
-    }
-    if ($scope.showPlaces) {
-      height += 80;
+    if (entry.additions) {
+      height += 40 * entry.additions.length;
```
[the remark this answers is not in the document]
JGreenlee: LGTM

www/templates/diary/trip_list_item.html
```diff
@@ -85,14 +85,14 @@
...
-                            <linkedsurvey element-tag="{{surveyOpt.elementTag}}" class="col" trip="trip"></linkedsurvey>
+                            <linkedsurvey element-tag="{{surveyOpt.elementTag}}" class="col" timeline-entry="trip"></linkedsurvey>
                         </div>
 
                         <verifycheck linkedtag="{{surveyOpt.elementTag}}" style="padding: 8px" class="col-10 diarycheckmark-container center-vert center-horiz"></verifycheck>
                     </div> 
                 </div>
             </div>
-            <enketo-notes-list timeline-entry="trip" addition-entries="trip.tripAddition" timezone="trip.start_local_dt.timezone"></enketo-notes-list>
+            <enketo-notes-list ng-if="trip.additions.length" timeline-entry="trip" addition-entries="trip.additions"></enketo-notes-list>
```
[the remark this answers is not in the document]
JGreenlee: Yes, `trip.additions` is an array. If it's empty (length=0), then we have no additions to display and we don't need to render the "notes list"

www/templates/diary/infinite_scroll_list.html
```diff
@@ -32,22 +32,10 @@
...
-                    <h2 class="hr-lines" translate> {{trip.display_date}}</h2>
-                </div> Commented out as we are not using prevTrip anymore. We should come back to 
-                this and implement some sort of horizontal line between trips-day-changing later on-->
-
-                <infinite-scroll-trip-item trip="trip" config="ui_config" surveys="surveys" map-limiter="mapLimiter"></infinite-scroll-trip-item>
-                <infinite-scroll-place-item ng-if="showPlaces" trip="trip" config="ui_config" surveys="surveys"></infinite-scroll-place-item>
-
+            <div collection-repeat="entry in data.displayTimelineEntries" item-height="getCardHeight(entry)">
+                <infinite-scroll-trip-item ng-if="entry.start_ts" trip="entry" config="ui_config" surveys="surveys" map-limiter="mapLimiter"></infinite-scroll-trip-item>
+                <infinite-scroll-place-item ng-if="showPlaces && entry.enter_ts" place="entry" config="ui_config" surveys="surveys"></infinite-scroll-place-item>
```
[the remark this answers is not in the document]
JGreenlee: We can actually take this out. From an earlier draft it's the `surveys_info.surveys` field from the config, but it's no longer being used by 'trip item' nor 'place item', since they receive the whole config now

www/js/survey/input-matcher.js
```diff
@@ -26,47 +27,59 @@ angular.module('emission.survey.inputmatcher', ['emission.plugin.logger'])
...
         }
         // https://github.com/e-mission/e-mission-docs/issues/476#issuecomment-747222181
-        const startChecks = userInput.data.start_ts >= trip.start_ts &&
-            userInput.data.start_ts <= trip.end_ts;
-        var endChecks = (userInput.data.end_ts <= trip.end_ts ||
-            (userInput.data.end_ts - trip.end_ts) <= 15 * 60);
+        const startChecks = userInput.data.start_ts >= entryStart &&
+            userInput.data.start_ts < entryEnd;
+        var endChecks = (userInput.data.end_ts <= entryEnd ||
+            (userInput.data.end_ts - entryEnd) <= 15 * 60);
```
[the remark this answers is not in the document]
JGreenlee: Start checks:
`
// For this input to match, it must begin after the start of the timelineEntry (inclusive) but before the end of the timelineEntry (exclusive)
`

End checks:
`
// A matching user input must also finish before the end of the timelineEntry, or within 15 minutes.
`

If start checks pass but end checks fail, there is a second level of end checks (a few lines below in the code)

www/js/survey/enketo/enketo-notes-list.js
JGreenlee: This is the same value that $scope.timzone had before, but I removed the scope variable because it was redundant to have as a scope variable when we can just grab the timezone from the timlineEntry

But yes, the overall reason this is necessary is that we want the timestamps of 'notes' to show up in the timezone of the trip/place they belong to.
The only reason using `start_local_dt` could be misleading is for trips that go across timezones. Not sure if we want to dig into that (fairly uncommon) scenario right now.

When the survey is filled out, we assume that the times entered are in the same timezone. So I think it's fair to assume that we will also display them as such, according to the starting timezone.

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: We don't have a use case for `filled-in-label` right now, so we haven't actually implemented the logic for the label to change.

I'm not sure what the use case would be exactly - filling in the label makes sense for 1:1 user inputs, but not for 1:n additions because I imagine the label will always be something like "Add X".

After one X is added, would there ever be a situation where we want it to change to say "Add More X"?

www/js/diary/infinite_scroll_list.js
JGreenlee: Yes, I think we can remove `mapLimiter` altogether

www/js/diary/infinite_scroll_list.js
JGreenlee: No, we should take this out

www/js/diary/services.js
JGreenlee: I'll refactor this

www/js/diary/services.js
JGreenlee: Correct, we don't need a separate request to retrieve the locations from the phone because they are included in the `composite trip` obj

When composite trips are created on the server, it samples 100 raw locations and they are put into the `locations` field. We still have to format the locations into geojson on the phone, though

www/js/diary/services.js
JGreenlee: Filling trajectories is no longer happening asynchronously, so we aren't using Promise syntax (`.then()`, `.catch()`)

The catch block was there in case of a network timeout. We don't have a network call anymore so unless you see another call that's unsafe I don't think we need to add a try/catch

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: Ah, yes you're right. Your suggested change is the intended behavior.
I misunderstood before, when I was looking at this on my phone

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: It was intended to be coming from the server, but now it's just called "additions" so it's coming from nowhere. I'll fix this soon

JGreenlee: fixed 8b2ef77e459bc381e0660c10ef36bd1fad7db512

www/templates/diary/infinite_scroll_list.html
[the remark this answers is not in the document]
JGreenlee: taken out in 6e49ea91ecafa761a47df38703c4125536424119

www/js/diary/infinite_scroll_list.js
JGreenlee: removed in 6e49ea91ecafa761a47df38703c4125536424119

www/js/diary/infinite_scroll_list.js
JGreenlee: taken out 5bc36d62b799f10a396a343f940dea5b688118e7

www/js/diary/services.js
JGreenlee: refactored bd12d1d5ac7d33d63ec2b410360b073a8d1c8497
Way simpler this way, as this is all that was needed

JGreenlee:
I will check again, but no, I didn't have that issue.
Let me try to reproduce again following those exact steps

JGreenlee:
I was able to reproduce this time, since I tested both time use and trip confirm. This looks like a regression specific to the matching of trip user inputs. Additions are working fine.

Trying to find the cause. Debugging now.

JGreenlee:
Actually, I think it has nothing to do with user inputs at all.

The issue occurs any time the Jul 27 2016 test data is loaded on the phone. That day has a ton of trips (22) which is way more than other days I have tested with. I'm guessing there's an issue with the UI and it can't handle too many trips.
My next guess is a memory leak that isn't noticeable with only a few trips, but causes a freeze once a lot are loaded.

Going to take a break from this for a little bit since this is giving me a headache

sebastianbarry:
Good find. I'll continue testing without using Jul 27 2016 data and see if there's anything else I can come up with

shankari:
I really doubt this is a memory issue, since we have been able to load Jul 27 2016 data in the old label screen without any problems. On my android test phone, I have the staging version of NREL OpenPATH loading 24 trips without any issues (see screenshot). There's been a regression elsewhere; I'm guessing in the code which decides when to stop loading.

JGreenlee:
I have found this really difficult to debug because Safari dev tools become unresponsive whenever the issue occurs.

I resorted to systematically commenting out different parts to narrow the scope of the issue. After a few hours, I found that it only happens when the 'leaflet' map is shown. Specifically, only when a trip is passed in as the geojson.

I'm guessing the trip>geojson conversion is somehow invalid - in such a way that leaflet doesn't know what to do with it. So next I'll look at how our geojson is being formatted before passed into the leaflets.

JGreenlee:
Pretty sure I've found the issue (I should have known). For every trip and place, its nextEntry is stored (like nextTrip used to be). But I did this in a bad way, and so each entry has circular references to the next entry, which we obviously don't want

Leaflet is probably trying to parse that and chewing on it forever because of the circular references

JGreenlee:
With 566a64604a57077635fe90e0e4358261b88b9844, I think that ordeal is finally over. Phew!
I'm going to call it a day, and hopefully @sebastianbarry's testing goes better tomorrow 😅

sebastianbarry:
## RE-TESTING:
All testing uses the following configuration
https://github.com/sebastianbarry/e-mission-phone/tree/places-survey
https://github.com/sebastianbarry/e-mission-server/tree/add-composite-trip

### Test 1:
iPhone 11 emulator
Program: `{name: "Devapp Program (Tripconfirm survey + Timeuse Survey)", joinurl: "emission://join_study?label=dev-emulator-program-tripconfirm&source=github" }`

| Trial | Notes |
:---:|:---:
| Trial 1: Load in jul-25-2016 data, label both trips and add timeuse to both trips | Data loads properly and time additions work properly | 
| Trial 2: Load in jul-27-2016 data, label some trips and add timeuse to some trips | New data loads properly and additions have stayed on the previously loaded trips and places. Trip/Place additions adding works perfect |
| Trial 3: Load in jul-29-2016 data | Data loads fine, does not affect the previously added notes additions for trips and places. |
| Trial 4: Run "end trips + force sync" and the run pipeline multiple times in succession, to simulate multiple hours/days going by | No apparent issue with the pipeline being ran multiple times in a row, notes additions stay |
| Trial 5: Observe the prefill times for Places/Trip additions (timeuse survey) | Trip additions timeuse survey works fine with prefill times; HOWEVER Place additions timeuse survey prefills time for the **_previous place_**, see this video: https://user-images.githubusercontent.com/61334340/226664750-0640a675-5041-4c72-b599-5aeb4b5185fd.mov |

#### Results of Test 1:
* Prefill times for places is incorrect, it is prefilling times based on the previous place object in the list
* Re-running the pipeline multiple times does not negatively affect the notes additions staying

### Test 2: 
iPhone 11 emulator
Program: `{name: "Devapp Program (Tripconfirm survey + Timeuse Survey)", joinurl: "emission://join_study?label=dev-emulator-program-tripconfirm&source=github" }`

| Trial | Notes |
:---:|:---:
| Trial 1: Load in jul-25-2016 data, label both trips and add timeuse to both trips | Data loads properly and time additions work properly | 
| Trial 2: Load in jul-27-2016 data, label some trips and add timeuse to some trips | New data loads properly and additions have stayed on the previously loaded trips and places. Trip/Place additions adding works perfect |
| Trial 3: Load in jul-29-2016 data | Data loads fine, does not affect the previously added notes additions for trips and places. |
| Trial 4: Run "end trips + force sync" and the run pipeline multiple times in succession, to simulate multiple hours/days going by | No apparent issue with the pipeline being ran multiple times in a row, notes additions stay |
| Trial 5: Observe the prefill times for Places/Trip additions (timeuse survey) | Trip additions timeuse survey works fine with prefill times; HOWEVER Place additions timeuse survey prefills time for the **_previous place_**, see this video: https://user-images.githubusercontent.com/61334340/226664750-0640a675-5041-4c72-b599-5aeb4b5185fd.mov |

#### Results of Test 2:
* Prefill times for places is incorrect, it is prefilling times based on the previous place object in the list
* Re-running the pipeline multiple times does not negatively affect the notes additions staying

### Test 3:
Android Pixel (Google Playstore API) emulator
Program: `{name: "Devapp Program (Tripconfirm survey + Timeuse Survey)", joinurl: "emission://join_study?label=dev-emulator-program-tripconfirm&source=github" }`

| Trial | Notes |
:---:|:---:
| Trial 1: Load in jul-25-2016 data, label both trips and add timeuse to both trips | Same issue is happening here, here is a screenshot of it: Screen Shot 2023-03-15 at 12 19 39 AM | 
| Trial 2: Load in jul-27-2016 data | After loading in the jul-27 data, I ran the pipeline and INSTEAD of pressing the "force trips + sync" button, the same significant error occurred again. When I tried to load in data from the server the app becomes unresponsive, and upon app reload the app remains unresponsive at the same point |
| Trial 3: Load in jul-29-2016 data | I was not able to test this due to the above issue |
| Trial 4: Run the pipeline multiple times, to simulate multiple hours/days going by | I was not able to test this due to the above issue |

JGreenlee:
Of the 4 action items listed above, I got around to fixing (2) and (4) today.
I wasn't able to reproduce (1) and I'm not sure what might have caused it.
I have seen (3) happen before, but I can't reproduce it anymore. Maybe it was fixed along with (2) and (4)

shankari:
I am fine with deferring (3) since I have also seen it happen before, and it will hopefully be fixed by #916

www/js/diary/infinite_scroll_place_item.js
```diff
@@ -30,18 +29,6 @@ angular.module('emission.main.diary.infscrollplaceitem',
     console.log("Place Item Controller called");
     console.log('config is ', $scope.config);
 
-    // timebounds is used in js/survey/enketo/enketo-add-note-button.js getPrefillTimes() function
-    // this allows us to pre-fill time and date in surveys that have 'Date', 'Start_time', and 'End_time' fields
-    $scope.timeBounds = () => {
```
shankari, line 35: so we don't need this any more? Could this be the cause of issue (1) that Sebastian found?

www/js/survey/input-matcher.js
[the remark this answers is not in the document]
shankari: @JGreenlee can you actually edit to include this?

www/templates/diary/place_list_item.html
```diff
@@ -2,23 +2,23 @@
...
                     <!-- Place (Destination of Prev. Trip) -->
                     <i class="icon ion-location"></i>
-                    <b>{{trip.end_display_name}}</b>
+                    <b>{{place.display_name}}</b>
                 </div>
                 <div style="margin: 10px">
-                    <enketo-add-note-button notes-config="configPlaceNotes" surveys="surveys" trip="trip"
-                        time-bounds="timeBounds" dataKey="manual/place_addition_input" style="margin: auto">
+                    <enketo-add-note-button notes-config="configPlaceNotes" timeline-entry="place"
+                        dataKey="manual/place_addition_input" style="margin: auto">
```
shankari, line 15: don't we want to show this only if place time use is configured?
and shouldn't there be a place_user_input button to match the server side code?

shankari (approved):
Again, I'm going to merge this, but I have some high-level cleanup suggestions below

www/js/diary/infinite_scroll_place_item.js
JGreenlee: Yes, I consolidated the logic into `enketo-add-note-button.js` > `getPrefillTimes()` instead of having it split across files.
I haven't been able to reproduce that issue yet, but `getPrefillTimes()` is certainly the first place I'll check.

www/templates/diary/place_list_item.html
JGreenlee: This seems like a low priority now, but I agree we should be consistent.

We use `enketo-add-note-button` for both trip-level and place-level notes.
We use `enketo-trip-button` for trip user inputs. I don't think it makes sense to make a new `enketo-place-button`.
Instead we should refactor `enketo-trip-button` into `enketo-userinput-button` or some other generic name

In terms of UI, it is hard for me to imagine where the place userinput button should go, since we don't have a use scenario yet. But for now I could just put whichever buttons are configured on the same row.

www/js/survey/input-matcher.js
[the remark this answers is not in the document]
JGreenlee: will address in a cleanup

www/js/diary/infinite_scroll_place_item.js
shankari: Resolving this for now until you can reproduce

PR #925, opened 2023-02-17
:fire: Remove the transition notify plugin

shankari:
Per https://github.com/e-mission/e-mission-docs/issues/838#issuecomment-1356461966 This is the first step in which I remove all the controls related to it

```
$ grep -ri notify www/js/control | wc -l
       0
$ grep -ri notify www/templates/control/ | wc -l
       0
```

Testing done:
Profile screen loads correctly


shankari:
| Top of control screen | Bottom of control screen (with notifications settings removed)
|--------|-------|
| Screenshot_1676593767 | Screenshot_1676593758 |

PR #929, opened 2023-03-12
Bump up versions of the dev chain and all other plugins

shankari:
- android CLI version to the latest
- android sdk packages to the latest
- cordova, ionic and bower to the latest
- all non-OpenPATH plugins to the latest versions

Some of these (`cordova-plugin-badge`/`phonegap-plugin-barcodescanner`) ran into an issue with using `compile` when they actually needed `implementation` for the new gradle.

The badge is a dependency of the local notification plugin, so I migrated from `timkellypa` to `levanax` who has some commits that have fixed it and some other issues. This fixes it for the local notification plugin but not for the `cordova-plugin-badge` which is pulled in as a dependency from elsewere

I fixed that and the barcodescanner by adding a new hook that renames `compile` to `implementation` before build.

Testing done: builds

This is the next step in the upgrade and addresses https://github.com/e-mission/e-mission-docs/issues/838#issuecomment-1465021443

### DIFF
--- a/.github/workflows/android-automated-sdk-install.yml
+++ b/.github/workflows/android-automated-sdk-install.yml
@@ -40,6 +40,7 @@ jobs:
 
     - name: Install to a new SDK root
       run: |
+        export JAVA_HOME=$JAVA_HOME_11_X64
         export ANDROID_SDK_ROOT=$NEW_ANDROID_SDK_ROOT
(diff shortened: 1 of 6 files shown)


shankari:
Built apps run, without updating the targetSDK

PR #933, opened 2023-03-17
Survey refactor expansion UI changes

sebastianbarry:
I previously had this PR pointed to `places-survey` branch (https://github.com/sebastianbarry/e-mission-phone/pull/7) but after a short discussion with Shankari I am redoing this PR for `survey_refactor_expansion` branch:

# Testing:
Part 1: **Left justify entry labels**

Part 2: **Confirm delete entry popup**
https://user-images.githubusercontent.com/61334340/226062221-34999dea-69f1-422f-af9d-5e48ccb71265.mov

### DIFF
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -494,14 +494,13 @@ enketo-notes-list, .notes-list {
 }
 
 .notes-list {
-  padding: 0 5px 5px 5px;
+  padding: 0 15px 5px 15px;
   text-align: center;
 }
 
 .notes-list-entry {
   display: grid;
   grid-template-columns: 10fr 9fr 1fr;
-  place-items: center;
   overflow: hidden;
   font-size: 13px;
   height: 40px;
@@ -509,4 +508,8 @@ enketo-notes-list, .notes-list {
 
 .notes-list-entry .icon {
   font-size: 24px;
+}
+
+.notes-list-label {
+  text-align: left;
 }
--- a/www/css/main.diary.css
+++ b/www/css/main.diary.css
@@ -501,6 +501,7 @@ enketo-notes-list, .notes-list {
 .notes-list-entry {
   display: grid;
   grid-template-columns: 10fr 9fr 1fr;
+  align-items: center;
   overflow: hidden;
   font-size: 13px;
   height: 40px;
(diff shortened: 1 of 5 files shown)


sebastianbarry:
@shankari does this PR look correct?

shankari:
At least it is to the right branch?!

www/css/main.diary.css
```diff
@@ -494,19 +494,22 @@ enketo-notes-list, .notes-list {
 }
 
 .notes-list {
-  padding: 0 5px 5px 5px;
+  padding: 0 15px 5px 15px;
```
JGreenlee, line 497: To me, this puts the trash can a little too far in. How about 15 on the left and 10 on the right?
```suggestion
  padding: 0 10px 5px 15px;
```
Here's what that looks like
image

JGreenlee:
Real quick, I just patched up the CSS so that the vertical alignment was retained

Left the pop-up as-is since it looks great. We will want to internationalize the text though.

www/templates/diary/place_list_item.html
```diff
@@ -4,7 +4,7 @@
         <!-- Start Time Tag -->
         <!-- <div class="start-time-tag">{{trip.display_end_time}}</div> -->
         <div class="diary-card short" style="padding-top: 10px;">
-            <div style="width: 100%; margin: auto; padding: 5px">
```
JGreenlee, line 7: Why was this padding removed?
We can do without it - but if we do remove it, then we have to update the card height values in `infinite_scroll_list.js > getTripHeight()`

shankari:
I am going to revert the padding change and merge tonight.

@sebastianbarry you need to create a follow-on change that:
- discusses/resolves the padding
- addresses the i18n of the new strings

www/templates/diary/place_list_item.html
```diff
@@ -4,7 +4,7 @@
...
+            <div style="width: 100%;">
```
shankari, line 7:
```suggestion
            <div style="width: 100%; margin: auto; padding: 5px">
```

www/templates/diary/place_list_item.html
shankari: Reverting. @sebastianbarry please address this comment on Monday

shankari (approved):

PR #934, opened 2023-03-25
Additions and composite trip code cleanup

JGreenlee:


www/js/diary/infinite_scroll_list.js
shankari, line 1: you mean "use composite trip" instead of "confirmed trip" in 
https://github.com/e-mission/e-mission-phone/pull/934/commits/72970addc939e7daf02df14ac7e29b54c2778f97
right?

shankari (approved):
LGTM! Merging now but will wait until Monday to see if there is any other feedback from UNSW that we need to address

shankari:
I just added a place in the time use on one of the test phones and 
(1) it didn't show up, and
(2) when I tried to reload, I got this error



I think that this may be fixed by https://github.com/e-mission/e-mission-phone/pull/934/files#diff-7aa8765674a8626980229fb8d83dae25fada654f680981389f72ff5b406ba9eaR81

so I'm going to pull the merged code and test some more

@sebastianbarry did you add time use to places in your testing? Did they show up correctly?

shankari:
Activity that doesn't span the entire time period does not seem to be added (without these changes, see below).
Will retry with changes, but am not sure it will fix anything given that the matching logic has not changed substantially
@sebastianbarry did you test this?
https://user-images.githubusercontent.com/2423263/227750237-b7561ae3-1e72-4159-ace4-48777239c911.mov

www/js/diary/infinite_scroll_list.js
JGreenlee: correct https://github.com/e-mission/e-mission-phone/commit/72970addc939e7daf02df14ac7e29b54c2778f97#commitcomment-106091114

JGreenlee:
> I think that this may be fixed by https://github.com/e-mission/e-mission-phone/pull/934/files#diff-7aa8765674a8626980229fb8d83dae25fada654f680981389f72ff5b406ba9eaR81

I think so too. While refactoring, I noticed there was something weird in that branch of the code so I fixed it.
(One of the things I hate about JS - you don't catch these things until runtime. I wouldn't mind using TypeScript for the upcoming rewrite if that's an option..)

PR #935, opened 2023-03-28
Display untracked time on Label Screen

JGreenlee:
In light of https://github.com/e-mission/e-mission-docs/issues/870 (and just because it is a good idea for clarity to the user) the UI will show when there is untracked time on the label screen.

As a bonus, this PR also includes a couple of quick fixes for the diary screen. There were some side effects from the 'composite trip' implementation.

### DIFF
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -409,7 +409,9 @@
         "download-apply": "Apply"
     },
     "login":{
-        "save-your-opcode":"Make sure to save your OPcode!",
+        "make-sure-save-your-opcode":"Make sure to save your OPcode!",
+        "cannot-retrieve":"NREL cannot retrieve it for you later!",
+        "save":"Save",
         "continue": "Continue",
         "enter-existing-token": "Enter the existing token that you have",
         "button-accept": "OK",
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -166,6 +166,7 @@
         "replaces": "Replaces",
         "purpose": "Purpose",
         "survey": "Details",
+        "untracked-time-range": "Untracked: {{start}} - {{end}}",
         "unlabeled": "All Unlabeled",
         "invalid-ebike": "Invalid",
         "to-label": "To Label",
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -433,7 +433,9 @@
         "powered-by": "Powered by",
         "dismiss": "Dismiss",
         "return-to-beginning": "Return to beginning",
-        "go-to-end": "Go to End"
+        "go-to-end": "Go to End",
+        "enketo-form-errors": "Form contains errors. Please see fields marked in red.",
+        "enketo-timestamps-invalid": "The times you entered are invalid. Please ensure that the start time is before the end time."
     },
     "join": {
         "welcome-to-nrel-openpath": "Welcome to NREL OpenPATH",
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -77,15 +77,15 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
   $scope.getCardHeight = function(entry) {
     let height = 0;
     if (entry.key == 'analysis/confirmed_place') {
-      height = 188;
+      height = 178;
     } else if (entry.key == 'analysis/cleaned_untracked') {
-      height = 130;
+      height = 164;
     } else if (entry.key == 'analysis/confirmed_trip') {
       // depending on if ENKETO or MULTILABEL is set, or what mode is chosen,
       // we may have 1, 2, or 3 buttons at any given time
-      // 272 is the height without any buttons, and each button adds 54 pixels
+      // 242 is the height without any buttons, and each button adds 54 pixels
       const numButtons = entry.INPUTS?.length || 1;
-      height = 272 + (54 * numButtons)
+      height = 242 + (54 * numButtons)
     }
 
     if (entry.additionsList) {
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -76,17 +76,18 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
 
   $scope.getCardHeight = function(entry) {
     let height = 0;
-    if (entry.enter_ts) { // entry is a place
+    if (entry.key == 'analysis/confirmed_place') {
       height = 188;
-    } else if (!entry.locations) { // entry is untracked time
+    } else if (entry.key == 'analysis/cleaned_untracked') {
       height = 130;
-    } else if (entry.start_ts) { // entry is a trip
+    } else if (entry.key == 'analysis/confirmed_trip') {
       // depending on if ENKETO or MULTILABEL is set, or what mode is chosen,
       // we may have 1, 2, or 3 buttons at any given time
       // 272 is the height without any buttons, and each button adds 54 pixels
       const numButtons = entry.INPUTS?.length || 1;
       height = 272 + (54 * numButtons)
     }
+
     if (entry.additionsList) {
       height += 40 * entry.additionsList.length; // for each trip/place addition object, we need to increase the card height
     }
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -77,7 +77,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
   $scope.getCardHeight = function(entry) {
     let height = 0;
     if (entry.enter_ts) { // entry is a place
-      height = 106;
+      height = 188;
     } else if (!entry.locations) { // entry is untracked time
       height = 130;
     } else if (entry.start_ts) { // entry is a trip
@@ -319,7 +319,15 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
     $scope.data.displayTrips.forEach((cTrip) => {
       const place = cTrip.confirmed_place;
       $scope.data.displayTimelineEntries.push(cTrip);
-      if ($scope.showPlaces && place?.duration >= 60) {
+      if ($scope.showPlaces && place) {
+        // Places with duration less than 60 seconds will not be displayed
+        if (place.duration && place.duration < 60) return; 
+
+        if (!place.display_end_time) {
+          // If a place does not have a display_end_time, it is the last place
+          // We will set display_end_time to the end of the day
+          place.display_end_time = moment(place.enter_fmt_time).parseZone().endOf('day').format("h:mm A");
+        }
         $scope.data.displayTimelineEntries.push(place);
       }
     });
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -20,6 +20,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
                                       'emission.plugin.logger',
                                       'emission.main.diary.infscrolltripitem',
                                       'emission.main.diary.infscrollplaceitem',
+                                      'emission.main.diary.infscrolluntrackedtimeitem',
                                     ])
 
 .controller("InfiniteDiaryListCtrl", function($window, $scope, $rootScope, $injector,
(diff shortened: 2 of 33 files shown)


www/js/diary/services.js
```diff
@@ -761,7 +761,7 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
     }
 
     var confirmedPoints2Geojson = function(trip, locationList) {
-      var sectionCoordinates = locationList.map(function(point) {
+      var sectionCoordinates = locationList?.map(function(point) {
```
shankari, line 764: Instead of all the checks to see if the location list is present and the user input is present, maybe we should continue to assume that they always exist and just ensure that the untracked time has a zero length list and an empty user input.

www/templates/diary/untracked_time_list_item.html
```diff
@@ -0,0 +1,24 @@
+<div>
+    <ion-item class="unified-diary-item" style="background-color: transparent;" class="list-item">
+        <div class="diary-card short greyed" style="padding-top: 10px;">
+            <div style="width: 100%; margin: auto; padding-block: 5px">
+                <div class="row place-title">
+                    <b translate=".untracked-time-range"
+                        translate-value-start="{{ triplike.display_start_time }}"
+                        translate-value-end=" {{ triplike.display_end_time }}">
```
shankari, line 8: Given the comments from UNSW, we should probably add the start/end times for this as well.

www/js/diary/infinite_scroll_list.js
```diff
@@ -75,14 +76,16 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
 
   $scope.getCardHeight = function(entry) {
     let height = 0;
-    if (entry.start_ts) { // entry is a trip
+    if (entry.enter_ts) { // entry is a place
+      height = 106;
+    } else if (!entry.locations) { // entry is untracked time
+      height = 130;
+    } else if (entry.start_ts) { // entry is a trip
```
shankari, line 83: I think it would be better to put a type into the entry instead of using these indirect inferences.  It will ensure that we have flexibility for different types of places and trips down the road, and will ensure that there is a central location where we set the types.

Having such indirect inferences sprinkled through the code is going to make it super hard to change in the future.

shankari: This is true for the `isPlace` checks as well. The data model on the server is already typed. Why don't we just propagate that type throughout instead of inferring place vs. trip in every location where we need to handle them.

www/js/diary/services.js
JGreenlee: According to the data model, `user_input` exists on confirmed objects but not cleaned objects.

So wouldn't it be strange to insert an empty `user_input` on the CLEANED_UNTRACKED type?

www/templates/diary/untracked_time_list_item.html
JGreenlee: The start and end times are listed here in bold text. Do you think it is necessary that they are moved to the badges?

We decided a while ago for places that it would be visually redundant.

shankari: Yes, we did. And UNSW disagreed 😄 
https://github.com/e-mission/e-mission-docs/issues/871#issuecomment-1486148982

www/js/diary/services.js
shankari: Good point. We should leave the `user_input` out of `CLEANED_UNTRACKED` time.
We could still have a zero array for locations though...

www/js/diary/infinite_scroll_list.js
JGreenlee: I definitely agree. I'll make sure every triplike or place object on the label screen has a 'key' describing what data type it comes from.
Currently when composite trips are unpacked / "flattened",  the metadata is sort of just thrown away. so I'll make sure to grab the key beforehand

JGreenlee:
The button styles have been unified. The start and end time tags have also been unified and will show on every card.

JGreenlee:
Survey timestamps are also validated as requested.

www/i18n/en.json
```diff
@@ -166,6 +166,7 @@
         "replaces": "Replaces",
         "purpose": "Purpose",
         "survey": "Details",
+        "untracked-time-range": "Untracked: {{start}} - {{end}}",
```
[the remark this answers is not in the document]
JGreenlee: I didn't get to this today.

The "Timestamps are invalid" message should also be internationalized https://github.com/e-mission/e-mission-phone/pull/935#issuecomment-1489711407
The default Enketo error message ("Form contains errors. Please see fields marked in red.") does not appear to be internationalized either.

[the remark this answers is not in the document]
JGreenlee: I internationalized all 3 of these. 4be125a049ce44e8957e6a13268c6764acca4cc2  https://github.com/e-mission/e-mission-translate/pull/23

Because I was not sure, I targeted the PR at 'master'. Let me know if it needs to be targeted at 'master_for_platform' instead.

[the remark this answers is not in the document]
shankari: I've rebased to `master_for_platform` and merged

www/js/metrics-factory.js
```diff
@@ -192,7 +192,7 @@ angular.module('emission.main.metrics.factory',
       mode = 'WALKING';
     }
     let currentMETs = cc.getMETs();
-    if (!currentMETs[mode]) {
+    if (!currentMETs?.[mode]) {
```
shankari, line 195: I don't think that this is the correct fix. Instead, we should just disable the dashboard screen completely unless we are in the MULTILABEL mode.

Please see https://github.com/e-mission/e-mission-phone/pull/938, https://github.com/e-mission/e-mission-phone/pull/940 and https://github.com/e-mission/e-mission-phone/pull/941 for the final fixes

www/js/metrics-mappings.js
```diff
@@ -389,7 +389,7 @@ angular.module('emission.main.metrics.mappings', ['emission.plugin.logger',
       }
     }
 
-    $ionicPlatform.ready().then(function() {
+    $ionicPlatform.ready().then(() => {
       DynamicConfig.configReady().then((newConfig) =>
         this.init()
       );
```
shankari: again, I don't see how this fixes anything. `this` is always the window, which does not have an `init` function.
See https://github.com/e-mission/e-mission-phone/pull/938

www/templates/diary/infinite_scroll_list.html
```diff
@@ -5,7 +5,7 @@
         <button class="button button-icon" ng-click="increaseHeight()">
             <i class="icon ion-arrow-resize"></i>
         </button>
-        <div class="row buttons labelfilterlist">
+        <div class="buttons labelfilterlist">
```
shankari, line 8: why does this fix the issue "buttons collapsing" issue? Why did it happen in the first place?

www/js/splash/startprefs.js
```diff
@@ -206,7 +206,7 @@ angular.module('emission.splash.startprefs', ['emission.plugin.logger',
             $rootScope.redirectParams = undefined;
             return {state: redirState, params: redirParams};
           } else {
-            return {state: 'root.main.inf_scroll', params: {}};
+            return {state: 'root.main.control', params: {}};
```
shankari, line 214: I don't think this is the correct fix. This is the "default" screen that we use, and it should go to the label screen by default.
`getPendingOnboardingState` is an artifact of a previous version of the app in which people installed the app but then signed up along with their friends using the now defunct habitica integration.

I don't think this is the reason for the profile -> label switch, which is more likely to be an error somewhere else in the codebase.

Let's not include extraneous fixes in here unless they are 100% reproducible given that we are getting to the point where we want to push to production.

shankari: Comment on the follow-on commit (97ed8f0369c59acb117fdd23f8fa5fd5304eeac0) as well; let's just leave this off for now. I don't want to change the core app functionality unless fixing a 100% broken and reproducible issue because I am afraid that it will introduce additional regressions (like https://github.com/e-mission/e-mission-phone/commit/f863264f19f61ba333c8392ad91c588c94bdff7c did)

www/js/survey/multilabel/trip-confirm-services.js
```diff
@@ -57,15 +57,16 @@ angular.module('emission.survey.multilabel.services', ['ionic', 'emission.i18n.u
...
           console.log("Promise list ", omPromises);
-          Promise.all(omPromises).then((omObjList) =>
-              ch.INPUTS.forEach(function(item, index) {
-                  inputParams[item] = omObjList[index];
-              })).catch((err) => {
-                    Logger.displayError("Error while loading input params in "+ch.INPUTS, err)
-                    reject(err);
-              });
-              console.log("Read all inputParams, resolving with ", inputParams);
-              resolve(inputParams);
```
shankari, line 68: I guess this is a good change; but I don't see what it fixes, and whether it is necessary after removing the dashboard completely on non-multilabel surveys.

shankari (changes requested):
Looks good at a high level.

Main fixes:
- merge my dashboard fixes properly; revert the changes you don't need
   - I turn off the dashboard completely for ENKETO configs, so we don't need a lot of the changes
   - I would prefer not making unnecessary/unrelated changes in the core functionality this late in the game since I am concerned about unavoidable regressions
- I saw your fixes to the startprefs and I am not sure that at least the first one works
   - Again, let's not change anything in the splash screen unless it is to fix a 100% reproducible error. Again, worried about unexpected regressions, and that code is fairly fragile and crusted over with old, obsolete functionality. And does it even happen once **my** dashboard fixes are in?

JGreenlee:
Ok, I'll revert what's been requested. I had been squashing bugs in the dashboard since before it was turned off for ENKETO, but those fixes are irrelevant now.
I merged from `survey-refactor-expansion` when I thought it was done, but it looks like there were more changes the next day, which I never knew about.

www/js/metrics-mappings.js
JGreenlee: Actually, when I was fixing the dashboard bugs myself it did fix the issue. However, with your changes this is probably irrelevant

(In traditional JS functions, `this` is always the window. But in arrow functions, `this` is inherited from the parent scope.)

www/js/splash/startprefs.js
JGreenlee: reverted 93959b540f5e52d67cbb66019e5c5cf5a9c16ee5

www/js/metrics-factory.js
JGreenlee: reverted 924fdd38d142aa52150366a507dddab4123b7702

www/js/metrics-mappings.js
JGreenlee: reverted 75e9468f90af16667ace1fb96df06fe8dc7e1bdb

www/js/survey/multilabel/trip-confirm-services.js
JGreenlee: The old code looks misleading and I don't think it does what it is intended to do. If you fix the intentation/alignment:
```js
Promise.all(omPromises).then((omObjList) =>
    ch.INPUTS.forEach(function (item, index) {
        inputParams[item] = omObjList[index];
    })
).catch((err) => {
    Logger.displayError("Error while loading input params in " + ch.INPUTS, err)
    reject(err);
});
console.log("Read all inputParams, resolving with ", inputParams);
resolve(inputParams);
```
You see that `resolve()` is called regardless of what happens to the inner `omPromises`.

However, now that the dashboard is disabled for non-MULTILABEL configurations, it doesn't actually cause problems anymore.
If you want me to revert back to the old code, which has evidently stood the test of time thus far, I will.

www/templates/diary/infinite_scroll_list.html
JGreenlee: I don't have a specific answer, but I do know that "row" applied a `width: 100%` to the `<div>`, which tends to cause problems with flexbox layouts.
The parent is a flexbox `<span>` and it is not big enough to fit all the buttons anyway, so they overflow. There's a lot of weird stuff happening in the CSS here.

shankari: Thanks for letting me know. Good thing we are going to rewrite in a non-classic CSS framework...

www/js/survey/multilabel/trip-confirm-services.js
shankari: Yes, let's please revert to the old code. Before changing, I would want to look at the blame to verify that there was no reason for it. I do remember that the input params loading was tricky when I wrote it.

shankari:
@JGreenlee I am happy to merge once the change in [www/js/survey/multilabel/trip-confirm-services.js](https://github.com/e-mission/e-mission-phone/pull/935/files/c60cf924686c5e940375f7d1dfb67c39737c9d14..fb35823fe2b5cb61d3135493369c61f453613632#diff-7c8a53603c832a6e37c0a96f369a84372f3189d10bef1beb80ab1997db586255) is reverted

www/js/survey/multilabel/trip-confirm-services.js
JGreenlee: reverted 094b677167cab8d6aacd6ae06609168b9401f505

JGreenlee:
Good to go.

shankari (approved):

PR #937, opened 2023-03-29
Add start time, end time, and date to each Place card

sebastianbarry:
infinite_scroll_list.js
- Increased the default height of the place cards in getCardHeight() by 82px
- Added a check that if a place does not have a display_end_time, we create a display_end_time to be the end of the day

place_list_item.html
- Added place start time tag
- Added place display date to be centered above the place name
- Added place end time tag


sebastianbarry:
Testing: 

Simulator Screen Shot - iPhone 11 - 2023-03-29 at 10 19 11
Simulator Screen Shot - iPhone 11 - 2023-03-29 at 10 19 13

shankari (approved):
LGTM!

PR #938, opened 2023-03-29
Make `CustomDataHelper` a factory instead of a service

shankari:
So that calling `this.init` actually works.
Since the `ionicPlatform` code is a callback, `this` within it is actually a reference to the `window` so we can't call `this` on it

https://github.com/e-mission/e-mission-docs/issues/875#issuecomment-1488972566

Testing done:
- Before this change, got a couple of error popups in the multilabel view
- With this change, the error popups are gone

### DIFF
--- a/www/i18n/en.json
+++ b/www/i18n/en.json
@@ -408,7 +408,9 @@
         "download-apply": "Apply"
     },
     "login":{
-        "save-your-opcode":"Make sure to save your OPcode!",
+        "make-sure-save-your-opcode":"Make sure to save your OPcode!",
+        "cannot-retrieve":"NREL cannot retrieve it for you later!",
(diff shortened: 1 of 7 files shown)


shankari:
At this point, there are no errors while using the multilabel option.
https://user-images.githubusercontent.com/2423263/228630849-010bb82f-3883-43a7-9a87-1610473c80a6.mov

shankari:
https://user-images.githubusercontent.com/2423263/228715832-708e46db-4fa2-4869-9b54-09d6955aa2bd.mov

shankari:
Here's what `study-program` currently looks like
simulator_screenshot_1574201E-6E55-429E-AF68-840AA6F7BB72

PR #939, opened 2023-03-29
Input first place card

sebastianbarry:
@shankari Here is the PR for visibility, since we want this done by today afternoon but I have to go to school now.

This PR is related to @MaliheTabasi comment https://github.com/e-mission/e-mission-docs/issues/871#issuecomment-1486160076

> It would be helpful to include a Place card before the first tracked trip that starts at 12:00 AM on the same day. This will allow users to enter activities they have done before that trip on that day. As of now, there doesn't seem to be any Place card above the first trip, as can be seen in the attached screenshot.

Here's my first draft:

infiinite_scroll_list.js
- Gather start and stop times for the scratch place item based on whether or not there is other timelineEntries in the list
- Push the custom scratch place item into the list of allTrips
  - The custom scratch place item only contains a few variables that I was comparing other place items to
- concat ctList to allTrips, instead of the other way around so that the object we pushed stays at the top

Testing:
| | |
:---:|:---:
| Simulator Screen Shot - iPhone 11 - 2023-03-29 at 12 42 55 | Place card shows up in list in correct position |
| Simulator Screen Shot - iPhone 11 - 2023-03-29 at 12 42 59 | Place card survey is able to be opened, however date is incorrectly filled and time is in the wrong timezone |
| Simulator Screen Shot - iPhone 11 - 2023-03-29 at 12 43 12 | Notes additions show up in the list |
| Screen Shot 2023-03-29 at 12 43 26 PM | When clicking refresh, the inputmatcher does not re-insert the notes-additions into this place |

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -171,8 +171,11 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
     });
     Timeline.readAllCompositeTrips(currEnd, ONE_WEEK).then((ctList) => {
         Logger.log("Received batch of size "+ctList.length);
+        if (!$scope.showPlaces) {
+          ctList.splice(0, 1); // Remove the scratch first place if we are not using places
+        }
         ctList.forEach((ct, i) => {
-          
+          if (ct.scratchPlace) {return;} // Skip the scratch first place
           if ($scope.showPlaces && ct.confirmed_place) {
             const cp = ct.confirmed_place;
             cp.getNextEntry = () => ctList[i + 1];
@@ -192,42 +195,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
             fillPlacesForTripAsync(trip);
             fillTrajectoriesForTrip(trip);
         });
-        // for first place object
-        let startTimestamp = moment().startOf("day");
-        let loc = {coordinates: [0,0], type: "Point"}
-        let endTimestamp = moment().endOf("day");
-        let tz = moment.tz.guess();
-        if (ctList.length > 0) {
-          startTimestamp = moment(ctList[0].start_fmt_time).parseZone().startOf("day");
-          endTimestamp = moment(ctList[0].start_fmt_time).parseZone();
-          loc = ctList[0].start_loc;
-          tz = ctList[0].locations[0].metadata.time_zone;
-        }
-        let firstPlaceObj = {
-          additions: [],
-          additionsList: [],
-          background: "bg-light",
-          display_date: moment(endTimestamp).format(),
-          display_start_time: startTimestamp.format("h:mm A"),
-          display_exit_time: endTimestamp.format("h:mm A"),
-          end_display_name: " ",
-          end_loc: loc,
-          ending_trip: {$oid: "6423b9fdd9ac0bff48c04bf8"},
-          enter_fmt_time: moment(startTimestamp).format(),
-          enter_local_dt: moment2localdate(startTimestamp, tz),
-          enter_ts: moment(startTimestamp).unix(),
-          exit_fmt_time: moment(endTimestamp).format(),
-          exit_local_dt: moment2localdate(endTimestamp, tz),
-          exit_ts: moment(endTimestamp).unix(),
-          listCardClass: "list card list-card bg-light list-card-lg",
-          location: {type: "Point", coordinates: [-122.09519, 37.3915317]},
-          scratchPlace: true,
-          start_display_name: " ",
-          start_loc: loc
-        };
-        fillPlacesForTripAsync(firstPlaceObj);
-        $scope.data.allTrips.push(firstPlaceObj);
-        $scope.data.allTrips = $scope.data.allTrips.concat(ctList);
+        $scope.data.allTrips = ctList.concat($scope.data.allTrips);
         Logger.log("After adding batch of size "+ctList.length+" cumulative size = "+$scope.data.allTrips.length);
         Timeline.setInfScrollCompositeTripList($scope.data.allTrips);
         const oldestTrip = ctList[0];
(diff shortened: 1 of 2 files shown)


sebastianbarry:
I was able to get the prefill time and date to work correctly, but there is still one problem still occurring:

- [x] Display name never shows up
- [x] Prefill da |te/time is incorrect
- [x] Notes-additions come in as the wrong timezone
- [ ] While the name for the `start_display_name` and `end_display_name` come back in the promise for the correct names (for the firstPlaceObj), the `$scope.$apply` is not changing the title for the object in the DOM

| | | | |
:---:|:---:|:---:|:---:
| Simulator Screen Shot - iPhone 11 - 2023-03-30 at 05 54 25 | Simulator Screen Shot - iPhone 11 - 2023-03-30 at 05 54 29 | Simulator Screen Shot - iPhone 11 - 2023-03-30 at 05 54 34 | Screen Shot 2023-03-29 at 12 43 26 PM |

shankari:
@sebastianbarry wrt

I don't think this works. Fundamentally, I don't think you are even testing the correct use case.
You need to make sure that the first place is added **as the first trip in the timeline**.
But the timeline that you are testing with has only 3 trips, so it has the check mark immediately upon load.
You are not even testing with the use case where the user downloads more data.

Once you test it, I think you will find two issues.
- "concat ctList to allTrips, instead of the other way around so that the object we pushed stays at the top" is going to end up with out-of-order trips
- you are concating the trip on **every** call to `readDataFromServer` so you are going to run into duplicate trips anyway

sebastianbarry:
>     * you are concating the trip on **every** call to `readDataFromServer` so you are going to run into duplicate trips anyway

I tested this and you are correct: 
Simulator Screen Shot - iPhone 11 - 2023-04-03 at 07 25 06

with each click of the download button, a new out-of-order custom place object is pushed to the bottom of the current display

sebastianbarry:
Even after moving the scratch first place to the timeline, duplicate scratch first places were able to be input by clicking the download button, so I figured I would remedy this with this commit: https://github.com/e-mission/e-mission-phone/pull/939/commits/e6890caebe312a61216a217da895ed41653489fd

The problem, is that I can see a case where this would not be effective, which brings up a larger issue with this design:

If we are in the filter "to-label" and we label the first trip, that first trip and it's associated place object (in this case it would be the second place in the list) will get removed from the to-label screen, but can still be displayed in the "all-trips" filter. The first place will remain on the to-label screen indefinitely, since there is not associated trip object to label in order to "remove it" from the to-label filter. If I label all of my trips that were generated for that batch in the timeline, and attempt to download more trips to label, this old depreciated place object will remain, so now I might have the scratch first place object with timestamps from 1 month ago, but now that I've labeled all of the trips from the last month, the next trip and place objects are from today, generating a large gap in time between these 2 objects.

I believe this is okay, since there still does exist place/trip add notes additions fields between these; they just exist on the all-trips screen now instead of the to-label.

shankari:
at a high level, why are we not switching to having composite trips have both starting and ending place, consistent with https://github.com/e-mission/e-mission-docs/issues/871#issuecomment-1489502892
https://github.com/e-mission/e-mission-docs/issues/871#issuecomment-1490938758
https://github.com/e-mission/e-mission-docs/issues/871#issuecomment-1490971326

sebastianbarry:
I have been attempting to add in the starting and ending composite trip into the server, but I am getting quite stuck and need assistance.

I am following the commit of @JGreenlee https://github.com/sebastianbarry/e-mission-server/commit/9eb812d6e48a6abf15824be4e05722ef8c72c7d6 to have an idea of how to
1. Add in a `start_confirmed_place` object into compositetrip.py
2. Modify the name of `confirmed_place` to be `end_confirmed_place`

This not only requires all the respective changes on e-mission-server, but reflective changes on e-mission-phone which is a substantial amount of changes that will take a significant amount of time. I also anticipate that I will want to ask @JGreenlee  for help with this if possible!

I can continue trying to make this change, but I wanted to check first; @shankari should I continue with trying to make these changes? I don't see this getting done until at least 2 weeks if I am the one working on it. Also, I know that this is the approach that we discussed before, but is this the correct way of going about the problem (i.e. adding in `start_confirmed_place` and renaming the current confirmed_place to `end_confirmed_place`)?

shankari:
2. at a high level, I think that composite should in fact have both places. I thought so all along (which is why the other trip-like objects have start and end places in the data model), but was willing to go with one place for the export. But almost immediately, we run into a use case where both places would be helpful. So I think we should support it by default.
3. if we don't do it before we push to production, we will need to handle backwards compatibility - e.g. apps that have not updated to support start and end confirmed_place will have to be updated to do so.
   - I would be fine with only renaming `confirmed_place` to `end_confirmed_place` for now to avoid the backwards compat issues and to make the other changes later; what is the ETA for that change?
   - It seems like it would be a simple find/replace and should be pretty easy to implement...

sebastianbarry:
>    - I would be fine with only renaming `confirmed_place` to `end_confirmed_place` for now to avoid the backwards compat issues and to make the other changes later; what is the ETA for that change?
> 
>    - It seems like it would be a simple find/replace and should be pretty easy to implement...

This seems very doable! I will have this change done before 1pm MST tomorrow and then we can evaluate the `end_confirmed_place` timeline after that.

shankari:
> I will have this change done before 1pm MST tomorrow and then we can evaluate the end_confirmed_place timeline after that.

You mean "evaluate the `start_confirmed_place` timeline, right?
By 1pm MST tomorrow, we will have the `end_confirmed_place` done.

JGreenlee:
@sebastianbarry We can look at it together tomorrow. I don't think it will take as much work as you think.

shankari:
@JGreenlee I want to make sure that at least the `end_confirmed_place` is done and tested by tomorrow. We can make other changes later, incrementally.

shankari:
superceded by #951, closing

PR #944, opened 2023-04-05
Fixed positioning issue on Android emulator

sebastianbarry:
style.css
- "translate: ..." is not a property detected by the old Android webkit, so we will use the classic notation "transform: translate(...)"

Testing: 
| Before | After |
:---:|:---:
| Screen Shot 2023-04-05 at 10 37 21 AM | image |

@shankari for visibility

### DIFF
--- a/www/css/style.css
+++ b/www/css/style.css
@@ -947,13 +947,13 @@ button.button.back-button.buttons.button-clear.header-item {
   background: #80D0FF;
   color: black;
   top: 0;
-  translate: -50% -50%;
+  transform: translate(-50%, -50%);
 }
 .stop-time-tag {
   background: #0088ce;
   color: white;
   bottom: 0;
-  translate: -50% 50%;
+  transform: translate(-50%, 50%);
 }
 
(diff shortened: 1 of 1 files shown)


shankari:
@sebastianbarry @JGreenlee

I don't even see what the issue is here, and we don't need to support the old android webkit anyway since the old webkit doesn't even show places properly.

Do I need to merge this change, given that we are at RC1 or can we skip?

JGreenlee:
It probably doesn't matter in real life, but the shorthand is a *little* new (widespread adoption around 2020). The safest thing would be to merge this

shankari:
I will merge it after pushing to production. If it is not broken, it isn't going to be fixed before the point release.

JGreenlee:
Note that the issue of places not showing is for **very** old webkit versions (like >5 years out of date)

This issue could affect **moderately** old versions (just a few years out of date).
Sebastian's emulator was showing this issue, while places displayed fine.

The first is because of ES6 support, the second is because of new CSS shorthands that I've gotten used to and instinctually use now.

PR #947, opened 2023-04-05
fix overlapping trip/place cards on android

JGreenlee:
Because of rendering differences between Android and iOS webviews, the hardcoded values here do not necessarily match the exact rendered height. It seems that all UI elements render much bigger on some, but not necessarily all, Android devices (it depends on screen size).

Ideally, our UI would be responsive and accommodate for this rather than relying on specific pixel values.
But for right now, I'm just going to leave an extra 15px as a buffer. This is a compromise between A) too close together on some Android and B) too far apart on iOS

<table>
<tr>
<td>
Android after fix:

</td>
<td>
iOS after fix:

</td>
</td>
</table>

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -75,17 +75,17 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
   }
 
   $scope.getCardHeight = function(entry) {
-    let height = 0;
+    let height = 15; // 15 pixels of padding to account for iOS/Android rendering differences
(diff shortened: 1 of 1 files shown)


shankari:
I hate hardcoded values and tried really hard to push back against them when previous interns tried to propose them.
But they couldn't figure out alternatives, and we had to finish the changes, so I reluctantly agreed.
Hopefully this will get better when we finally switch to React Native.

JGreenlee:
>Hopefully this will get better when we finally switch to React Native.

It will!

PR #956, opened 2023-04-24
Sort additions coming from the server

sebastianbarry:
input-matcher.js
- Instead of just returning the list of MatchingAdditions, we now sort first by start_ts and then by end_ts of the additions and return the list of sorted matching additions
- Added a compare function for this

PR #959, opened 2023-04-26
Sort additions

sebastianbarry:
This is an updated PR for the correct base from https://github.com/e-mission/e-mission-phone/pull/956

@shankari I can't seem to get the `compareTimestamps` function to work for `ng-repeat`'s `orderBy`. I followed this guide https://docs.angularjs.org/api/ng/filter/orderBy but it looks like it chooses to just sort by start_ts regardless of if I add in a function or not

### DIFF
--- a/www/js/survey/enketo/enketo-notes-list.js
+++ b/www/js/survey/enketo/enketo-notes-list.js
@@ -22,15 +22,6 @@ angular.module('emission.survey.enketo.notes-list',
   .controller("NotesListCtrl", function ($scope, $state, $element, $window, EnketoSurveyLaunch, $ionicPopup) {
     console.log("Notes List Controller called");
 
-    $scope.compareTimestamps = (a, b) => {
-      if (!a || !b) 
-        return;
-      if (a.data.start_ts == b.data.start_ts) {
-        return a.data.end_ts - b.data.end_ts;
-      }
-      return a.data.start_ts - b.data.start_ts;
-    }
-
     const getScrollElement = function() {
       if (!$scope.scrollElement) {
           const ionItemElement = $element.closest('ion-item')
--- a/www/js/survey/enketo/enketo-notes-list.js
+++ b/www/js/survey/enketo/enketo-notes-list.js
@@ -22,6 +22,15 @@ angular.module('emission.survey.enketo.notes-list',
   .controller("NotesListCtrl", function ($scope, $state, $element, $window, EnketoSurveyLaunch, $ionicPopup) {
     console.log("Notes List Controller called");
 
+    $scope.compareTimestamps = (a, b) => {
+      if (!a || !b) 
+        return;
+      if (a.data.start_ts == b.data.start_ts) {
+        return a.data.end_ts - b.data.end_ts;
+      }
+      return a.data.start_ts - b.data.start_ts;
+    }
+
     const getScrollElement = function() {
       if (!$scope.scrollElement) {
           const ionItemElement = $element.closest('ion-item')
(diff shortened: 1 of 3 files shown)


sebastianbarry:
image

See how this newly added entry **should** go here but it only gets sorted by start_ts

shankari:
@sebastianbarry did you look up the documentation/search stackoverflow for `orderBy`?
I don't think you even need your function.

shankari:
Since we are running out of time here:
https://stackoverflow.com/questions/17037524/orderby-multiple-fields-in-angular

shankari:
@sebastianbarry can you please revert the other changes since they are not necessary?

sebastianbarry:
(commented in wrong PR before: ) This change is complete, for both
1. Unprocessed additions
2. Additions coming from the server

Testing: 

https://user-images.githubusercontent.com/61334340/234988415-697d6dcd-95a3-4db2-949c-6dc8f35cca85.mov

sebastianbarry:
> can you please revert the other changes since they are not necessary?

@shankari I'm not quite sure what this means, can you explain how I would revert the changes? I don't have anything committed to this branch that doesn't belong

shankari:
@sebastianbarry not sure if you saw my comment - you need to remove the javascript changes since they are not required.

shankari:
> @shankari I'm not quite sure what this means, can you explain how I would revert the changes? I don't have anything committed to this branch that doesn't belong

There are three changes here. The HTML changes are all that is required. You are not using the javascript changes - they are just dead code and should be removed.

www/js/survey/enketo/enketo-notes-list.js
```diff
@@ -22,6 +22,15 @@ angular.module('emission.survey.enketo.notes-list',
...
 
+    $scope.compareTimestamps = (a, b) => {
+      if (!a || !b) 
+        return;
+      if (a.data.start_ts == b.data.start_ts) {
+        return a.data.end_ts - b.data.end_ts;
+      }
+      return a.data.start_ts - b.data.start_ts;
+    }
+
```
sebastianbarry, line 33:
```suggestion
```

sebastianbarry:
Ah I see what you're saying! Removed the unneccesary `enketo-notes-list.js` code, the other javascript in `input-matcher.js` is needed to process the additions coming in from the server, and the unprocessed additions are sorted in the template `notes-list.html`

shankari:
> the unprocessed additions are sorted in the template notes-list.html

This is not correct. *all* additions are sorted in the `template notes-list.html`

I don't see any distinction in `ng-repeat="entry in additionEntries | orderBy: ['data.start_ts', 'data.end_ts']">` between processed and unprocessed entries.

sebastianbarry:
I tested this as well:

| https://user-images.githubusercontent.com/61334340/235009347-0a48f592-e3b1-4fd8-afe7-ae3f6b5c44a8.mov | FIrst addition getting added, start time is changed and it gets sorted to the correct location |
:---:|:---:
| https://user-images.githubusercontent.com/61334340/235009404-4ae6267b-b0f6-4ca9-990c-adfe125e5825.mov | Second addition getting added, start time and end time are changed and it is sorted to the correct location: **Unprocessed additions get sorted** |
| https://user-images.githubusercontent.com/61334340/235009449-8b3b8e3c-39fd-4de9-b171-1c2a4ea8bca7.mov | After refreshing, additions stay sorted properly despite being input in the "incorrect" order: **Processed additions get sorted** |

Success!

sebastianbarry:
Also testing:

| | |
:---:|:---:
| Simulator Screen Shot - iPhone 11 - 2023-04-27 at 17 10 14 | After `end trip + sync` and then running the pipeline again, additions stay in the correct spots and are ordered correctly! |

shankari (approved):

PR #960, opened 2023-04-28
Descriptive error 403 messages

sebastianbarry:
services.js
- Added processErrorMessages function which only checks for error 403 at the moment
- Added catch blocks for most of the functions that perform server calls

### DIFF
--- a/www/js/diary/infinite_scroll_list.js
+++ b/www/js/diary/infinite_scroll_list.js
@@ -212,11 +212,7 @@ angular.module('emission.main.diary.infscrolllist',['ui-leaflet',
         $scope.$broadcast('scroll.infiniteScrollComplete');
 
     }).catch((err) => {
-        if (err.includes("403")) {
-          Logger.displayError("Invalid OPcode: while reading confirmed trips", err)
-        } else {
-          Logger.displayError("while reading confirmed trips", err);
-        }  
+        Logger.displayError("while reading confirmed trips", err);
         Logger.log("Reached the end of the scrolling");
         $scope.infScrollControl.reachedEnd = true;
         Logger.log("Broadcasting infinite scroll complete");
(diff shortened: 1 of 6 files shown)


sebastianbarry:
This is the updated PR to https://github.com/e-mission/e-mission-phone/pull/914

sebastianbarry:
These are the only error messages I was able to replicate with the `error 403: Token not found on the server` error:

| While updating user, | While getting pipeline range timestamps, | While updating user, (separate location) |
:---:|:---:|:---:
| Simulator Screen Shot - iPhone 11 - 2023-04-28 at 10 07 15 | Simulator Screen Shot - iPhone 11 - 2023-04-28 at 10 07 35 | Simulator Screen Shot - iPhone 11 - 2023-04-28 at 10 08 03 |

shankari:
@sebastianbarry are you planning to finish this PR before you stop working for NREL?

sebastianbarry:
Here are the changes for the title of the error message to display `"Invalid OPcode: [original error title]"`

| | | |
:---:|:---:|:---:
| Simulator Screen Shot - iPhone 11 - 2023-04-28 at 15 43 23 | Simulator Screen Shot - iPhone 11 - 2023-04-28 at 15 43 25 | Simulator Screen Shot - iPhone 11 - 2023-04-28 at 15 43 34 |

@shankari  what do you think of this change?

shankari:
@sebastianbarry LGTM! go ahead and push those changes. I think that's it for this PR

sebastianbarry:
Now the code is much cleaner while being modular and displaying the same result and functionality

shankari:
@sebastianbarry this doesn't affect you any more, but FYI, I am squash-merging this, so you won't be able to use this branch any more. Hopefully, by the time you are back, the phone codebase will be completely different.

shankari:
For the record, since we have made the change in `displayError`, I don't think we need all the catch blocks in `service.js`.
but it doesn't hurt anything, and we will be changing this anyway as part of the rewrite.
Maybe I will try to remove them in a future commit.

shankari (approved):

PR #961, opened 2023-04-28
Dynamic Config Start Page

sebastianbarry:
startprefs.js
- Added a check that if the study/program is using ENKETO, we will automatically take them to inf_scroll
- Otherwise, we will read the config for the startup_page and take them to that
- If no startup_page field exists in the config, we will take them to inf_scroll

### DIFF
--- a/www/js/splash/startprefs.js
+++ b/www/js/splash/startprefs.js
@@ -205,8 +205,14 @@ angular.module('emission.splash.startprefs', ['emission.plugin.logger',
             $rootScope.redirectTo = undefined;
             $rootScope.redirectParams = undefined;
             return {state: redirState, params: redirParams};
-          } else {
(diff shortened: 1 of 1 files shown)


sebastianbarry:
Here is how I setup the config file https://raw.githubusercontent.com/sebastianbarry/nrel-openpath-deploy-configs/starup-page-for-surveys-info-and-surveys-data/configs/dev-emulator-study.nrel-op.json

I placed the new field `startup_page` as a child of `display_config`. The field should be filled in as the `root.main` child, either `metrics`, `diary`, `inf_scroll`, etc.

shankari:
> However this doesn't seem to be working with MULTILABEL, and reveals a larger issue:

What is the larger issue?

> I am able to get this to work for ENKETO, since:

It might just be working for enketo since there is no metrics screen unless we are in multi-label

shankari:
Closing this since we have embarked on the new rewrite

PR #962, opened 2023-05-03
Support overnight activities

JGreenlee:
This PR allows overnight additions to be added, by resolving both "Start Date" and "End Date" from survey responses instead of only one "Date". Note that this only works if the survey itself includes the new fields.

This also fixes a few regressions that were introduced last week, and makes some styling tweaks since the cards now list out the dates of multi-day entries.

<table>
<tr>
<td>
Overnight place activities
<br>

</td>
<td>
Overnight untracked
<br>

</td>
</tr>

<tr>
<td>
Survey for overnight trip/place
<br>

</td>
<td>
Survey for same-day trip/place
<br>

</td>
</tr>
</table>

### DIFF
--- a/www/js/diary/services.js
+++ b/www/js/diary/services.js
@@ -25,9 +25,17 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
   //   document.querySelector('#hidden-' + id.toString()).parentElement.parentElement.parentElement
   //   .setAttribute('style', 'width: '+oldVal2);
   // }
-  dh.getFormattedDate = function(ts) {
-    var d = moment(ts * 1000).format("DD MMMM YYYY");
-    return d;
+
+  dh.isMultiDay = function(beginTs, endTs) {
+    if (!beginTs || !endTs) return false;
+    return moment(beginTs * 1000).format('YYYYMMDD') != moment(endTs * 1000).format('YYYYMMDD');
+  }
+
+  dh.getFormattedDate = function(beginTs, endTs) {
+    if (dh.isMultiDay(beginTs, endTs)) {
+      return `${moment(beginTs * 1000).format('ddd LL')} - ${moment(endTs * 1000).format('ddd LL')}`;
+    }
+    return moment((beginTs || endTs) * 1000).format('ddd LL');
   }
   dh.isCommon = function(id) {
     var ctrip = CommonGraph.trip2Common(id);
@@ -144,14 +152,22 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
       return background;
   }
 
-  dh.getLocalTimeString = function (dt) {
+  dh.getLocalTimeString = function (dt, includeDay=false) {
+    if (!dt) return;
     //correcting the date of the processed trips knowing that local_dt months are from 1 -> 12 and for the moment function they need to be between 0 -> 11
     let mdt = angular.copy(dt)
     mdt.month = mdt.month - 1
+    if (includeDay) {
+      const opts = { weekday: 'short', month: 'short', day: 'numeric' };
+      const date = Intl.DateTimeFormat($translate.use(), opts)
+        .format(moment(mdt).toDate());
+      return `${moment(mdt).format("LT")} (${date})`;
+    }
     return moment(mdt).format("LT");
   };
 
   dh.getFormattedTime = function(ts_in_secs) {
+    if (isNaN(ts_in_secs)) return;
     if (angular.isDefined(ts_in_secs)) {
       return moment(ts_in_secs * 1000).format('LT');
     } else {
@@ -159,11 +175,13 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
     }
   };
   dh.getFormattedTimeRange = function(end_ts_in_secs, start_ts_in_secs) {
+    if (isNaN(end_ts_in_secs) || isNaN(start_ts_in_secs)) return;
     var startMoment = moment(start_ts_in_secs * 1000);
     var endMoment = moment(end_ts_in_secs * 1000);
     return endMoment.to(startMoment, true);
   };
   dh.getFormattedDuration = function(duration_in_secs) {
+    if (isNaN(duration_in_secs)) return;
     return moment.duration(duration_in_secs * 1000).humanize()
   };
   dh.getTripDetails = function(trip) {
--- a/www/templates/diary/place_list_item.html
+++ b/www/templates/diary/place_list_item.html
@@ -1,11 +1,11 @@
 <div ng-if="configPlaceNotes">
     <!-- Place Item Template  -->
     <ion-item class="unified-diary-item" style="background-color: transparent;" class="list-item">
-        <div class="diary-card short" style="padding-block: 5px;">
+        <div class="diary-card short" style="padding-block: 12px">
             <div class="start-time-tag">{{place.display_start_time}}</div>
-            <div style="width: 100%; margin: auto; padding-block: 5px">
+            <div style="width: 100%; margin: auto">
                 <div class="row" style="font-size: 12px; place-content: center; padding: 8px">
-                    <b><u>{{place.display_date}}</u></b>
+                    <b class="card-date">{{place.display_date}}</b>
                 </div>
                 <div class="card-title">
                     <!-- Place (Destination of Prev. Trip) -->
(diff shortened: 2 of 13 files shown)


www/js/diary/services.js
```diff
@@ -144,26 +174,33 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
       return background;
   }
 
-  dh.getLocalTimeString = function (dt) {
+  dh.getLocalTimeString = function (dt, includeDay=false) {
+    if (!dt) return;
```
shankari, line 178: I don't like adding on the day to what is supposed to be the time string. We should handle this in the HTML. If there is a reason why that won't work/is awkward, please document it.

shankari: Where do we even pass in `includeDay=true`? The only two calls seem to be around here 
https://github.com/e-mission/e-mission-phone/pull/962/files#diff-3692dc7185f771351204f053be118ba7122c58fc2a51f2c0d50ae838e3a97351R436

and neither of them seem to pass in this argument.

shankari: What am I missing here?

www/templates/survey/enketo/notes_list.html
```diff
@@ -1,9 +1,12 @@
...
     <ion-item class="notes-list" style="background-color: transparent">
-        <ul>
-            <li class="notes-list-entry" ng-repeat="entry in additionEntries | orderBy: ['data.start_ts', 'data.end_ts']">
+        <ul class="notes-list-grid">
+            <li ng-repeat="entry in additionEntries | orderBy: ['data.start_ts', 'data.end_ts']">
                 <b class="two-lines" ng-click="editEntry(entry)">{{entry.data.label}}</b>
-                <i ng-click="editEntry(entry)">{{entry.displayTime || setDisplayTime(entry)}}</i>
+                <i ng-click="editEntry(entry)">
+                    <p>{{ entry.displayDt.date }}</p>
+                    <p>{{ entry.displayDt.time || setDisplayDt(entry).time }}</p>
```
shankari, line 8: maybe I should have noticed this earlier, but I don't quite understand what 
`{{ entry.displayDt.time || setDisplayDt(entry).time }}` does

It looks like it lazily populates the displayDt, but then why don't you need it for

`{{ entry.displayDt.date }}` as well?

Is it because the date is only displayed for multi-day cards while the time is always displayed? And when we call `setDisplayDt`, the date and time will both be displayed.

I think that is it; but you might want to add a comment to make it easier for others to read the code in the future.

shankari (changes requested):
There's only one change that I really want you to make - the one in getFormattedTimeString. The others are mainly requests for comments, or optional refactoring.

www/js/diary/services.js
```diff
@@ -25,10 +25,40 @@ angular.module('emission.main.diary.services', ['emission.plugin.logger',
...
+    let t = beginTs || endTs;    // whichever is defined. may be timestamp or dt object
+    if (typeof t == 'number') t = t*1000; // if timestamp, convert to ms
+    if (!t._isAMomentObject) t = moment(t);
+    return t.format('ddd LL');
+  }
+
+  /* returns a formatted range if both params are defined, 
+    one formatted date if only one is defined */
+  dh.getFormattedDateAbbr = function(beginTs, endTs=null, dayOfWeek=true) {
+    if (!beginTs && !endTs) return;
```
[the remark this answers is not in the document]
JGreenlee: Yeah, I originally had it differently and settled on this.
For the rewrite I suspect we'll be using something other than Moment anyway, and hopefully will need fewer helper functions like this

www/js/diary/services.js
JGreenlee: You're right, I can take out that unused parameter.
it's because I originally used this option to just append the date, but then decided to refactor that functionality into `getDisplayDateAbbr()` and store it in separate fields.

www/templates/survey/enketo/notes_list.html
JGreenlee: yes, that's why. Will add comment

www/templates/diary/place_list_item.html
```diff
@@ -1,11 +1,14 @@
...
     <!-- Place Item Template  -->
     <ion-item class="unified-diary-item" style="background-color: transparent;" class="list-item">
-        <div class="diary-card short" style="padding-block: 5px;">
-            <div class="start-time-tag">{{place.display_start_time}}</div>
-            <div style="width: 100%; margin: auto; padding-block: 5px">
+        <div class="diary-card short" style="padding-block: 12px">
+            <div class="start-time-tag">
+                <b>{{place.display_start_time}}</b>
+                <i ng-if="place.display_start_date_abbr">&nbsp;({{place.display_start_date_abbr}})</i>
+            </div>
```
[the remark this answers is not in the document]
JGreenlee: I was thinking about that too. if not now, will definitely do it for the rewrite

JGreenlee:
Addressed the review, except https://github.com/e-mission/e-mission-phone/pull/962#discussion_r1184556881 which will be simplified later

JGreenlee:
Because the new survey has different time/date fields ("Start Date" and "End Date"), it is not compatible with the old phone code, and this PR is not compatible with the old survey.

So, when this is ready to merge, how should we go about upgrading the survey? (And in general, how do we expect survey updates to work, when necessary?)

If I upgrade the survey in-place, (ie. [update it on sebastian's branch](https://github.com/sebastianbarry/nrel-openpath-deploy-configs/blob/surveys-info-and-surveys-data/survey-resources/data-json/time-use-survey-form-v9.json)), the old phone code will load the new survey but no longer be able to record responses.

And if I upload the new survey somewhere else and update the config to point to the new location, the new config won't be loaded unless the app is deleted and reinstalled (unless it has the `autoreload` dev setting).

If need be, I can make this PR backwards-compatible with the old timeuse survey, but we still have to address the issue of when the survey is going to be upgraded.

shankari:
> So, when this is ready to merge, how should we go about upgrading the survey? (And in general, how do we expect survey updates to work, when necessary?)

I think that at this point, we assume that there will not be survey updates. Note that the current time-use survey is just a placeholder while @MaliheTabasi designs a new survey in conjunction with Taha and the rest of the UNSW team.
The time use/survey use case is to spend a lot of time designing the perfect survey and then deploying it for a week
Unlike the multilabel use case where the use case is to collect longitudinal data.

@asiripanich for visibility

> If I upgrade the survey in-place, (ie. [update it on sebastian's branch](https://github.com/sebastianbarry/nrel-openpath-deploy-configs/blob/surveys-info-and-surveys-data/survey-resources/data-json/time-use-survey-form-v9.json)), the old phone code will load the new survey but no longer be able to record responses.

The survey functionality has only been currently deployed on staging, and we don't make any backwards compatibility guarantees on staging. Once we push this change out, we can just send a notification to all staging users to delete and reinstall the app if they want to see the new functionality (or just send email to the UNSW team).

JGreenlee:
Ok, then tomorrow I'll make a PR to [nrel-openpath-deploy-configs](https://github.com/e-mission/nrel-openpath-deploy-configs) to complement this one, which will update the config to reference the new version of the survey.

shankari:
Also, as I was thinking about this today evening, it seems like just updating the survey would also work.

So the scenarios are:
- right now:
   - old app, old survey: works
- push out new app, upgrade the survey **at the same location** to the new one
- after new app upgrade:
   - config is cached, but it downloads the survey on the fly and does not cache it locally IIRC
   ```
      .then(() => $http.get(_state.formLocation))
      .then(formJson => {
        _state.loaded.form = formJson.data.form;
        _state.loaded.model = formJson.data.model;
      });
   ```
   = new app, new survey: works

This will only be broken for the really (hopefully really small) window between when we push out the release and when apple/google apply the rolling updates.

JGreenlee:
Sure, that works fine if we are comfortable assuming that the app will update timely / automatically.

Then, there's no need to reload config via delete/reinstall

shankari:
> we are comfortable assuming that the app will update timely / automatically

I am fine with a staging app being broken for a day or two; and asking people to check and update the app if it is broken.
And as I said, on production, we don't anticipate that the survey will change during the course of a study.
If the survey does change, then old responses will not be loaded anyway (form and data will be incompatible).

shankari (approved):

PR #964, opened 2023-05-04
Add popup to label screen if incorrect permissions

JGreenlee:
Strings need to be internationalized


JGreenlee:
I modified `permissioncheck.js` so that it will give a give a promise-based callback when 'recomputeAppStatus' is broadcast.

While doing this, one significant issue I noticed is that in `permissioncheck.js`, `$scope.backgroundRestrictionChecks` is set for Android but not for iOS.
This meant that every time `recomputeBackgroundRestrictionStatus()` was called, it would hit an error. Thus, I added a check for the existence of `$scope.backgroundRestrictionChecks`.

JGreenlee:
Tests performed:

-- Install app with correct settings, open label screen. No warning appears.

-- Close app, go to system settings and deny any permission.
-- Reopen app. Warning popup appears. Clicking "Fix" directs to "Permissions" popup over profile tab.

JGreenlee:
Issue noted: after following the above and fixing permissions, any subsequent time navigating to the Profile tab opens the Permissions modal again.

JGreenlee:
Spanish translations
https://github.com/e-mission/e-mission-translate/pull/25

shankari:
> Issue noted: after following the above and fixing permissions, any subsequent time navigating to the Profile tab opens the Permissions modal again.

@JGreenlee Does this happen even after the app is killed and started?

shankari:
>  This meant that every time recomputeBackgroundRestrictionStatus() was called, it would hit an error. Thus, I added a check for the existence of $scope.backgroundRestrictionChecks.

I also noticed this while catching every single error (so that I could see where the notification was being swallowed). I fixed it too, but in a slightly different way and will use mine while merging.

JGreenlee:
Fixed anyway

shankari:
@JGreenlee this is still fairly ugly; primarily because the permissioncheck code is not structured very well. It has all the business logic in the controller instead of a service. I am merging this now, but I think we should consider whether we want to restructure it before the rewrite. But there is a long-term issue for this anyway, so let me get this out first.

JGreenlee:
Agreed. I didn't want to go through the process of making it a service today since it's a quick fix

shankari:
There is another minor issue, which I want to document before I forget.
The following sequence doesn't work as expected on iOS:
- launch the app with incorrect settings
- click on the popup, go to the profile screen, fix it
- go back to the settings and make them incorrect
- get the "click to fix me" popup
- click on it; we go back to the app on the profile screen
   - since it is the profile screen, we don't check the status
   - since we already closed the app status popup, we don't open it again

This is a very unlikely case, and I am very reluctant to make more changes at the last minute. So we will punt on this for now.

PR #965, opened 2023-05-04
Minor fixes found while testing local notifications

shankari:
Testing done: Refreshed the app with a debugger connected and breakpoints on all exceptions

Before these changes:
- hit exceptions

After these changes:
- did not hit exceptions


shankari:
@JGreenlee would you like to review this given that we are going to push it out fairly quickly?

JGreenlee:
fix in `permissioncheck.js`, LGTM

fix in `pushnotify.js`, I don't fully understand because I didn't see those exceptions. Seems fine just by looking though

shankari:
> fix in pushnotify.js, I don't fully understand because I didn't see those exceptions. Seems fine just by looking though

you will only get them if you turn off permissions, get the "click to fix" notification and click on it

PR #967, opened 2023-05-05
⬆️ 🚸 Improve usability 

shankari:
- remove extra unnecessary error (OPGeofenceCfg)
- make the app settings message more positive


shankari:
Related UI spanish change: https://github.com/e-mission/e-mission-translate/commit/9fd85eb684a1553390e901fe1e6ba7279bab6eb5
Related native code changes in English: https://github.com/e-mission/e-mission-data-collection/pull/205
Related native code changes in Spanish: https://github.com/e-mission/e-mission-translate/pull/26

I am not bumping up the plugin version at this time since I do not want to go through the hassle of making a new release, and when I plan to have a new update for the bluetooth integration soon anyway.

shankari:
@JGreenlee would you like to review this before I merge?
Would be good to have reviews for *all code changes* at least the ones during business hours!

www/js/app.js
```diff
@@ -48,6 +48,11 @@ angular.module('emission', ['ionic',
       StatusBar.styleDefault();
     }
     cordova.plugin.http.setDataSerializer('json');
+    // backwards compat hack to be consistent with
+    // https://github.com/e-mission/e-mission-data-collection/commit/92f41145e58c49e3145a9222a78d1ccacd16d2a7#diff-962320754eba07107ecd413954411f725c98fd31cddbb5defd4a542d1607e5a3R160
+    // remove during migration to react native
+    storage.remove("OP_GEOFENCE_CFG");
```
JGreenlee, line 54: Not sure I understand this.
Are we removing the geofence config because, as referenced, we now "turn OPGeofence on all the time"?

www/i18n/en.json
```diff
@@ -56,7 +56,7 @@
         "username": "Username {{usernamedata}}",
         "tracking": "Tracking",
         "app-status": "App Status",
-        "incorrect-app-status": "Incorrect app settings",
+        "incorrect-app-status": "Please update permissions",
```
JGreenlee, line 59: 👍

JGreenlee (commented):
The string change is good, not sure I understand the OPGeofence change though

www/js/app.js
shankari: Yes. Before this, the geofence was opt-in with a control toggle in the app (https://github.com/e-mission/e-mission-phone/commit/6f544cf793c460116106edc154a37e3052bc7a0f). The config was only "true" or "false". After field testing with it on, we decided that it was actually better and did not raise battery drain too much. So we removed it.

shankari: The commit message has a lot more detail.

PR #968, opened 2023-05-05
🚸 ✨  Multiple time-use and usability fixes

shankari:
- ✨  Add start place
- ✨  Order activities
- 🚸 403 error checking
- 🐛 fix regressions in error checking and start place
- ✨  handle overnight activities
- 🔧 actually decide when to show the replaced mode button based on the dynamic config
- ⬆️ remove the OPGeofenceCfg from the local storage as well so it doesn't get copied over every time
- 🚸 make the permission error message more positive

This deploys https://github.com/e-mission/e-mission-docs/issues/887 and 
https://github.com/e-mission/e-mission-docs/issues/884 and https://github.com/e-mission/e-mission-docs/issues/889
and a couple of other minor fixes without issues.


shankari:
Testing done:
- Merged everything, loaded some data, everything worked
- Pushing to staging for additional testing

PR #973, opened 2023-05-19
♻️ ✨ Reconfigure token flow

shankari:
[♻️ Change the join page to support entering the token directly](https://github.com/e-mission/e-mission-phone/commit/054f95a8e190035f3a08c3dfd8ee43999f46414b)

Testing done:

| Reconfigured join page | Entering the token |
|-------|------|
| Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 55 08 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 55 39 |


shankari:
[✨ configure the app using a study name inferred from the token](https://github.com/e-mission/e-mission-phone/pull/973/commits/e7105752e27891503935479051ffcbc75b4fb1f0)

Testing done:

|  |  |  |
|------|-------|------|
| Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 23 27 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 23 47 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 24 08 | 
| Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 25 44 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 26 05 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-18 at 22 27 37 |

shankari:
@JGreenlee I'd like a second opinion on usability. Concretely, from 
[✨ configure the app using a study name inferred from the token](https://github.com/e-mission/e-mission-phone/pull/973/commits/e7105752e27891503935479051ffcbc75b4fb1f0)

> Question:
> - Should we even support links any more? The link option is the only one that doesn't have a button in the app, and it won't work to send the link via email, text or printing out anyway.
>   - The only case where the link is useful is in the autogenerated option on
      the join webapp, and then users still need to go back to the webpage to
      click on the link
>   - Instead, we can just display the token as text with a "copy" button right
      next to it and simplify the codebase...

JGreenlee:
I agree that having 3 options is unnecessary and we should simplify – but to me the link still seems preferable to copy-pasting. It is a bit more user-friendly and cleaner in terms of presentation.
Not everyone knows how to paste text on their phone (e.g. my parents, who are not tech savvy in the slightest)

If we want to simplify these 3 options down to 2, I would suggest removing the manual entry text box and only using QR codes / links.

And don't those 2 methods share the same logic anyway?

shankari:
@JGreenlee the reason that we don't support the link is because you cannot email/SMS it to people. Due to security reasons, at least gmail, and most likely other email software, will not open non-HTTPS URLs in email messages. The only option is to create a website with the URL and have people click it there, which was the reason for the join page in the first place. I didn't want to make people type out long strings, and there wasn't an alternative approach.

You can try this - send nrelopenpath://login_token?nrelop_denver-casr_unrestricted-regular_i8BwZTNmSMpII7WV4453fV47jQPAgTdv to your personal gmail and try to access it from your phone.

If this has changed, or you can get it to work somehow, I am fine with removing the copy-paste option.

JGreenlee:
Ah, TIL

(I did find out the actual way to do this is "Universal Links" on iOS and "App Links" on Android - which is what social media / other big apps do
https://www.branch.io/resources/blog/deep-linking-from-url-and-uri-schemes-to-universal-links-app-links-and-beyond/)

But that's fairly involved and not worth our time, so we'll have to hope people can figure out at least 1 of the 2 options between QR and copy-pasting. Still a big improvement over the old flow!

JGreenlee:
> Any other text that needs to be changed/clarified?

It's a little confusing to me that the blue button says "Copy and paste the OPCode". I think it should just say "Paste the OPCode", because the copying happens elsewhere / beforehand

shankari:
> But that's fairly involved and not worth our time

Certainly not at this time. We can revisit after the big rewrite.

> so we'll have to hope people can figure out at least 1 of the 2 options between QR and copy-pasting. Still a big improvement over the old flow!

Just to clarify, I should remove the link and only support QR code (displayed on laptop or piece of paper) or copy-paste (from email or SMS on mobile phone), right?

> It's a little confusing to me that the blue button says "Copy and paste the OPCode". I think it should just say "Paste the OPCode", because the copying happens elsewhere / beforehand

Will do!

shankari:
Testing done:

| Join page | Incorrect login | Correct login |
|------|------|------|
| simulator_screenshot_96FA3D73-A5B4-436B-AA4A-B17E72F4849C | simulator_screenshot_BC63802A-4FA9-4EC5-A59B-83F7FD5D68D7 | simulator_screenshot_4FA80DC7-3167-4595-B609-85920A26353D |

After logging in

```
START 2023-05-19 16:52:49.030228 POST /profile/create
END 2023-05-19 16:52:49.041183 POST /profile/create  0.010889053344726562
START 2023-05-19 16:52:49.135295 POST /datastreams/find_entries/timestamp
END 2023-05-19 16:52:49.156530 POST /datastreams/find_entries/timestamp 6afb9865-c48b-4d35-a64c-c94f5a9da5cb 0.021172046661376953
```

shankari:
@JGreenlee would you like to review this?

www/js/config/dynamic_config.js
```diff
@@ -164,23 +164,93 @@ angular.module('emission.config.dynamic', ['emission.plugin.logger'])
...
+            if (config.opcode.subgroups) {
+                if (config.opcode.subgroups.indexOf(tokenParts[2]) == -1) {
+                // subpart not in config list
+                    throw new Error("Invalid opcode, subgroup '"+tokenParts[2]+"' not found in list '"+config.opcode.subgroups+"'");
+                } else {
+                    console.log("subgroup "+tokenParts[2]+" found in list");
+                    return tokenParts[2];
+                }
+            } else {
+                if (tokenParts[2] != "default") {
```
JGreenlee, line 207: So all OPCodes from now on **must** include a subgroup?
But if subgroups aren't being used, it must be "default"?

www/js/config/dynamic_config.js
```diff
@@ -164,23 +164,93 @@ angular.module('emission.config.dynamic', ['emission.plugin.logger'])
...
+                }
+            }
+        } else {
+            /* old style study, expect token without subgroup
+             * nothing further to validate at this point
+             * only validation required is `nrelop_` and valid study name
+             * first is already handled in extractStudyName, second is handled
+             * by default since download will fail if it is invalid
+            */
+            console.log("Old-style study, expecting token without a subgroup...");
```
JGreenlee, line 222: Ah, I see - If `config.opcode` is not present then we skip.

www/js/config/dynamic_config.js
```diff
@@ -164,23 +164,93 @@ angular.module('emission.config.dynamic', ['emission.plugin.logger'])
         }
     }
 
+    /*
+     * We want to support both old style and new style tokens.
```
JGreenlee, line 168: So old-style tokens have 3 parts, and new style tokens have 4 parts?

JGreenlee (commented):
LGTM

www/js/config/dynamic_config.js
shankari: All new style tokens, yes.

www/js/config/dynamic_config.js
shankari: Yes. Because there are tokens out there (e.g. to smart commute, pueblo, etc) that are in old format.
Once all the old style programs are done, we can remove that backwards compatibility

shankari:
@shankari  (note to self) don't forget to translate all the text changes to Spanish

shankari:
These are the phone changes for https://github.com/e-mission/e-mission-docs/issues/850#issuecomment-1532420403

PR #977, opened 2023-05-21
Hide untracked time in "To Label" if previous or next trip has been labeled

JGreenlee:
https://github.com/e-mission/e-mission-docs/issues/907
We'd like for untracked time to not remain in "To Label" after trips have been labeled.
With this PR, untracked time will only show if the trips before and after it are showing. Once either is labeled, the untracked time will disappear from "To Label"


shankari (approved):
Good job taking care of the number of trips as well!

PR #976, opened 2023-05-21
🧹 Configurable Notification Schedules (follow-up)

JGreenlee:
- ensures NotificationScheduler is initialized from the label screen, not just the profile screen
- ensures scheduling only happens if `reminderSchemes` is present in dynamic config
- hides 'Time of Day' setting from Profile tab if NotificationScheduler is not being used
- allows notification text to be internationalized

PR #979, opened 2023-05-23
Usability Fixes (May 2023)

JGreenlee:
Addresses the following issues:

- https://github.com/e-mission/e-mission-docs/issues/908
- https://github.com/e-mission/e-mission-docs/issues/909
- https://github.com/e-mission/e-mission-docs/issues/914

Also includes:
- https://github.com/e-mission/e-mission-phone/pull/978. I am incorporating it into this PR and closing the original.


JGreenlee:
Also addressed:
- https://github.com/e-mission/e-mission-docs/issues/911

#### And I included a logout confirmation popup

<br>
<br>

The new strings for that are internationalized here:
- https://github.com/e-mission/e-mission-translate/pull/29

www/js/config/dynamic_config.js
```diff
@@ -99,14 +99,15 @@ angular.module('emission.config.dynamic', ['emission.plugin.logger'])
...
-                .catch((storeError) => Logger.displayError($translate.instant('config.unable-to-store-config'), storeError));
+            return storeConfigPromise.then((result) => {
+                logSuccess(result);
+                dc.saveAndNotifyConfigChanged(downloadedConfig);
+                dc.saveAndNotifyConfigReady(downloadedConfig);
+                if (thenGoToIntro)
+                    $state.go("root.intro");
+                return true;
+            }).catch((storeError) =>
+                Logger.displayError($translate.instant('config.unable-to-store-config'), storeError));
```
shankari, line 110: Good point. Not sure why I wrote it this way; maybe it used to return a promise at some point...

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -70,8 +70,6 @@ angular.module('emission.survey.enketo.add-note-button',
     if (momentStop) {
       prefills.End_date = momentStop.format('YYYY-MM-DD');
       prefills.End_time = momentStop.format('HH:mm:ss.SSSZ');
-    } else {
-      prefills.End_date = momentBegin.format('YYYY-MM-DD');
```
shankari, line 74: @JGreenlee I am not sure this is the fix. If we don't have a defined end date, which version of the survey will be displayed? Can you add a screenshot of what it would like like?

shankari (changes requested):
I've just blindly accepted all the css fixes, because they are not related to functionality. Could you add a few screenshots that show what is different, if it is indeed visible to the naked eye?

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: There is only one version of the survey, but redundant fields are hidden.
If the Start and End Date are both supplied and they are the same, then we they don't show up.

The problem was that I was filling an End Date, even when there was no End Time. So both Date fields would be hidden.

Here's how it now shows on the last place, where there is no exit_ts



Clicking the empty End Date field allows you to pick, initializing on 'today'

JGreenlee:
For styling, here are a couple before and afters (on my Pixel 5)

- unified the look of "To Label" / "Unlabeled" / "All Trips" to be consistent (or mostly consistent) between iOS/Android
- fixed collisions of diary cards on smaller screens
- fixed jumbled profile tab on smaller screens https://github.com/e-mission/e-mission-phone/pull/978

<table>
<tr>
<td>



</td>
<td>


</td>
</tr>
</table>


<table>
<tr>
<td>



</td>
<td>



</td>
</tr>
</table>

JGreenlee:
- https://github.com/e-mission/e-mission-phone/pull/976
never made its way back into this branch (`may_2023_point_release`)

Do you want me to pull it in here, or do you want to just merge upstream? Don't know if there will be conflicts.

shankari (approved):

PR #980, opened 2023-05-30
🧹 Migration Prep Follow-Up / Fixes

JGreenlee:
Addresses the review from https://github.com/e-mission/e-mission-phone/pull/974#pullrequestreview-1444813979.
Since #974 was so big, we are breaking the new changes into a separate follow-up PR for review to be more manageable in size.


JGreenlee:
I merged in `master_for_platform` here. There were a lot of merge conflicts and I wanted to sort through those myself.

JGreenlee:
This branch is not clean yet. I would call consider it 'experimental' or 'alpha'.
There are at least a couple things that are still broken / not working as expected.

But most critically: surveys. I reworked a lot of things just to get it to where it is now (it's using `enketo-core` v6), but it's still a little broken. Prefilling the time-use survey gives timezone issues. I need to re-implement something to the effect of this: https://github.com/e-mission/e-mission-phone/commit/0f1051f469cd1bc68e4133e459fb92ef04cdb8e3, but hopefully more elegantly - and probably on a fork of enketo-core. My goal would be to work with Enketo to get on enketo-core@latest (now v7), but there's a couple issues to work through. A forked v6 may have to do for now.

www/index.html
```diff
@@ -7,7 +7,7 @@
     <title></title>
 
     <!-- cordova script (this will be a 404 in the browser) -->
-    <script src="cordova.js"></script>
+    <script type="text/javascript" src="cordova.js"></script>
```
shankari, line 10: Yup. Originally added in 8893a329806c5d4a2113e280ce3f7a180e4011ac

shankari (commented):
This is a review upto the end of May.
June changes will be in the next review because I don't want to complicate this more than I need to

shankari:
> I merged in `master_for_platform` here. There were a lot of merge conflicts and I wanted to sort through those myself.

I tried to merge `master_for_platform` to the parent branch `framework_migration_prep` to reduce the number of changes due to previously merged code and you are right that there were a bunch of merge conflicts. Hopefully, my memory is good enough that I can go through them fairly quickly...

www/js/diary/LeafletView.jsx
```diff
@@ -3,21 +3,29 @@ import { angularize } from "../angular-react-helper";
 import { object, string } from "prop-types";
 import { View } from "react-native";
 
+const mapSet = new Set();
+export function invalidateMaps() {
+  mapSet.forEach(map => map.invalidateSize());
+}
+
```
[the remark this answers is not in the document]
JGreenlee: I referred back to the issue from before and looked at this reproduction case: https://github.com/e-mission/e-mission-docs/issues/111 - (instead of switching screens to "Diary", which no longer exists, I used "Dashboard")
Without this fix, the reproduction case was always the same as you saw. With the fix, the maps load fine every time. I haven't seen weird maps since then; so thus far, it seems like no hacks will be needed.

I basically just did what you wanted to do in the first place https://github.com/e-mission/e-mission-docs/issues/111#issuecomment-233184618, but because we were using `ui-leaflet` then, you didn't have access to the underlying maps. Now we do.

www/js/control/general-settings.js
```diff
@@ -103,6 +105,15 @@ angular.module('emission.main.control',['emission.services',
             $scope.qrp.show($event);
         } else {
             $ionicPopover.fromTemplateUrl("templates/control/qrc.html", {scope: $scope}).then((q) => {
+                // const qr = new QRCode({
+                //     content: $scope.tokenURL,
+                //     padding: 0
+                // });
+                // q.el.querySelector('#qr-opcode').innerHTML = qr.svg();
+                // QRCode.toCanvas(q.el.querySelector('#qr-opcode'), $scope.tokenURL, { width: 500 }, (e, canv) => {
+                //     e && console.error("Error while generating QR Code", e);
```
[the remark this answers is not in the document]
JGreenlee: Just removed it.
I left it there in case I changed my mind. That was before I realized <canvas> was a bad idea for the future.

www/templates/intro/saveTokenFile.html
```diff
@@ -1,28 +1,29 @@
 <ion-content>
 <div class="consent-title">
-<center><h2 ng-i18next>{{'login.make-sure-save-your-opcode'}}</h2></center>
+    <h2 ng-i18next>{{'login.make-sure-save-your-opcode'}}</h2>
```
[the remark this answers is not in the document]
JGreenlee: It looks exactly the same as before. I just made sure text was centered by methods other than the deprecated `<center>` tag.
It has technically never been supported in HTML5. I guess all the major browsers are scared to drop support for it (for legacy 2000s pages?) but they could at any point

JGreenlee:
All surveys should be working as expected now, using Enketo v6.
v7 is affected by this issue:
- https://github.com/enketo/enketo-core/issues/979

If Enketo merges my fix, we can upgrade and be on Enketo's latest

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -66,16 +65,14 @@ angular.module('emission.survey.enketo.add-note-button',
     // Enketo requires these specific date/time formats
     if (momentBegin) {
       prefills.Start_date = momentBegin.format('YYYY-MM-DD');
-      prefills.Start_time = momentBegin.format('HH:mm:ss.SSSZ');
+      prefills.Start_time = momentBegin.format('HH:mm:ss.SSS');
```
shankari: what is this change for? It was introduced in https://github.com/JGreenlee/e-mission-phone/commit/abdb5c640b71229026d4af32f77c95b65e5f7455 but doesn't seem to be related..

www/css/style.css
```diff
@@ -1,5 +1,29 @@
...
+.fill-container div[class*='css-view'] {
+  height: 100%;
+  width: 100%;
+  position: absolute;
+}
 
+.fill-container div[class*='css-view'] > div[class*='css-view'] {
+  position: static;
+}
+
```
shankari: For my own edification, these are adding fill-container to existing `div` styles.
But where are those div styles used? I have searched through the entire codebase and I don't see them. And I searched for css-view on the internet, but there don't seem to be any matches, so I am not sure it is standard terminology.

www/index.js
```diff
@@ -4,11 +4,7 @@ import './css/intro.css';
 import './css/appstatus.css';
 import './css/main.recent.css';
 import './css/main.diary.css';
-import './manual_lib/enketo/css/formhub.css';
-import './manual_lib/enketo/css/formhub-print.css';
 import './manual_lib/fontawesome/css/all.min.css';
-
-import 'leaflet/dist/leaflet-src.js';
```
shankari, line 11: I understand that we don't need this any more since the leaflet hack is no longer required. However, can't we also then remove the leaflet.css file (next line)?

www/js/diary/infinite_scroll_list.js
```diff
@@ -456,7 +456,7 @@ angular.module('emission.main.diary.infscrolllist',[
     if (!alreadyFiltered) {
         $scope.data.displayTrips = $scope.data.allTrips;
     };
-
+    $scope.data.displayTrips.count = $scope.data.displayTrips.length;
```
shankari, line 459: This was added in https://github.com/JGreenlee/e-mission-phone/commit/129004ee0f542974c0e977c99484e539d9973234 but I don't see the other, related changes - e.g. 

`$scope.data.displayTrips.forEach((cTrip) => {` -> `$scope.data.displayTrips.forEach((cTrip, i, displayTrips) => {`  on line 461.

Double check the merge conflict resolution.

www/templates/survey/enketo/add-note-button.html
```diff
@@ -2,5 +2,5 @@
   <!-- <button class="button ion-plus diary-btn" ng-click="openPopover($event, timelineEntry, 'NOTES')">
     {{displayLabel}}
   </button> -->
-  <diary-button text="displayLabel"></diary-button>
-</div> 
\ No newline at end of file
+  <diary-button text="displayLabel" ng-click="openPopover($event, timelineEntry, 'NOTES')"></diary-button>
+</div>
```
shankari, line 6: Note to self - the `ng-click` was removed in
https://github.com/JGreenlee/e-mission-phone/commit/f4346d0e5b154bae3b081258156a9971bff931de#diff-7699c6da2de00352d238d5a4eb12ff678300c7ab006a60dfe94e1068c1417da5

while switching to react.

Added back now that enketo is fixed

shankari (commented):
Second round of review: catching up with the changes after
https://github.com/e-mission/e-mission-phone/pull/980/commits/2666db453132c9c40c88dd303349373a1321f6d3

www/js/diary/infinite_scroll_list.js
JGreenlee: Ah, it's because this branch displays the totals differently. It doesn't have the "Displaying X/T trips" text - instead, it puts this up in the filter `<select>` --- e.g. `To Label (9/20)`. These counts reflect the number of the list entries shown out of total total, regardless of whether they are trips, places, untracked, etc.
So we don't actually need to have a `displayTrips.count` on this branch.

www/css/style.css
JGreenlee: It comes from React Native Web. The way it converts React Native markup into HTML is kind of messy. It adds a bunch of wrappers with randomized class names in order to apply the RN stylesheets as CSS.
Read here- https://giuseppegurgone.com/twitter-html
If you go to Twitter web and inspect, I think you'll see what I mean. The resulting HTML is not pretty, but it works.

In our case, React Native Web isn't really meant to be injected into an existing DOM structure. We need to control the behavior of the wrappers so they play nice with their AngularJS parents. To do this, I am selecting anything with a class that starts with `css-view`.

Once everything is rewritten we won't have to do this.

webpack.config.js
```diff
@@ -74,6 +74,7 @@ module.exports = {
     // to load jQuery and moment globally
     new webpack.ProvidePlugin({
       $: 'jquery',
+      jQuery: 'jquery',
```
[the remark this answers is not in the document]
JGreenlee: For Enketo, I don't know. But with jQuery libraries and anything that uses jQuery, there seems to be little consistency for which is used.
Exposing both globally seems to be the recommendation for Webpack. https://webpack.js.org/plugins/provide-plugin/#usage-jquery

Maybe someday not so far away, we won't have to do this. AngularJS will be gone, and I saw that the Enketo team is gradually removing its reliance on jQuery.

www/index.js
JGreenlee: No, we still have to get the Leaflet styles from somewhere.
It's not coming from `manual_lib`; this is the official Leaflet package from npm (latest version). We don't have anything for Leaflet in `manual_lib` anymore.

But one good thing to do sometime is refactor all uses of the `L` global into modular imports. Leaflet 2.0, when it is released, is not going to provide the `L` global.

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: Ah, that change should have been in ab6f7adee12ee929ef380da554df5b1d523f3daa.
Regardless, I changed it again last night to actually fix the prefilled timezones.

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -60,40 +60,27 @@ angular.module('emission.survey.enketo.add-note-button',
...
+    // the current, local time offset (e.g. -07:00)
+    const currOffset = moment().toISOString(true).slice(-6);
+
     const prefills = {}
     // Fill in only the fields that are present
     // Enketo requires these specific date/time formats
     if (momentBegin) {
       prefills.Start_date = momentBegin.format('YYYY-MM-DD');
-      prefills.Start_time = momentBegin.format('HH:mm:ss.SSS');
+      prefills.Start_time = momentBegin.format('HH:mm:ss.SSS') + currOffset;
```
shankari: I am not 100% sure that this is correct. Shouldn't we add the offset before formatting as both date and time. What happens if the offset causes the date to change? How have you tested this?

Or if you want the original date and time, but just pretend to be in the current timezone, do you even need the currOffset? What happens if you don't set it?

shankari (commented):
I think that this is the last of my reviews

www/index.js
shankari: So webpack automatically includes js through tree shaking but not css?

www/js/diary/infinite_scroll_list.js
shankari: So you should remove this line?

shankari:
I am going to merge this as well, so that all the final cleanup fixes come in a separate PR

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: Date and time are stored in separate fields, so the time "overflowing" to another day isn't a worry here.
Note that I'm not adding time to the Moment object. I'm just concatenating a fake offset to the string that's going to Enketo.

> Or if you want the original date and time, but just pretend to be in the current timezone, do you even need the currOffset?

Yes; we have to pretend, for the purpose of the survey, that the trip times are in the current local time. Otherwise, Enketo will forcibly convert them to the local timezone.
If I don't put any offset (how it was before this change), the prefilled times display like we want them to, but it fails validation and the survey can't be submitted.

I tried everything I could think of and this is the only workaround, short of modifying Enketo again.

www/index.js
JGreenlee: Here's my best understanding of it. Webpack, when it analyzes our project, is only able to build its dependency graph based on ES `import` statements or CommonJS `require()` statements (in our project I've been sticking to the former).

In the case of Leaflet, we don't use either of these; we rely on a global variable `L`. I have to specifically provide this `L` in `webpack.config.js` https://github.com/e-mission/e-mission-phone/pull/980/files#diff-1fb26bc12ac780c7ad7325730ed09fc4c2c3d757c276c3dacc44bfe20faf166f

Then the CSS has to be separately imported. I could do it here or I could do it somewhere else, like `styles.css`. But it doesn't happen automatically.

www/js/diary/infinite_scroll_list.js
shankari: This is another TODO to be handled?

JGreenlee: Irrelevant in the new branch since the filters were organized into their own component

PR #982, opened 2023-06-01
add short demo survey locally

JGreenlee:
Here I am locally checking in the short form of the demographics survey.

It was also checked in here: https://github.com/e-mission/nrel-openpath-deploy-configs/pull/22

> This shortened demographics survey has only 9 questions and will be used for the Denver CASR study.
> 
> The questions are:
> 
> > In which year were you born? (must be 1900-2005 inclusive)
> > What is your gender?
> > What is your race/ethnicity?
> > Do you have a valid drivers’ license?
> > Highest level of education completed?
> > What is your home type?
> > Income
> > How many people live in the home
> > Number of motor vehicles


shankari:
@JGreenlee I am merging this now, but for cleanup/clarity, please:
- check the json and xml in to the `survey_resources` directory
- download the xls file as well and add to the `survey_resources` directory

My understanding is that to edit a survey in kobotoolbox, we need the xls file. So we should have it checked in as well, to support editing it further in the future.

JGreenlee:
I meant I checked it into `survey-resources` on https://github.com/e-mission/nrel-openpath-deploy-configs/pull/22.
The Excel file is there too.

JGreenlee:
In the morning, I'll update the responses for the gender question and the driver's license question.

shankari:
ok, so I will plan on pushing out the release tomorrow morning as well.

PR #981, opened 2023-05-31
:children_crossing: Decouple app required/not required from study/pro…

shankari:
…gram

Denver CASR is a program (with `mode_of_interest` = `e-bike`) which has participants opt in for data collection. So let's decouple the app required/not required and make it explicit instead of implicit with study/program.

We also decouple the flag for whether the token is auto-generated or not from the study or program flag.

The `program_or_study` option is currently only used:
- in `trip_confirm_services` to see whether to look at the `mode_of_interest` or not
- to fill in "program" or "study" in the consent text

We might be able to simplify the use in `trip-confirm-services`, rename all the text to `study` and then remove the option completely later.

```
$ grep -r program_or_study www/
www//js/intro.js:        // old-style "program_or_study"
www//js/intro.js:        $scope.ui_config.app_required = $scope.ui_config?.intro.app_required || $scope.ui_config?.intro.program_or_study == 'program';
www//js/intro.js:        $scope.ui_config.opcode.autogen = $scope.ui_config?.opcode.autogen || $scope.ui_config?.intro.program_or_study == 'study';
www//js/survey/multilabel/trip-confirm-services.js:        Logger.log("About to start initializing the confirm helper for " + ui_config.intro.program_or_study);
www//js/survey/multilabel/trip-confirm-services.js:        if (ui_config.intro.program_or_study == 'program') {
www//templates/intro/consent-text.html:{{ui_config.intro.program_or_study}} and for research to inform public policy.
www//templates/intro/consent-text.html:    the {{ui_config.intro.program_or_study}}, send personalized surveys or provide
www//templates/intro/consent-text.html:Participation in the {{ui_config.intro.program_or_study}} is completely
www//templates/intro/consent-text.html:participate in the {{ui_config.intro.program_or_study}}.
```

PR #984, opened 2023-06-01
:bug: Customize the privacy policy when displayed from the profile

shankari:
Per Google requirements, we display the privacy policy from the control screen as well, and pass in the `$scope` variable so that it can be configured correctly. However, most of the customizable text in the policy is from the `$scope.translated_text` variable, which needs to be filled in with the correct locale.

And then we added several backward compatibility hacks in https://github.com/e-mission/e-mission-phone/pull/981

This commit copies over the `translated_text` selection and the backward compatibility hacks so that the in-app privacy policy is also generated properly

### DIFF
--- a/www/js/control/general-settings.js
+++ b/www/js/control/general-settings.js
@@ -174,6 +174,24 @@ angular.module('emission.main.control',['emission.services',
     $ionicPlatform.ready().then(function() {
         DynamicConfig.configReady().then(function(newConfig) {
             $scope.ui_config = newConfig;
+            // backwards compat hack to fill in the raw_data_use for programs that don't have it
+            const default_raw_data_use = {
+                "en": `to monitor the ${newConfig.intro.program_or_study}, send personalized surveys or provide recommendations to participants`,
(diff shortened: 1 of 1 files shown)


shankari:
Testing done:

| After | Before |
|----------|----------|
| simulator_screenshot_4A37F8A5-00E5-4C2D-B3FF-9C463F881FDA | simulator_screenshot_CB36E3E8-A482-4CD3-AD4F-549FE793362C |
| simulator_screenshot_794E1DEC-7631-4EC1-9851-F8D7AC96B2D6 | simulator_screenshot_D36982D1-A862-4E2F-B85C-C61065F805E6

shankari:
Testing done:
- Loaded a non Denver CASR program

|  app required works  |  for smart commute |
|--------|---------|
| simulator_screenshot_D497E58D-BFA4-4DE7-8E0B-C369AA08C085 | simulator_screenshot_4AC5720F-A384-4753-B633-DCCA58C5CED1 |

PR #990, opened 2023-06-12
Finish Migration Prep

JGreenlee:
This should be the last round of fixes before we begin the bulk of the migration and actually rewrite the UI into React components.


www/js/angular-react-helper.jsx
```diff
@@ -38,18 +38,18 @@ export function angularize(component, modulePath) {
...
-      if (!root) {
-        root = createRoot($element[0]);
-        roots.set($element[0], root);
-      }
+    controller: function($element) {
+      /* TODO: once the inf scroll list is converted to React and no longer uses
+        collection-repeat, we can just set the root here one time
+        and will not have to reassign it in $onChanges. */
+      /* Until then, React will complain everytime we reassign an element's root */
+      let root;
```
shankari, line 50: Just to clarify, this reverts https://github.com/JGreenlee/e-mission-phone/commit/39ddf90840833397dbb8fc7c289895c8709cad2d and we are just going to live with the errors 😄

shankari (approved):
LGTM. Is this now ready to merge and push out, at least for testing?

www/js/angular-react-helper.jsx
JGreenlee: Correct. I don't see a way around the errors for now.

Hopefully this will only be for a few weeks, until the list itself is rewritten in React.

JGreenlee:
> LGTM. @JGreenlee Is this now ready to merge to `master_for_platform` and push out, at least for testing?

@shankari Nothing seems majorly broken anymore, so I think yes; let's do it!

I'd like to get this on staging and get some real-world testing out of it so that we can better track issues as they appear.

PR #992, opened 2023-06-13
settings tables migration to React - Abby's first migration task

Abby-Wheelis:
In line with what Jack suggested, I have created a DataTable component that is used now for both the collection and sync settings details. The image shows what the sections look like after the switch.
Screenshot 2023-06-13 at 2 20 28 PM

Jack also suggested we could then use this component to show scheduled notifications so I'll start looking at that next.


www/js/control/ControlDataTable.jsx
```diff
@@ -0,0 +1,28 @@
...
+    </DataTable>
+    );
+  };
+ControlDataTable.propTypes = {
+    controlData: array
+  }
+ 
+// need call to angularize to let the React and Angular co-mingle
+  //second argument is "module path" - unsure of exact usage, could require update later?
+angularize(ControlDataTable, 'emission.main.control.dataTable');
```
JGreenlee, line 27: In AngularJS, everything is registered under a module.

For a typical AngularJS directive/component, you register it under some module path, and then in the places where you need to use it, you list dependencies by their module path.

For React components that are angularized, I made this a little easier - the module path is added to the component itself and can be accessed via `ExampleComponent.module` after `ExampleComponent` is imported. This way, you only have to write the modulepath in one place, and if you make a typo or something it doesn't really matter.

www/js/control/general-settings.js
```diff
@@ -19,7 +20,8 @@ angular.module('emission.main.control',['emission.services',
                                         'emission.survey.enketo.demographics',
                                         'emission.plugin.logger',
                                         'emission.config.dynamic',
-                                        QrCode.module])
+                                        QrCode.module,
+                                        ControlDataTable.module])
```
JGreenlee, line 24: This is where the modulepath is being used.
`ControlDataTable.module` is `'emission.main.control.dataTable'`

www/js/control/ControlDataTable.jsx
Abby-Wheelis: Ok, that makes sense, thank you for explaining! I was just emulating what I'd seen other places in the code, now I understand why it works.

www/js/control/general-settings.js
```diff
@@ -19,7 +21,9 @@ angular.module('emission.main.control',['emission.services',
                                         'emission.survey.enketo.demographics',
                                         'emission.plugin.logger',
                                         'emission.config.dynamic',
-                                        QrCode.module])
+                                        QrCode.module,
+                                        ControlDataTable.module,
+                                        ControlListItem.module])
```
shankari, line 26: For my own edification: do we need to import `ControlListItem` here?
I don't see a `ControlListItem.jsx`; is it missing?

www/js/splash/notifScheduler.js
```diff
@@ -65,6 +65,14 @@ angular.module('emission.splash.notifscheduler',
             if (!notifs?.length)
                 return Logger.log(`${prefix}, there are no scheduled notifications`);
             const time = moment(notifs?.[0].trigger.at).format('HH:mm');
+            cordova.plugins.notification.local.scheduledNotifs = notifs.map((n) => {
```
shankari, line 68: I don't think we should put this into the plugin - the plugin is typically read-only in javascript. This should go into the rootScope or a service

shankari, line 68: Can we also print these logs - change the `Logger` message below?

www/templates/control/main-control.html
```diff
@@ -5,7 +5,13 @@
...
     <enketo-demographics-button></enketo-demographics-button>
-    <div class="control-list-item">
+    <!-- each list item is very similar:
+      contains a text and a button icon or switch (for the most part) 
+      moving to react -- a text component and an icon button component that live in a list item comoponent? -->
+    <!-- <div class="control-list-item">
+      <control-list-item text="View Privacy" icon-name = "eye"></control-list-item>
+    </div> -->
+      <div class="control-list-item">
       <div class="control-list-text" ng-i18next>{{'control.view-privacy'}}</div>
```
shankari, line 15: This change is inadvertent, right?

www/templates/control/main-control.html
```diff
@@ -120,6 +126,13 @@
         <div class="control-list-text" ng-i18next>Dummy Notification in 5 Seconds</div>
         <div ng-click="dummyNotification()" class="control-icon-button"><i class="ion-ios-bell"></i></div>
       </div>
+      <!-- maybe notif schedule preview could go here? -->
+      <div class = "control-list-item">
+        <div class="control-list-text" ng-i18next>{Upcoming Notifications}</div>
```
shankari, line 131: we were avoiding spaces and full text strings for the i18n keys before this. Let's discuss if we should change that.

By default, we shouldn't change it

www/templates/control/main-control.html
```diff
@@ -140,23 +153,31 @@
...
+      <!-- replacing collection and sync with React DataTable-->
+      <!-- collection li and data table -->
       <div class="control-list-item">
         <div class="control-list-text" ng-i18next>{{'control.collection'}}</div>
         <div ng-click="editCollectionConfig($event)" class="gray-icon control-icon-button"><i class="ion-edit"></i></div>
       </div>
-      <ion-list>
-
+      
+      <!-- not sure what if any class(es) are appropriate here, chose these based off existing code, no particular change in appearance -->
```
shankari, line 163: nit: haven't heard any requests for change in appearance either :)

www/templates/control/main-control.html
```diff
@@ -140,23 +153,31 @@
...
+      
+      <!-- sync li and data table -->
       <div class="control-list-item">
         <div class="control-list-text" ng-i18next>{{'control.sync'}}</div>
         <div ng-click="editSyncConfig($event)" class="gray-icon control-icon-button"><i class="ion-edit"></i></div>
       </div>
-      <ion-list>
+      <control-data-table control-data="settings.sync.show_config"></control-data-table>
+      
+      <!-- <ion-list>
```
shankari, line 180: nit: we should remove commented out code before the final commit 🔥, to avoid confusion down the road

shankari (changes requested):
Great first change! Just a few small comments....

www/js/splash/notifScheduler.js
JGreenlee: Ah, this was my fault, not @Abby-Wheelis

Yes, `$rootScope`, or a variable at the root level of `scheduler`, would be a better choice

www/js/control/general-settings.js
Abby-Wheelis: You're right, that's unnecessary. It was a part of something else, I had started to tinker with changing a different piece of that page

www/templates/control/main-control.html
Abby-Wheelis: Yes, part of the same thing you noted in the `general-settings.js`

www/js/control/ControlDataTable.jsx
```diff
@@ -6,11 +6,12 @@ import { array } from "prop-types";
...
 // val with explicit call toString() to resolve bool values not showing
 const ControlDataTable = ({ controlData }) => {
-  console.log("Printing data trying to tabulate", controlData);
+  // console.log("Printing data trying to tabulate", controlData);
   return (
+    //rows require unique keys!
     <DataTable style={styles.table}>
       {controlData.map((e) =>
-        <DataTable.Row>
+        <DataTable.Row key={e.key}>
```
shankari, line 14: At this point, we expect at most one notification per day so this should be fine. Feel free to resolve once you have seen this.

www/templates/control/main-control.html
```diff
@@ -123,13 +120,12 @@
         <div ng-click="checkConsent()" class="control-icon-button"><i class="ion-checkmark-circled"></i></div>
       </div>
       <div class="control-list-item">
-        <div class="control-list-text" ng-i18next>Dummy Notification in 5 Seconds</div>
+        <div class="control-list-text" ng-i18next>{{'control.dummy-notification'}}</div>
         <div ng-click="dummyNotification()" class="control-icon-button"><i class="ion-ios-bell"></i></div>
       </div>
-      <!-- maybe notif schedule preview could go here? -->
+      <!-- notif scheudle preview (react component) -->
```
shankari, line 126: nit: scheudle 😄  I am fine with merging as is and having you fix later...

shankari (changes requested):
Looking much better.
Just need to make the changes to `notifSched` now...

www/js/control/ControlDataTable.jsx
```diff
@@ -6,11 +6,12 @@ import { array } from "prop-types";
 // Note the camelCase to dash-case conventions when translating to .html files!
 // val with explicit call toString() to resolve bool values not showing
 const ControlDataTable = ({ controlData }) => {
-  console.log("Printing data trying to tabulate", controlData);
+  // console.log("Printing data trying to tabulate", controlData);
   return (
+    //rows require unique keys!
```
JGreenlee, line 11: This is something I encountered as well (in https://github.com/e-mission/e-mission-phone/pull/993)
I see you've already figured it out; I think this will be the case anywhere we use `.map()` or otherwise have repeated instances of a component one after another

www/templates/control/main-control.html
```diff
@@ -8,6 +8,10 @@
     <!-- each list item is very similar:
       contains a text and a button icon or switch (for the most part) 
       moving to react -- a text component and an icon button component that live in a list item comoponent? -->
+
+    <!-- here is the html related to my (failing) attempts to convert more of the profile screen -->
+    <setting-row text-key = "control.view-privacy" icon-name = "eye"></setting-row>
```
shankari, line 13: At a high level, when you get to refactoring the profile screen, I think it might be worthwhile creating a parent React Component for the profile screen as well. If they all follow a similar pattern, we could even define the i18n label and the function to call in a map and then create the elements by iterating over the map!

www/js/control/ControlDataTable.jsx
shankari: wait, what is related to `map()`? I thought this would be because the values don't "stick" until the `$apply()` is called

shankari (changes requested):
so if I understand correctly, this PR is still marked as draft because 

> I started to convert the data storage, data is fine in general-settings.js but is failing to make it through main-control.html to ControlDataTable.jsx

www/js/control/ControlDataTable.jsx
shankari: Ah I see that this comment was about the requiring unique keys

www/js/control/general-settings.js
```diff
@@ -131,6 +133,7 @@ angular.module('emission.main.control',['emission.services',
         $scope.settings.notification.prefReminderTime = m.format('LT'); // display in user's locale
         if (storeNewVal)
             NotificationScheduler.setReminderPrefs({ reminder_time_of_day: m.format('HH:mm') }); // store in HH:mm
+        $scope.settings.notification.scheduledNotifs = cordova.plugins.notification.local.scheduledNotifs;
```
JGreenlee, line 136:
```suggestion
        $scope.settings.notification.scheduledNotifs = NotificationScheduler.scheduledNotifs;
```

@Abby-Wheelis Here's your fix! It needed to be changed here as well.

shankari (approved):
LGTM! @JGreenlee if you approve it too, I will merge

JGreenlee:
@shankari LGTM too

Abby-Wheelis:
I accidentally brought over some changes from MoreProfileMigrations, but confirmed that I reverted those successfully, this PR represents and conversation of the data tables only, MoreProfileMigrations is on its way to being a whole-page conversion

shankari:
@Abby-Wheelis I am going to squash merge this to avoid all the churn. Make sure you deal with it properly while pulling from `framework_migration_prep` next time.

www/templates/control/main-control.html
```diff
@@ -118,9 +118,15 @@
         <div ng-click="checkConsent()" class="control-icon-button"><i class="ion-checkmark-circled"></i></div>
       </div>
       <div class="control-list-item">
-        <div class="control-list-text" ng-i18next>Dummy Notification in 5 Seconds</div>
+        <div class="control-list-text" ng-i18next>{{'control.dummy-notification'}}</div>
```
shankari, line 121: Ah I see that this has been added in
https://github.com/e-mission/e-mission-translate/pull/30/files

shankari (approved):

PR #993, opened 2023-06-14
🏷️ Migration of Label Screen Components

JGreenlee:
This PR begins the work of rewriting the label screen components as React components.

Specifically, `TimestampBadge.jsx` and `MultilabelButtonGroup.jsx` are new. (The popup menus that `MultilabelButtonGroup` creates need to be reworked.)

(once `framework_migration_prep` is merged into `master_for_platform`, I will change the target of this PR to `master_for_platform`)


www/js/survey/multilabel/ModeSelectMenu.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+import React, { } from "react";
+import { angularize } from "../../angular-react-helper";
+import { func } from "prop-types";
+import { ScrollView } from "react-native";
+import { Menu } from "react-native-paper";
+import { useTranslation } from "react-i18next";
+
+const ModeSelectMenu = ({ anchor, visible, inputParams, chooseFn }) => {
+  const { t } = useTranslation();
```
JGreenlee, line 11: @Abby-Wheelis here is an example of how internationalization will work in the new React components we're writing.

`t` is the translation function and it's received using `useTranslation()`.

www/js/survey/multilabel/ModeSelectMenu.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <Menu visible={visible} anchor={anchor}>
+      <ScrollView>
+        {inputParams?.MODE?.options.map((mode) => {
+          return (
+            <Menu.Item
+              key={mode}
+              onPress={() => chooseFn('MODE', mode)}
+              title={t(mode.text)}
```
JGreenlee, line 21: @Abby-Wheelis Then, wherever a translation needs to happen, we'll just pass the translation ID to `t()`

Another example: `t('general-settings.are-you-sure')` will return `"Are you sure?"` if the locale is English, or `"¿Estás seguro?"` if the locale is Spanish.

www/js/app.js
```diff
@@ -43,7 +44,8 @@ locales.forEach(locale => {
   }
 });
 
-i18next.init({
+i18next.use(initReactI18next)
```
JGreenlee, line 47: @Abby-Wheelis
Note that our use of i18next in React is dependent on the changes here in `app.js` - I also included this in https://github.com/e-mission/e-mission-phone/pull/990 and I believe @shankari plans to merge that soon

shankari:
I have now merged https://github.com/e-mission/e-mission-phone/pull/990
I am not going to merge anything to master_for_platform until we have tested on staging and are ready to push to production since other teams are pulling from it for their own custom instances of the platform.

JGreenlee:
The multilabel buttons, and the logic necessary for their behavior, has been consolidated into `MultilabelButtonGroup.jsx`.
I also made the decision to include the checkmark icon into that component, rather than rendering it separately.

I did not create a separate component for the checkmark, since it's just a one-liner `IconButton`.



As you can see, I opted to make the checkmark button show with an outline (this way it actually looks like something you can press); and I also shifted the color palette with the goal of having a better contrast ratio for readability

JGreenlee:
Much of the hardest work is out of the way. `MultilabelButtonGroup` is most likely the most complex component we will have inside the label screen, other than perhaps the list component itself.

I think that creating components for the diary cards will not be difficult, as the only non-React elements left are just text. Similarly with the details page.

Migrating the label screen may not take as long as I had anticipated. But first, I want to stop and write unit tests for what's here.

www/js/app.js
shankari: This has now been merged

JGreenlee:
I have not been able to set up component testing.

I burned the better part of yesterday trying to find the right configuration. I exhausted all the help I can find on StackOverflow + GitHub, and I'm not sure that me continuing to wrangle with it is going to be beneficial anymore.

It seems darn near impossible to get Jest to work in our project at this time - there are too many complications: TypeScript, Ionic/Angular legacy packages, the fact that we are using React Native Web in a nonstandard way..

It's not going to be easy to implement quality testing until we are free of Cordova and are working in an Expo ecosystem.

That said, I think we can still test simple, individual functions easily; i.e we can make sure functions give the correct output when given a fake input. But I'm not sure how valuable that kind of testing is going to be in this codebase, because at that point we have to mock almost everything.
I was hoping to test components as a whole, using tools that virtually render a component in the React DOM - but I have not been able to get that part working.

I know that testing is really important because of the large-scale nature of the project, but unless someone else can get component testing working, we may have to settle for the more primitive kind of unit tests.

shankari:
> That said, I think we can still test simple, individual functions easily; i.e we can make sure functions give the correct output when given a fake input. But I'm not sure how valuable that kind of testing is going to be in this codebase, because at that point we have to mock almost everything.

I think that this is still valuable and we should implement it. I have two main reasons:
1. Theoretically, unit testing includes what you have described above (https://en.wikipedia.org/wiki/Unit_testing)

> In procedural programming, a unit could be an entire module, but it is more commonly an individual function or procedure. By writing tests first for the smallest testable units, then the compound behaviors between those, one can build up comprehensive tests for complex applications.

2. I want to get into the habit of writing unit tests for UI code. Having a test framework and examples makes that more lilkely

3. We can implement CI/CD and code coverage on the phone as well

Finally, wrt:

> It's not going to be easy to implement quality testing until we are free of Cordova and are working in an Expo ecosystem.

Jest is supposed to work for ReactJS as well, so it seems like we should be able to use the more comprehensive tests that you wanted ("snapshot tests?") once we are fully in ReactJS but have not migrated the plugins from Cordova to ReactNative yet. Let's re-explore once the javascript migration is done.

shankari:
@JGreenlee is this ready to merge? Still marked as draft.
No rush, I can just deploy the previous version until you finish this.

JGreenlee:
> I think that this is still valuable and we should implement it

Ok, I will try to at least add Jest for individual functions, and not worry about configuring it for component testing.

> it seems like we should be able to use the more comprehensive tests that you wanted ("snapshot tests?") once we are fully in ReactJS but have not migrated the plugins from Cordova to ReactNative yet. Let's re-explore once the javascript migration is done.

Ah, true. There should be a point where we are no longer using AngularJS but are still using Cordova, and I'll try again then.

JGreenlee:
> @JGreenlee is this ready to merge? Still marked as draft. No rush, I can just deploy the previous version until you finish this.

Ready to be pushed for staging. I expect we will encounter some minor problems but I think the best way to find them and track them is to go forward and push it out.

JGreenlee:
Here's a status report of where this branch leaves off:

With these changes, all the buttons on the diary card, relevant to any of the app configurations, have been rewritten in React.

So there's `MultilabelButtonGroup`, `UserInputButton`, and `AddNoteButton`.
To display additions in the case of `AddNoteButton`, there is also `AddedNotesList`.

The `LeafletView` was already done beforehand.

All that leaves on the diary cards that isn't in React is text (date, place names, and/or distance traveled).

I began the initial work of rewriting the diary cards themselves, but it's not ready. I'll do it next week.

JGreenlee:
I am curious what we will find in our testing in the way of performance.

Each diary card can currently have as many as 5 React components embedded in it, independently from each other, so basically React DOM has multiple 'roots' or 'entry points' to keep track of. I think this could cause some slowness.

Once the diary cards themselves are converted to React, this is unified to one 'root' because it's just a matter of embedding React components inside React components.

So if the app is slow right now, I would expect it to get better soon.

JGreenlee:
@shankari Actually I don't think you will be able to easily deploy from this branch.
Since moving all dependencies to npm and adding Webpack, there are a couple other steps that have to be added to the build process before we can deploy.

Since this hasn't been merged, I have been working ahead on a new branch, where I restructured the Webpack config to have separate `prod` and `dev` configurations. The 'prod' bundle will be smaller and more optimized.
The implementation of `angularize()` also needed to be tweaked because bundling with minification tends [to break AngularJS apps, apparently, unless you use explicit annotations (which we don't)] - https://github.com/JGreenlee/e-mission-phone/commit/e96cdcd9acd93338521c013f8ac71f86e266740e https://docs.angularjs.org/guide/di#implicit-annotation.
Our other AngularJS code is fed through a babel plugin that automatically adds the annotations for us. https://github.com/JGreenlee/e-mission-phone/commit/11fc98c17e73224d7e259a0b1f0d44359c649530#diff-d3465f3a71d5cc473d32d3919bb34796e47f195e78b453d88c727c3f5ad80e6e

I was going to let you figure out how to get a deployment out, but it required a lot more reworking than I thought. This way, I hopefully saved you a few hours.

JGreenlee:
So I can either cherry-pick the changes necessary for deployment and bring them into this branch so that we can test this branch on staging, or we can just wait until the new branch is ready (probably end of this week) and test on it.

The new branch rewrites almost all of the label screen.

shankari:
I was hoping to at least move up to the unified diary and label (https://github.com/e-mission/e-mission-phone/pull/974), but I think that already has all the webpack changes, so it probably won't work 😦 

In that case, I think we should wait until the end of the week. My husband also found a bug between diary and label screens that I will file an issue for. I don't want to remove the diary screen until we know that is fixed, since we don't know whether the label screen is actually working properly yet 😦

JGreenlee:
> In that case, I think we should wait until the end of the week.

Just merged everything in. Here's where we're at:
 - Every* UI element accessible from the Label tab has been converted to React
 -- *The only exceptions are popups, which are still using `$ionicPopup` / `$ionicPopover`

- `LabelTab` contains all the logic that does the heavy lifting of loading, populating, and filtering trips. This used to be in `infinite_scroll_list.js`, a file that was very big and messy. I think I have made progress simplifying it / extracting logic out to other places, but there's a lot more that should be done to clean that file

- `/diary/services.js` is still in Angular. There's really no rush to convert it because we can easily use Angular services alongside React components

- The buttons in the header, for filtering and for choosing dates, still use HTML. They are written in React, but they aren't fully native and we will want to revisit them at some point

- The details screen is missing the sections list, as well as the speed graph

www/js/diary/cards/TripCard.tsx
```diff
@@ -0,0 +1,173 @@
...
+      console.log("DiaryCard: timelineEntry changed, force update");
+      setRerender(!rerender);
+    }
+  }, []);
+
+  useEffect(() => {
+    const surveyOptKey = appConfig?.survey_info?.['trip-labels'];
+    setSurveyOpt(SurveyOptions[surveyOptKey]);
+  }, [appConfig, loading]);
+
```
shankari, line 51: will we still need `useEffect` when the migration is complete? I assume that the answer is "yes", since we are syncing the config from outside the react ecosystem (over http). Just wanted to check to deepen my understanding.

www/js/diary/list/TimelineScrollList.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+
+const separator = () => <View style={{ height: 8 }} />
+const bigSpinner = <ActivityIndicator size="large" style={{margin: 15}} />
+const smallSpinner = <ActivityIndicator size="small" style={{margin: 5}} />
+
+const TimelineScrollList = ({ listEntries, loadedRange, pipelineRange, loadMoreFn, isLoading }) => {
+
+  // The way that FlashList inverts the scroll view means we have to reverse the order of items too
+  const reversedListEntries = listEntries ? [...listEntries].reverse() : [];
```
shankari, line 28: would it be more performant to maintain the list in reversed order in the `LabelTab`?

www/js/diary/list/TimelineScrollList.tsx
```diff
@@ -0,0 +1,72 @@
...
+                      { reachedPipelineEnd ? "No more travel" : "Show More Travel"}
+                  </LoadMoreButton>;
+
+  if (isLoading=='replace') {
+    return bigSpinner;
+  } else if (listEntries?.length == 0) {
+    return "No travel to show";
+  } else {
+    return (
+      <FlashList inverted
```
shankari, line 48: I looked around for `inverted` and there is not much documentation for it 
"Reverses the direction of scroll. Uses scale transforms of -1."
Have to try it out to see if it is intuitive.

www/js/diary/list/TimelineScrollList.tsx
```diff
@@ -0,0 +1,72 @@
...
+        data={reversedListEntries}
+        renderItem={renderCard}
+        estimatedItemSize={240}
+        keyExtractor={(item) => item._id.$oid}
+        /* TODO: We can capture onScroll events like this, so we should be able to automatically
+            load more trips when the user is approaching the bottom or top of the list.
+            This might be a nicer experience than the current header and footer buttons. */
+        // onScroll={e => console.debug(e.nativeEvent.contentOffset.y)}
```
shankari, line 56: I had originally thought that as well but the ion-list didn't support refreshing from both sides. We used to have "pull to refresh" on the diary screen, but it used to be a bit confusing to people because they didn't know that they needed to pull to get more information, so I took it out and added the refresh button instead. 

It can also lead to scrolling that people don't really want.

The buttons may be clunky but they are also clear.

Let's merge this for now and then experiment with that on staging before deciding whether to keep it or not.

www/js/diary/services.js
```diff
@@ -1,22 +1,23 @@
...
-  WALKING: {name: "WALKING", icon: "ion-android-walk", color: "brown"},
-  RUNNING: {name: "RUNNING", icon: "ion-android-walk", color: "brown"}, // TODO: do we need this or is it always converted to WALKING?
-  CAR: {name: "CAR", icon: "ion-android-car", color: "red"},
-  AIR_OR_HSR: {name: "AIR_OR_HSR", icon: "ion-plane", color: "red"},
+  IN_VEHICLE: {name: "IN_VEHICLE", icon: "speedometer", color: "purple"},
+  ON_FOOT: {name: "ON_FOOT", icon: "walk", color: "brown"},
+  BICYCLING: {name: "BICYCLING", icon: "bike", color: "green"},
+  UNKNOWN: {name: "UNKNOWN", icon: "help", color: "orange"},
+  WALKING: {name: "WALKING", icon: "walk", color: "brown"},
+  RUNNING: {name: "RUNNING", icon: "walk", color: "brown"}, // TODO: do we need this or is it always converted to WALKING?
```
shankari, line 9: I think we convert it to WALKING

shankari: I think that we also need an `ON_FOOT` just in case we end up getting cleaned trips instead of inferred trips

www/js/survey/enketo/enketo-add-note-button.js
shankari, line 1: This is only still around for the service, right?

www/js/survey/multilabel/trip-confirm-services.js
```diff
@@ -147,26 +147,6 @@ angular.module('emission.survey.multilabel.services', ['ionic', 'emission.i18n.u
         }
     }
 
-    ch.checkOtherOption = function(inputType, onTapFn, $scope) {
```
shankari, line 150: why remove the "other" option?

www/js/useAppConfig.ts
```diff
@@ -0,0 +1,24 @@
...
+  const [loading, setLoading] = useState<boolean>(true);
+
+  const DynamicConfig = getAngularService('DynamicConfig');
+  const $ionicPlatform = getAngularService('$ionicPlatform');
+
+  useEffect(() => {
+    $ionicPlatform.ready().then(() => {
+      DynamicConfig.configReady().then((config) => {
+        setAppConfig(config);
+        setLoading(false);
```
shankari, line 16: are we sure we are done with all loading here? Or is this flag only used to track when the app config has loaded? Where do we actually check for this?

www/templates/diary/infinite_scroll_detail.html
shankari, line 1: why do we still need this html file? We have removed `infinite_scroll_details.js`...

www/js/survey/multilabel/trip-confirm-services.js
shankari: Ah I see that it has been moved to `MultiLabelButtonGroup.tsx`

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -0,0 +1,132 @@
...
+    });
+  }
+
+  const inputKeys = Object.keys(trip.inputDetails);
+  return (
+    <View style={{flexDirection: 'row', alignItems: 'center'}}>
+      <View style={{width: 'calc(100% - 20px)'}}>
+        {inputKeys.map((key, i) => {
+          const input = trip.inputDetails[key];
+          const inputIsConfirmed = trip.userInput[input.name];
```
shankari, line 99: do we still have the distinction between `trip.userInput` and `trip.user_input`? I have frequently felt that we should rename one or the other just to avoid confusion among future maintainers.

shankari (changes requested):
This includes everything EXCEPT `LabelTab` and `TimelineScrollList`. Those are large and complex, so I will defer them until tomorrow morning. Note that I have also not reviewed the CSS/SCSS in any level of detail since the proof of their correctness is in the visual representation.

package.cordovabuild.json
```diff
@@ -140,14 +146,16 @@
...
     "react": "^18.2.*",
     "react-dom": "^18.2.*",
     "react-i18next": "^12.3.1",
     "react-native-paper": "^5.8.0",
-    "react-native-safe-area-context": "^4.5.3",
+    "react-native-safe-area-context": "^4.6.3",
+    "react-native-screens": "^3.22.0",
     "react-native-vector-icons": "^9.2.0",
-    "react-native-web": "^0.19.*",
+    "react-native-web": "^0.18.10",
```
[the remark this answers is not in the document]
JGreenlee: https://github.com/react-navigation/react-navigation/issues/11356

package.cordovabuild.json
```diff
@@ -120,10 +127,9 @@
     "cordova-plugin-inappbrowser": "5.0.0",
     "cordova-plugin-ionic-keyboard": "2.2.0",
     "cordova-plugin-ionic-webview": "5.0.0",
-    "cordova-plugin-local-notification-12": "github:e-mission/cordova-plugin-local-notification-12#v0.1.4-fix-android-action",
+    "cordova-plugin-local-notification-12": "0.1.4",
```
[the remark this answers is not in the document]
JGreenlee: Will fix. bad merge conflict

package.cordovabuild.json
```diff
@@ -100,15 +108,14 @@
...
-    "core-js": "^2.5.7",
     "cordova-android": "11.0.0",
     "cordova-ios": "6.2.0",
     "cordova-plugin-advanced-http": "3.3.1",
     "cordova-plugin-androidx-adapter": "1.1.3",
     "cordova-plugin-app-version": "0.1.14",
     "cordova-plugin-customurlscheme": "5.0.2",
     "cordova-plugin-device": "2.1.0",
-    "cordova-plugin-em-datacollection": "git+https://github.com/e-mission/e-mission-data-collection.git#v1.7.8",
+    "cordova-plugin-em-datacollection": "git+https://github.com/e-mission/e-mission-data-collection.git#v1.7.7",
```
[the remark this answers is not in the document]
JGreenlee: Will fix. bad merge conflict

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  "compilerOptions": {
+    "outDir": "./dist/",
+    "sourceMap": true,
+    "module": "es6",
+    "target": "es5",
```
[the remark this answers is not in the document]
JGreenlee: Good point! I was thinking that everything would be transpiled at this point, but now that you point it out I do think it's just .ts and .tsx files.

I can't confirm at this moment that all affected files have been migrated. Let's definitely investigate that before a release.

If not, we can use webpack and babel to add yet another transpilation step, until everything that needs to be transpiled is written in TypeScript.

www/index.js
```diff
@@ -28,17 +28,11 @@ import './js/intro.js';
 import './js/main.js';
 import './js/survey/survey.js';
 import './js/survey/input-matcher.js';
-import './js/survey/one-click-button.js';
 import './js/survey/multilabel/infinite_scroll_filters.js';
 import './js/survey/multilabel/trip-confirm-services.js';
 import './js/survey/multilabel/multi-label-ui.js';
```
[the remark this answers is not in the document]
JGreenlee: Yes; this PR migrates all UI components. The remaining code in those files is just services, and I'm not in a rush to rewrite those.

Leaving the services in Angular for now doesn't block the progress of anything else because they can be accessed from anywhere.

www/js/appTheme.ts
```diff
@@ -0,0 +1,80 @@
...
+      level4: '#e0f0ff', // lch(94% 50 250)
+      level5: '#d6ebff', // lch(92% 50 250)
+    },
+  },
+  roundness: 5,
+};
+
+/* Next, we'll set up 'flavors' of the theme, which are variations on the theme
+    with any number of properties overridden.
+  This is used to automatically style the different types of diary cards
```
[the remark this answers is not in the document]
JGreenlee: I think so. In our meeting, @ctyrrellnrel hinted towards a slightly varied color palette for different sections of the Dashboard.
I think that would be a good use for this mechanism.

webpack.config.js
```diff
@@ -15,59 +12,47 @@ module.exports = {
       // to load CSS and SCSS (enketo-core only supplies SCSS)
       {
         test: /\.(scss|css)$/,
+        include: [path.resolve(__dirname, 'www/css'),
+                  path.resolve(__dirname, 'www/manual_lib'),
+                  path.resolve(__dirname, 'node_modules/enketo-core'),
+                  path.resolve(__dirname, 'node_modules/leaflet')],
```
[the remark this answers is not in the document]
JGreenlee: Yes, it's annoying to be this specific but much more efficient for bundling.
If we can tell Webpack it only has to check these particular directories for X plugin, the bundling goes a lot faster.

It's still slower than I'd like, but when you are using 2 different frameworks at the same time, I can imagine you end up with lot of JS to bundle.

webpack.prod.js
```diff
@@ -0,0 +1,29 @@
...
+  devtool: 'source-map',
+  module: {
+    rules: [
+      /* In production, Webpack minifies JS files and randomizes variable names.
+          This causes problems with AngularJS unless you use explicit annotations,
+          which we don't.
+          https://docs.angularjs.org/error/$injector/strictdi
+          (The syntax we use is like the 'bad' example: implicit annotations)
+          So rather than change every file in our codebase, I'm adding this
+          babel plugin which basically preprocesses our 'bad' code into 'good' code.
```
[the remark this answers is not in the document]
JGreenlee: It is complicated for sure. I was definitely in the weeds for a few hours on getting this plugin to work.

Hopefully many of the gnarly JS issues will be solved by TypeScript and we won't need Babel for much except .jsx resolution

www/js/components/LeafletView.jsx
```diff
@@ -29,8 +29,21 @@ const LeafletView = ({ geojson, opts }) => {
   }, []);
 
   return (
-    <View>
-      <div id="map" ref={mapRef} style={{width: '100%', height: '100%'}}></div>
+    <View {...otherProps}>
```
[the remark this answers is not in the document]
JGreenlee: Yes, in this instance it's only used for styles.

I have been using `...otherProps` in some other places too. Anywhere I need to pass props into the underlying child (grandchild).

www/js/diary/LabelScreen.tsx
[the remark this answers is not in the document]
JGreenlee: Any name suggestions?

I originally had this unified as one, but when I added Navigation it was necessary to separate the Tab from the Screen.

www/js/diary/cards/TripCard.tsx
JGreenlee: I am not 100% sure, but I think once DynamicConfig is converted we won't have to listen for config changes like this.
It depends how we do it, I think.

www/js/diary/list/TimelineScrollList.tsx
JGreenlee: In usage, it feels totally fine. However, if you inspect the HTML and look at what it does, it's pretty gross.

Basically, it uses CSS transforms to flip the entire view along the horizontal axis, like a mirror.
Then, it applies the same transformation to each individual list entry to put them right side up again.

I thought this might be an accessibility concern. As far as a screen reader is concerned, the most recent trips are at the top of the page, not the bottom.
But then again, aren't the most recent trips what a blind / visually impaired user would want to be read first anyway?

JGreenlee: I did see either an issue or early PR where someone planned to rework FlatList/FlashList to handle inversion "the right way". I don't know if that will come to fruition though, as it seems like a major change.

www/js/diary/services.js
JGreenlee: ON_FOOT is there; I experienced that when testing without an inference model and receiving cleaned sections

www/templates/diary/infinite_scroll_detail.html
JGreenlee: I was only keeping it for my own reference because the new Details screen was still missing some elements.
Now it has everything except the Speed chart, which I am deferring until I get a new plotting library up and running.

Probably fine to remove the file now.

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
JGreenlee: I didn't rename them. I've also thought it would be a good idea; for additions as well.

Currently, the server gives us `user_input` and `additions`, and the phone organizes these into `userInput` and `additionsList`

Maybe the phone can do `userInputList` and `additionsList` -- or a better descriptive name if we can think of one

www/js/useAppConfig.ts
JGreenlee: `loading` is scoped to this hook and only refers to the appConfig not being ready yet.

It's used like this in a few components including TripCard and PlaceCard:
`const { appConfig, loading } = useAppConfig()`

When `loading` changes, we know it's safe to access `appConfig`.

www/js/diary/list/TimelineScrollList.tsx
JGreenlee: Doesn't the server give us oldest -> newest?

We could do the reversal sooner if that's what you're asking. But I think the reversal has to happen at some point?

www/js/survey/enketo/enketo-add-note-button.js
JGreenlee: Yes, this and `multilabel-ui` are only around for their services

package.cordovabuild.json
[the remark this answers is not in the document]
JGreenlee: *TODO before merge*

package.cordovabuild.json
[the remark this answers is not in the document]
JGreenlee: *TODO before merge*

www/templates/diary/infinite_scroll_detail.html
JGreenlee: *TODO later*

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
JGreenlee: *TODO later*

www/js/diary/services.js
JGreenlee: *TODO later* (remove RUNNING)

tsconfig.json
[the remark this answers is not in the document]
JGreenlee: *TODO later* (try running a build on an old phone / outdated WebView)

package.cordovabuild.json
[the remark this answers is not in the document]
JGreenlee: *TODO later* (add a comment here)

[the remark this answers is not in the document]
JGreenlee: actually nevermind; you can't put comments in JSON

package.cordovabuild.json
```diff
@@ -7,6 +7,9 @@
     "type": "git",
     "url": "git+https://github.com/e-mission/e-mission-phone.git"
   },
+  "scripts": {
+    "build": "npx webpack --config webpack.prod.js && npx cordova build"
```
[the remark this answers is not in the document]
JGreenlee: *TODO later*

package.cordovabuild.json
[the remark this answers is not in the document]
shankari: Can you also double-check the merge to make sure that you didn't inadvertently overwrite anything else/all fixes were included? The changes to `master_for_platform` were fairly simple but might get lost in the churn of this large and complex rewrite.

www/index.js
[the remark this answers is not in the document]
shankari: Yeah I figured this out later in the review but forgot to come back and modify this comment.

www/js/diary/LabelScreen.tsx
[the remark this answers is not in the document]
shankari: Maybe `LabelListScreen` to distinguish from `LabelDetailScreen`?
You might also want to add a high level comment indicating the tab wraps the screen.

www/js/diary/list/TimelineScrollList.tsx
JGreenlee: Ah, I see how it would make a difference for loading subsequent batches into view
*TODO later*

www/js/diary/LabelScreen.tsx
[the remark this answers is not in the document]
JGreenlee: *TODO later*

package.cordovabuild.json
```diff
@@ -17,9 +20,10 @@
     "@babel/preset-typescript": "^7.21.4",
     "@ionic/cli": "6.20.8",
     "babel-loader": "^9.1.2",
+    "babel-plugin-angularjs-annotate": "^0.10.0",
     "babel-plugin-optional-require": "^0.3.1",
     "concurrently": "^8.0.1",
-    "cordova": "^11.1.0",
+    "cordova": "11.1.0",
```
JGreenlee, line 26: needs revert, TODO before merge

JGreenlee: c0cf40030cc0d90168a035fdfb00aaa8adf37ad4

package.cordovabuild.json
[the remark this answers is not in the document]
JGreenlee: c0cf40030cc0d90168a035fdfb00aaa8adf37ad4

shankari (approved):
LGTM! Everything else seems fixable in a future polishing round

shankari:
I'm merging, building, and pushing to staging.
ETA ~ noon assuming no surprises when I build on native

shankari:
@JGreenlee while building the native app, I have the following error

```
angular.js:14794 Error: transition superseded
    at $StateProvider.$get (angular-ui-router.js:2827:44)
    at Object.invoke (angular.js:5097:1)
    at angular.js:4884:1
    at getService (angular.js:5038:1)
    at injectionArgs (angular.js:5063:1)
    at Object.invoke (angular.js:5089:1)
    at angular.js:4894:37
    at forEach (angular.js:408:1)
    at createInjector (angular.js:4894:1)
    at doBootstrap (angular.js:1955:1) 'Possibly unhandled rejection: {}'
(anonymous) @ angular.js:14794

angular.js:14794 Error: transition prevented
    at $StateProvider.$get (angular-ui-router.js:2828:43)
    at Object.invoke (angular.js:5097:1)
    at angular.js:4884:1
    at getService (angular.js:5038:1)
    at injectionArgs (angular.js:5063:1)
    at Object.invoke (angular.js:5089:1)
    at angular.js:4894:37
    at forEach (angular.js:408:1)
    at createInjector (angular.js:4894:1)
    at doBootstrap (angular.js:1955:1) 'Possibly unhandled rejection: {}'
(anonymous) @ angular.js:14794
angular.js:14794 Error: transition aborted
    at $StateProvider.$get (angular-ui-router.js:2829:41)
    at Object.invoke (angular.js:5097:1)
    at angular.js:4884:1
    at getService (angular.js:5038:1)
    at injectionArgs (angular.js:5063:1)
    at Object.invoke (angular.js:5089:1)
    at angular.js:4894:37
    at forEach (angular.js:408:1)
    at createInjector (angular.js:4894:1)
    at doBootstrap (angular.js:1955:1) 'Possibly unhandled rejection: {}'
(anonymous) @ angular.js:14794
angular.js:14794 Error: transition failed
    at $StateProvider.$get (angular-ui-router.js:2830:40)
    at Object.invoke (angular.js:5097:1)
    at angular.js:4884:1
    at getService (angular.js:5038:1)
    at injectionArgs (angular.js:5063:1)
    at Object.invoke (angular.js:5089:1)
    at angular.js:4894:37
    at forEach (angular.js:408:1)
    at createInjector (angular.js:4894:1)
    at doBootstrap (angular.js:1955:1) 'Possibly unhandled rejection: {}'
```

and the app hangs at the splash screen

shankari:
Upgrading to ui-router to 0.3.1
per https://stackoverflow.com/a/41515812/4040267

```
$ npm list | grep ui
├── angular-ui-router@0.3.1
├── babel-plugin-optional-require@0.3.1
```

did not fix the issue

JGreenlee:
Does it hang on splash immediately on launch, or after onboarding?

shankari:
on splash immediately after launch

JGreenlee:
The instance of `ui-router` we are using is in `manual_lib`, so upgrading it from npm probably won't do anything.

When I added Webpack, I did try using other versions of `ui-router` from npm and it didn't work. I think ionic wants this specific version - I believe it was part of the ionic bundle

JGreenlee:
I'll try building later

shankari:
For the record, I also tried 

```
diff --git a/www/js/app.js b/www/js/app.js
index eeb65a0e..5c95bfcc 100644
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -106,9 +106,10 @@ angular.module('emission', ['ionic', 'jm.i18next',
   console.log("Ending run");
 })

-.config(function($stateProvider, $urlRouterProvider, $compileProvider) {
+.config(function($stateProvider, $qProvider, $urlRouterProvider, $compileProvider) {
   console.log("Starting config");
   // alert("config");
+  $qProvider.errorOnUnhandledRejections(false)

   // Ionic uses AngularUI Router which uses the concept of states
   // Learn more here: https://github.com/angular-ui/ui-router
```

from https://stackoverflow.com/a/43777868/4040267
but got the same error

> I'll try building later

I don't think you need to build, just uninstall + reinstall the devapp

shankari:
For the record, this also happens on iOS (as would be expected)

JGreenlee:
@shankari 
I was unable to reproduce the issue. I see the same error messages, but I am not getting hanging on the splash screen.

Using `framework_migration_prep`:
First, I tried just uninstalling + reinstalling the devapp in the Simulator (Devapp version 3.2.5)
After reinstalling and connecting, the join page still appeared correctly. I was able to log in and everything seemed fine.

Next, I tried reproducing the issue on a build.
I setup and activated the native configuration, preparing for an android build

Then `npm run build`, which bundled JS with `webpack` and then built the app with `cordova`.
I installed the generated apk on a test phone with `adb install`

Result:

image

JGreenlee:
Could the issue somehow be specific to the 'NREL OpenPATH' deployment?

I only built the 'emission' deployment (because that's all I know how to do)

shankari:
wait, so you are still getting the error, but the join page still launches?!
Reading carefully through your message, I think that the problem was that I wasn't running `npm run build`
I was running `npx webpack` manually before, but might have forgotten to run it before this build. Oops!
Stay tuned...

shankari:
I think we will also need to change `bin/sign_and_align_keys.sh` to match - let me do that as well

www/js/diary/LabelTab.tsx
```diff
@@ -0,0 +1,374 @@
...
+        // Only display places with duration >= 60 seconds, or with no duration (i.e. currently ongoing)
+        if (isNaN(end_place.duration) || end_place.duration >= 60) {
+          newListEntries.push(end_place);
+        }
+      }
+    });
+    setListEntries(newListEntries);
+  }, [displayTrips]);
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
JGreenlee: Definitely something to consider.
I would guess that roughly half of this file is just for retrieving, filtering, and formatting the trips.
And that's code that shouldn't really be in LabelTab; it can live elsewhere

www/js/diary/LabelTab.tsx
```diff
@@ -0,0 +1,374 @@
...
+    if (!listEntries?.length) return;
+    invalidateMaps();
+    setIsLoading(false);
+  }, [listEntries]);
+
+  function checkPermissionsStatus() {
+    $rootScope.$broadcast("recomputeAppStatus", (status) => {
```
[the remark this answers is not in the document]
JGreenlee: Yes, it really should be a new component that displays a modal

Modals are tricky until the UI is fully migrated though. I can have Ionic pop up over React screens, but I can't have React pop up over Ionic screens.
So any popup/ modals will have to be some of the last things to convert

www/js/diary/LabelTab.tsx
```diff
@@ -0,0 +1,374 @@
...
+      const allFalseFilters = tripFilterFactory.configuredFilters.map((f, i) => ({
+        ...f, state: (i == 0 ? true : false) // only the first filter will have state true on init
+      }));
+      setFilterInputs(allFalseFilters);
+    }
+    loadTimelineEntries();
+  }, [appConfig, loading, refreshTime]);
+
+  useEffect(() => {
+    Timeline.setInfScrollCompositeTripList(allTrips);
```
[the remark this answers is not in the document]
JGreenlee: > to share the list between all the components

Actually I'm not sure that we even have to do this anymore. Label screen and LabelDetails screen have access to `allTrips` by sharing a Context.

When I go to rewrite services.js, I'll take out anything that isn't needed from Timeline.

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
JGreenlee: `aggUserInput` and `aggAdditions` ?
I'm thinking it's basically an aggregation of processed input(s) from server and unprocessed input(s) on phone

www/templates/diary/infinite_scroll_detail.html
JGreenlee: removed in #1000

www/js/diary/services.js
JGreenlee: removed in #1000

www/js/diary/LabelScreen.tsx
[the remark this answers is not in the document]
JGreenlee: renamed in #1000

package.cordovabuild.json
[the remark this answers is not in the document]
JGreenlee: added `build-dev` script in #1000

www/js/diary/LabelTab.tsx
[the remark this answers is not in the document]
JGreenlee: I refactored this flow in e8090a6ddc1c99eba8e570090802c13997896063. I did not explicitly structure it as a pipeline, but I do think it is simpler and more comprehensible now

www/js/diary/LabelTab.tsx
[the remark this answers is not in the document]
JGreenlee: Deferring for now. We will handle all popups in a future cleanup

www/js/diary/list/TimelineScrollList.tsx
JGreenlee: I still want to consider this, but I think it is lower priority and can be deferred until a later cleanup

PR #994, opened 2023-06-16
More profile migrations

Abby-Wheelis:
I've started a new branch on my end, to try and separate the data table conversions, which are almost good to go, from other conversions on the profile page, which are presenting more trouble than I expected. 

What I am having an issue with right now, is passing the i18n keys from the markup in general-settings.html into the react component defined in SettingRow.jsx - the key is currently coming through as the number 0


www/js/control/SettingRow.jsx
```diff
@@ -0,0 +1,50 @@
...
+import { useTranslation } from "react-i18next";
+import { string, func, object} from "prop-types";
+
+/*
+biggest struggle here is how to pass i18n keys as parameters (or strings in general)
+*/
+
+const SettingRow = ({textKey, iconName}) => {
+    const { t } = useTranslation(); //this accesses the translations
+    console.log("row created!!!!!", "translate hard-code" + t('general-settings.are-you-sure') + "translate via param" + t(textKey)+ " and the icon (plain string) " + iconName);
```
Abby-Wheelis, line 13: @JGreenlee this log statement highlights the issue I've been having (no rush, I'm about to be 98% offline till Tuesday anyways)

www/templates/control/main-control.html
```diff
@@ -5,6 +5,9 @@
       <div ng-click="logOut()" id ="switch-user" class="control-icon-button"><i class="icon ion-log-out"></i></div>
     </div>
     <enketo-demographics-button></enketo-demographics-button>
+    <!-- here is the html related to my (failing) attempts to convert more of the profile screen -->
+    <setting-row text-key = "control.view-privacy" icon-name = "eye"></setting-row>
```
JGreenlee, line 9:
```suggestion
    <setting-row text-key="'control.view-privacy'" icon-name="'eye'"></setting-row>
```

@Abby-Wheelis Put single quotes inside the double quotes and the strings will pass in like you want them to.
I believe this AngularJS thing, not a React thing.

JGreenlee:
Next week, or whenever you resume this, you will probably want to figure out how to handle the popups that many of these `SettingRow`s need to launch.

When I ran into this issue on the Label screen, I tried using [Menu, Dialog, and Modal from React Native Paper](https://callstack.github.io/react-native-paper/docs/components/Modal) to no luck. These components are supposed to sit on top of everything else and actually get rendered higher up in the DOM tree. We won't be able to use these until the upper-level components (i.e. the entire Profile screen) are written in React.

So I decided we can just keep using the existing popovers, which are done using `$ionicPopover`, and launch them from the React component. (I already had a function in `angular-react-helper.jsx` that gets any particular Angular service for use in React). We can swap out the popovers for Menus, Dialogs, or Modals at a later point in the migration.

Here are the relevant places in the code showing how I used `$ionicPopover`:
> https://github.com/e-mission/e-mission-phone/blob/d71eb5fdf296f8a9880f0ff2eddc5d0d381f8b10/www/js/survey/multilabel/MultiLabelButtonGroup.tsx#L24
> https://github.com/e-mission/e-mission-phone/blob/d71eb5fdf296f8a9880f0ff2eddc5d0d381f8b10/www/js/survey/multilabel/MultiLabelButtonGroup.tsx#L41-L47

www/templates/control/main-control.html
```diff
@@ -1,18 +1,32 @@
 <ion-view view-title="{{'control.profile' | i18next}}" ng-class="ion-view-background">
   <ion-content>
-    <div class="control-list-item">
+    
+    <!-- opcode and logout - this row has two clicks! (and an argument for one of them) -->
+    <!-- <div class="control-list-item">
       <div ng-click="copyToClipboard(settings.auth.opcode)" class="control-list-text"><pre>{{settings.auth.opcode}}</pre></div>
```
JGreenlee, line 6: `copyToClipboard()` is not super important functionality that users depend on (honestly, I really only added it to make my own life easier), so if it doesn't make sense to include it I don't think anyone will care.

That said, it might be nice anyway to support actions on the settings labels as well as on the icons. I'll leave it up to you

www/templates/control/main-control.html
```diff
@@ -1,18 +1,32 @@
...
       <div ng-click="logOut()" id ="switch-user" class="control-icon-button"><i class="icon ion-log-out"></i></div>
-    </div>
+    </div> -->
+    <setting-row text-key="settings.auth.opcode" icon-name="'clipboard'" action = "copyOpToClipboard"></setting-row>
+    <setting-row text-key="settings.auth.opcode" icon-name="'exit'" action = "logOut"></setting-row>
+
+    <!-- custom piece already? -->
     <enketo-demographics-button></enketo-demographics-button>
```
JGreenlee, line 13: Yes, this one is its own directive because it handles some more complex logic to load the survey.
I think it should be refactored before we try to convert it, so I would leave it as-is for right now

www/templates/control/main-control.html
```diff
@@ -24,151 +38,200 @@
           style="position: absolute; height: 100%; inset: 0; display: flex; opacity: 0;">
       </div>
     </div>
-    <div class="control-list-item">
+    <!-- <setting-row text-key="'control.reminders-time-of-day' | i18next: {time: settings.notification.prefReminderTime}" icon-name="'clock'" action = "">
+      <input type="time" name="timeOfDay" value="{{settings.notification.prefReminderTimeOnLoad}}"
```
JGreenlee, line 42: I would also leave this one how it was, for now

We need to find a library for time/date pickers
I've been eyeing this one: [react-native-paper-dates](https://github.com/web-ridge/react-native-paper-dates) but haven't implemented it anywhere yet.

For the date picker on the label screen I just left it as HTML and wrapped it in a React component

Abby-Wheelis:
In case anyone's curious, and for organizing my thoughts for myself... Where this stands now is that all but two of the items on the profile page have been converted, at least in a rough-draft format. Some issues that I plan to address next: 
- the toggle-switches just being icons (and not interactive)
- the drop-downs being less interactive (no arrow flipping) I think things like "developer zone" should eventually be ```List.Accordian``` elements, since that's a part of the list system being used now
- putting the ```List.Item``` elements in a ```List``` rather than free-standing in the html: complication with this being the complexity added by the need for ```List.Accordian``` in some places and also ```ControlDataTable``` elements (ie not just ```List.Items```, but I'm sure someone has made a complicated list with these frames before
- as of now ```onPress``` applies to the entire ```List.Item``` or row and I worry this would create confusion so I'd like the icon to be the only thing to press, or make it an ```IconButton```
- needs some styling generally to make it look less like a sterile white screen, maybe make the icons stand out more like they used to

Abby-Wheelis:
Updates today, made good progress, but theres a couple of things I might need some help/advice on:
> the toggle-switches just being icons (and not interactive)

Now they are switches, but I'm struggling with getting the state of the switch to start and persist in the right place (ie on when the setting is on, off when the setting is off). I think likely has most to do with my lack of familiarity with React's "state" but I think there could be a way to tie the state of the switch to the existing way it was being tracked in the angular code... @JGreenlee have you run into anything similar in migration?

> the drop-downs being less interactive (no arrow flipping) I think things like "developer zone" should eventually be List.Accordian elements, since that's a part of the list system being used now

"developer zone" and "user data" are now `List.Accordian` components, defined in `ExpandMenu`. That component leverages map, and currently has a custom data array coded into the html - should this be moved into `general-settings.js` or somewhere else?

> as of now onPress applies to the entire List.Item or row and I worry this would create confusion so I'd like the icon to be the only thing to press, or make it an IconButton

The desired `onPress` is now the action for an icon button embedded in each setting row. However, due to a bug https://github.com/callstack/react-native-paper/issues/3898, each `List.Item` must also have an `onPress` and therefore still acts clickable. Looks like it will be resolved soon, but for now my concern is only partially resolved - users can't accidentally take action when they mean to scroll, but will probably be confused

The "normal" (non collapsing) settings are still individually placed in the .html file, and I haven't started styling yet

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,81 @@
...
+
+//any pure functions can go outside
+const ProfileSettings = ({ settingsScope, settingsObject }) => {
+    // anything that mutates must go in --- depend on props or state... 
+
+    // settingsScope is the $scope of general-settings.js
+    // grab any variables or functions you need from it like this:
+
+
+    //why is settings not defined but everything else is fine?
```
JGreenlee, line 17: I don't know why you would have had this problem, but it looks like you found a workaround?

I think that's fine for now. This way of accessing the $scope variables is temporary anyway.

www/js/control/SettingRow.jsx
```diff
@@ -0,0 +1,59 @@
...
+import { StyleSheet } from 'react-native';
+import { List, IconButton, Switch} from 'react-native-paper';
+import { useTranslation } from "react-i18next";
+import { string, func, bool} from "prop-types";
+
+const SettingRow = ({textKey, iconName, action, desc, switchValue}) => {
+    const { t } = useTranslation(); //this accesses the translations
+
+    let rightComponent;
+    // will using switchValue here only render when the value is true?
```
JGreenlee, line 12: if `switchValue` is not given, it's `undefined`; the condition will evaluate as false and the interpreter will move to the `else` block

www/js/control/ProfileSettings.jsx
Abby-Wheelis: Settings is still undefined, even with the workaround that I was trying. I think it might be a timing issue, since settings is undefined in the moment ProfileSettings needs it, but the functions are fine. I've started moving as much as I can from general-settings into ProfileSettings, since the pass-through is temporary anyways. I'm thinking that will eventually resolve the issues I'm having here when everything is in one place.

www/js/control/SettingRow.jsx
Abby-Wheelis: I've kept working with the switches, and they're consistently showing up now, but the values aren't always changing when expected - when I toggle Tracking from "off" to "on" the value shown on the switch is not changing, but it changes fine when I toggle from "on" to "off".

www/templates/control/main-control.html
Abby-Wheelis: I did end up eliminating copy to clipboard, since it was the only row that had multiple functions

JGreenlee:
@Abby-Wheelis I tried out this branch just now. Things are coming along really well! I can see there are things that are broken, but we can work on those.

With the dialogs, I can see that the `Portal` isn't really working the way you want it to. There's no `Portal.Host` higher up in the component hierarchy, and I don't think we can add one until the migration is further along.

I have a workaround for you. Instead of wrapping the `<Dialog>` in a `<Portal>`, try wrapping it in a `<Modal>` (Modal comes from `react-native`) and give the Modal the same visibility. Also give it transparent=true so we can still see the other settings in the background behind the dialog

```jsx

{/* menu for "nuke data" */}
<Modal visible={nukeSetVis} onDismiss={() => setNukeVis(false)}
    transparent={true}>
    <Dialog visible={nukeSetVis} onDismiss={() => setNukeVis(false)}>
        <Dialog.Title>{t('general-settings.clear-data')}</Dialog.Title>
        <Dialog.Content>
            ...
```

When I added `react-native-paper-dates` to the Label screen, I discovered that `Modal` works without any kind of Portal mechanism.
Docs for Modal: https://reactnative.dev/docs/modal

JGreenlee:
I looked into why the switches are broken. It's definitely AngularJS and React not playing nice with each other.

When a switch setting is changed, there's a popup and then `$scope.refreshScreen()` is called. `refreshScreen` repopulates the settings object, first initializing it as empty.

```js
$scope.settings = {};
$scope.settings.collect = {};
$scope.settings.sync = {};
$scope.settings.notification = {};
$scope.settings.auth = {};
$scope.settings.connect = {};
```
... it then goes on to fill it back in.

React realizes that `settings` has changed and does not wait for the new values to be filled in.
At the time that React checks for updates, the `settings` object looks like 
```js
{
  auth: {}
  clientAppVer: "3.2.5"
  collect: {}
  connect: {}
  notification: {}
  sync: {}
}
```

`collect` is empty, so `settingsObject?.collect?.trackingOn` evaluates to false.
So any time the settings are updated, we will get a false result. This is why you appear to be able to flip the switch off, but never on.

JGreenlee:
@Abby-Wheelis Let's talk tomorrow about a few ways we can navigate the above issue

www/js/control/SettingRow.jsx
Abby-Wheelis: I was able to make some progress with the switches using a dedicated state and useEffect() watch method, the tracking toggle now changes whenever it is clicked!

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,286 @@
...
+                transparent={true}>
+                <Dialog visible={carbonDataVis}
+                    onDismiss={() => setCarbonDataVis(false)}>
+                    <Dialog.Title>{t('general-settings.choose-dataset')}</Dialog.Title>
+                    <Dialog.Content>
+                        {carbonOptions.map((e) =>
+                            <Button key={e.text}
+                            onPress={() =>  {
+                                console.log("changeCarbonDataset(): chose locale " + e.value);
+                                CarbonDatasetHelper.saveCurrentCarbonDatasetLocale(e.value); //there's some sort of error here
```
Abby-Wheelis, line 262: There is an error being thrown on this line, and I actually found the same error being thrown in the same place on master_for_platform. Should I try to change CarbonDatasetHelper to resolve the issue? I can see how changing this value is meant to change the amount of emissions emitted by each mode, is this feature still important?
Screenshot 2023-07-06 at 2 14 53 PM

Abby-Wheelis:
An update on this PR - it's nearing a stable state!

Everything has been moved out of `main-control.html` and now resides in `ProfileSettings.jsx`. Some recent highlights include migrating collection settings (thanks @JGreenlee) and later notification settings fully out of `general-settings.js` mostly out of a need to circumnavigate Reach and Angular not cooperating well. 

>We need to find a library for time/date pickers
I've been eyeing this one: [react-native-paper-dates](https://github.com/web-ridge/react-native-paper-dates) but haven't implemented it anywhere yet.

The prefRemiderTime picker is now implemented as a `react-native-paper-dates TimePickerModal` in `ReminderTime.tsx`

Here's what the page is looking like now - 


What needs to be done before reaching a stable stopping point: 

- [ ] determine why the upcoming notifications aren't showing
- [ ] get the dummy notification popup to work
- [ ] investigate carbon dataset (see earlier comment about error)
- [ ] check if refresh is working (it's not doing anything right now, but I forget the correct behavior)
- [ ] formatting for collect and sync edit settings? (toggles are too big)
- [ ] i18n for force state options (currently hardcoded - @shankari is this something we want to do now?)

shankari:
>  i18n for force state options (currently hardcoded - @shankari is this something we want to do now?)

I am fine with leaving this (and anything else in the developer options) hardcoded for now.

JGreenlee:
Great work!

> determine why the upcoming notifications aren't showing

It is probably because the OPCode you are using has already run its course. Make a new OPCode and see if notifications show up then.
(the `dev-emulator-timeuse` configuration only schedules out a few days, starting when an OPCode is first signed into)

> get the dummy notification popup to work

Not important. It does work, just not when the app is in the foreground.
Click it and quickly minimize the app, then you should get the notification. If it's still broken, it's fine to just remove.

JGreenlee:
A few other UI tasks for this:

* The OPCode should display in monospace font. This is necessary to distinguish `0` from `O` and `l` from `I`. And preferably, if the OPCode is really long, we should show 4 or 5 lines of it before truncating to ellipsis.
  - Hint for monospace: have `SettingRow` accept an optional `descriptionStyle` and feed it through to the underlying `List.Item`.
  - Hint for 4 or 5 lines: The CSS property is `-webkit-line-clamp`, but maybe you need [titleNumberOfLines](https://callstack.github.io/react-native-paper/docs/components/List/ListItem/#props) instead
* Can you lower font sizes by 1 increment?
  - Titles and descriptions are too big. The modals are fine how they are.

Abby-Wheelis:
This branch is now stable, as far as I know. There was a little bit of refactoring required in order to have the notifications work properly - as documented by recent commits. Essentially, increased reliance on promises allowed for finer-grained control to ensure the notifications existed and were up to date before pulling them for display.

Two notes of things I was unable to "fix"
- the formatting of the "edit sync" and "edit select" popups is not ideal (notably the toggles get cut off) -> the code for this is not exposed to this repo
- there is an error thrown when the carbon dataset is changed (see earlier comment) -> this same error is thrown in `master_for_platform` and the behavior is the same as on my test phones (the text changes in settings but no other effect to my knowledge)

Files in .jsx will need to move to .tsx eventually, but this PR is already lengthy and everything is stable as is.

www/js/splash/notifScheduler.js
```diff
@@ -200,7 +250,7 @@ angular.module('emission.splash.notifscheduler',
             return;
         }
         setUpActions();
-        update();
+        update.then();
```
JGreenlee, line 253: Can you add more explanation for what this does?
I read the commit message but still don't see why we don't just call `update()` here

Abby-Wheelis: Just calling `update()` would mean that the "ready" method would return without waiting for the promise that `update()` returns to resolve. If we don't wait here, we have the same messiness as before with the notifications not being scheduled before we get them to populate the table

JGreenlee: If you change it, do you actually observe a difference? Even if you do, I don't think it's for the reason you think it is.

Adding a `.then()` block doesn't block the current scope from continuing. It attaches a function to be called asynchronously once the promise resolves.
The block inside `ready().then(...)` is going to finish executing whether you have a `.then` or not.

Secondly, none of the code here is determining when `ready` returns.
It's just for declaring the things you want to do *after* Ionic is ready and the app config is loaded.

Abby-Wheelis: That makes sense, I think making that change must have been a lucky coincidence that the loading timed out right the few times I tried it then. Taking it out leaves a situation where sometimes it loads in the right timing and sometimes does not. 

I think I'm going to have to find some condition to cause a refresh once that data is ready, kind of like the `appConfig` worked well for the sync and collect settings. I have `refreshNotifications` called on 'appConfig' now, but I think it's too soon since it sounds like that's the same time the `ready` in `notifScheduler` starts

Abby-Wheelis: I ended up refactoring the way that `getScheduledNotifs` works so that if `isScheduling` (an existing boolean in `NotificationScheduler` was true, it used `scheduledPromise.then(*code to get notifications*)` to ensure the scheduling was completed before the fetching. This works consistently. 

It does feel a little clunky to me, and I'm a little worried that it might slow things down (though I haven't noticed that yet) so I'm definitely open to feedback on better ways to ensure the notifs are displayed consistently. I also think this is something that might get smoothed out once we're fully in React.

www/js/splash/notifScheduler.js
```diff
@@ -79,6 +79,55 @@ angular.module('emission.splash.notifscheduler',
...
+    //get scheduled notifications from cordova plugin and format them
+    const getNotifs = function() {
+        return new Promise((resolve, reject) => {
+            cordova.plugins.notification.local.getScheduled((notifs) => {
+                if (!notifs?.length){
+                    console.log("there are no notifications");
+                    resolve([]); //if none, return empty array
+                }
+                
+                const notifSubset = notifs.slice(0, 5); //prevent near-infinite listing
```
Abby-Wheelis, line 116: I realized that some opcodes might have more notifications than we want to show at once, so this caps it at 5 to be displayed!

shankari:
@JGreenlee have you reviewed the final version of this? I will take a look but I am not as familiar with the react codebase/migration patterns.

JGreenlee:
This PR is good and has been for a while.

The only thing I'd comment on here are organizational / tidiness things, like the abundance of comments and console.log statements about the parts that are still "under construction" where this PR leaves off, but since #1009 extends on this PR, those things are being, and continue to be, cleaned up there.

I think it would be confusing to make changes to this PR at this point, so I'd say we can merge this and anything that needs to be refined should just go in #1009

www/js/control/ProfileSettings.jsx
shankari: If it is not a regression, I am fine with moving this to production. It is not super important, and we can fix it while fixing the dashboard.

We should also file a more generic issue for it, because arguably we want to automatically figure out the CO2 emissions based on the location of the trip, instead of having the user choose it. The grid in WA state has a lot of hydropower, for example, while the grid in the Southern states uses more coal.

Note also that this is not just location dependent, it is also time dependent. We are currently using values from 2020, so are not taking into account the impact of efforts to decarbonize the grid by using wind and solar.

EIA API integration and caching, @nataliejschultz?

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,351 @@
...
+           <SettingRow textKey="control.view-qrc" iconName="grid" action={viewQRCode}></SettingRow>
+           {timePicker}
+           <SettingRow textKey="control.tracking" action={userStartStopTracking} switchValue={collectSettings.trackingOn}></SettingRow>
+           <SettingRow textKey="control.app-status" iconName="check" action={fixAppStatus}></SettingRow>
+           <SettingRow textKey="control.medium-accuracy" action={toggleLowAccuracy} switchValue={collectSettings.lowAccuracy}></SettingRow>
+           <SettingRow textKey={carbonDatasetString} iconName="database-cog" action={() => setCarbonDataVis(true)}></SettingRow>
+           <SettingRow textKey="control.force-sync" iconName="sync" action={forceSync}></SettingRow>
+           <SettingRow textKey="control.share" iconName="share" action={share}></SettingRow>
+           <SettingRow textKey="control.download-json-dump" iconName="calendar" action={openDatePicker}></SettingRow>
+           {logUploadSection}
```
shankari, line 230: Good job building a generic `SettingRow` component...

www/js/control/general-settings.js
```diff
@@ -347,57 +280,28 @@ angular.module('emission.main.control',['emission.services',
...
-        navigator.clipboard.writeText(textToCopy).then(() => {
-            ionicToast.show('{Copied to clipboard!}', 'bottom', false, 2000);
-        });
-    }
+    //this feature has been eliminated (as of right now)
+    // $scope.copyToClipboard = (textToCopy) => {
+    //     navigator.clipboard.writeText(textToCopy).then(() => {
+    //         ionicToast.show('{Copied to clipboard!}', 'bottom', false, 2000);
+    //     });
+    // }
```
shankari, line 302: Do we want to restore it? It seems like it would be a good way to get the opcode in text. But it seems like it is a bit of an easter egg since it is non-obvious that clicking it will copy the opcode. And the button in that row is already used for logout.

We could split it into two rows in the future - one with the opcode and a "copy button" and one with the text "logout" and the logout button.

Thoughts?

www/js/splash/notifScheduler.js
shankari: Yes, for the year-long studies, there are definitely more notifications than we want to show at once!
You can resolve this after you have read it.

shankari (approved):
I have read that the react style is to create components for everything, but it is still a bit of surprising to see how many there are when we get to our project's code 😄 We really do have a  separate component for everything (or at least for anything that is not just a `<SettingsRow>`

I am fine with merging this as an intermediate step since the expanded rewrite is ongoing.
I've added some comments to think about as part of the expanded rewrite or future work.
Let's resolve them after they have been included in the expanded rewrite, or have had a new issue filed for tracking them.

www/js/control/general-settings.js
JGreenlee: It only worked on iOS and yeah, I used it as an Easter egg / dev convenience feature.
To copy text on Android we would need an extra permission

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,351 @@
...
+    const editCollectionConfig = ControlCollectionHelper.editConfig;
+    const editSyncConfig = ControlSyncHelper.editConfig;
+
+    //states and variables used to control/create the settings
+    const [opCodeVis, setOpCodeVis] = useState(false);
+    const [nukeSetVis, setNukeVis] = useState(false);
+    const [carbonDataVis, setCarbonDataVis] = useState(false);
+    const [forceStateVis, setForceStateVis] = useState(false);
+    const [collectSettings, setCollectSettings] = useState({});
+    const [notificationSettings, setNotificationSettings] = useState({});
```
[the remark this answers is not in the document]
Abby-Wheelis: Right, and I've actually run into the issues that can be caused by NOT doing it the way react expects in the PR that follows this. I think I wrote a comment about it there but I'm storing an array of Checks that the device is identified to need (moved to one list instead of one per category) and at one point I was mutating the objects in a way that React was not recognizing (because the reference to the check wasn't changing) and so the UI wasn't updating, this is definitely something I personally need to be careful about doing the "react way" so I don't introduce bugs (again) when working with complex states.

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,351 @@
...
+        
+        const collectionPluginState = await ControlCollectionHelper.getState();
+        newCollectSettings.state = collectionPluginState;
+        newCollectSettings.trackingOn = collectionPluginState != "local.state.tracking_stopped"
+                                        && collectionPluginState != "STATE_TRACKING_STOPPED";
+
+        // I am not sure that this is actually needed anymore since https://github.com/e-mission/e-mission-data-collection/commit/92f41145e58c49e3145a9222a78d1ccacd16d2a7
+        const geofenceConfig = await KVStore.get("OP_GEOFENCE_CFG");
+        newCollectSettings.experimentalGeofenceOn = geofenceConfig != null;
+
```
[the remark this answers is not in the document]
Abby-Wheelis: That was Jack's catch actually! I just took those lines out [56cdd83](https://github.com/e-mission/e-mission-phone/pull/1009/commits/56cdd832ca4b36a4db8889c71070e8f6d8262584)

www/js/control/ProfileSettings.jsx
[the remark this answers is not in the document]
JGreenlee: Exactly! To summarize, we should treat all state as immutable

If we want to update an object which is stored as state, we should make a shallow copy with the change and pass that to the `set` function.
Which is how Abby resolved her issue in the other PR, the `...` operator makes a shallow copy

www/js/control/PopOpCode.jsx
```diff
@@ -0,0 +1,49 @@
...
+        <Modal visible={visibilityValue} onDismiss={() => setVis(false)}
+        transparent={true}>
+            <Dialog visible={visibilityValue} 
+            onDismiss={() => setVis(false)}
+            style={styles.dialog(colors.elevation.level3)}>
+                <Dialog.Title>{t("general-settings.qrcode")}</Dialog.Title>
+                <Dialog.Content style={styles.content}>
+                    <QrCode value={tokenURL}></QrCode>
+                    <Text>{t("general-settings.qrcode-share-title")}</Text>
+                    <IconButton icon="share" onPress={() => action()} style={styles.button}/>
```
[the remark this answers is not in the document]
Abby-Wheelis: I ended up setting `prepopulateQRMessage.message`, because I did not find a "text" option in that readME, but adding "message" and keeping "url" meant that the opCode now shows in the share sheet and can be saved to files. See [this comment](https://github.com/e-mission/e-mission-phone/pull/1009#issuecomment-1686686220) for a demo and more explanation. I also added the text to the modal!

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,351 @@
...
+    const toggleLowAccuracy = function() {
+        ControlCollectionHelper.toggleLowAccuracy();
+        refreshCollectSettings();
+    }
+
+    const shareQR = function() {
+        var prepopulateQRMessage = {};  
+        var qrAddress = "emission://login_token?token="+settings?.auth?.opcode;
+        prepopulateQRMessage.files = [qrAddress];
+        prepopulateQRMessage.url = settings.auth.opcode;
```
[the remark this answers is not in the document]
Abby-Wheelis: added a `message` attribute, to `prepopulateQRMessage` see my other comment here and more info in the other PR

www/js/control/ProfileSettings.jsx
```diff
@@ -0,0 +1,351 @@
...
+
+    //angular services needed
+    const CarbonDatasetHelper = getAngularService('CarbonDatasetHelper');
+    const UploadHelper = getAngularService('UploadHelper');
+    const EmailHelper = getAngularService('EmailHelper');
+    const ControlCollectionHelper = getAngularService('ControlCollectionHelper');
+    const ControlSyncHelper = getAngularService('ControlSyncHelper');
+    const CalorieCal = getAngularService('CalorieCal');
+    const KVStore = getAngularService('KVStore');
+    const NotificationScheduler = getAngularService('NotificationScheduler');
```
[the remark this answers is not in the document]
Abby-Wheelis: Yep! It's unused now, so no need to get the service

www/js/control/ProfileSettings.jsx
Abby-Wheelis: I opened an issue to track this [#954](https://github.com/e-mission/e-mission-docs/issues/954)

www/js/control/general-settings.js
Abby-Wheelis: If we wanted to split into two rows, I could absolutely do that -- the copy function itself would be simple, it was just that I couldn't make the description have a separate "onClick" from the row itself. 

If Android requires other permissions is that something we want to pursue? We already have a way to save the opcode as a document, so I don't think it's critical to users.

One way I can think of to keep it as an "easter egg" would be to display the opcode in one row, with no function, and maybe a "profile" icon, and then have the logout row just below that. Then the copy click can remain a dev/ios bonus. @JGreenlee @shankari what do you think of 2 rows with the function only available to iOS?

PR #995, opened 2023-06-17
Simple React element definition in metrics.jsx

ctyrrellnrel:
-- Small component to test creation of angular friendly react components -- Will be expanded later
-- added as a module to metrics.js


ctyrrellnrel:
This is just a really basic thing, but I wanted to have a pr open

ctyrrellnrel:
I have some questions, and will post them after next commit tomorrow 🚀

shankari (commented):
You might want to add the `<metrics-card` to the metrics HTML and verify that it does show the HELLO before making additional changes.

www/templates/main-metrics.html
```diff
@@ -131,48 +131,7 @@ <h4 class="dashboard-headers" ng-i18next>{{'main-metrics.calories'}} </h4>
...
-                        {{ summaryData.defaultSummary.count[dIndex + i].key }}
-                      </div>
-                      <div class="metric-summary-content">
-                        {{ formatCount(summaryData.defaultSummary.count[dIndex + i].values).slice(-1)[0] }}
-                      </div>
-                    </div>
-                  </div> 
-                </div>
-            </div>
+            <metrics-card class = "detail item-text-wrap control-info" metric-card-heading = 'main-metrics.duration' default-summary-arr='summaryData.defaultSummary.distance'></metrics-card>
```
shankari, line 134: why do you need the `detail` and `control-info` classes here?

shankari (commented):
Please add screenshots indicating what this looks like as you change it.

www/js/metrics.jsx
```diff
@@ -1,27 +1,79 @@
...
+import { string ,object, array,any} from "prop-types";
 import { Card } from "react-native-paper";
 
 // Note the camelCase to dash-case conventions when translating to .html files!
 // val with explicit call toString() to resolve bool values not showing
-const metricsCard = ({ metricCardHeading, defaultSummaryVar}) => {
+
+
+
+const metricsCard = ({ mainMetrics, mainMetricsProp, summaryData}) => {
```
ctyrrellnrel, line 11: Is this not a good way to pass in these variables? They seem to not behave like objects, even though they have properties? Is it specific to angular? Will keep looking into this. Putting them in console.log gives NaN and zero

ctyrrellnrel, line 11: Is this not a good way to pass in these variables? They seem to not behave like objects, even though they have properties? Is it specific to angular? Will keep looking into this. Putting them in console.log gives NaN and zero

ctyrrellnrel:
> Please add screenshots indicating what this looks like as you change it.

# Screenshots before first commit
Screenshot 2023-06-21 at 5 05 36 PM

ctyrrellnrel:
[pervious screenshot](https://github.com/e-mission/e-mission-phone/pull/995#issuecomment-1601881869)
# Screenshot at [third commit](https://github.com/e-mission/e-mission-phone/pull/995/commits/672729c6ad67b2c3ec6eb94908e6c76aae970baa)


Screenshot 2023-06-21 at 5 34 30 PM

www/templates/main-metrics.html
ctyrrellnrel: I don't, just changed it

PR #998, opened 2023-06-29
💅 Label Screen cleanup & polishing 

JGreenlee:
After converting the Label Screen UI to React components with https://github.com/e-mission/e-mission-phone/pull/993, there are several items to address.

This PR will fix regressions, tidy, and continue to improve usability.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,6 +5,7 @@
     "module": "es6",
     "target": "es5",
     "esModuleInterop": true,
+    "resolveJsonModule": true,
     "jsx": "react",
     "allowJs": true,
     "moduleResolution": "node"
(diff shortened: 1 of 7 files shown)


JGreenlee:
The above commits are minor cleanups I already had, but didn't want to tack onto #993 while it was being reviewed

shankari:
> The most notable thing is that I implemented caching for the address names we fetch from OSM. I've always found it annoying that they're retrieved again on every app load.

Note that we are also signing up for a paid subscription to geofabrik.de. Once that is done, we can fill in the addresses on the server. Regardless, we will still need direct client calls for draft trips, so this is a useful fix. Just might be less cool down the road.

JGreenlee:
Sounds cool!

I think this was still largely worthwhile because now we have a lightweight storage method to use instead of `localstorage` (on a true native app, `localstorage` will not exist) and I have figured out a process for caching, which we could do for other network calls down the line

shankari (approved):
LGTM!

PR #999, opened 2023-06-30
Adding Lao to survey

Abby-Wheelis:
Alongside the work in e-mission translate, [e-mission translate PR#31](https://github.com/e-mission/e-mission-translate/pull/31), the survey file needed to be updated to include Lao language support


shankari:
High level comments:
- Please also update the xls file - our survey builder (kobotoolbox) only supports xls. This would be in `survey-resources/data-xls`
- You then need to copy the final json file to `www/json` - note that the survey is loaded from
https://github.com/e-mission/nrel-openpath-deploy-configs/blob/main/configs/usaid-laos-ev.nrel-op.json#L41

That should be it! You can then change this from draft and I can merge.

shankari:
Let's merge this also to framework_migration_prep for now since I will be pulling from there for the next release

Abby-Wheelis:
I eventually update just the xls file, the extra commits are because I accidentally committed changes from locales (which I did not want to do). Sorry! I know I'm making things more confusing, I'm still learning GitHub

shankari:
No worries. Learning is what you do when you are an intern.
If this is ready to merge, move it out from draft, and I will squash merge

shankari:
Don't forget to copy over the final json into `www/json`!

shankari:
@Abby-Wheelis squash merged this as well to avoid blame churn

PR #1005, opened 2023-07-20
Jul 2023 fixes - related to label screen react migrations

Abby-Wheelis:
This PR works to address issues in issue [#932](https://github.com/e-mission/e-mission-docs/issues/932 ) relating to the migration of the label screen into React


Abby-Wheelis:
I believe P0/P1 issues have now been addressed. The three unstruck notes in [this comment](https://github.com/e-mission/e-mission-docs/issues/932#issuecomment-1636574012) show that we needed to fix:

- QR code - addressed React migration issues by grabbing the svg of the screen and downloading it as a png as shown in the demo for the react-qr-code
- Duplicate activities - the problem that I identified was that the existing activities were being pushed AGAIN into the list for display alongside the new activity, I've addressed by clearing the existing list before the updated list is pushed into it: I have some concerns about the universality so I do plan to keep testing / keep and eye on this behavior
- Dashboard errors - I believe this was a result of a server update and testing happening at the same time, I'm no longer able to recreate in the emulator or on a test phone, where the issues originally arose

www/js/diary/cards/TripCard.tsx
```diff
@@ -46,7 +46,7 @@ const TripCard = ({ trip }) => {
     navigation.navigate("label.details", { tripId: trip._id.$oid });
   }
 
-  const mapOpts = { zoomControl: false, dragging: false };
+  const mapOpts = { minZoom: 4, maxZoom: 17, dragging: false };
```
JGreenlee, line 49: Please revert this change.

We only want zooming to be enabled in the details view. We don't want zooming enabled on the `TripCard`s.

JGreenlee (commented):
Almost ready for merge.

Once ready, maybe we can put an RC2 on staging for a couple days to make sure everything is fixed and ready for production?

www/js/components/LeafletView.jsx
```diff
@@ -65,9 +65,12 @@ const LeafletView = ({ geojson, opts, ...otherProps }) => {
   );
 };
 
-const startIcon = L.divIcon({className: 'leaflet-div-icon-start', iconSize: [18, 18], html: '<div class="leaflet-div-ionicon leaflet-div-ionicon-start">★</div>'});
-const stopIcon = L.divIcon({className: 'leaflet-div-icon-stop', iconSize: [18, 18], html: '<div class="leaflet-div-ionicon leaflet-div-ionicon-stop">⚑</div>'});
-const pointToLayer = (feature, latlng) => {
+const startIcon = L.divIcon({className: 'leaflet-div-icon-start', iconSize: [18, 18], 
+  html: '<svg class="div-icon-start" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><path d="M12 2C8.1 2 5 5.1 5 9C5 14.2 12 22 12 22S19 14.2 19 9C19 5.1 15.9 2 12 2M14.5 13L12 11.5L9.5 13L10.2 10.2L8 8.3L10.9 8.1L12 5.4L13.1 8L16 8.3L13.8 10.2L14.5 13Z" /></svg>'});
```
JGreenlee, line 69: Where did these SVG paths come from?

Abby-Wheelis: I downloaded the svgs from [here](https://pictogrammers.com/library/mdi/) and when I inserted the files this what what was inside. I was thinking directly putting them in was simpler than routing to a file path? But I could put the files in so that it's clearer where things came from

JGreenlee: Ok. Ideally we'd get them from the "MaterialCommunityIcons" font-family (registered in `angular-react-helper.jsx`). That's the underlying font that React Native Paper is using when it renders icons

But I think if you just add comments in the code explaining where the hardcoded SVGs are from, this is ok for now

Abby-Wheelis: They are rendered with the "MaterialCommunityIcons" font now, but I had trouble fully centering them in the circles. They are looking fine left-right but not up-down

shankari (approved):

www/js/diary/diaryHelper.ts
```diff
@@ -40,49 +52,61 @@ export function motionTypeOf(motionName: MotionTypeKey | `MotionTypes.${MotionTy
...
  * @returns true if the start and end timestamps fall on different days
+ * @example isMultiDay("2023-07-13T00:00:00-07:00", "2023-07-14T00:00:00-07:00") => true
  */
-export function isMultiDay(beginTs: number, endTs: number) {
-  if (!beginTs || !endTs) return false;
-  return moment(beginTs * 1000).format('YYYYMMDD') != moment(endTs * 1000).format('YYYYMMDD');
+export function isMultiDay(beginFmtTime: string, endFmtTime: string) {
+  if (!beginFmtTime || !endFmtTime) return false;
+  return moment.parseZone(beginFmtTime).format('YYYYMMDD') != moment.parseZone(endFmtTime).format('YYYYMMDD');
 }
```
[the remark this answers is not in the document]
JGreenlee: I chose the `fmt` strings in consideration for the future - Luxon cannot directly consume the `dt` objects (a few things would need to be renamed)

shankari:
Tested against the iOS 13 emulator since @JGreenlee couldn't do it - works (no WSOD)

shankari:
The Lao translations seem to be working, with the caveat that a couple of buttons are not getting translated.

Screenshot 2023-07-25 at 5 45 09 PM


They are in the lo.json

```
    "consent": {
        "button-accept": "ຂ້ອຍຍອມຮັບ",
        "button-decline": "ຂ້ອຍປະຕິເສດ"
    },
```

but are not showing up correctly in the translation

JGreenlee:
I bet the templates in `e-mission-translate` are still trying to use ng-translate.

This would affect the other languages too.

JGreenlee:
What's the point of having different templates for different languages anyway?
Are they functionally different in some way?

shankari:
> I bet the templates in e-mission-translate are still trying to use ng-translate.

I already tried


```
diff --git a/lo/i18n/intro/consent-lo.html b/lo/i18n/intro/consent-lo.html
index 286a97c..c431766 100644
--- a/lo/i18n/intro/consent-lo.html
+++ b/lo/i18n/intro/consent-lo.html
@@ -8,10 +8,10 @@

 <div class="row" style="margin-top: 20px;" translate-namespace="consent">
   <div class="col" style="margin-right: 5px">
-      <button class="button button-block button-assertive button-outline" ng-click="disagree()" id="consent-button"><span translate>{{'.button-decline'}}</span></button>
+      <button class="button button-block button-assertive button-outline" ng-click="disagree()" id="consent-button"><span translate>{{'consent.button-decline'}}</span></button>
   </div>
   <div class="col" style="margin-left: 5px">
-      <button class="button button-block button-positive button-outline" ng-click="next()" id="consent-button"><span translate>{{'.button-accept'}}</span></button>
+      <button class="button button-block button-positive button-outline" ng-click="next()" id="consent-button"><span translate>{{'consent.button-accept'}}</span></button>
   </div>
 </div>
 </div>
``` 

And it didn't work. Will poke around further.

> What's the point of having different templates for different languages anyway?
> Are they functionally different in some way?

No, they are not. They are an artifact of the time when we were thinking of OpenPATH more as a framework - a set of tools that people would use to build their own apps. So we wanted to make it easy for people to write their own consent form and their own summary and so on. And FabMob (the group that contributed the i18n) chose to leave these files as files because (I guess) they were large chunks of text that would be easier to translate if they were all together (e.g. you could copy paste from some other document

As I said in the issue https://github.com/e-mission/e-mission-docs/issues/938#issuecomment-1650200978 we can clean this up

> the summary basically has no actual text in it any more; everything comes in from the dynamic config. (as an aside, we can probably remove it from the list of i18n files in intro.js (var allIntroFiles) since all the summary-* files are the same :)

shankari:
Fixed!

Screenshot 2023-07-25 at 7 30 06 PM

shankari:
The survey works fine and autopopulates to Lao
Screenshot 2023-07-25 at 7 32 43 PM

HOWEVER, there's a weird issue with the step through.
I'm attaching a video to the issue

Issue #40, opened 2016-06-20
User heatmap in addition to aggregate heatmap


shankari:

shankari:
@adambrown, do I assume you are working on this?

shankari:
@adambrown, sorry, wrong Adam Brown.
@adambrown13, do I assume you are working on this?

adambrown13:
I am. I have a few questions about the data collection for the aggregate

shankari:
Go ahead and ask them in here, and I'll respond.
Note that there's no separate data collection for the aggregate - the aggregate is basically all the data collected, treated as a giant pool with the `user_id` stripped out.

PR #1009, opened 2023-07-28
Migration of Profile Logic

Abby-Wheelis:
This PR continues where #994 leaves off, working towards fully migrating the logic pieces of the Profile/Settings Page


Abby-Wheelis:
A few issues that have come up during migration that I'm working on solving next:

- [ ] position Snackbar alerts properly - they are appearing off the visible part of the screen
- [ ] userData is not loaded in the right time frame, need some sort of listener for the change in the service probably
- [ ] "go to states" need to be migrated - like those used for reconsent, checking logs, etc
- [ ] permissions elements not currently in the status modal (`<permissionscheck>` element?) - need to understand this more so I can include and/or migrate it
- [ ] privacy policy is yet to be migrated - this relies on the complex template html files rather than i18n, so hard to migrate
- [ ] few remaining parts of general-settings to be migrated

Once these are handled, all the functionality in `general-settings` will have been moved to `ProfileSettings`, but there is still heavy reliance on Angular services, and ProfileSettings is getting to be a long file, so I think I will want to refactor it into some more parts at some point.

Abby-Wheelis:
Continuing to migrate away from angular, and ionic popups in particular, I've been trying to convert the Permissions screen that pops up into a React element. This is what it looks like so far. It is controlled in the `AppStatusModal` element, but doesn't quite work yet. 
Screenshot 2023-08-09 at 2 55 47 PM
When I pull up the Modal by clicking "check app status" an error is thrown in the console about the icon name being blank (there should still be status icons by all the line items). I think this is caused by the fact that "setting up" is not complete by the time the elements need to be displayed, so the variables that control the icon names are blank. 

@JGreenlee do you know of a way to get this setting up/loading to happen before the element is displayed? I'm not sure how React handles that when it comes to hidden elements, I'm showing/hiding it the same way I am with other popups in `ProfileSettings` and haven't had this problem yet, so maybe it's an issue with the AppStatusModal itself and not the way React handles hidden things.

Abby-Wheelis:
I have discovered that it is not a timing problem, it must be a different bug, as the icon names are set to blank in the log statement I placed after the set up functions are called, none of which are async. 
Screenshot 2023-08-09 at 3 26 29 PM

Abby-Wheelis:
I ended up setting a default of the x icon, which resolved that error. I've now found another error where the OS version is being set and working to debug that - printing "window.device" to the console showed an object with a version, so it might have to do with the application of the split function.

Abby-Wheelis:
I'm also seeing an issue where the `locChecks` are not getting set. I haven't figured out why yet, but maybe the way I'm declaring the objects and/or setting them is wrong? The console.log statement printing the description tags is fine, but I can't get a console.log statement after setting the `locChecks` to run.

Abby-Wheelis:
@JGreenlee and I met today, and talked through some of the remaining issues in the Profile page, including the trouble I'd been having with the permissions. He suggested a refactor that might be possible and I think would help streamline the file `AppStatusModal` which is basically a clone of `permissioncheck.js` in React form and includes the UI pieces of the modal. What Jack and I were talking about was beginning by setting up a list of the checks that apply to that device, then updating them all when something changes rather than updating specific checks individually. This would work a little better with the way React tracks state, but we wanted to make sure there wasn't a known performance or other reason to keep them individual.  Before proceeding down that road, I wanted to ask @shankari is there any reason that the checks in that file can't be run as a batch?

Abby-Wheelis:
I have also had some issues with the UserData (height, weight, etc) syncing between the dashboard and profile. I realized that this is predominantly used for the calorie calculations, which we have been talking about removing in the dashboard anyways. I looked through the service that handles the information, `CalorieCal` and I can see that it's used for the calculations related to MET and calories. If we move to counting minutes, then we might not need the user's height/weight data. I think there's a couple choices here:
- take out user data in the profile it it's leaving dashboard soon anyhow
- put the user data back to the old way (I can't recall if this worked great either, but should check)
- find a way to listen for changes in `CalorieCal` to update the user data proactively, as right now it's behind but catches up when manually refreshed 1-2 times

Abby-Wheelis:
> position Snackbar alerts properly - they are appearing off the visible part of the screen

This is only an issue when developer zone is open, at least. The alerts position themselves in the center of the view now, so if the view is expanded by the dev zone then it shows up off the screen. I want to figure out how to position them relative to the viewport/phone screen and not the view contents, but I haven't yet.

shankari:
> What Jack and I were talking about was beginning by setting up a list of the checks that apply to that device, then updating them all when something changes rather than updating specific checks individually. This would work a little better with the way React tracks state, but we wanted to make sure there wasn't a known performance or other reason to keep them individual. 

No performance reason as to why they can't run in a batch.

I'm just a bit confused because they do, in fact, run in a batch.
We set them up when the app starts
https://github.com/e-mission/e-mission-phone/blob/e91c8d05a5e84cd1a538a5e76dbb5515fba748cc/www/js/appstatus/permissioncheck.js#L397
and then recompute them in a batch
https://github.com/e-mission/e-mission-phone/blob/e91c8d05a5e84cd1a538a5e76dbb5515fba748cc/www/js/appstatus/permissioncheck.js#L405
on resume and restart

- were you thinking about recompute the whole state in a batch when the user "fixed" a value instead?
- when I wrote this, I originally wanted to make a generic "status" directive and then just use it 'n' times. That is why the code is structured properly to have check functions that have the same interface. I ran out of time to actually refactor into directives - there was some other firefighting task that came up - but I would encourage you to *do it right* this time.

shankari:
> take out user data in the profile it it's leaving dashboard soon anyhow

I am fine with this approach. No point in trying to preserve functionality which has a short shelf life.
In fact, I would even go one step further and say that I am fine with removing the setting of the weight and height from the dashboard as well if that makes your life easier.

I just want all functionality that *is* visible to users to be working (aka not broken).

Abby-Wheelis:
I have been trying to get the "alerts" about syncing, docs, etc. to be formatted such that they show up on the visible parts of the page. I've gone back and forth between `Modal` and `Portal` as a wrapper and eventually moved them out of being a separate component and into `ProfileSettings` where I'm getting what's pictured. I think using the component I made for this `AlertBar` would be better, but this is the most usable way of formatting I've tried. I think it might have to do with the context from some of the [websites](https://www.developerway.com/posts/positioning-and-portals-in-react) I've read, which sort of explains how having separate components is making formatting more difficult. This is what it looks like now: I might try working with context now, but this might be something we clean up once we have less mixing of Angular and React.

Abby-Wheelis:
RE the new Permissions flow, here are some drawings roughly outline what we have and what I'm planning to try migrating to. The current flow isn't working well because the status is trying to update based on the check that's not done being set, so setting then refreshing will hopefully help. Also, developing a list of checks at the beginning will facilitate listening for changes so the status can be updated.

Abby-Wheelis:
I'm making progress with these as I start to implement a more "batched" workflow, and now have the checks showing up. (I'll get the descriptions filled in soon)
 
Looking at making these changes and the way it looks now, what if we moved to more of a singular list rather than a binned list? Each permission would still have it's name and description, and I think we could put the "section" descriptions in a "why do we need this" Modal/Dialog of some sort. Then we could also have a single refresh button that checked all of them, which was something I believe Jack mentioned when we talked. I feel like having a single list and single refresh button would help "declutter" this screen, in a sense, while keeping the information available to the user if they want it.

Abby-Wheelis:
I went ahead and implemented it as a singular list, with just one refresh button. I think this simplified the screen a lot, as well as the code. There's still a few things I need to do, especially with formatting, but it's starting to work well.


- [ ] incorporate the `statusClass` to style the icons and/or text with red/green indication of status
- [ ] ensure "scrollability" for cases with more permissions
- [ ] make descriptions fully visible to each permission
- [ ] add the "why we need this" popup and make it visibly accessible
- [ ] continue to refactor / streamline code (use `useEffect`, lots of repeated fcn defs for fix/refreshes)
- [ ] resolve issue encountered with needing to press "refresh" to get overall status to be true, even when all permissions are green (I think `useEffect` will help here)

Abby-Wheelis:
Update on visual state after incorporating some of the changes that @JGreenlee recommended this morning:
                         
The full descriptions are now visible, with 5 lines seeming to be a comfortable maximum. The icons are now color-coded, but I still need to figure out how to pull the colors from the `appTheme.ts file`. The Panel is now scrollable, which will be needed it seems mostly on Android devices (they have the background permissions as well). The scroll bar will be permanently visible on android platforms. The "What are these used for?" Button opens the Modal on the right, where permission categories and their purposes are now listed. The text for this panel is somewhat customized by platform where it is compiled in `setupPermissionText()`. 

The `AlertBar` components are now formatted well thanks to Jack's formatting suggestions and the realization that I should be using `react native modal` over `react native paper modal`. Now the alerts pop up for a couple seconds at the bottom of the screen, and not relative to the content!

Abby-Wheelis:
Now I'm encountering an issue where the the icons aren't updating visually on time, but from what I can tell, they are getting updated in the list of checks stored in state at the right moments. The video shows the app with two wrong permissions, fixing one, refreshing, and then fixing the other and having both change at the same time. What should happen is each turn green upon refresh after fixing. I'm suspecting it's a timing issue, where the changes haven't gone through when we're checking that variable or it could be that React isn't quite smart enough to always know when an attribute of an object in a list is updated, but I need to keep investigating. 
https://github.com/e-mission/e-mission-phone/assets/54848919/3d6a8f7d-9e0e-4a96-b236-3ac309e75bc2

Abby-Wheelis:
At least part of the problem was that I was "mutating" the state rather than changing it, so React was not picking up on the need to re-render the components ([answered here](https://stackoverflow.com/questions/70010619/react-why-are-my-components-not-re-rendering-when-controlled-by-an-array-state/70010826#70010826), [and here](https://stackoverflow.com/questions/70104545/react-ui-not-updating-although-im-changing-the-state)). Now, the state is more responsive to change, there's a lag with the notification settings, but I think that might be the time it takes ios to update... keeping an eye on that. 

Something I hope to address next is that neither `settingsScope.$on("recomputeAppStatus", function() ` nor `$ionicPlatform.on("resume", function()` is getting tripped, since the console.log statements are never showing up in the console. The way I'm retrieving both scope and $ionicPlatform might be the issue, since that's where I'm having to interface with angular. I might try passing them in from `ProfileSettings` since I know it works there rather than fetching them in the `AppStatusModal`

Abby-Wheelis:
>Something I hope to address next is that neither settingsScope.$on("recomputeAppStatus", function()  nor $ionicPlatform.on("resume", function() is getting tripped, since the console.log statements are never showing up in the console. 

Now the `SettingsScope` is passed in, and `$ionicPlatform` is still retrieved as an Angular service, but nothing changed. Then I checked on the onboarding permissions screen, which I haven't, to my knowledge, changed, and this was the same behavior that I saw there, with no prompting of recompute or resume. I watched the console for messages beginning with "PERMISSION", since both listeners have that console statement. From the Permissions screen, I went into settings, changed location from "Always" to "never" and then went back to the devapp. I never saw the expected console statements and the permission did not turn red until I clicked refresh. At this point, my React version and this Angular version are behaving the same, so I think I need to investigate further into `$ionicPlatform.on("resume"...)` and see how it works / where it comes from.

Abby-Wheelis:
I just checked on `master_for_platform`, and I see the message from "recompute app states" which is the the one controlled by the angular `$scope.$on(recomputeAppStatus...)` but not seeing the `app has resumed, should refresh` message from `$ionicPlatform.on("resume"...)` at all. So there must be some sort of issue with the resume listener here, thought not everywhere because the `storage.js` resume message is thrown.

JGreenlee:
I have 2 potential ideas:

1) We can try to detect app resume events using `AppState` from React Native / RN Web instead of relying on `$ionicPlatform`. I experimented with this a little bit before - it does generally work but I decided to revisit it later
2) We can just refresh checks every 1-2 seconds using `setInterval()` any time the popup is visible. (However note that `setInterval()` is deceptively tricky to do properly in React)

Abby-Wheelis:
Using React Native's `AppState` works great! At @JGreenlee's suggestion, I incorporated it using a [custom hook](https://react.dev/learn/reusing-logic-with-custom-hooks) `useAppStateChange` which executes a passed function when the state goes from !active to active. Testing in the devapp, the function gets called if I go to the home page then back to the app, and whenever I go back into the app after "fixing" a setting. It also is immediately reflecting the breaking/fixing of settings when I navigate away and return to make those changes -- in other words, it works great!

Jack and I also talked about how this custom hook and `AppStatusModal` can be incorporated down the line to check the settings each time the user launches the app, and to show them permissions immediately if something is broken. 

I *think* I'm moving into a polishing stage with the permissions now, starting with some text (header on explanations, disappearing location permissions explanation, ect) so feel free to make suggestions if something sticks out or you have wording suggestions for the "what are these used for?" button!

Abby-Wheelis:
I do have a question about one of the permissions, probably for @shankari:

The samsung background permission does not seem to be used anywhere, I can see it set in [`permissioncheck.js`](https://github.com/e-mission/e-mission-phone/blob/b6da3bbaa1b14f19937c079bef975dc6752c427e/www/js/appstatus/permissioncheck.js#L383) but as far as I can tell it's never shown to the user in [`permissioncheck.html`](https://github.com/e-mission/e-mission-phone/blob/b6da3bbaa1b14f19937c079bef975dc6752c427e/www/templates/appstatus/permissioncheck.html) like the other checks are. And I don't see it used anywhere, other than setting it, when I [searched all of e-mission](https://github.com/search?q=org%3Ae-mission+backgroundRestricted&type=code). Is this something that we need to make sure Samsung users are prompted to fix?

If we need these permission from Samsung users, I can make a check that ends up in the list managed for Samsung users, else, can I remove this code?

shankari:
@Abby-Wheelis I think we discussed this in person

> is there any reason that the checks in that file can't be run as a batch?

No. In fact, they are effectively run as a batch on every startup and resume

```
        console.log("app is launched, should refresh");
        $scope.platform = $window.device.platform;
        $scope.osver = $window.device.version.split(".")[0];
        $scope.setupPermissionText();
        $scope.setupLocChecks($scope.platform, $scope.osver);
        $scope.setupFitnessChecks($scope.platform, $scope.osver);
        $scope.setupNotificationChecks($scope.platform, $scope.osver);
        $scope.setupBackgroundRestrictionChecks($scope.platform, $scope.osver);
    });
```

shankari:
> The samsung background permission does not seem to be used anywhere, 

Ah the infamous samsung background permission checks. So Samsung is the worst offender on https://dontkillmyapp.com/ (although OnePlus is second). And as you can also see, there is no programmatic way to allow users to turn the settings off. So back before we had the app status screen, we had a generic popup which, if the user was on a Samsung, would tell you to turn off the background optimizations. 

But we had no way of "popping up" the screen with the optimizations (which was different for every version BTW) and worse, no way of checking that the permissions were indeed changed. So it didn't fit into the pattern for the app status screen and I removed it.

I am not 100% sure what to do about the Samsung/OnePlus issue going forward. 
- we are not really allowed to link to non-.gov sites, so can't link to dontkillmyapp
- all the instructions on dontkillmyapp don't apply to us anyway - e.g. the app works perfectly well with Doze mode
- I am not sure if this is still a problem after adding the "Do not optimize" option to deal with Android 12
  - I have had some minor issues with the OnePlus, but only after some kind of weird restart
  - @nataliejschultz I think you have a Samsung test phone - have you noticed any issues with the app running on it?

Abby-Wheelis:
TODO from feedback in the PR preceding this:

- [ ] add opcode in text form to the share for people who can't scan [1](https://github.com/e-mission/e-mission-phone/pull/994#discussion_r1299196951)
- [ ] remove geofence stuff [2](https://github.com/e-mission/e-mission-phone/pull/994#discussion_r1299208385)
- [ ] maybe reinstate opcode copying (convenient esp for dev, but not a 100% needed)

Abby-Wheelis:
https://github.com/e-mission/e-mission-phone/assets/54848919/0c9a8cb1-e49b-4cd2-9600-8afb0214bc23

See the video for an example of the opCode being saved as a text when "files" is the selected share mode. "save to photos" also works, but only saves the image. The "url" option shows the opcode text in the share sheet and the "message" option is what I found to work best saving the text. 

The opCode is now displayed in the `PopOpCode` modal as well, which can be seen in the first few seconds of the screen recording. 

While making these changes, I discovered that the image was not actually being scraped and set up for saving (I think I fixed it pre-migration and didn't bring that fix through to migration? Not sure, but it's good now!).

Abby-Wheelis:
>I think we need more data to decide what to do about the Samsung checks going forward.

Should the line about Samsung or Huawei phones be in the permissions as well as the join page? I could add it as a check or just a line on the check page?

I do run a Samsung test phone but it is only on Android 7, and [https://dontkillmyapp.com/samsung](https://dontkillmyapp.com/samsung) cites Android 9 and newer as the problem versions. If there's no way to pop-up the fix, I could try adding it to the screen as just a piece of text, if that might work? Or if we find out that most people haven't had problems I can take it out entirely, at least for now.

Abby-Wheelis:
While I'm in settings, should I update this share to reflect NREL OPENPath and an nrel-hosted download site, or is this information still acurate?

Abby-Wheelis:
If we want to keep the copy function for the opcode, I've gone ahead and made an attempt at the two-row idea @shankari had in the other PR, I think this might be a good way to do it, since it presents the opCode as "your profile", since it is your main identifier (open to icon feedback, this just looked like a good representation of "credentials"), then the logout function is also presented clearly. I don't think it would be too confusing for Android users to not have an action happen on the profile row. What do others think?

JGreenlee:
Here is what I would recommend instead:


And the OPCode popup can contain a "Copy OPCode" button underneath the QR (maybe next to the share button)

Abby-Wheelis:
It makes so much sense to have the copy OPCode with the rest of the OPCode functions, to keep everything in one place. Thank you for the suggestion! I've added a button here: 



I'll work on profile and Log Out as a header, though I think that sliver of the UI is still controlled by angular, so that might have to wait - but I'm seeing you customized that header in the Label screen so I'll go look at that to see how to hopefully populate that header.

JGreenlee:
> * Privacy Policy is the last ionic popup, so I think I might go ahead and migrate it. In the process I plan to refactor away from the html files to a keyed system, which will also facilitate the transition if we start using something like [locize](https://locize.com/) -- any thoughts on dropping these into a section in the existing i18n files as a dedicated category vs having them be their own file?

That sounds like a good plan to me. We just have to keep track of which text is language-specific (managed in e-mission-translate) and which text is both program- and language-specific (managed in the config file)

Abby-Wheelis:
I have the Privacy Policy popup mostly converted now, when I push you'll be able to see the difference between the `e-mission-translate` and `appConfig` text pretty easily in the code. I'm just needing to find a viable way to incorporate the links  in React (they are currently html `<a>` tags and the formatting and function are lost. 

https://github.com/e-mission/e-mission-phone/assets/54848919/a5c9acf5-3477-4d70-aca6-43ce5854f094

shankari:
> share button title and link look like they might be pre-NREL, are we ready to change those? https://github.com/e-mission/e-mission-phone/pull/1009#issuecomment-1686783716

Yes

> Privacy Policy is the last ionic popup, so I think I might go ahead and migrate it. In the process I plan to refactor away from the html files to a keyed system, which will also facilitate the transition if we start using something like [locize](https://locize.com/) -- any thoughts on dropping these into a section in the existing i18n files as a dedicated category vs having them be their own file?

Yes, the files are a vestige of an older time. I think that I have an issue or a PR somewhere that points out that, after the dynamic config rewrite the summary.html basically has no text. So yes, we can definitely move everything to a keyed system.

Abby-Wheelis:
Here is the documentation for the way these hyperlinks are handled, but currently they do not work: [react-native Linking](https://reactnative.dev/docs/linking#openurl), adapting to this formatting based on [this example](https://snack.expo.dev/@aboutreact/react-native-click-on-a-text-link-to-open-it-in-browser?platform=web) that @JGreenlee sent me allowed me to improve the formatting (the spacing was bad before). In checking on my phone/test phones I've discovered that the links don't work anywhere (production, staging, or this new way). 

@shankari  do you know anything about when the links last worked or why they don't work now? 

If we can't get the links to actually redirect, I could expose the urls to the user (some are shown, some are clickable text describing where the link goes) as a temporary solution. 

This and the samsung permissions are the last things I know of for this PR, so I think it's time I start working on dashboard. 
(there will also be an e-mission-translate PR to come out of these changes, marking that down so I don't forget!)

shankari:
@nataliejschultz any updates on the samsung testing?

Abby-Wheelis:
Some accessibility observations, since I just got to look at this with the devapp on my LG phone using the screen reader. (NREL network seems to interfere with connecting a physical device to the devapp, and you have to have both devices on the same network, so easier at home!) Overall, the React Profile is navigable with the screen reader, to the extent that each ListItem (setting row) is showing up as something clickable and you can scroll between them, but that's a low bar. The modals take focus, so that's encouraging. There definitely some changes that would help:
- getting the names of the icons associated to be read (especially with logout, all it sees is "button")
- figuring out why it reads the name of each row twice
- getting the dev zone dropdown coded as an expansion and not just a "button"
- the permissions are better, but they scroll as the whole item then each subcategory: wondering if there's a way to present it as one entity, click to fix?
- it's reading out that lots of things are "disabled" - thinking this might mean that the screen reader just can't interpret them even enough to tell what it is?

I'm sure there are other accessibility improvements we could make, but those are the things that stood out to me in sort of a "first pass". I think I might prefer to dedicate a PR to these improvements (and others we find) in order to keep documenting those changes clear and since this PR is already getting extensive.

shankari:
>  when I push you'll be able to see the difference between the e-mission-translate and appConfig text pretty easily in the code.

Now that we have webpack with the DEV flag, we could also choose to draw an outline around the text from the dynamic config or show it a different color during development.

> While I'm in settings, should I update this share to reflect NREL OPENPath and an nrel-hosted download site, or is this information still acurate?

We could also just remove it. That share button was added when we were doing the recruitment. It is now clear that we will never do recruitment by ourselves. If we do share it, what would we include? Not necessarily the program that people are in, because their friend may not be part of the same program. Not necessarily. Open Access? Maybe, but then you won't really be part of your friend's cohort. So it might be easiest to just remove it altogether and simplify.

> I went ahead and implemented it as a singular list, with just one refresh button. I think this simplified the screen a lot, as well as the code. There's still a few things I need to do, especially with formatting, but it's starting to work well.

As an aside, from an accessibility perspective, the UX with "opening" the titles to see the list of details is apparently bad. When we had the meeting with UW, they were not able to get past the permission screen, so we couldn't even test out all the accessibility improvements that @JGreenlee had made.

> edit - just checked on production in Profile > View Privacy Policy and none of those links take me anywhere either... so the formatting is the only regression here, but it would be good to get the links working again, particularly since they're tied to our transparency on how things work and where the data goes.

So the current production already has the react native migration started. But as I pointed out, we have archived `.app.zip` files from previous production releases as well. If you pick one from early this year (March/April) and try, we can figure out if the links really did not work before this migration. 

Even if they did work, though, we can't go back to angular1 just for this.
I do agree that it would be good to have the links work, for transparency.
Can we display the privacy policy in an embedded webview, the way we do with the leaflet code?
Or maybe even just open a new webview/browser window?

I added that popup webview to meet the google play developer guidelines.

Looks like FabMob has had luck just replacing it with a link to their web-based privacy policy
https://github.com/e-mission/e-mission-docs/issues/611#issuecomment-843920662

But the problem is that the web-based privacy policy is generic and will not have the dynamic config elements for this study. But maybe we could show the privacy policy without links but have a "more info" at the button, which opens up the web page for additional linked info

Abby-Wheelis:
>So it might be easiest to just remove it altogether and simplify.

Done! The settings keep getting simpler! They're now <1 page on the emulator before expanding dev zone.

>But maybe we could show the privacy policy without links but have a "more info" at the button, which opens up the web page for additional linked info

This seems like it could work for us, the way the privacy policy and the text from the config are woven together works well as it stands, I'm just struggling with the links, I can look into your suggestions about a webview or a browser window.

JGreenlee:
> So the current production already has the react native migration started. But as I pointed out, we have archived `.app.zip` files from previous production releases as well. If you pick one from early this year (March/April) and try, we can figure out if the links really did not work before this migration.

To do this, I think we would also need old QR codes from when it was a 2-step process

shankari:
> I used an old OPCode (I had to remember how it used to work) and tried an old release from March 8 (version 1.1.4).

And I assume the new flow is better?

shankari:
> This seems like it could work for us, the way the privacy policy and the text from the config are woven together works well as it stands, I'm just struggling with the links, I can look into your suggestions about a webview or a browser window.

Does it work in the intro onboarding? I wonder if this is a popup issue/popover issue (e.g. some level of "sanitize"), and we can just open the custom privacy policy in a webview instead.

Abby-Wheelis:
They don't work in onboarding on staging or production, just checked on my personal and test phones. Not sure about the older version of the app. 

If we don't see a clear way to patch this now, would it be an ok patch to include the urls as text (half of them already are) for now, so that people still know how to access the information, and work out a more elegant way to handle this later? I'm not sure about what adding a webview / opening a browser would entail, so I could be missing that as an obvious solution, I'll look into it more when I get the chance.

JGreenlee:
> And I assume the new flow is better?

Way better! I had forgotten how confusing the old flow was with the 2 separate codes.
(I still think we can do even better though. Have you had a chance to look at https://github.com/e-mission/e-mission-docs/issues/956?)

Abby-Wheelis:
>Yup, if it doesn't work in onboarding either, I was going to suggest that we just make them not look like URLs. We can change the text to "something like the e-mission-server repo, linked from the NREL OpenPATH page" because if something starts with https://... people are going to want to click it. Let's run that by legal really quickly as a stopgap measure.

I went through this morning and replaced the links with guidance to the NREL website. The text for repos reads "... you can inspect the data that OpenPATH collects in the background, the analysis pipeline, and the dashboard metrics through links on the NREL OpenPATH website." And for the TSDC reads "...Further information on the procedures is available through links on the NREL OpenPATH website." This seemed like the simplest way to bridge the gap, since we hope to have functional links again at  some point. Once all the text is final, I can make sure [translate #40](https://github.com/e-mission/e-mission-translate/pull/40) matches.

I don't know of anything else pressing that this PR needs to address, so I'm ready for review and tracking future work here: [issue#946](https://github.com/e-mission/e-mission-docs/issues/946)

JGreenlee:
I went to test this in an emulator and noticed 2 big issues:

1) Unable to scroll down in the Profile tab (open developer zone and swipe, nothing happens)
2) After a reload, the app seemed broken and there was an error in the console about `window['device'].platform` being undefined. I haven't been able to consistently reproduce this.

Abby-Wheelis:
I've also noticed those things, for (1) it usually takes 1-2 seconds to be able to scroll, if you know what's causing that, I'd love to fix it, since it's a really annoying behavior. 

I thought I had fixed (2), as I also occasionally saw that, and my best guess was that it depended on timing of the call in setting up the list of permissions in AppStatusModal vs `window['device']` being ready. I made the change [78d971f](https://github.com/e-mission/e-mission-phone/pull/1009/commits/78d971fa85033a5adb44b97e124328093177b97f) in an attempt to address that, but clearly I didn't actually fix it.

JGreenlee:
Ah, I see that you did this in that commit:
```js
if (appConfig && window['device'].platform) {
    setPlatform(window['device'].platform.toLowerCase());
    ...
```
If the error is still throwing, it would suggest that `window['device']` is not even ready. If `window['device']` was ready but `window['device'].platform` was not ready, the condition would evaluate false and it would just skip.

If you want to skip when either of `window['device']` or `window['device'].platform` are not ready, you can use optional chaining

```js
window['device']?.platform
```

JGreenlee:
> for (1) it usually takes 1-2 seconds to be able to scroll, if you know what's causing that, I'd love to fix it, since it's a really annoying behavior.

I haven't been able to get it to scroll at all. We had this issue on the dashboard.
I think if you implement something to the effect of https://github.com/JGreenlee/e-mission-phone/commit/b6dfb4f9cc543720774eab1400c798ecf424474e, the issue will go away.

JGreenlee:
You don't need `main-control.html` anymore as an intermediate between main.js and ProfileSettings. You can put `<profile-settings>` directly as 'template' in `main.js`

JGreenlee:
> You don't need `main-control.html` anymore as an intermediate between main.js and ProfileSettings. You can put `<profile-settings>` directly as 'template' in `main.js`

Actually, I take this back. If you are still using anything from `settingsScope`, you still need `main-contro.html`

Abby-Wheelis:
>If you want to skip when either of window['device'] or window['device'].platform are not ready, you can use optional chaining

Made this change and refreshed a few times, monitored both window['device'] and window['device']?.platform with a console.log statement, confirmed that it does not break even when one or both are null/undefined.

>Or do you? I'm not sure. You can try and see what happens.

I do, for `showLog` and `showSensed ` because of how I get the `settingsScope`, can eliminate  when we convert those pages and use React's routing. 

>I haven't been able to get it to scroll at all because ProfileSettings is growing taller than the Angular parent, so it doesn't know it needs to scroll. We had this issue on the dashboard.

This was new, from when I took out the `ion-content` tags in `main-control.html` because I thought I did not need them any more, which broke the scrolling. I've added the tags back, and tried adding `class="fill-container"` but that tries to cram it into the heading. Unfortunately, the way I've gotten it to work pushes my React profile header down a bit, so it it looks a tad awkward. The scrolling currently works after a small delay.

JGreenlee:
fixed that in 95202a9b551aca88e60c07482706cd3d36ce6c3f

I found that we were able to take out `main_control.html` after all; `settingsScope` still works it's just attached to the parent when it used to be attached to the child. So this is actually a lot simpler now.

No more scrolling issues and no weird gap.

JGreenlee:
- OPCode text grows wider than the dialog and gets cut off (also, it should be in a monospace font but it's currently in Roboto)
    - 
    - Why not show "You can save..." above the QR code, and show the OPcode text below it (right above the copy button)?
- Log out button does not have text and its function may be unclear
    - Although we can't add text to the `<Appbar.Action>`, we can put any component we want inside the `AppBar`
- Most dialogs are full-width
    - Not a deal-breaker but a bit of horizontal margin would probably look nicer
- "Medium accuracy" is not working

Abby-Wheelis:
I think I've now addressed the first three comments, and thank you for fixing the scroll issue!

> "Medium accuracy" is not working

I am able to toggle the switch, and checking with breakpoints I can see that `lowAccuracy` is true when the switch is off and false when the switch is on. Is this not the expected behavior?

Also, if this switch is to go from low-> medium accuracy, should we rename it? I don't find "turning on medium" very intuitive to know if that's more or less accurate. What about "increased accuracy"?

Abby-Wheelis:
I've discovered an issue: when you log out and log back in as a different opcode, the Profile still displays the previous opcode because it is fetched when the dynamic config is ready (early on in the login flow) and the opcode is not flushed when you log out. The opcode is gotten from a plugin [`OPCodeAuth`.](https://github.com/e-mission/cordova-jwt-auth#v1.7.2) Refreshing would I think fix the issue.

Potential solution: force a refresh when the user opens the Profile page? React seems to have [`useFocusEffect`](https://reactnavigation.org/docs/use-focus-effect/) for this purpose, but it requires a navigation object, which I'm unsure how to get since we're not fully using React's navigation yet

JGreenlee:
> I am able to toggle the switch, and checking with breakpoints I can see that `lowAccuracy` is true when the switch is off and false when the switch is on. Is this not the expected behavior?

It was not toggling when I tried before.  I just tried again and was able to toggle it off, but I can't toggle it back on.

It flashes for a split second and seems like it might be toggling on and then immediately back off. Might the click event be duplicated?

JGreenlee:
> I've discovered an issue: when you log out and log back in as a different opcode, the Profile still displays the previous opcode because it is fetched when the dynamic config is ready (early on in the login flow) and the opcode is not flushed when you log out. The opcode is gotten from a plugin [`OPCodeAuth`.](https://github.com/e-mission/cordova-jwt-auth#v1.7.2) Refreshing would I think fix the issue.
> 
> Potential solution: force a refresh when the user opens the Profile page? React seems to have [`useFocusEffect`](https://reactnavigation.org/docs/use-focus-effect/) for this purpose, but it requires a navigation object, which I'm unsure how to get since we're not fully using React's navigation yet

We already trigger a refresh on logout, once the config is wiped. (see `resetConfigAndRefresh()`)
I'm not sure why the old config is still hanging around

Abby-Wheelis:
>It flashes for a split second and seems like it might be toggling on and then immediately back off. Might the click event be duplicated?

The `onPress` *is* getting called twice sometimes, I can't figure out exactly why, but I have been able to reproduce. It seems like this is an issue that sometimes comes up in React, and it may be something that react's "StrictMode" is causing to happen as a debugging feature [(React issue #12856)](https://github.com/facebook/react/issues/12856#issuecomment-390206425) though the whole `onPress` (toggle low accuracy in `ControlCollectionHelper` and not just `setState` (part of the refresh at the end) is called twice, so maybe I need to debounce? [(preventing double clicks)](https://medium.com/@daveford/prevent-double-click-dups-in-react-83fcbc475704)

Abby-Wheelis:
I was especially puzzled by the fact that the tracking toggle worked, but the accuracy toggle did not. The only difference was that the tracking toggle broadcasted an "update complete" event, and accuracy did not. I updated my local copy of `collect-settings.js` to broadcast the event once the accuracy updates had taken place, and this resolves the UI issue. Sometimes the function is still called twice, but the UI is only updated once, preventing flickering. I think ultimately it was an asynchronous vs synchronous context issue. Sometimes it's slow, but the toggle consistently behaves as expected.

Abby-Wheelis:
In testing, I was able to see this fix prevent the function from being called a second time, and therefore prevent the switch from flickering. I pressed the toggle once, it tried to call the function twice (as seen by double log statement), but the time limit prevented the function from running the second time. 
Screenshot 2023-09-05 at 11 08 04 AM

JGreenlee:
I noticed that when I commented out the call to `ControlCollectionHelper.toggleLowAccuracy();` it didn't call `toggleLowAccuracy` a second time. So something in there is blocking the flow of events.

The debouncing fix is kind of a workaround, but I think it's the only thing we can do for now without actually modifying ControlCollectionHelper (which is at https://github.com/e-mission/e-mission-data-collection/blob/c21f8354f133498dedec37aeaf98ec3afddaefec/www/ui/ionic/js/collect-settings.js#L170)

Abby-Wheelis:
>>when you log out and log back in as a different opcode, the Profile still displays the previous opcode

>We already trigger a refresh on logout, once the config is wiped. (see resetConfigAndRefresh())
I'm not sure why the old config is still hanging around

This is the last issue that I know of that this PR needs to address - it's consistently reproducible for me, but does fix when I hit refresh. I'm thinking I either need to have it automatically refresh at some point, or find a way to actually clear out the old config so it doesn't hang around.` resetConfigAndRefresh()` is the method that is called when the user logs out.

JGreenlee:
To me, it doesn't seem like a blocking issue - it's certainly the unexpected behavior, but it doesn't fall in the domain of how we expect end users to interact with the app (users are given 1 OPCode and should not need to switch out).

The only people this should really affect are us as developers (who frequently test out many different OPCodes).

www/js/control/AppStatusModal.tsx
```diff
@@ -0,0 +1,450 @@
...
+        console.log("recomputing checks", checkList);
+        checkList.forEach((lc) => {
+            lc.statusIcon = iconMap(lc.statusState);
+            lc.statusColor = colorMap(lc.statusState)
+        });
+        recomputeOverallStatus();
+    }
+
+    //custom hook executes function on app resuming foreground
+    useAppStateChange( function() {
```
Abby-Wheelis, line 367: This is the code responsible for launching the model if a permission is wrong. I just noticed that it also launches when the permissions are correct, so I am working to debug and find a fix.

Abby-Wheelis: I think, by refactoring a bit, I've resolved this issue. Tested performance by closing and reopening app with good permissions, logging out and logging back in (had issues showing modal before consent at one point), and leaving and returning to app (background -> foreground) without the modal popping up. Also left, turned off location permission, and reopened, which did launch the modal. 

Closing the app, breaking a permission, and re-opening the app opened the modal only when I navigated to the profile page. If someone is using the app as directed, this wouldn't really ever happen, so I'm not sure it's a big concern.

shankari:
> Closing the app, breaking a permission, and re-opening the app opened the modal only when I navigated to the profile page. If someone is using the app as directed, this wouldn't really ever happen, so I'm not sure it's a big concern.

@Abby-Wheelis This is not correct. Various android versions can reset permissions automatically, without user intervention (see https://github.com/e-mission/e-mission-docs/issues/967)  So the user can end up with a broken app for no fault of theirs. We do prompt them to indicate that there is a problem, but then when they click on the prompt, they want to see a way to fix it. I don't know that they will remember that they have to navigate to the profile screen to fix it.

Abby-Wheelis:
@shankari By "closing the app" I don't just mean removing it from the foreground, I mean killing the app entirely (which sends the notification to keep the app running in the background), I'm sorry if that was confusing. If the app is in the background and a permission breaks, it launches the modal when you bring the app to the foreground!

I'm hoping to make a fix such that if a permission breaks while the app is killed, it at least launches the modal the next time the app is brought from background -> foreground (so 2nd open), but still working on it.

Abby-Wheelis:
I recorded testing most of the features on the new profile page, to help make the changes a little easier to see: [view on google drive](https://drive.google.com/file/d/1DclGFvptNz1XLQG6PJ4-R8yljgJKCQfi/view?usp=sharing) since it was too large of a file to embed directly into this PR

JGreenlee:
> If someone is using the app as directed, this wouldn't really ever happen, so I'm not sure it's a big concern.

@Abby-Wheelis I understand what you mean, but I do think it's a big concern.
People might miss the "fix your permissions" notification (or ignore it) and just open the app from their home screen / app drawer. In that situation, we shouldn't allow them to do anything on the Label screen until they fix their permissions. And I wouldn't expect people to go to the Profile tab regularly.

I'm wishing we had that `App` component by now - it would make this so much cleaner. But not possible until this PR and dashboard PR are done+merged.

Abby-Wheelis:
>we shouldn't allow them to do anything on the Label screen until they fix their permissions

>This is because the Profile tab is not initialized on app launch. [comment](https://github.com/e-mission/e-mission-docs/issues/967#issuecomment-1711053474)

I agree that we need to make sure that permissions are correct at any opportunity, but it sounds like I might not be able to control what happens after a fresh launch if the Profile (and therefore AppStatusModal) have not been initialized yet. Though, I did see it popping up during onboarding at least a few times, so maybe I can revert the change I made to prevent that and go from there. 

If there is no way to address the close app entirely -> break permission -> fresh launch flow by adjusting what I have now, I can see two ways forward:

1. push this code to a branch somewhere to save it for later and revert to the old appStatus page for now, since it will be cleaner to implement once we have the `App` component
2. add an instance of `AppStatusModal` to the Label screen, as a temporary fix to ensure permissions are monitored and fixed, since that's the "main" (most viewed) component for now

Abby-Wheelis:
Almost there...
- works for a fresh launch on broken permission
- works for bringing into foreground on a broken permission
- shows modal when label screen is refreshed on a broken permission

However, if the user has visited both profile and label, there are two copies of the modal, so if they break a permission with the app in the background, there's two modals and you have to fix and accept on the first one, then refresh and accept the second one. This is annoying, certainly, but the only fix I've thought of so far is to have the modal close when the permissions are all green, but this  bypasses having the user click "I accept" which is bad.

shankari:
> so if they break a permission with the app in the background, there's two modals and you have to fix and accept on the first one, then refresh and accept the second one. 

If you are still using angular for the routing, then this function (from `www/js/controllers.js`) should be called on every state change. So can you close the modal every time we switch between screens?

```
 23   $rootScope.$on('$stateChangeSuccess', function(event, toState, toParams, fromStat    e, fromParams){
 24     console.log("Finished changing state from "+JSON.stringify(fromState)
 25         + " to "+JSON.stringify(toState));
 26     ClientStats.addReading(ClientStats.getStatKeys().STATE_CHANGED,
 27       fromState.name + '-2-' + toState.name).then(function() {}, function() {});
 28   });
```

Abby-Wheelis:
>So can you close the modal every time we switch between screens?

The duplicated modals are appearing one stacked on top of the other on the same screen, so there's no changing screens between having to deal with the two modals. I also don't see the console message after closing the first one, the modal's visibility is controlled by a boolean held in state. 

This is what that behavior looks like:

https://github.com/e-mission/e-mission-phone/assets/54848919/3773f10f-eed6-421b-8222-b03bef37a550

shankari:
At what point in that video do you see two modals? I am confused. That seems fine-ish to me? Unless there are some clicks that are not obvious...

Abby-Wheelis:
> That seems fine-ish to me? Unless there are some clicks that are not obvious...

At worst, it's a little annoying. There's two copies of the modal (one coming from ProfileSettings and one from LabelTab) and they both get tripped to open when there's a broken permission. They're one on top of the other, so I have to close them both, you can see in the video when the background color shifts is when the first modal closes. 

around 12s I close the first one, then at 14 I refresh & close the second modal -- so 2 "extra" clicks

shankari:
I am a bit tempted to merge with this given that it is currently broken very badly now and two clicks are better than no popup. And this is not a very frequent use case anyway

Abby-Wheelis:
>I am a bit tempted to merge with this given that it is currently broken very badly now and two clicks are better than no popup. And this is not a very frequent use case anyway

We'll be able to fix the duplication as soon as we can implement React routing, but I'm not sure if there's a good way to share the state across components that are essentially independent of each other while we're in this middle zone. I've tried a custom hook, but that has separate instances for each place it is used I believe, and when I've tried to research most of what I've found would work if we were fully React, or we could potentially use someone's library ([see first answer](https://stackoverflow.com/questions/53451584/is-it-possible-to-share-states-between-components-using-the-usestate-hook-in-r)), but I haven't come up with anything that works in this context.

The other option would be to merge the new modal on JUST the label screen, which I actually already tried earlier this morning [on this branch](https://github.com/Abby-Wheelis/e-mission-phone/tree/permission-launch-fix). Let me know if that's preferred, and I'll test it more thoroughly and prepare a PR. Then we could maybe merge this PR right after the patches are on staging which I believe would maintain the timeline for React routing we were talking about earlier.

shankari:
> The other option would be to merge the new modal on JUST the label screen, which I actually already tried earlier this morning [on this branch](https://github.com/Abby-Wheelis/e-mission-phone/tree/permission-launch-fix). Let me know if that's preferred, and I'll test it more thoroughly and prepare a PR. Then we could maybe merge this PR right after the patches are on staging which I believe would maintain the timeline for React routing we were talking about earlier.

So with this option, they would see the status modal popup on the label screen but would have to click on the button to launch it from the profile screen?

That actually sounds like a good idea given that the label screen is (currently) the hardcoded default for when the app is launched. It is also consistent with the pre-react migration. I don't think we popped up the status screen if people changed to profile.

JGreenlee:
> At worst, it's a little annoying. There's two copies of the modal (one coming from ProfileSettings and one from LabelTab) and they both get tripped to open when there's a broken permission. They're one on top of the other, so I have to close them both, you can see in the video when the background color shifts is when the first modal closes.

On resume, only open the app status modal if `!window.appStatusModalOpened`.
Then when it is opened, set `window.appStatusModalOpened = true;`

I think that would be good enough. Variables on the `window` are global and preserved within a session. It's a temporary workaround but we can do it all the right way next release

Abby-Wheelis:
>So with this option, they would see the status modal popup on the label screen but would have to click on the button to launch it from the profile screen?

@shankari yes. When something is broken, the modal will pop up regardless of screen, as long as they've visited the label screen before. And it works on a fresh launch because the label screen is the default. In Profile, clicking the button takes them to the old permissions page, but still works fine. 

>On resume, only open the app status modal if !window.appStatusModalOpened.
Then when it is opened, set window.appStatusModalOpened = true;

Ok, global is what I was looking for, let me try that out.

Abby-Wheelis:
That worked! Thank you @JGreenlee

Now, we have exactly 1 modal open 
- on launch (at label screen) if something is broken
- on resume if something is broken if never visited profile screen before (2nd not even intialized)
- on resume if something is broken if visited profile before (2nd is initialized, I watched that global prevent the 2nd from opening)

This resolved the modal-launching-if-broken oddness we've been chasing around, and so this should be good now and a fix to [issue#967](https://github.com/e-mission/e-mission-docs/issues/967). The translate PR is also ready, and includes keys from the label screen [translate#40](https://github.com/e-mission/e-mission-translate/pull/40)

www/js/control/DataDatePicker.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import React from "react";
+import { DatePickerModal } from 'react-native-paper-dates';
+import { useTranslation } from "react-i18next";
+import { getAngularService } from "../angular-react-helper";
+
+const DataDatePicker = ({date, setDate, open, setOpen}) => {
+  const { t } = useTranslation();
+  const ControlHelper = getAngularService("ControlHelper");
+  const { i18n } = useTranslation('general-settings.choose-date'); //able to pull lang from this
```
JGreenlee, line 12: `useTranslation()` does not need this argument

www/js/control/PopOpCode.jsx
```diff
@@ -1,30 +1,48 @@
...
     const { t } = useTranslation(); 
     const { colors } = useTheme();
+
+    const opcodeList = tokenURL.split("=");
+    const opcode = opcodeList[opcodeList.length - 1];
+    
+    const [copyAlertVis, setCopyAlertVis] = useState(false);
+
+    const copyText = function(textToCopy){
+        navigator.clipboard.writeText(textToCopy).then(() => {
```
JGreenlee, line 18: Is this still iOS-only ? Last I checked we couldn't copy text on Android without extra permissions.
What happens when Android users click the copy button? Nothing?

Abby-Wheelis: You're right, I forgot about that. I just changed it so that nothing happens on Android, though now that I think of it, it would be better for the text to be clickable, and to maybe not have a copy button since a button that does nothing is confusing.

www/js/control/DataDatePicker.tsx
Abby-Wheelis: I'm not sure why I thought I needed that, I just took it out

www/js/control/PopOpCode.jsx
JGreenlee: Or just only show the button on iOS...

Abby-Wheelis: that was a better idea, made that change

shankari:
ok, so I'm going to start reviewing this now...

www/js/control/ProfileSettings.jsx
```diff
@@ -120,6 +179,38 @@ const ProfileSettings = () => {
         console.log("notification settings before and after", notificationSettings, newNotificationSettings);
         setNotificationSettings(newNotificationSettings);
     }
+
+    async function getSyncSettings() {
+        console.log("getting sync settings");
+        var newSyncSettings = {};
+        ControlSyncHelper.getSyncSettings().then(function(showConfig) {
+            newSyncSettings.show_config = showConfig;
+            setSyncSettings(newSyncSettings);
```
shankari, line 188: Just to clarify: this sets the in-memory value of the sync settings, but does not update the native code value in the plugin, correct? Because it seems bad to update the native code value on every `get`. Similar for the other `getXXX` calls based on native code below.

www/js/control/ProfileSettings.jsx
```diff
@@ -150,54 +255,206 @@ const ProfileSettings = () => {
           So we don't need to call refreshCollectSettings here. */
     }
 
-    const toggleLowAccuracy = function() {
+
+    const safeToggle = function() {
+        if(toggleTime){
+            const prevTime = toggleTime.getTime();
```
shankari, line 261: if I followed the thread correctly, this is a workaround until the plugin helpers are also migrated to react
https://github.com/e-mission/e-mission-phone/pull/1009#issuecomment-1707002882
https://github.com/e-mission/e-mission-phone/pull/1009#issuecomment-1707011173

www/js/control/ProfileSettings.jsx
```diff
@@ -150,54 +255,206 @@ const ProfileSettings = () => {
...
     }
+    
+    const clearNotifications = function() {
+        window.cordova.plugins.notification.local.clearAll();
+    }
 
-    var prepopulateMessage = {
-        message: t('general-settings.share-message'),
-        subject: t('general-settings.share-subject'),
-        url: t('general-settings.share-url')
```
shankari, line 179: wrt https://github.com/e-mission/e-mission-phone/pull/1009#issuecomment-1686783716

> While I'm in settings, should I update this share to reflect NREL OPENPath and an nrel-hosted download site, or is this information still acurate?

we decided to just remove that functionality, correct?

www/js/control/ProfileSettings.jsx
```diff
@@ -332,14 +594,110 @@ const ProfileSettings = () => {
...
+                            // $state.go("root.reconsent"); //don't know how to do this yet
+                            }}>
+                                {t('general-settings.confirm')}
+                        </Button>
+                    </Dialog.Actions>
+                </Dialog>
+            </Modal>
+
+            {/* handle consent */}
+            <Modal visible={consentVis} onDismiss={()=>setConsentVis(false)} transparent={true}>
```
shankari, line 671: Future: I think we can remove this in the future. The consent "number" is currently an IRB protocol number from UC Berkeley's Committee for the Protection of Human Subjects (CPHS). Since I am no longer at Berkeley, this no longer applies to us. And I had some vision initially that the CPHS protocols could change, and we would require people to consent to the new protocol, but that has never yet happened in practice.

We should think about changing terms and conditions midway through a data collection period, but I don't think this is the solution.

We should remove this and redesign it better.

www/js/control/ProfileSettings.jsx
```diff
@@ -215,42 +472,46 @@ const ProfileSettings = () => {
...
+                <SettingRow textKey="control.nuke-all" iconName="delete-forever" action={() => setNukeVis(true)}></SettingRow>
+                <SettingRow textKey={parseState(collectSettings.state)} iconName="pencil" action={() => setForceStateVis(true)}></SettingRow>
+                <SettingRow textKey="control.check-log" iconName="arrow-expand-right" action={showLog}></SettingRow>
+                <SettingRow textKey="control.check-sensed-data" iconName="arrow-expand-right" action={showSensed}></SettingRow>
+                <SettingRow textKey="control.collection" iconName="pencil" action={editCollectionConfig}></SettingRow>
+                <ControlDataTable controlData={collectSettings.config}></ControlDataTable>
+                <SettingRow textKey="control.sync" iconName="pencil" action={editSyncConfig}></SettingRow>
+                <ControlDataTable controlData={syncSettings.show_config}></ControlDataTable>
+                <SettingRow textKey="control.app-version" iconName="application" action={()=>console.log("")} desc={appVersion}></SettingRow>
```
shankari, line 510: nit + future: should we move this out of the developer zone? I am not sure why it needs to be hidden away.

It seems like fairly important information to know which version of the app you have, and it is surprisingly hard to figure our from other locations such as the play store. I think that the original thought by @sebastianbarry was that most control/settings screens hide away the app version under something like "General" (we should verify this by looking up the related commit)

However, after having deployed it for a while, I will point out that "Developer Zone" != "General". I would be OK with creating some other "versions" expansion section if needed, but I don't think it should be in the Developer Zone.

www/js/control/ProfileSettings.jsx
shankari, line 1: I looked through this briefly and don't see anything obvious missing, but @Abby-Wheelis, please flag the high-level changes, if any, so I can comment on whether they are useful to retain or not.

www/js/control/AppStatusModal.tsx
```diff
@@ -0,0 +1,458 @@
...
+    const [osver, setOsver] = useState(0);
+    const [platform, setPlatform] = useState<string>("");
+
+    const [error, setError] = useState<string>("");
+    const [errorVis, setErrorVis] = useState<boolean>(false);
+
+    const [explainVis, setExplainVis] = useState<boolean>(false);
+
+    const [backgroundRestricted, setBackgroundRestricted] = useState<boolean>(false);
+    const [allowBackgroundInstructions, setAllowBackgroundInstructions] = useState<Array<any>>([]);
```
shankari, line 29: why are these two called out specially instead of being included with the other errors?
I don't see specific errors for location permissions, for example.

www/js/control/AppStatusModal.tsx
```diff
@@ -0,0 +1,458 @@
...
+
+        setExplanationList(tempExplanations);
+  
+        //waiting on samsung feedback, need more information
+        setBackgroundRestricted(false);
+        if(window['device'].manufacturer.toLowerCase() == "samsung") {
+          setBackgroundRestricted(true);
+          setAllowBackgroundInstructions(t("intro.allow_background.samsung"));
+        }
+
```
shankari, line 322: so these are not actually used right now pending "dealing with Samsung", correct?

www/js/control/AppStatusModal.tsx
shankari: Ah I see that these are for handling the samsung issue, but are currently unused
https://github.com/e-mission/e-mission-phone/pull/1009/files#diff-2c9b8aaf6321e412caf37d08c26037408558d62ed592f16fb0847b6cccd8639fR317

shankari (changes requested):
This looks good overall and I have no qualms with merging and pushing to staging.
I would like to have non-future comments in here addressed before we move to production at the end of next week.

shankari (approved):
Going to merge this now because there is only one non-future change (removing the non-functional action button to change the notification time)

shankari:
One more non-future change from:
https://github.com/e-mission/e-mission-docs/issues/934#issuecomment-1681375028

We need to make sure that the instruction text is correct for Android 13. e.g. for "unused apps disabled" are the instructions on how to resolve correct on android 13, or do we need to add a new `android-disable-gte-13`?


```
            "unusedapprestrict": {
                "name": "Unused apps disabled",
                "description": {
                    "android-disable-lt-12": "On the app settings page, go to 'Permissions' and ensure that the app permissions will not be automatically reset.",
                    "android-disable-gte-12": "On the app settings page, turn off 'Remove permissions and free up space.'",
                    "ios": "Please allow."
                }
            },
```

www/js/control/ProfileSettings.jsx
Abby-Wheelis: correct, the `setSyncSettings` call is to set the React state that holds the sync settings -- which is what the UI depends on

www/js/control/ProfileSettings.jsx
Abby-Wheelis: Correct that it is a workaround to prevent the function from being called twice in a row and causing the "bouncing" behavior that was our concern with the switch. When we migrate plugin helpers we should revisit this to see if it can be resolved.

www/js/control/ProfileSettings.jsx
Abby-Wheelis: correct, we removed "sharing the app" from the profile mostly because we're not doing our own recruiting anymore

www/js/control/AppStatusModal.tsx
Abby-Wheelis: Correct, this is pending update based on ["dealing with Samsung"](https://github.com/e-mission/e-mission-docs/issues/535)

www/js/control/general-settings.js
```diff
@@ -9,546 +9,36 @@ angular.module('emission.main.control',['emission.services',
...
-                $scope.ui_config.opcode.autogen = $scope.ui_config?.intro.program_or_study == 'study';
-            }
-            $scope.refreshScreen();
-        });
-    });
+    //used to have on "afterEnter" that checked for 2 things
+        //modal launch -> migrated into AppStatusModal w/ use of custom hook!
+        //stateParams.openTimeOfDayPicker -> functionality lost for now
+            //to change reminder time if accessing profile by specific android notification flow
+            //would open the date picker
```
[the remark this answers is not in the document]
Abby-Wheelis: removed in [0184847](https://github.com/e-mission/e-mission-phone/pull/1028/commits/01848471d8e5f8670a803f377f505552fad28cd3) to be restored after routing has been migrated

Abby-Wheelis:
future comments have been noted in: [Next Phase of migrating Profile to React](https://github.com/e-mission/e-mission-docs/issues/946#top)

www/js/control/AppStatusModal.tsx
Abby-Wheelis: done in [426c440](https://github.com/e-mission/e-mission-phone/pull/1028/commits/426c44069ef3446ba8ec33bad3dbdf5fec335adf)

Abby-Wheelis:
Cleanup of now / near future comments: [🧹 Profile cleanup](https://github.com/e-mission/e-mission-phone/pull/1028#top) including the android text ([3435e12](https://github.com/e-mission/e-mission-phone/pull/1028/commits/3435e122f5b2ec9553ce7f581711efefa8ac6002))

www/js/control/ProfileSettings.jsx
```diff
@@ -10,6 +10,10 @@ import DemographicsSettingRow from "./DemographicsSettingRow";
 import PopOpCode from "./PopOpCode";
 import ReminderTime from "./ReminderTime"
 import useAppConfig from "../useAppConfig";
+import AlertBar from "./AlertBar";
+import DataDatePicker from "./DataDatePicker";
+import AppStatusModal from "./AppStatusModal";
+import PrivacyPolicyModal from "./PrivacyPolicyModal";
```
Abby-Wheelis, line 16: Addition of the `AlertBar` -- which shows a "snackbar" at the bottom of the screen with a message in replacement of many ionic alerts -> cases include text copied, data pushed, etc.
`DataDatePicker` used for choosing the data that json dump download is used for, replaces the ionic datepicker with react native paper's which is cleaner and works better
`AppStatusModal` and `PrivacyPolicyModal` we have talked about in the PR, so I'm sure you're up to speed on those, but were major changes here.

www/js/control/ProfileSettings.jsx
```diff
@@ -21,24 +25,25 @@ const ProfileSettings = () => {
...
+    const settingsScope = angular.element(mainControlEl.querySelector('profile-settings')).scope();
+    console.log("settings scope", settingsScope);
+    
     // grab any variables or functions we need from it like this:
-    const { settings, logOut, viewPrivacyPolicy,
-        fixAppStatus, forceSync, openDatePicker,
-        eraseUserData, refreshScreen, endForceSync, checkConsent, 
-        dummyNotification, invalidateCache, showLog, showSensed,
-        parseState, userDataSaved, userData, ui_config } = settingsScope;
+    const { showLog, showSensed } = settingsScope;
```
Abby-Wheelis, line 33: significantly less dependent on the `settingsScope` variables -- which means decreasing dependence on angular code, and approaching being able to use React routing

www/js/control/ProfileSettings.jsx
```diff
@@ -150,54 +255,206 @@ const ProfileSettings = () => {
...
     }
 
-    //conditional creation of setting sections
-    let userDataSection;
-    if(userDataSaved())
-    {
-        userDataSection = <ExpansionSection sectionTitle="control.user-data">
-                            <SettingRow textKey="control.erase-data" iconName="delete-forever" action={eraseUserData}></SettingRow>
-                            <ControlDataTable controlData={userData}></ControlDataTable>
-                        </ExpansionSection>;
```
Abby-Wheelis, line 198: removed userData section in anticipation of eliminating calories from dashboard, and thus need to collect/maintain/show this data

www/js/control/ProfileSettings.jsx
```diff
@@ -215,42 +472,46 @@ const ProfileSettings = () => {
...
-           <SettingRow textKey="control.download-json-dump" iconName="calendar" action={openDatePicker}></SettingRow>
-           {logUploadSection}
-           <SettingRow textKey="control.email-log" iconName="email" action={emailLog}></SettingRow>
-
-            {userDataSection}
+        <Appbar.Header statusBarHeight={12} elevated={true} style={{ height: 46, backgroundColor: 'white', elevation: 3 }}>
+            <Appbar.Content title={t("control.profile")} />
+            <Text>{t('control.log-out')}</Text>
+            <IconButton icon="logout" onPress={() => setLogoutVis(true)}></IconButton>
+        </Appbar.Header>
```
Abby-Wheelis, line 479: addition of `Appbar` is what allowed the logout button to go all the way to the top of the screen, and matches the label screen, characteristic of UI making progress into React

www/js/control/ProfileSettings.jsx
```diff
@@ -332,14 +594,110 @@ const ProfileSettings = () => {
...
+                    </Dialog.Actions>
+                </Dialog>
+            </Modal>
+
+            <DataDatePicker date={dumpDate} setDate={setDumpDate} open={dateDumpVis} setOpen={setDateDumpVis}></DataDatePicker>
+
+            <AlertBar visible={dataPushedVis} setVisible={setDataPushedVis} messageKey='all data pushed!'></AlertBar>
+            <AlertBar visible={invalidateSuccessVis} setVisible={setInvalidateSuccessVis} messageKey='success -> ' messageAddition={cacheResult}></AlertBar>
+            <AlertBar visible={noConsentMessageVis} setVisible={setNoConsentMessageVis} messageKey='general-settings.no-consent-message'></AlertBar> 
+
```
Abby-Wheelis, line 691: all of the `Modal` and `AlertBar` components replace ionic popups and their visibility is controlled by a state variable

Abby-Wheelis (commented):
For the most part, the changes represent moving away from angular and ionic code in order to implement things in React. We still use a lot of angular services, and the two log screens haven't been migrated yet.

AppStatusModal and PrivacyPolicyModal are two large pieces that are in Profile and Onboarding, so right now both versions are maintained, but it was my hope these would serve as a starting point when we get to the Onboarding migration.

www/js/control/ProfileSettings.jsx
Abby-Wheelis: I went ahead and did this as a review: https://github.com/e-mission/e-mission-phone/pull/1009#pullrequestreview-1620420136

PR #1013, opened 2023-08-14
🐛 Fix extra pipeline range request every time a trip is labeled + handle network error better

JGreenlee:
Resolves a bug tracked in https://github.com/e-mission/e-mission-docs/issues/948

The pipeline range should only be fetched when the label screen is initialized or refreshed. When a trip is labeled, we should not need to fetch the pipeline range - this was happening because the old `getUnprocessedLabels` function called `getPipelineRangeTs` every time it was called.

This function was rewritten into `getUnprocessedInputs` in `timelineHelper.ts`. It will accept the pipelineRange as an argument.

Now in LabelTab, only on init / refresh will we request the pipelineRange. Subsequent calls to `getUnprocessedLabels` will use the already-fetched pipelineRange.

We will also handle errors better - if retrieving pipeline range fails (network error), we should show error popup and stop loading spinner.

Testing done:
-Labeled several trips - all labels appeared within ~1 second. -Disconnected network, tried labeling trips - all labels appeared within ~1 second. -While network disconnected, tried refreshing label screen - popup "Error while loading pipeline range". Same result on app launch while network disconnected.

extra tidying note:
- removed getAngularService("DiaryHelper") from LabelTab because it no longer used (the functions that were previously used there had already been moved to the new Typescript file diaryHelper.ts)

### DIFF
--- a/www/js/diary/LabelTab.tsx
+++ b/www/js/diary/LabelTab.tsx
@@ -17,7 +17,7 @@ import LabelListScreen from "./LabelListScreen";
 import { createStackNavigator } from "@react-navigation/stack";
 import LabelScreenDetails from "./LabelDetailsScreen";
 import { NavigationContainer } from "@react-navigation/native";
-import { compositeTrips2TimelineMap, getUnprocessedInputs, populateBasicClasses, populateCompositeTrips } from "./timelineHelper";
+import { compositeTrips2TimelineMap, getAllUnprocessedInputs, getLocalUnprocessedInputs, populateBasicClasses, populateCompositeTrips } from "./timelineHelper";
 import { fillLocationNamesOfTrip, resetNominatimLimiter } from "./addressNamesHelper";
 
 let labelPopulateFactory, labelsResultMap, notesResultMap, showPlaces;
@@ -103,7 +103,7 @@ const LabelTab = () => {
   async function loadTimelineEntries() {
     try {
       const pipelineRange = await CommHelper.getPipelineRangeTs();
-      [labelsResultMap, notesResultMap] = await getUnprocessedInputs(pipelineRange, labelPopulateFactory, enbs);
+      [labelsResultMap, notesResultMap] = await getAllUnprocessedInputs(pipelineRange, labelPopulateFactory, enbs);
       Logger.log("After reading unprocessedInputs, labelsResultMap =" + JSON.stringify(labelsResultMap)
                                                 + "; notesResultMap = " + JSON.stringify(notesResultMap));
       setPipelineRange(pipelineRange);
@@ -235,7 +235,7 @@ const LabelTab = () => {
   const timelineMapRef = useRef(timelineMap);
   async function repopulateTimelineEntry(oid: string) {
     if (!timelineMap.has(oid)) return console.error("Item with oid: " + oid + " not found in timeline");
-    const [newLabels, newNotes] = await getUnprocessedInputs(pipelineRange, labelPopulateFactory, enbs);
+    const [newLabels, newNotes] = await getLocalUnprocessedInputs(pipelineRange, labelPopulateFactory, enbs);
     const repopTime = new Date().getTime();
     const newEntry = {...timelineMap.get(oid), justRepopulated: repopTime};
     populateBasicClasses(newEntry);
--- a/www/js/diary/LabelTab.tsx
+++ b/www/js/diary/LabelTab.tsx
@@ -17,7 +17,7 @@ import LabelListScreen from "./LabelListScreen";
 import { createStackNavigator } from "@react-navigation/stack";
 import LabelScreenDetails from "./LabelDetailsScreen";
 import { NavigationContainer } from "@react-navigation/native";
-import { compositeTrips2TimelineMap, populateBasicClasses, populateCompositeTrips } from "./timelineHelper";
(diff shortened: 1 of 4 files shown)


shankari:
it seems like we can optimize this one level further.

If we have pulled remote labels at the beginning, during the first retrieval, any new unprocessed labels must only be on the phone side. So we don't have to make the server calls to retrieve them. I think this will translate into using `BEMUserCache` directly instead of the `UnifiedDataLoader`. That way, we won't make *any* network calls while labeling.

In fact, when we wrote this out initially, I believe we didn't even read the unprocessed inputs again from the user cache again - since the user was labeling through the app, we just kept the matched inputs cached (I think in `userInput` versus `user_input`) so we didn't need to re-read the user cache ever.

Why do we need to read the inputs again after the first load?

JGreenlee:
> Why do we need to read the inputs again after the first load?

We don't strictly have to, but I found this to be a more React-friendly way of doing things.
My understanding of the old approach was to record the response in storage, then separately cache a "fake" label in the trip object. The React paradigm would be to store the response in only one place and receive it downstream.

It does simplify the codebase; I didn't realize it would cause any extra server requests though - definitely need to make sure that doesn't happen.

JGreenlee:
> any new unprocessed labels must only be on the phone side. So we don't have to make the server calls to retrieve them. I think this will translate into using `BEMUserCache` directly instead of the `UnifiedDataLoader`. That way, we won't make _any_ network calls while labeling.

Is there a reason that the original `getUnprocessedLabels()` function used `UnifiedDataLoader`?
https://github.com/e-mission/e-mission-phone/blob/b2d0a21c66a48a7829e76ee3402e8dd9ceae5ebc/www/js/diary/services.js#L141-L150

shankari:
> Is there a reason that the original getUnprocessedLabels() function used UnifiedDataLoader?

The original `getUnprocessedLabels` was only called on initial load. At that point, the labels could be:
- in the phone cache and not pushed up OR
- pushed up and in the server cache, but not yet processed

The `UnifiedDataLoader` is a utility library that reads from both the local cache and the remote cache and unified them with with proper de-duplication.

JGreenlee:
> * pushed up and in the server cache, but not yet processed

In this situation, are the labels gone from the local phone cache?
I thought they stayed in the phone cache until they were processed.

shankari:
> In this situation, are the labels gone from the local phone cache?
Yes.

The flow is:
- label in phone cache
- sync phone to server: label gone from phone cache, in server cache
- pipeline runs: labels moved into timeseries and incorporated into confirmed trips

JGreenlee:
I was struggling with this and making it more complicated than it needed to be, but I think my latest commit accomplishes the following: (see the commit message)

> If we have pulled remote labels at the beginning, during the first retrieval, any new unprocessed labels must only be on the phone side. So we don't have to make the server calls to retrieve them. I think this will translate into using `BEMUserCache` directly instead of the `UnifiedDataLoader`. That way, we won't make _any_ network calls while labeling.

___

But I am worried that we could be getting duplicates now, so I need to investigate a little bit and test different configurations.
I think with mode & purpose labels we wouldn't necessarily notice duplicates - but for Activities, duplicates would be problematic.

JGreenlee:
Actually, I'm not seeing duplicates activities upon entering one, like I expected to see.
I didn't add anything to remove duplicates myself. I don't see a problem, fortunately, but I want to verify that this is for a benign reason and not some other weirdness

JGreenlee:
We account for duplicated activities in https://github.com/e-mission/e-mission-phone/blob/b2d0a21c66a48a7829e76ee3402e8dd9ceae5ebc/www/js/survey/enketo/enketo-add-note-button.js#L68-L85
so it's ok.

No weirdness when adding activities (and fortunately, it's fairly snappy now), but some weird behavior when *editing* activities.

If I add an activity and immediately edit it without refreshing, the new activity shows up as a fresh entry instead of replacing the old one. But then upon refreshing, only the new one shows.

JGreenlee:
We edit by simultaneously adding a new entry and deleting the old one.

https://github.com/e-mission/e-mission-phone/blob/b2d0a21c66a48a7829e76ee3402e8dd9ceae5ebc/www/js/survey/enketo/AddedNotesList.tsx#L79-L90

___

`repopulateTimelineEntry` is called before `deleteEntry`. Both are async calls. maybe we haven't recorded the deletion yet when we are repopulating.
This wasn't an issue before because it took some period of time to label / add activities, and the deletion would have happened by the time we actually repopulated. 
We should await deletion before calling `repopulateTimelineEntry`

JGreenlee:
Testing all configurations and types of inputs, I am not seeing any other issues.
I also verified that labeling trips or completing surveys does not trigger any server calls by watching the server logs while testing.

shankari (approved):
This looks fine to me for now. I still don't understand why we need to re-read the data from the local phone cache every time - it is a local DB, but it is still a DB call instead of just reading the value from memory.

The fastest and most performant approach must be to store a copy in local memory. Even if React prefers to have a single source of truth, they must support in-memory write-through caches of data stored on disk/in a database. Otherwise, their performance will be really bad or they will have significant load on their server.

So I think, it would be good, before we go further down the rewrite route, to understand react policies around caching and single source of truth. That will help us we build out a decent distributed system architecture that supports high throughput and low latency.

PR #1016, opened 2023-08-19
🔧 ⬆️ 🚸 Upgrade android to API 33

shankari:
+ ⬆️ Upgrade android to API 33
+ 🔧 build script updates
+ 🚸 permission usability updates


shankari:
@niccolopaganini for visibility into the state of this change

shankari:
Cleared the decks before starting further changes; to ensure no pollution with old state

```
$ rm ./hooks/before_prepare/download_translation.js.bak
$ rm -rf platforms/
$ rm -rf plugins/
$ rm -rf node_modules/
$ diff package.cordovabuild.json package.json
$ diff config.cordovabuild.xml config.xml
```

shankari:
We will first implement the upgrade to the cordova platform and the cordova file plugin

shankari:
Bumped up 

```
        <preference name="android-minSdkVersion" value="24" />
        <preference name="android-targetSdkVersion" value="33" />
```

But still getting the same error

```
       7.  Dependency 'androidx.annotation:annotation-experimental:1.3.0' requires libraries and applications that
           depend on it to compile against version 33 or later of the
           Android APIs.

           :app is currently compiled against android-32.

           Recommended action: Update this project to use a newer compileSdkVersion
           of at least 33, for example 33.
```

I can set the compileSdkVersion, but I thought that android 12 was the more recent SDK version by default.

> This release has increased the minimum supported SDK version to 24 which is Android 7.0. It also has increased the target SDK to 33, Android 13.

> To use cordova-android@12, SDK Platform 33 and SDK Build Tools 33.0.2 must be installed. Older build tools version can be uninstalled if older versions of cordova-android is no longer used in your projects.

and they are installed
https://github.com/e-mission/e-mission-docs/issues/934#issuecomment-1685090087

Are any of the plugins overriding this?

shankari:
Aha!

```
$ grep -r compileSdkVersion plugins/
plugins//cordova-plugin-em-datacollection/src/android/enable_api_31.gradle:            compileSdkVersion 32
```

and 

```
node_modules//react-native/template/android/build.gradle:        compileSdkVersion = 33
node_modules//react-native/ReactAndroid/hermes-engine/build.gradle:    compileSdkVersion 31
node_modules//react-native/ReactAndroid/build.gradle:    compileSdkVersion 33

node_modules//@shopify/flash-list/android/build.gradle:def _compileSdkVersion = _ext.has('compileSdkVersion') ? _ext.compileSdkVersion : 30
node_modules//@shopify/flash-list/android/build.gradle:    compileSdkVersion _compileSdkVersion
node_modules//react-native-safe-area-context/android/build.gradle:    compileSdkVersion getExtOrDefault('compileSdkVersion', 30)
node_modules//react-native-vector-icons/android/build.gradle:  compileSdkVersion safeExtGet('compileSdkVersion', 28)

node_modules//react-native-screens/android/build.gradle:            if (safeExtGet('compileSdkVersion', rnsDefaultCompileSdkVersion) >= 33) {
node_modules//react-native-gesture-handler/android/build.gradle:    compileSdkVersion safeExtGet("compileSdkVersion", 28)
```

The final version is

```
$ cat build-extras.gradle
ext.postBuildExtras = {
    android {
        defaultConfig {
            compileSdkVersion 31
        }
    }
}
```

We _could_ change the version in data-collection, but now that we have a plethora react native components that specify a `compileSdkVersion`, let's start by setting it at the top level

shankari:
Setting it at the top level returns the same error

```
        <preference name="android-minSdkVersion" value="24" />
        <preference name="android-compileSdkVersion" value="33" />
        <preference name="android-targetSdkVersion" value="33" />
```

```
$ npm run build-dev-android > /tmp/android_build.log 2>&1
       1.  Dependency 'androidx.appcompat:appcompat-resources:1.6.1' requires libraries and applications that
           depend on it to compile against version 33 or later of the
           Android APIs.

           :app is currently compiled against android-32.
```

shankari:
Ok. I am currently betting that the `compileSdkVersion` is coming from the data collection plugin because:
- it is the only plugin with a `compileSdkVersion`; the others are in `node_modules`
- it is set to 32, while the ones in `node_modules` are all over the map

Let's check on why we set this in the plugin instead of the top level `compileSdkVersion` last time...

shankari:
Ah, we added it while upgrading to API 31
https://github.com/e-mission/e-mission-data-collection/commit/300336c3258b3a6ed6aecd4e7ea1eb49a5bfc7e4
because cordova-android had removed the top level `compileSDKVersion` setting

I filed <a href="https://github.com/apache/cordova-android/issues/1373">an issue</a> for cordova-android


but added this is a workaround

They re-introduced the flag in android 11, with the
<a href="https://github.com/apache/cordova-android/issues/1373#issuecomment-1127157694">following behavior</a>

- If the target & compile sdk is not configured, it will default to the cordova-android default.
- If the user sets the target sdk, the compile SDK will match.

which are what we want, depending on whether we want to force the targetSDK or not.

So we should remove the workaround....

shankari:
Removed `$ rm src/android/enable_api_31.gradle` from the directory and the `plugin.xml`

```
$ npx cordova plugin rm cordova-plugin-em-datacollection
$ npx cordova plugin add ../plugins/e-mission-data-collection
```

That worked! Compile was successful!

```
$ npm run build-dev-android > /tmp/android_build.log 2>&1
BUILD SUCCESSFUL in 59s
52 actionable tasks: 46 executed, 6 up-to-date
Built the following apk(s):
        /Users/kshankar/e-mission/native_code_upgrade/platforms/android/app/build/outputs/apk/debug/app-debug.apk
```

shankari:
Let us now try to run in on an android emulator.

I have an android-32 emulator

```
---------
    Name: Pixel_2_API_32
  Device: pixel_2 (Google)
    Path: /Users/kshankar/.android/avd/Pixel_2_API_32.avd
  Target: Google Play (Google Inc.)
          Based on: Android 12L (Sv2) Tag/ABI: google_apis_playstore/x86_64
    Skin: pixel_2
  Sdcard: 512M
```

I don't have an android-33 avd, so I'll copy this one over and test on that as well.

shankari:
The app launches, but is stuck in the start screen. On connecting via the chrome debugger, looks like `deviceready` has not fired. Checking adb, might update this comment.

```
deviceready has not fired after 5 seconds.
​ Channel not fired: onDOMContentLoaded
```

Screenshot 2023-08-19 at 2 48 45 PM

shankari:
Ah, this is much _snappier_ (pun intended)

```
$ ./emulator/emulator -avd Pixel_2_API_33 -no-snapstorage
```

| Launched! | Got to permissions correctly | but then hung while returning from setting permissions |
|------|-------|------|
| Screenshot 2023-08-19 at 3 19 11 PM | Screenshot 2023-08-19 at 3 20 58 PM | Screenshot 2023-08-19 at 3 22 23 PM |

shankari:
On trying to re-launch, ran into the application timeout again.
Then tried to re-re-launch, this time permission page was fine.
App launch was fine.

Phew! It really does appear to be almost as smooth as I had hoped.
Let's check this in and then start polishing.

shankari:
Created a new branch called `maint_upgrade_2023` so that we can run the tests on PRs
https://github.com/e-mission/e-mission-phone/blob/b6da3bbaa1b14f19937c079bef975dc6752c427e/.github/workflows/android-build.yml#L11

and rebased to it

shankari:
Removed all the overrides, build still succeeded.
```
-        <preference name="android-minSdkVersion" value="23" />
-        <preference name="android-targetSdkVersion" value="32" />
```

shankari:
upgrading the android SDK fails, probably need to upgrade the java version

```
Downloading the android SDK. This will take a LONG time and will require you to agree to lots of licenses.
Do you wish to continue? (Y/N)BEGIN: About to start android SDK download
Error: LinkageError occurred while loading main class com.android.sdklib.tool.sdkmanager.SdkManagerCli
        java.lang.UnsupportedClassVersionError: com/android/sdklib/tool/sdkmanager/SdkManagerCli has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 55.0
END: Done with android SDK download, exiting script
```

shankari:
Yep! It appears to need 16+
https://stackoverflow.com/questions/69674255/how-to-fix-linkageerror-occurred-while-loading-main-class-error-using-an-other

the most recent version is 17.0.8
https://adoptium.net/temurin/releases/?os=mac

@Abby-Wheelis @niccolopaganini  how hard was it to get IT to install a new version of java for you? I am tempted to bump this up now, but don't want to block/affect your work

shankari:
@niccolopaganini will wait for your review until Monday morning, otherwise I will merge since all the tests as passing
will address android command line tools upgrade in separate PR

niccolopaganini (commented):
Hello ma'am. These changes look good to me.

niccolopaganini:
> how hard was it to get IT to install a new version of java for you? I am tempted to bump this up now, but don't want to block/affect your work

One of the problems that I faced during the initial build was that the JDK version was installed differently. I was not able to build. I can try it with the updates on a separate computer and see how the changes affect.

Abby-Wheelis:
> how hard was it to get IT to install a new version of java for you? I am tempted to bump this up now, but don't want to block/affect your work

It took a while, but the actual install was very simple. It took 11ish days from filing the request until I got it installed, but if I had it to do over again I'd follow up sooner, I asked for an update after the request hadn't moved for about a week and ended up getting to install later that day. Since I just went through the process, I'd be happy to help anyone else that needs get the request filed. Once installed I was able to set the path and run the android emulator easily.

PR #1017, opened 2023-08-19
accidental key removals

Abby-Wheelis:
I found these as deleted in comparing to the PR in e-mission translate, I must have taken them out by mistake


shankari:
So there was a merge conflict in this due to the "Refresh" having a space. I resolved it, but now there is only that change left.

The aria branch does still have "Fix"
https://github.com/e-mission/e-mission-phone/blob/3d2e1b98d8bfbb1377a4690e801c8260e344aa39/www/i18n/en.json#L224

So I am not sure why you needed to add it in your commit
https://github.com/e-mission/e-mission-phone/pull/1017/commits/a6a1a5461540d58c622ffb12ad5ffe06fc441eea#diff-a9d7f252f809b38fc33b5713e1572d24240a54a17589256affdf78c37ff9999dR224

I think it might be because you hadn't merged back from aria to your branch?

Abby-Wheelis:
I think you're right, these changes seem to be moot, I'll close this PR, no need to merge

PR #1002, opened 2023-07-12
Update Dashboard UI

ctyrrellnrel:
- Very basic, just replaced normal details view of 'My distance' with chart,
- next steps 
    - switch 'details' view to a react component 
    - create button to switch between chart and details view in 'my distance' card 
    - Figure out why all data is showing up unlabled 
    - Allow the axis to switch easily 
         - right now, setting isHorizontal doesn't completely switch data, only thedirection of the bars 
         - would be nice to flip data axis and bar axis simultaneously

PR #1018, opened 2023-08-21
📊 Rewrite of the New Dashboard Tab

JGreenlee:
- This PR will extend on https://github.com/e-mission/e-mission-phone/pull/1002 (which can now be closed) to rewrite the Dashboard tab in React components, while implementing the new design as discussed in https://github.com/e-mission/e-mission-docs/issues/922


JGreenlee:
When rewriting the other screens, we have taken a bottom-up approach where we take the existing UI and gradually refactor it into components, starting from building blocks and working up to the larger pieces - and eventually we get to the main views or screens.

However, the existing Dashboard is harder to do this for because it is not organized into components at all. It exists as one very dense view+controller. The template `main-metrics.html` is 250 lines long and contains almost everything for the entire Dashboard tab. `metrics.js` is 1450 lines long and is largely opaque to me in my understanding of how it works.

I am starting to wonder if it would be more efficient to tackle this in a more 'top-down' way and not get caught up in how the old dashboard did things.
At the end of all this, we will want a Screen component that is a function of some input data, and its output is the rendered UI that we came up with in https://github.com/e-mission/e-mission-docs/issues/922.

That input data definitely includes whatever we receive from the server - plus *maybe* a few things that are stored on the phone. (But in my experience using the Dashboard, it doesn't seem like unprocessed labels are considered for the computation of metrics, so maybe server data is the *only* input)

Maybe, instead of gradually refactoring into UI components under the existing Angular architecture, we should start with what data is received from the server and then figure out how to represent that in a component-based architecture.

Thoughts? @abby-wheelis @shankari

JGreenlee:
For each of 1) user metrics and 2) aggregate metrics, I can see that we receive an array with 4 entries, each of which contain an array of objects.

image

I think the 4 entries of the outer array are the four metrics we are evaluating: 1) distance, 2) trip count, 3) duration, and 4) speed.
And I think the objects that each of these contain correspond to a particular day:
image
August 7, August 8, August 9.

Abby-Wheelis:
>tackle this in a more 'top-down' way and not get caught up in how the old dashboard did things

I think it makes sense to start with the data we bring in from the server (and maybe the phone) and then build the pieces we want to display off of that. In a sense we will still be able to build one component at a time, but just based off of what we want to end up with, not what we started with. I've barely looked at the code that controls the Dashboard currently, but agree that we want to avoid getting bogged down in old code, in this case in particular because our end goal is more of a redesign and less of a translation (at least compared to the Profile redesign I've been doing).

JGreenlee:
At a high level, the new Dashboard screen has 3 clusters of information that it needs for its 3 main sections:
- "My Footprint" needs computed values of 'kilograms CO2'
    - **Where do these currently come from? Do these computations require any input other than the above server data?**
- "Active minutes" just needs a list of which modes are considered "active" (and potentially a mapping of what is high intensity vs low intensity), along with the above server data.
- "My Distance / Trips / Duration / Speed" just needs the above server data (and which units of measurement should be used for display)

shankari:
> Maybe, instead of gradually refactoring into UI components under the existing Angular architecture, we should start with what data is received from the server and then figure out how to represent that in a component-based architecture.

I am fine with this. I edited the metrics screen most recently (to support the use of user labels in the dashboard calculations) and it was an unholy mess. I refactored it to some extent in the time that I had, and it is better than before (e.g. https://github.com/e-mission/e-mission-phone/commit/6e88d8e1292dcc93464a5c93acd716f5c80f6a7a !!) but I remember that I wanted to just throw it away and rewrite but not having the time to do so.

> Where do these currently come from? Do these computations require any input other than the above server data? Is this going to change anyway with https://github.com/e-mission/e-mission-docs/issues/954?

For user labels, these come from the trip_confirm options! the `"co2PerMeter"` field.
For the sensed modes (if we are not returning user labels, these come from the CarbonFootprint settings)
Both of these should be handled in `FootprintHelper`

JGreenlee:
Some progress on implementing the new design

<table><td>

</td><td>

</td></table>

Units of measurement are not handled yet

Abby-Wheelis:
Another progress screenshot - Carbon Footprint card is making progress

Bar colors to better represent uncertainty and potential good/bad indication, see [issue#961(comment)]( https://github.com/e-mission/e-mission-docs/issues/961#issuecomment-1696798475) still to be done

Abby-Wheelis:
**Examples of color-by-mode:**




Drove Alone and Shared Ride, as well as Bus and Train would be the same color, but repeated colors are darkened or lightened to prevent duplication, giving up to 5 total options.

JGreenlee:
**color-by-mode** is great and I think it does a lot for the UX of the app because it conceptually links the mode labels between the Label tab and the Dashboard.

In the above example, it starts to become difficult to tell the modes apart from one another. We have noted that will be exacerbated in large studies with even more modes.
I think it is necessary to refactor this, and then modify it so that the colors are more distinct even when many duplicates occur.

Here are some screenshots as I am reworking this and making the de-dup function more balanced:

<table>
<tr>
<td>With a moderate number of modes, the colors will be easily distinguished from each other and clearly identified as the color of the base mode (e.g. see that Bike and E-bike are both green, but Bike is dark green while E-bike is light green)</td>
<td></td>
</tr><tr>
<td>As the number of modes increases, and we have more user-defined labels, it becomes challenging to give them all distinct colors. At a certain point, colors with many duplicates will start to appear washed-out or muddy. Although this is not as aesthetically pleasing, it does ensure each label has a distinct color.
<td></td>
</tr>
</table>

shankari:
I have two high-level thoughts around this:
- if we are going to hardcode color mappings, let's be consistent in the public dashboard as well @iantei
- but should we hardcode color mappings in the first place, especially given that users can now support custom labels? We do have them specify the basemode in the custom config, but it sounds like you are actually coding the rich modes, not just the base modes. Would it make sense to have the rich mode colors be generated from basemode colors by perturbing the base modes?

If we do end up using perturbations, we may want to write a common library, with implementations in both python and javascript to implement the same code so that it can be reused in the public dashboard and the individual dashboard.

JGreenlee:
> * if we are going to hardcode color mappings, let's be consistent in the public dashboard as well @iantei

Absolutely - we can use the same base mode colors in both places.

> * but should we hardcode color mappings in the first place (...)

For base modes, I think yes, because the semantic color association of modes, and the consistency of these colors throughout the UI, seems important to users.
(e.g. car = red because red = bad; and red means car whether we are on the Label tab or Dashboard tab)

> * it sounds like you are actually coding the rich modes, not just the base modes.

The basemode colors are the only thing currently hardcoded.

> Would it make sense to have the rich mode colors be generated from basemode colors by perturbing the base modes?

We are currently doing this. But the perturbations (i.e. tinting the base mode color lighter or darker) only happen when duplicates occur on the charts, and only happen on a case-by-case basis.
Because when a chart has to display 5 modes with the same base mode, more drastic perturbations are needed than when there are only 2 with the same base mode.

On the JS side of things, we achieve this with:
- `darken` and `lighten` from the [`color`](https://www.npmjs.com/package/color) library (it is internally used by RN Paper so we might as well mooch off of it)
- a simple de-duplication function that Abby and I wrote that applies `lighten` or `darken` to the duplicates according to how many duplicates there are.

To do the same on the public dashboard, we would just need a Python library that provides `darken` and `lighten` (or similar), and re-implement the de-duplication function.

JGreenlee:
Since #1009 has been merged into `label_dashboard_profile_sept_2023`, I rebased, and this branch now has all 3 tabs (Label, Dashboard, Profile) in React.

This means now is a good time to create the `App` component that allows us to navigate between these 3 tabs. I think we can accomplish the tabs/routing by putting a `BottomNavigation` (from React Native Paper) inside our `App` component.

If we need more complex routing (which I am leaning towards thinking we don't) we can set that up in `React Navigation` (a different library) and use  `BottomNavigation.Bar` (which is meant to easily hook up to `React Navigation`).

But with the exception of onboarding (which is one-time and separate from the tabs), the whole app is pretty much just 3 tabs, each of which have a set of different modals they can launch. We don't have a lot of complex routing between screens or across tabs. If we do need _basic_ routing between tabs, it looks like we can do so by setting the index of the `BottomNavigation`.

So unless anyone objects, I say we try to accomplish this with the tools that `React Native Paper` provides by default, and only consider `React Navigation` if we find the default tools insufficient.

JGreenlee:
The new dashboard is starting to feel more polished:

<table><td>

image

</td><td>

image

</td></table>

JGreenlee:
Only a couple tasks left that I can think of:

- In "My Footprint", try color gradient bars and decide whether it's better than the plain colors w/ dashed lines
- Have "Unlabeled" bars show as outlined and/or dashed on the Distance, Trips, Duration cards as well, and as grey instead of brown

shankari:
I think that this looks great. Looking forward to pushing this out to production in the next release train (the current one leaves tonight) and getting feedback from users!

Abby-Wheelis:
I've been really hesitant to get on board with gradient bars because I've been worried that they might be visually overwhelming, but I think you're right that it communicates the information more clearly. The contrast between the two sets of images really helps me see that, with the green starting to show on the lower week in the second row, even though the threshold has not been reached. The relationship between transparency and "unlabeled" is clear to me, I don't feel like we would have to use the lines to show that, and it could make sense to show "unlabeled" as partially transparent in the other charts as well?

shankari:
@JGreenlee @Abby-Wheelis where are we showing the actual footprint from this week? 
I don't see a bar for it to compare with all the other bars.
Also, my daughter is complaining that the gradients should all start at the same color so that you can see how far along the gradient the bar is

JGreenlee:
How about we **only** start the gradient at the 2050 guideline?

If emissions are below the 2050 line, the bar will be solid green.

If they pass the 2050 line, the bar will begin as yellow/orange and only start transitioning to red the further it gets past the 2050 line. By the 2030 line, it is red - past this it can be darker red.

This will ensure that all bars are following the same gradient - 15kg CO2 will always mean yellow, 35 kg CO2 will always mean orange, and 55 kg CO2 will always mean red.

It will be more similar to the first example I referenced here: https://github.com/e-mission/e-mission-docs/issues/961#issuecomment-1699969668

JGreenlee:
@Abby-Wheelis Would that also help to keep it from becoming visually overwhelming?

Abby-Wheelis:
@JGreenlee I think so, the example you linked seems like it would be easy to understand. Keeping the gradient with a consistent starting point gives people one less thing to try and understand the first time they look at the chart.

shankari:
I would also suggest using something around "7 days" because there is ambiguity around "week"
e.g. is it the last 7 days or the "week" starting on Saturday/Sunday/Monday?

JGreenlee:
What's ambiguous about "past week"? Is "past 7 days" any better?

"Past week" / "Prev. week" or "Past 7 days" / "Prev. 7 days"
Also, note that we're very much limited on space for that text. Exceeding 10 or so characters might force the graph to be smaller.

We chose "past week" because "this week" was ambiguous. And then we explicitly included the dates underneath -- because if the user selects a different date range with the datepicker, it is "past week" *relative to* the chosen end date.

shankari:
I think that "past 7 days" is better (although it is longer) because then there is no confusion on when the "week" starts.
We can still show the date range under the text - that is definitely a good idea. But I don't see that "past 7 days" with a custom range is more confusing that "past week" with a custom range. Can you clarify further?

JGreenlee:
I don't think "Past 7 days" is more confusing with dates - just slightly wordier and harder to digest at first.

JGreenlee:
It was mentioned earlier that `If all 'taxi'` is not definitely not clear either.

The only other idea I can come up with, without getting too long, is

>Worst case
>(All 'taxi')

We could also just say `Worst case⁴` and put another footnote to keep the axis labels concise.

Abby-Wheelis:
>Worst case
(All 'taxi')

I like "worst case", and it might not need a subtitle, since "worst case" is what that number means. In `CarbonFootprintCard` what's being calculated is the "highest footprint" given the "worst distance". If a footnote would help, we could write out something along the lines of "worst case emissions estimate based on distance in the past week/7 days".

JGreenlee:
I think it is a lot better with the gradients aligned. It's also simpler in the code.

<table>
<td>

image

</td>
<td>

image

</td></table>

shankari:
@JGreenlee can you resolve the merge conflict? I took a look and the difference is basically in the colors used. I assume that the colors in the `dashboard-rewrite` branch are correct, but would appreciate your taking a look.

www/js/metrics/CarbonFootprintCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+            if(userLastWeekSummaryMap[0]) {
+                userPrevWeek = {
+                    low: FootprintHelper.getFootprintForMetrics(userLastWeekSummaryMap, 0), 
+                    high: FootprintHelper.getFootprintForMetrics(userLastWeekSummaryMap, FootprintHelper.getHighestFootprint())
+                };
+                graphRecords.push({label: t('main-metrics.unlabeled'),  x: userPrevWeek.high - userPrevWeek.low, y: `${t('main-metrics.prev-week')}\n(${formatDateRangeOfDays(lastWeekDistance)})`})
+                graphRecords.push({label: t('main-metrics.labeled'), x: userPrevWeek.low, y: `${t('main-metrics.prev-week')}\n(${formatDateRangeOfDays(lastWeekDistance)})`});
+            }
+
+            //calculate low-high and format range for past week
```
shankari, line 58: can we clarify that past week = past 7 days? Maybe this is just my verbal blind spot, but whenever I see "past week", I think that it is the last week, not the "current week". And since this is code, we have a lot of space for clarification.

www/js/metrics/CarbonTextCard.tsx
```diff
@@ -0,0 +1,151 @@
...
+const CarbonTextCard = ({ userMetrics, aggMetrics }: Props) => {
+
+  const { colors } = useTheme();
+  const { t } = useTranslation();
+  const FootprintHelper = getAngularService("FootprintHelper");
+
+  const userText = useMemo(() => {
+    if(userMetrics?.distance?.length > 0) {
+        //separate data into weeks
+        const [thisWeekDistance, lastWeekDistance] = segmentDaysByWeeks(userMetrics?.distance, 2);
```
shankari, line 20: instead of recomputing this separately for `CarbonTextCard` and `CarbonFootprintCard`, is it possible to compute it once in `metricsHelper`? This may not be too much of a performance optimization but I want to get the sense of whether it is possible in react and is "the react way".

www/js/metrics/CarbonTextCard.tsx
```diff
@@ -0,0 +1,151 @@
...
+          let userPrevWeek = {
+              low: FootprintHelper.getFootprintForMetrics(userLastWeekSummaryMap, 0), 
+              high: FootprintHelper.getFootprintForMetrics(userLastWeekSummaryMap, FootprintHelper.getHighestFootprint())
+          };
+          const label = `${t('main-metrics.prev-week')} (${formatDateRangeOfDays(lastWeekDistance)})`;
+          if (userPrevWeek.low == userPrevWeek.high)
+            textList.push({label: label, value: Math.round(userPrevWeek.low)});
+          else
+            textList.push({label: label + '²', value: `${Math.round(userPrevWeek.low)} - ${Math.round(userPrevWeek.high)}`});
+        }
```
shankari, line 49: seems like all this computation can also become common.

www/js/metrics/ChangeIndicator.tsx
```diff
@@ -0,0 +1,75 @@
...
+            return "+";
+        } else {
+            return "-";
+        }
+    };
+
+    const changeText = useMemo(() => {
+        if(change) {
+            let low = isNaN(change.low) ? '∞' : Math.round(Math.abs(change.low));
+            let high = isNaN(change.high) ? '∞' : Math.round(Math.abs(change.high));
```
shankari, line 26: do we have screenshots  of what this would look like?  I want to see if `∞` is the best representation, or whether we should use `-` or something else

www/js/metrics/metricsHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+export function segmentDaysByWeeks (days: DayOfMetricData[], nWeeks?: number) {
+  const weeks: DayOfMetricData[][] = [];
+  for (let i = days?.length - 1; i >= 0; i -= 7) {
+    weeks.push(days.slice(Math.max(i - 6, 0), i + 1));
+  }
+  if (nWeeks) return weeks.slice(0, nWeeks);
+  return weeks;
+};
+
+export function formatDate(day: DayOfMetricData) {
```
shankari, line 44: isn't this in/overlapping with diaryHelper?

www/js/metrics/metricsHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+  const dt = DateTime.fromISO(day.fmt_time, { zone: 'utc' });
+  return dt.toLocaleString({...DateTime.DATE_SHORT, year: undefined});
+}
+
+export function formatDateRangeOfDays(days: DayOfMetricData[]) {
```
shankari, line 49: Ditto (for overlapping entries)

www/js/metrics/metricsHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+  const firstDay = firstDayDt.toLocaleString({...DateTime.DATE_SHORT, year: undefined});
+  const lastDay = lastDayDt.toLocaleString({...DateTime.DATE_SHORT, year: undefined});
+  return `${firstDay} - ${lastDay}`;
+}
+
+/* formatting data form carbon footprint calculations */
+
+//modes considered on foot for carbon calculation, expandable as needed
+const ON_FOOT_MODES = ['WALKING', 'RUNNING', 'ON_FOOT'] as const;
+
```
shankari, line 62: as long as we use outputs beyond CLEAN_AND_RESAMPLE in the pipeline, we should not need this mapping. We should clarify that we only support cleaned/inferred/confirmed values in the dashboard so that we can simplify this logic.

www/js/metrics/metricsHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+        if (!(field in mode_bins)) {
+          mode_bins[field] = [];
+        }
+        //for all except onFoot, add to bin - could discover mult onFoot modes
+        if (field != "ON_FOOT") {
+          mode_bins[field].push([metric.ts, metricToValue(population, metric, field), metric.fmt_time]);
+        }
+      }
+      //this section handles user lables, assuming 'label_' prefix
+      if(field.startsWith('label_')) {
```
shankari, line 122: we should be able to split the parsing of inferred versus user label modes into two separate functions instead of this giant function - e.g. the two `if` statements

www/js/metrics/metricsHelper.ts
```diff
@@ -0,0 +1,212 @@
...
+  return summaryMap;
+}
+
+/*
+* We use the results to determine whether these results are from custom
+* labels or from the automatically sensed labels. Automatically sensedV
+* labels are in all caps, custom labels are prefixed by label, but have had
+* the label_prefix stripped out before this. Results should have either all
+* sensed labels or all custom labels.
+*/
```
shankari, line 178: for the record: this should eventually go into the dynamic config - whether people want to use sensed values or user values. Or it may be moot once "count every trip" is implemented.

www/js/metrics/MetricsTab.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+  if (type == 'user')
+    return CommHelper.getMetrics('timestamp', query);
+  return CommHelper.getAggregateData("result/metrics/timestamp", query);
+}
+
+function getLastTwoWeeksDtRange() {
+  const now = DateTime.now().startOf('day');
+  const start = now.minus({ days: 15 });
+  const end = now.minus({ days: 1 });
```
shankari, line 39: for the record, we might want to revisit this later if we ever change our labeling flow to be more real-time, so that the labels are reflected in the metrics. Note also that this is in UTC, which may cause some confusion among users that we could consider addressing in a future cleanup.

www/js/metrics/MetricsDateSelect.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Text, StyleSheet } from "react-native";
+import { DatePickerModal } from "react-native-paper-dates";
+import { Divider, useTheme } from "react-native-paper";
+import i18next from "i18next";
+import { useTranslation } from "react-i18next";
+import NavBarButton from "../components/NavBarButton";
+import { DateTime } from "luxon";
+
+type Props = {
+  dateRange: DateTime[],
```
shankari, line 19: we should set this to the first date of the study/program or the first date for the current user, similar to the "Download JSON dump" code in the profile.

shankari (changes requested):
I am fine with merging this once the merge conflict has been resolved, and tackling the other changes in a fairly quick cleanup. It would also be good to pause for a bit and add tests for the `metricsHelper` while cleaning up.

JGreenlee:
Confirmed that the merge conflict was just color values and that `dashboard-rewrite` had the most up-to-date palette.
I fixed the conflict and will address the rest of the review tomorrow.

www/js/metrics-factory.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 import angular from 'angular';
+import { getBaseModeByValue } from './diary/diaryHelper'
+import { labelOptions } from './survey/multilabel/confirmHelper';
```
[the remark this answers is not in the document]
JGreenlee: No we don't need. Leftover from a previous commit and those functions aren't used here anymore
TODO

www/js/survey/multilabel/confirmHelper.ts
```diff
@@ -13,21 +13,22 @@ type InputDetails<T extends string> = {
...
   }[]
 } & { translations: {
   [lang: string]: { [translationKey: string]: string }
 }};
 
 let appConfig;
-let labelOptions: LabelOptions<'MODE'|'PURPOSE'|'REPLACED_MODE'>;
-let inputDetails: InputDetails<'MODE'|'PURPOSE'|'REPLACED_MODE'>;
+export let labelOptions: LabelOptions<'MODE'|'PURPOSE'|'REPLACED_MODE'>;
+export let inputDetails: InputDetails<'MODE'|'PURPOSE'|'REPLACED_MODE'>;
```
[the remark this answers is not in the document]
JGreenlee: Yes, after refactoring they're imported in the metrics components for better encapsulation- this way the charting code doesn't have bits of 'label' code sprinkled in it

www/js/metrics/ActiveMinutesTableCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ACTIVE_MODES } from './WeeklyActiveMinutesCard';
+import { labelKeyToRichMode } from '../survey/multilabel/confirmHelper';
+
+type Props = { userMetrics: MetricsData }
+const ActiveMinutesTableCard = ({ userMetrics }: Props) => {
+
+  const { colors } = useTheme();
+  const { t } = useTranslation();
+
+  const cumulativeTotals = useMemo(() => {
```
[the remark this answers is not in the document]
JGreenlee: The memoized functions calculate the values that the component will display, no? (At least, that is the way I've been using them)

Once we're able to test this component, we'll be able to assert that given the metrics for example user A, the first row of the table should read: `"9/9 - 9/22 | 190 minutes | 60 minutes"` when it is rendered.

If we need to, we could also extract this to a named pure function whose input param is `userMetrics.duration` - and then unit test. That way we would able to find mistakes in the logic by checking the function result.

But then what if there was a mistake or error while rendering that result? That is why, to me, component testing seems much preferable - we test what actually gets shown to the user and not just some intermediate value

www/js/config/useImperialConfig.ts
```diff
@@ -41,7 +48,9 @@ export function useImperialConfig() {
   return {
     distanceSuffix: useImperial ? "mi" : "km",
     speedSuffix: useImperial ? "mph" : "kmph",
-    getFormattedDistance: useImperial ? getFormattedDistanceInMiles : getFormattedDistanceInKm,
-    getFormattedSpeed: useImperial ? getMph : getKmph,
+    getFormattedDistance: useImperial ? (d) => formatForDisplay(convertDistance(d, true))
+                                      : (d) => formatForDisplay(convertDistance(d, false)),
+    getFormattedSpeed: useImperial ? (s) => formatForDisplay(convertSpeed(s, true))
+                                   : (s) => formatForDisplay(convertSpeed(s, false)),
```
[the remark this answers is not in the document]
JGreenlee: Yes, I think speed values are likely to be shown when we add them back later, both on the dashboard and the label details screen

shankari:
There is now a new conflict in `main.js`, likely from my merge of #1029

www/js/metrics/CarbonFootprintCard.tsx
Abby-Wheelis: I've just added some comments to this section, referring to "how many days ago" the calculations are referencing, so the previous week is 14 days ago - 8 days ago, and the past week is 7 days ago - yesterday. Depending on user selection, this could be shifted up to include "today".

www/js/metrics/CarbonFootprintCard.tsx
```diff
@@ -0,0 +1,168 @@
...
+            //separate data into weeks
+            const [thisWeekDistance, lastWeekDistance] = segmentDaysByWeeks(userMetrics?.distance, 2);
+
+            //formatted data from last week
+            let userLastWeekModeMap = {};
+            let userLastWeekSummaryMap = {};
+            if(lastWeekDistance && lastWeekDistance?.length >= 7) {
+                userLastWeekModeMap = parseDataFromMetrics(lastWeekDistance, 'user');
+                userLastWeekSummaryMap = generateSummaryFromData(userLastWeekModeMap, 'distance');
+            }
```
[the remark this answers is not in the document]
Abby-Wheelis: Definitely never greater than 7 days, I was focused on weeks < 7 days (since that's not a week), but > days is an unexpected behavior as well, changing that to == 7 days.

www/js/config/useImperialConfig.ts
```diff
@@ -1,33 +1,40 @@
...
+  return distMeters / 1000;
+}
 
-const getMph = (metersPerSecond) =>
-  (KM_TO_MILES * Number.parseFloat(getKmph(metersPerSecond))).toFixed(2);
+const convertSpeed = (speedMetersPerSec: number, imperial: boolean): number => {
+  if (imperial)
+    return speedMetersPerSec * MPS_TO_KMPH * KM_TO_MILES;
+  return speedMetersPerSec * MPS_TO_KMPH;
+}
```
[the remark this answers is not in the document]
JGreenlee: I don't understand this. Do you mean the Angular service ImperialConfig? This file replaces that.

package.cordovabuild.json
```diff
@@ -109,7 +109,6 @@
     "angular": "1.6.7",
     "angular-animate": "1.6.7",
     "angular-local-storage": "^0.7.1",
-    "angular-nvd3": "^1.0.7",
```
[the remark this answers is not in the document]
JGreenlee: The speed graph should be fairly easy to add because we made `LineChart` in such a way that it can be easily re-used.
If we want, we could display plenty of other interesting things there too.

www/js/metrics/ActiveMinutesTableCard.tsx
```diff
@@ -0,0 +1,99 @@
...
+              <DataTable.Cell>{total['period']}</DataTable.Cell>
+              {ACTIVE_MODES.map((mode, j) =>
+                <DataTable.Cell key={j}>{total[mode]} {t('metrics.minutes')}</DataTable.Cell>
+              )}
+            </DataTable.Row>
+          )}
+          <DataTable.Pagination page={page} onPageChange={p => setPage(p)}
+            numberOfPages={Math.ceil(allTotals.length / 5)} numberOfItemsPerPage={5}
+            label={`${page * 5 + 1}-${page * 5 + 5} of ${allTotals.length}`} />
+        </DataTable>
```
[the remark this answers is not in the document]
JGreenlee: Right now this is our only paginated DataTable. But yeah, I thought the same thing.

I was kind of surprised that RN Paper doesn't handle the pagination calculations for us. But it wouldn't be much work to extract this to a generic reusable function anyway

www/js/metrics/CarbonTextCard.tsx
JGreenlee: Certainly possible. The body of a `useMemo` can call any function, including ones exported from generic Typescript modules ('helpers' as I've been naming them)
TODO cleanup

www/js/metrics/CarbonTextCard.tsx
JGreenlee: TODO cleanup

www/js/metrics/ChangeIndicator.tsx
JGreenlee: This is something we may want to address before a prod release.
I have not actually seen `∞` get displayed but I have seen the word `infinity` show up if the data is weird. So I don't know if this is really working. We should test and investigate

TODO now

www/js/metrics/metricsHelper.ts
JGreenlee: Kind of. This is all using Luxon (ie `DateTime`) but `diaryHelper` is still using Moment in a few places, which should be converted eventually

And they're not exactly the same date format because the day of week "Mon" or "Monday" is used on the label tab while just numeric dates are used on the dashboard.

But still at least some unification can be done, I think.

www/js/metrics/metricsHelper.ts
JGreenlee: TODO cleanup

www/js/metrics/metricsHelper.ts
JGreenlee: because after cleaning, all of these would turn into 'WALKING', right?

www/js/metrics/ChangeIndicator.tsx
Abby-Wheelis: Investigating now, and it definitely isn't set up right, I think I need to work on `calculatePercentChange` because it sometimes gives the string "infinity"

Abby-Wheelis: I found a fix - use `isFinite` instead of `isNaN` -- this will catch "Infinity" and "NaN", when I used {0,0} -> {0,100} as my two weeks low/high values, I was able to have this display: 

Screenshot 2023-09-25 at 2 33 54 PM

 If we like this behavior and '∞' I can push this change.

JGreenlee: Definitely looks better this way!

I think if both boundaries are infinity though, we could consider not showing the change indicator at all, since it's not going to have any useful information

Abby-Wheelis: I think, in this case, the difference should really be 0 -> infinity, so maybe we should try and represent that? Or just omit it altogether.

Abby-Wheelis: I have it set up now so that if the range is -inf to +inf, it doesn't show up, because as you said that doesn't mean much. However, if the change is infinite and we're sure of that, show that (ex. 0 emissions to 100 emissions would be infinite change)

www/js/config/useImperialConfig.ts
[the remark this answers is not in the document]
shankari: you are right. I added a bunch of comments when I thought this was outside `imperialConfig`, then noticed that it was inside, and deleted them, but I guess I missed one.

www/js/metrics/ActiveMinutesTableCard.tsx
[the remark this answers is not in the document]
shankari: well, theoretically, you do both. At a high level, the pure functions can be tested much more easily with a broader variety of inputs, including invalid values. The component tests can be more sophisticated and include more realistic inputs.

www/js/metrics/ChangeIndicator.tsx
shankari: I think we may want to skip the %  change display in the zero -> 100 case as well because people don't expect to encounter inf in the real world (even if it is mathematically correct). But you can clean that up in TODO

www/js/metrics/metricsHelper.ts
shankari: Yup. The `RUNNING` and `ON_FOOT` are really for data that comes in directly from the phone.

shankari (approved):

PR #1020, opened 2023-08-22
🚑 Fix Angular Annotations in logger.ts

JGreenlee:
fixes https://github.com/e-mission/e-mission-docs/issues/955#issuecomment-1688470738

We use a Babel plugin `angularjs-annotate` to fix our bad, implicitly annotated, Angular code. This way, it still works when bundled. But we only have Babel apply this to .js files.
I moved a little bit of Angular code to .ts (the "Logger" service). I don't expect we will usually put Angular code in .ts files, so I'm just going to explicitly type it in the one place we do.

### DIFF
--- a/www/js/plugin/logger.ts
+++ b/www/js/plugin/logger.ts
@@ -2,7 +2,8 @@ import angular from 'angular';
 
 angular.module('emission.plugin.logger', [])
 
-.factory('Logger', function($window, $ionicPopup) {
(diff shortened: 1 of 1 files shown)


shankari (commented):
@JGreenlee I assume you tested this with a prod webpack build and it worked?
Can you put that into the issue?

JGreenlee:
Yes, I tested it in the devapp with a prod webpack bundle.

That setup is where I observed the error before the change, and no error after

PR #1021, opened 2023-08-23
🚑 🐛 Staging dashboard error fix

Abby-Wheelis:
This PR aims to fix the errors popping up on the dashboard on staging right now, which are also breaking the dashboard. You can see more about the investigation that led to this fix in the thread starting [here](https://github.com/e-mission/e-mission-docs/issues/947#issuecomment-1689015188). 

I've now tested this fix on all 4 of my opCodes (including my personal nrel commute) and have not seen errors on any of them. I tested on NREL commute since the dashboard is working on my personal phone, and I wanted to make sure that the values were correct - the carbon and distances match exactly, calories are different, but that is because I have userData on my phone but not on the emulator.


shankari:
I think that this deserves an 🚑 !

shankari (approved):

PR #1022, opened 2023-08-23
Update Scripts: download settings controls during setup, add serve-prod

JGreenlee:
For native building, we will now call `download_settings_controls.js` during the 'native setup' script.

For devapp serving, we will add a new script (`serve-prod`) allowing us to test production JS (which is bundled+minified) on the devapp.

### DIFF
--- a/package.cordovabuild.json
+++ b/package.cordovabuild.json
@@ -8,6 +8,7 @@
     "url": "git+https://github.com/e-mission/e-mission-phone.git"
   },
   "scripts": {
+    "setup-native": "./bin/download_settings_controls.js",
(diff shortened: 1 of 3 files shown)


package.cordovabuild.json
```diff
@@ -8,6 +8,7 @@
     "url": "git+https://github.com/e-mission/e-mission-phone.git"
   },
   "scripts": {
+    "setup-native": "./bin/download_settings_controls.js",
```
shankari, line 11: note to self, this needs to be in `package.json` because it is a javascript hook. It is called from `setup/setup_shared_native.sh`

shankari (approved):

PR #1023, opened 2023-08-29
🏷️ Label UX: show labeled modes on map, update LabelOptions

JGreenlee:
This PR applies what was discussed in https://github.com/e-mission/e-mission-docs/issues/952:

- labelOptions will have a baseMode field that maps them onto a mode we recognize and have an icon+color for
- Once a trip is labeled, the mode shown on the map will the labeled mode; no longer the sensed or inferred mode
- Unlabeled trips will still show the sensed or inferred mode(s)
  -- If the trip is unimodal, there is no need to show percentages

Also changed in labelOptions, the rename of `co2PerMeter` -> `kgCo2PerKm`, in accordance with https://github.com/e-mission/nrel-openpath-deploy-configs/pull/30

In general, the code for the Label tab has been tidied, especially with regard to fixing typings.
Everything that Typescript and the VSCode linter picked up on has now been addressed.

Simulator Screen Recording - iPhone 12 - 2023-08-29 at 16 00 00


JGreenlee:
TODO: after a trip is labeled, we should still be able to access the sensed/ inferred modes in the details screen

JGreenlee:
> TODO: after a trip is labeled, we should still be able to access the sensed/ inferred modes in the details screen

I have made much progress on this. I reworked the label details screen to include a toggle between "Labeled Mode" and "Detected Modes" on trips that have the mode labeled. Unlabeled trips only show "Detected Modes".
<table><td>

</td><td>

</td><td>

</td></table>

JGreenlee:
For this to make sense, I reordered the Details screen altogether. See it in this demo:

Simulator Screen Recording - iPhone 12 - 2023-08-30 at 18 17 26

To me, this was the most logical way to frame it and make sense of the toggling. Open to feedback.

JGreenlee:
Ready for review and should be merged at the same time as https://github.com/e-mission/nrel-openpath-deploy-configs/pull/32 should be merged too.

www/js/diary/timelineHelper.ts
```diff
@@ -113,3 +152,56 @@ export function getAllUnprocessedInputs(pipelineRange, labelsFactory, notesFacto
...
+  if (!trip.sections) {
+    // this is a unimodal trip so we put all the locations in one section
+    sectionsPoints = [locationList];
+  } else {
+    // this is a multimodal trip so we sort the locations into sections by timestamp
+    sectionsPoints = trip.sections.map((s) =>
+      trip.locations.filter((l) =>
+        l.ts >= s.start_ts && l.ts <= s.end_ts
+      )
+    );
```
shankari, line 189: I'm confused. I see from the demo that when there is a user label, we essentially display the trip as unimodal.
However, in that case, we must be setting the `trip.sections` to `[]` somewhere so that we don't just fall through and execute the multi-modal trip use case. But I don't see where that is.

www/js/diary/services.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 import angular from 'angular';
-import { getFormattedTimeRange, motionTypeOf } from './diaryHelper';
+import { getBaseModeByKey, getBaseModeOfLabeledTrip } from './diaryHelper';
```
shankari, line 4: nit: presumably you don't need this any more?

www/js/survey/enketo/AddedNotesList.tsx
```diff
@@ -33,7 +35,7 @@ const AddedNotesList = ({ timelineEntry, additionEntries }: Props) => {
     if (isMultiDay(beginTs, stopTs)) {
       const beginTsZoned = moment.parseZone(beginTs*1000).tz(timezone);
       const stopTsZoned = moment.parseZone(stopTs*1000).tz(timezone);
-      d = getFormattedDateAbbr(beginTsZoned.unix(), stopTsZoned.unix());
+      d = getFormattedDateAbbr(beginTsZoned.toISOString(), stopTsZoned.toISOString());
```
shankari, line 38: Note to self: this was changed in 7eedb92ce96580238ca4444328560275ed4f84b1 because 

"--getFormattedDateAbbr actually takes an ISO string, not a unix timestamp. This was changed some time ago but not updated here (thanks typescript)"

www/js/diary/list/DateSelect.tsx
```diff
@@ -71,20 +71,20 @@ const DateSelect = ({ tsRange, loadSpecificWeekFn }) => {
       <Text>{dateRangeEnd}</Text>
     </NavBarButton>
     <DatePickerModal locale={i18next.resolvedLanguage || 'en'}
-      accessibilityLabel="Select a date"
```
shankari, line 74: Note to self, this is also handled in the cleanup commit
eedb92ce96580238ca4444328560275ed4f84b1

due to
> --DatePickerModal doesn't actually support the accessibilityLabel prop. However, looking through the source code revealed that RN Paper Dates does give much consideration to accessibility. So I think this is handled sufficiently, internally in the library.

www/js/diary/diaryHelper.ts
```diff
@@ -1,54 +1,74 @@
...
-import i18next from "i18next";
 import moment from "moment";
 import { DateTime } from "luxon";
 
 const modeColors = {
+  pink: '#d43678',        // oklch(59% 0.2 0)       // e-car
   red: '#b9003d',         // oklch(50% 0.37 15)     // car
   orange: '#b25200',      // oklch(55% 0.37 50)     // air, hsr
-  green: '#007e46',       // oklch(52% 0.37 155)    // bike
+  green: '#007e46',       // oklch(52% 0.37 155)    // bike, e-biek, moped
```
shankari, line 11: nit: e-bike

shankari (changes requested):
I am done with most of the review, except for ~ 10 of the more complex files.
Will finish review tomorrow morning and get this onto staging.

www/js/diary/details/OverallTripDescriptives.tsx
shankari, line 1: should this be in components? Do we plan to reuse it anywhere else?

shankari (changes requested):
This is the full review. I don't see anything that cannot be cleaned up later, so I am pulling this and getting it to staging.

www/js/diary/LabelTab.tsx
```diff
@@ -11,15 +11,15 @@ import { angularize, getAngularService } from "../angular-react-helper";
 import useAppConfig from "../useAppConfig";
 import { useTranslation } from "react-i18next";
 import { invalidateMaps } from "../components/LeafletView";
-import Bottleneck from "bottleneck";
```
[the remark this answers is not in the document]
JGreenlee: `Bottleneck` is still used in `addressNamesHelper.ts` but unused in `LabelTab.tsx`

www/js/diary/timelineHelper.ts
JGreenlee: `trip.sections` is never modified because we want to treat the trip object as immutable.

If a trip has been labeled with a mode, (ie `trip.userInput.MODE.value` exists), we keep `trip.sections` as it is and just don't use it.
That discrimination happens a few places including `ModesIndicator` and `TripSectionsDescriptives`

www/js/diary/services.js
JGreenlee: Yeah, none of these external functions are needed anymore

There are only a few functions left in this file though - basically just `readAllCompositeTrips` and `readUnprocessedTrips` and the smaller functions they depend on.

I think we can incorporate those into `timelineHelper.ts` on the next pass and `services.js` will no longer be needed

www/js/diary/details/OverallTripDescriptives.tsx
JGreenlee: I can't think of anywhere we would currently show details about a trip other than the Label Tab > Details screen

www/js/diary/diaryHelper.ts
```diff
@@ -1,54 +1,74 @@
...
-  CAR: { name: "CAR", icon: "car", color: modeColors.red },
   AIR_OR_HSR: { name: "AIR_OR_HSR", icon: "airplane", color: modeColors.orange },
-  // based on OSM routes/tags:
+  // END MotionTypes
+  CAR: { name: "CAR", icon: "car", color: modeColors.red },
+  E_CAR: { name: "E_CAR", icon: "car-electric", color: modeColors.pink },
+  E_BIKE: { name: "E_BIKE", icon: "bicycle-electric", color: modeColors.green },
+  E_SCOOTER: { name: "E_SCOOTER", icon: "scooter-electric", color: modeColors.periwinkle },
+  MOPED: { name: "MOPED", icon: "moped", color: modeColors.green },
+  TAXI: { name: "TAXI", icon: "taxi", color: modeColors.red },
```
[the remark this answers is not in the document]
JGreenlee: I thought we decided if a custom mode is defined in `labelOptions`, they would have to point to map it to one of these base modes?

PR #1026, opened 2023-09-08
🐛 Label Screen Fixes, Sept 2023

JGreenlee:
Fixes the following bugs:

- https://github.com/e-mission/e-mission-docs/issues/971
- https://github.com/e-mission/e-mission-docs/issues/969
- https://github.com/e-mission/e-mission-docs/issues/966
- https://github.com/e-mission/e-mission-docs/issues/968


JGreenlee:
TODO
- https://github.com/e-mission/e-mission-docs/issues/966
- https://github.com/e-mission/e-mission-docs/issues/968

JGreenlee:
It took a bit more work to get the details page to accommodate draft trips correctly.

<table><td>




</td><td>



</td><td>



</td></table>

Summary of the achieved behavior:
- All draft trips appear with a greyed-out color palette.
- Unlabeled draft trips show "UNPROCESSED" as their detected base mode (grey question mark)
- Labeled draft display their labeled mode and can be toggled back to the detected mode / "UNPROCESSED"
- Unlabeled draft trips will not show an empty "Detected: " in the ModesIndicator above the map
- Labeled draft trips _will_ show their labeled mode in the ModesIndicator

shankari:
Now that I have merged the profile changes, there is a merge conflict here.
Let's see if it is easy to resolve.

www/js/diary/cards/ModesIndicator.tsx
```diff
@@ -31,7 +31,7 @@ const ModesIndicator = ({ trip, detectedModes, }) => {
         </Text>
       </View>
     );
-  } else if (detectedModes?.length > 1 || detectedModes[0]?.mode != 'UNKNOWN') {
```
shankari, line 34: Ooops! I guess I didn't check it super carefully during the review.

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -55,7 +55,7 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
     if (isOther) {
       /* Let's make the value for user entered inputs look consistent with our other values
        (i.e. lowercase, and with underscores instead of spaces) */
-      chosenLabel = chosenLabel.toLowerCase().replace(" ", "_");
+       chosenLabel = readableLabelToKey(chosenLabel);
```
shankari, line 58: This is fine here, but a quick note for https://github.com/e-mission/e-mission-phone/commit/c480d67217eb18cf7c9fe24595a3bec04a9dc69e we should only use this for "other" labels. For the labels that have partner-provided readable labels, we should use them instead - e.g. `drove_alone` should become `Gas Car Drove Alone`, not `Drove Alone`

JGreenlee: I see. It will need to be changed on the new Dashboard, then

PR #1025, opened 2023-09-06
API Migration Automation script

niccolopaganini:
1. Complete Redesign/ Structure of script
- Structured code redesign
- creates multiple csvs -> unique_packages.csv & methods.csv (is something I am planning to retain as that seems to be useful as a source of reference)
- Pending tasks: - compare unique_packages.csv & methods.csv to find the packages (and their respective methods) - iterate the above - "simplify" them into a single python script - delete the unnecessary CSVs

2. updated YML file

PR #1031, opened 2023-09-13
API Migration Automation script - patch 2

niccolopaganini:
1. Converted the .ipynb file to python scripts (12 in total; each serving their own purpose; explained in README.md)
2. Added a shell script which executes the aforementioned python scripts
3. README.md file included which explaining the whole process
4. Screenshots directory included for README
5. Archive folder containing the .ipynb file (with updated code) and YML file for reference (included if in-case for future reference)

### DIFF
--- a/setup/export_shared_dep_versions.sh
+++ b/setup/export_shared_dep_versions.sh
@@ -11,4 +11,4 @@ export GRADLE_VERSION=7.6
 export OSX_EXP_VERSION=12
 
 export NVM_DIR="$HOME/.nvm"
-export RUBY_PATH=$HOME/.gem/ruby/$RUBY_VERSION.0/bin
+export RUBY_PATH=$HOME/.local/share/gem/ruby/$RUBY_VERSION.0/bin


niccolopaganini:
- Closed #1025 and reopened it here as I made multiple changes and wanted a separate PR for this patch. 
- Issue tracked in https://github.com/e-mission/e-mission-docs/issues/963 -> I will be compiling all the errors faced here (which will be done before 9/14/2023 9AM)

niccolopaganini:
> * Closed [API Migration Automation script #1025](https://github.com/e-mission/e-mission-phone/pull/1025) and reopened it here as I made multiple changes and wanted a separate PR for this patch.
> * Issue tracked in [Script to narrow down on changes (methods) that's considered for API migration e-mission-docs#963](https://github.com/e-mission/e-mission-docs/issues/963) -> I will be compiling all the errors faced here (which will be done before 9/14/2023 9AM)

Mentioned in issue - Also mentioning here for reference: Spotty internet connection (I couldn't connect my work computer to campus wifi); I will try to update the error logs sometime tonight

bin/API_migration_scripts/screenshots/CSVs.png
shankari, line 1: If something can be represented as text, it should be, to make it easier to search, copy/paste etc
In this case, this can be replaced by 

```
$ ls -al 
...
```

bin/API_migration_scripts/screenshots/csv order.png
shankari, line 1: Ditto

shankari (changes requested):
This is a lot better than the previous notebook, but still needs some cleanup.
High level comments below.

niccolopaganini:
1. Addressed issues regarding fragmentation (changing from 12 scripts to 4 "head" python scripts) which allows the modularity without being fragmented 
2. Paths have been changed top be relative 
3. API changes link previously had to be changed by the user. Now it can be inputted by the user.
4. Pending: README needs to be updated & setup script needs to be created.

niccolopaganini:
> 4\. README needs to be updated ✅

shankari:
> Addressed issues regarding fragmentation (changing from 12 scripts to 4 "head" python scripts) which allows the modularity without being fragmented

What is your rationale for even needing 4 scripts? Why not just combine into one script?

niccolopaganini:
Not sure what happened but when I deleted the repo on GitHub, the local dir was also deleted. I restored it from a recent HEAD and had to work on it afterward. I have now (hopefully) addressed all the comments. With this push, 
1. I have boiled it down to one python script 
2. Parallelized the process so now instead of taking 4 minutes to run, it takes 28 seconds on average to execute everything
3. Two bash files
     - One to execute everything 
     - One to remove the rest once the work is done
4. Updated README

README.md
shankari, line 1: why is this file included here?

bin/API_migration_scripts/API_changes_identifier.py
```diff
@@ -0,0 +1,248 @@
...
+import pandas as pd
+
+#for classes CSV
+def get_links(url):
+    response = requests.get(url)
+    soup = bs4.BeautifulSoup(response.content, "html.parser")
+
+    links = []
+    for a in soup.find_all("a"):
+        if a.has_attr("href") and a["href"].startswith("/sdk/api_diff/33/changes/"):
```
shankari, line 18: this is still hardcoded

bin/API_migration_scripts/API_mcf.sh
```diff
@@ -0,0 +1,45 @@
...
+
+sleep 4
+
+if [ -e classes.csv ]; then 
+    rm classes.csv
+fi
+
+if [ -e modified_csv.csv ]; then 
+    rm modified_csv.csv
+fi
```
shankari, line 15: why do we have to write csv files any more given that all the functions are in the same file? We can just pass the dataframes around and avoid the I/O overhead.

bin/API_migration_scripts/API_changes_identifier.py
```diff
@@ -0,0 +1,248 @@
+#API changes identifier
+import os
+import csv
```
shankari, line 3: I had this an earlier comment. Why do we need `csv` instead just using `pandas`?
https://github.com/e-mission/e-mission-phone/pull/1031#discussion_r1327899392

shankari: Also, where is the `environment.yml` with the new packages needed by this code?

bin/API_migration_scripts/API_changes_identifier.py
```diff
@@ -0,0 +1,248 @@
...
+if __name__ == "__main__":
+    main()
+try:
+    # Load unique packages into a set
+    unique_packages = set()
+    with open("unique_packages.csv", "r") as unique_file:
+        reader = csv.reader(unique_file)
+        for row in reader:
+            if row:
+                unique_packages.add(row[0].strip())  # Remove leading/trailing whitespace
```
shankari, line 214:
```suggestion
    output_df = match_csv_java(....)
    unique_packages = get_unique_packages(output_df)
```

bin/API_migration_scripts/API_changes_identifier.py
```diff
@@ -0,0 +1,248 @@
...
+    with open(csv_file, "r") as f:
+        reader = csv.reader(f)
+        for row in reader:
+            if row:
+                packages.add(row[0])
+
+    with open("unique_packages.csv", "w") as f:
+        csvwriter = csv.writer(f)
+        for package in packages:
+            csvwriter.writerow([package])
```
shankari, line 121:
```suggestion
return packages
```

bin/API_migration_scripts/API_changes_identifier.py
```diff
@@ -0,0 +1,248 @@
...
+    # Scrape changed content and save it
+    scraped_content_file = "changed_scraped.csv"
+    with open(scraped_content_file, "w", newline='') as f:
+        csvwriter = csv.writer(f)
+        csvwriter.writerow(['Method Name', 'Method URL', 'Description'])
+    
+    # parallelization process 
+    with concurrent.futures.ThreadPoolExecutor() as executor:
+        results = executor.map(get_changed_content, links)
+
```
shankari, line 177: Good job on experimenting with python concurrency.

Although from a design perspective, this is a bit of overkill for what should be a super simple script.
Four minutes is not that long for a script that is run once every year, and concurrency can add complexity that makes it harder to maintain simple scripts.

However, this appears to be embarrassingly parallel, so the complexity seems manageable for now.
So I will not request a change for now, but will just note that simplicity is also a virtue, especially for simple scripts that are not part of the production system.

shankari (changes requested):

PR #1032, opened 2023-09-18
👋 New Onboarding Flow + Switch to React Navigation

JGreenlee:
This PR unifies the 3 tabs we have migrated (Label, Dashboard, Profile) under a central component in React (the `App` component).

It utilizes `BottomNavigation` from React Native Paper, described here:
- https://github.com/e-mission/e-mission-phone/pull/1018#issuecomment-1712649328

This PR is also where we will rewrite the onboarding flow, since it will use React navigation. New flow discussed here:
- https://github.com/e-mission/e-mission-docs/issues/956

When this PR is finished, we should no longer have any reliance on Ionic.


JGreenlee:
This is what the new tabs looks like:

JGreenlee:
I have discovered that `dynamic_config.js` is fairly entangled with the Angular router, which makes it difficult to migrate the onboarding flow while keeping the Angular service as-is.

I think the most efficient thing to do here is just rewrite `dynamic_config` right now - it makes little sense to modify it in-place to work with the new routing, when we will just have to rewrite it again in a few weeks.

JGreenlee:
While removing all the old 'intro' code, I pruned a ton of other unused code too, in hopes that it will make it easier to clean up and rewrite all the Angular services left scattered around.

However, I'm going to stop here because this PR is already big and it will get harder to review the more I do.
I think once QR scanning is done, we should consider this ready for review.

Abby-Wheelis:
I'm working on wrapping up QR scanning now, but haven't been able to get to the onboarding flow to test. When I log out I just end up stuck on a loading label screen with a permissions modal with no permissions. I've tried setting a breakpoint in the onboarding stack if/else ladder, but I never hit the breakpoint.

JGreenlee:
I am pretty sure this is due to an error in the new `dynamicConfig.ts`. I think I made a typo and called a Cordova function that doesn't exist. fix incoming

Abby-Wheelis:
^I can confirm that the fix worked for me, I was able to ensure that QR scanning is working, as well as restore error handling. Also @JGreenlee's new styling of the Welcome page and the whole onboarding flow looks awesome!

JGreenlee:
Great! I can confirm that QR scanning is working in my testing too. Marked as ready!

JGreenlee:
I can't believe we got this done in less than 1 week! Things are really coming together as we approach the end of the UI rewrite.

I believe #1034 can be closed because it was absorbed into this PR.

shankari:
> I can't believe we got this done in less than 1 week! Things are really coming together as we approach the end of the UI rewrite.

Yes, you are definitely on a roll here! It took a lot of initial work, but I think that we can see the light at the end of the tunnel (and I don't think it is a train 😄 )

JGreenlee:
Because of the upstream fixes from earlier today (https://github.com/e-mission/e-mission-phone/pull/1041) we had some tricky merge conflicts here.
I carefully resolved these to make sure that this branch benefits from those fixes too, and then verified that this branch is in good, operational status.

www/templates/intro/reconsent.html
shankari, line 1: just for my edification, we have removed the reconsent logic completely, correct?
I remember discussing this with @Abby-Wheelis during the profile screen rewrite, but just wanted to check since I don't see a replacement.

www/js/onboarding/OnboardingStack.tsx
```diff
@@ -0,0 +1,56 @@
...
+  } else if (pendingOnboardingState.route == 'consent') {
+    return <ConsentPage />;
+  } else if (pendingOnboardingState.route == 'save-qr') {
+    return <SaveQrPage />;
+  } else if (pendingOnboardingState.route == 'survey') {
+    return <SurveyPage />;
+  } else {
+    return 'TODO'
+  }
+}
```
shankari, line 34: This seems like a change in behavior from the previous onboarding flow. Concretely, the previous flow only tracked whether the study was joined and the intro was done. If the study was joined, and the intro was not done, we would always go to the summary page. With this, I think we will go to the page that we left off.

This seems like better UX overall, so I have no complaints. But I do know, that during testing, I have frequently killed the app and restarted during the onboarding to go back to a previous screen, since the previous onboarding flow did not have a back button either.

Does the new onboarding flow support a back button? I don't seem to see it in the styles here.

www/js/onboarding/SummaryPage.tsx
```diff
@@ -0,0 +1,34 @@
...
+  function next() {
+    setSummaryDone(true);
+    refreshOnboardingState();
+  };
+
+  // summary of the study, followed by 'next' button
+  return (<>
+    <ScrollView contentContainerStyle={{flex: 1}}>
+      <Surface style={[onboardingStyles.page, {flex:1, gap: 16}]}>
+        <StudySummary />
```
shankari, line 25: why do we need to split `StudySummary` and `SummaryPage`? I thought we were not showing the `StudySummary` in the profile page, only the privacy policy.

www/js/onboarding/WelcomePage.tsx
```diff
@@ -0,0 +1,202 @@
...
+
+  const scanCode = function() {
+    window.cordova.plugins.barcodeScanner.scan(
+      function (result) {
+        console.debug("scanned code", result);
+          if (result.format == "QR_CODE" && 
+              result.cancelled == false) {
+                let text = result.text.split("=")[1];
+                console.log("found code", text);
+              loginWithToken(text);
```
shankari, line 31: weren't we checking here to see that the QR code actually did start with `emission://`?

www/js/onboarding/OnboardingStack.tsx
shankari: Ah wait, it looks like this is purely in-memory state, so this will retain the previous behavior.
Feel free to resolve if that is true.

shankari (approved):
Given that this change involved 146 files, my code review is going to focus on high level structure and functionality and not on sanity checking code. Although, kudos to @JGreenlee and @Abby-Wheelis - the code is very clean, and I don't have any high priority fixes.

I am merging this right now to get a release out tonight for the accessibility testing. You can address the comments in a cleanup fix.

www/css/style.css
```diff
@@ -44,11 +44,6 @@ label-tab > div {
   --accent-dark: hsl(200 100% 30%);
 }
 
-body.platform-ios {
-  padding-top: calc(env(safe-area-inset-top) / 2);
-  margin: 0 10px 0 0;
-}
-
```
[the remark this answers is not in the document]
JGreenlee: Not needed because we are using [SafeAreaView](https://reactnative.dev/docs/safeareaview)

www/js/onboarding/WelcomePage.tsx
Abby-Wheelis: That was somewhere in the older version of the QR scanning flow, you're right. I think I got focused on splitting the token from the rest of the url, I can add that back in the cleanup pass. Before, there was a line with `result.text.startsWith("emission://"))`

www/templates/intro/reconsent.html
Abby-Wheelis: The last update I recall from the consent logic was that we should remove reconsent at some point, but we first needed to brainstorm a good replacement for the event that we updated some sort of terms mid-data collection. I have that task tracked in my issue for the "future profile updates" but we still need to brainstorm and decide on a replacement.

www/js/survey/enketo/answer.js
```diff
@@ -1,11 +1,9 @@
...
 import MessageFormat from 'messageformat';
+import { getConfig } from '../../config/dynamicConfig';
 
-angular.module('emission.survey.enketo.answer', [
-  'ionic',
-  'emission.config.dynamic',
-])
-.factory('EnketoSurveyAnswer', function($http, DynamicConfig,) {
+angular.module('emission.survey.enketo.answer', ['ionic'])
+.factory('EnketoSurveyAnswer', function($http) {
```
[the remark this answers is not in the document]
JGreenlee: I don't think this file actually uses `ionic` and it could be removed as a dependency here.

But unfortunately, we are not fully done with `ionic`. After this PR we don't use any of the `ion-*` HTML elements. But many of the services that haven't been rewritten do still use `$ionicPopup` and `$ionicPlatform`.

www/js/App.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  const refreshOnboardingState = () => getPendingOnboardingState().then(setPendingOnboardingState);
+  useEffect(() => { refreshOnboardingState() }, []);
+
+  useEffect(() => {
+    if (!appConfig) return;
+    setServerConnSettings(appConfig).then(() => {
+      refreshOnboardingState();
+    });
+  }, [appConfig]);
```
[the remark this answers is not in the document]
JGreenlee: It has to be called from *somewhere*. The new `serverConn.ts` does not self-initialize - it is just a collection of functions.
Whether it is in the `App` component or not, we have to listen for config changes and call `setServerConnSettings`. Would you suggest a separate file for initialization logic? Besides this and the `i18n` initialization, I am not sure what would go in there.

The reason we do not have initialization logic here for `imperialConfig` is that it was made into a React hook and is used at the component-level.
`useImperialConfig` is used in a lot of different components. `serverConn` is just set up initially and not used by components.

www/js/splash/startprefs.js
```diff
@@ -179,28 +151,6 @@ angular.module('emission.splash.startprefs', ['emission.plugin.logger',
...
 
-    startprefs.getNextState = function() {
-      return startprefs.getPendingOnboardingState().then(function(result){
-        if (result == null) {
-          if (angular.isDefined($rootScope.redirectTo)) {
-            var redirState = $rootScope.redirectTo;
-            var redirParams = $rootScope.redirectParams;
-            $rootScope.redirectTo = undefined;
-            $rootScope.redirectParams = undefined;
-            return {state: redirState, params: redirParams};
```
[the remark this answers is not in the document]
JGreenlee: Needs to be reconsidered when we rewrite `localnotify`. There's no good way to do it right now with `localnotify` still being Angular and using `$rootScope`

For right now, if there's an issue with permissions it doesn't matter whether you open the app and end up on Label or Profile, because the AppStatusModal isn't specific to Profile anymore

PR #1033, opened 2023-09-18
Add Jest for Basic Testing of Pure Functions

JGreenlee:
One example test given here - simply asserting an expected output from a function of `diaryHelper`.

This will only allow us to test simple functions, not components

### DIFF
--- a/.github/workflows/serve-install.yml
+++ b/.github/workflows/serve-install.yml
@@ -59,7 +59,7 @@ jobs:
     - name: Run Jest tests
       shell: bash -l {0}
       run: |
-        npx jest www/__tests__
+        npx jest
 
 # TODO: figure out how to check that a server started correctly
 #    - name: Try starting it
--- a/.github/workflows/serve-install.yml
+++ b/.github/workflows/serve-install.yml
@@ -56,6 +56,11 @@ jobs:
         echo "ionic version"
         npx ionic --version
 
+    - name: Run Jest tests
+      shell: bash -l {0}
+      run: |
+        npx jest www/__tests__
+
 # TODO: figure out how to check that a server started correctly
 #    - name: Try starting it
 #      run: npx run serve
--- a/jest.config.json
+++ b/jest.config.json
@@ -1,0 +1,8 @@
+{
+  "transform": {
+    "^.+\\.(ts|tsx|js|jsx)$": "ts-jest"
+  },
+  "moduleNameMapper": {
+    "^react-native$": "react-native-web"
+  }
+}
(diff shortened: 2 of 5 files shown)


Abby-Wheelis:
I just got a chance to test this out, and am working on adding a few more tests. This is my first time working with this testing framework, so it took me a moment to figure out. I discovered that the tests can be run by setting up the phone server (I needed to do a full re-setup to have the updated dependencies) and then typing "npm test" at the command line. 

@JGreenlee can you take a look at the test I commented out and see if you think it should run? It's failing for me, but I haven't figured out why.

JGreenlee:
No fault of the test. That function isn't working as intended!
Look at the body of the function and you see that `key` is never used. So the function won't work with something like `MotionTypes.WALKING`.

All the more reason that we need tests to find unexpected behaviors like this!

```ts
/**
 * @param motionName A string like "WALKING" or "MotionTypes.WALKING"
 * @returns A MotionType object containing the name, icon, and color of the motion type
 */
export function motionTypeOf(motionName: MotionTypeKey | `MotionTypes.${MotionTypeKey}`) {
  let key = ('' + motionName).toUpperCase();
  key = key.split(".").pop(); // if "MotionTypes.WALKING", then just take "WALKING"
  return MotionTypes[motionName] || MotionTypes.UNKNOWN;
}
```

JGreenlee:
Runs correctly with fix, added commit

JGreenlee:
@shankari How can we integrate this with Github Actions so the tests run automatically?

shankari:
@JGreenlee you would create a new github actions workflow (`.github/workflows`) that runs the command. I would suggest starting with the install workflow  `osx-serve-install ` from this repo and adding on a test stage, similar to the way that the `test-with-manual-install` builds on `osx-ubuntu-manual-install` in the server code.

JGreenlee:
That seems to have worked! I think the tests ran twice though (once for the `www` folder, and once for the `platforms/browser/www` folder)
I'll see if I can ignore `platforms` so actions don't take longer than they need to

.github/workflows/serve-install.yml
shankari, line 1: In the server code, we have install and test in separate workflows.
**(future fix):** I am fine with having this in the same workflow (at least initially), but then we should rename the file to indicate that it is also testing.

shankari (approved):
I am merging this for now, since the workflow name does not affect the addition of new tests, and we want people to write tests as they are migrating the code.

@JGreenlee please create a new workflow or rename this one in a future PR

PR #1034, opened 2023-09-18
Split shared components from their container

Abby-Wheelis:
To increase reusability of the privacy policy and permissions elements, we need flexibility of how the component is formatted, achieve this by having the contents component, and the container as another.

This is related to the work started in [👋 New Onboarding Flow + Switch to React Navigation](https://github.com/e-mission/e-mission-phone/pull/1032#top), and could be better suited to be pulled in to that PR rather than be it's own, but drafting the changes here.


Abby-Wheelis:
It's occurring to me that I'm losing a lot of the blame here, so I might try and do this a little differently tomorrow to preserve more of the blame history, which I think is especially important for the permissions list -- the privacy policy is really just text

Abby-Wheelis:
To recap the way that permissions are now handled, for future reference and organizing my own thoughts:

- `usePermissionsStatus` holds all the "logic" for the permissions. The checks are refreshed every time the app resumes through leverage of `useAppStateChange` and set up when the config is ready using `useAppConfig` (this likely needs to change when we're ready to use this for onboarding). I decided a custom hook would be a good fit here because the "overallStatus" needed to be shared between `PermissionsControls` and `AppStatusModal`
- `PermissionsControls` is the UI for the Title, description, explanations, permissions, and refresh/accept buttons. It draws the information from `usePermissionStatus` -- designed to be used in onboarding and in the popup. It's `onAccept` parameter takes a function to execute when the "I accept" button is pressed
- `AppStatusModal` is still the popup on broken permissions and shown when you click the "AppStatus" row in the Profile screen. The dialog contents are now `<PermissionsControls />` where the modal is closed `onAccept`
- There are two instances of `AppStatusModal` still -- one in `LabelTab` and one in `ProfileSettings`. Soon to be one instance in the `App` component which will allow us to remove "window.appStatusModalOpened"

After making these changes, I re-verified that there is exactly one permissions popup if a permissions is broken - on startup, on the label screen, and if profile has been visited.

Abby-Wheelis:
This became a part of [👋 New Onboarding Flow + Switch to React Navigation](https://github.com/e-mission/e-mission-phone/pull/1032#top), and does not need to be merged independently.

PR #1038, opened 2023-09-27
Settings to enable android native code debugging in vscode

shankari:
These should be copied to `platforms/android` and then if you open the `build.gradle`, you should be able to connect to a running process and set breakpoints etc


shankari:
Superceded by https://github.com/e-mission/e-mission-docs/pull/1091

PR #1039, opened 2023-09-27
Register all ChartJS elements as workaround

JGreenlee:
Described in https://github.com/e-mission/e-mission-docs/issues/986#issuecomment-1738074971, we have a production-specific error where the needed ChartJS elements are not recognized as being registered. Likely an incompatibility between ChartJS and Webpack.

As a workaround, we will register all of the `registerables`. Unfortunately, we will no longer benefit from tree-shaking with this approach.

### DIFF
--- a/www/js/components/Chart.tsx
+++ b/www/js/components/Chart.tsx
@@ -2,23 +2,12 @@
 import React, { useEffect, useRef, useState, useMemo } from 'react';
 import { View } from 'react-native';
 import { useTheme } from 'react-native-paper';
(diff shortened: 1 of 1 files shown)


shankari:
@JGreenlee can you confirm that this change fixes the issues - maybe with a screenshot?

JGreenlee:
This is the dashboard for that `usaid-laos` user from yesterday, loading correctly in the devapp after `npm run serve-prod`.



(Not much to look at because this user doesn't have any trips.) But before this change, this was 100% reproducibly loading as a blank page

PR #1041, opened 2023-09-28
Fixes for End of Sept RC

JGreenlee:
In AppStatusModal, we listen for the value of overallStatus and if it is false, we trigger the permissions popup.

But there is an issue where the AppStatusModal sometimes shows up even if all the checks are green. This is because it can take some time for the checks to initialize and return back their value. During this time, we want to have a sort of 'undetermined' state for overallStatus, so that we don't show the popup prematurely.

We'll do this by allowing overallStatus to be either true/false OR undefined. If there are no checks yet, overallStatus will be undefined and we won't show the popup yet. Once the checks are initialized, we'll be able to determine the status and it will return true or false.

For the individual checks themselves, they also need to be able to report an 'undetermined' state. If their statusStates are false by default, and are false before they have actually been determined, then overallStatus will also be false while the check is still ongoing. So we should remove the initial 'false' condition from the few checks that had it, allowing the initial to be 'undefined', and when we iterate through the checks, we will explicitly check if `statusState === false` instead of just checking if `!statusState`.

Result of this is that AppStatusModal only shows if any permissions are broken.

Tested by loading the app, with all of the correct permissions, 20 times both before and after the change. Before the change, 2 of the attempts resulted in the AppStatusModal incorrectly appearing. After the change, the AppStatusModal did not appear on any of the 20 attempts. After this, I disallowed the Location permission and the modal did still appear as expected.

### DIFF
--- a/www/js/control/AppStatusModal.tsx
+++ b/www/js/control/AppStatusModal.tsx
@@ -1,5 +1,5 @@
 //component to view and manage permission settings
-import React, { useState, useEffect, useMemo } from "react";
+import React, { useState, useEffect, useMemo, useRef } from "react";
 import { Modal,  useWindowDimensions, ScrollView } from "react-native";
 import { Dialog, Button, Text, useTheme } from 'react-native-paper';
 import { useTranslation } from "react-i18next";
@@ -16,8 +16,8 @@ const AppStatusModal = ({permitVis, setPermitVis}) => {
     const { appConfig, loading } = useAppConfig();
 
     const { height: windowHeight } = useWindowDimensions();
-    const [osver, setOsver] = useState(0);
-    const [platform, setPlatform] = useState<string>("");
+    const osver = useRef(0);
+    const platform = useRef<string>("");
 
     const [error, setError] = useState<string>("");
     const [errorVis, setErrorVis] = useState<boolean>(false);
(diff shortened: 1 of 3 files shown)


Abby-Wheelis:
On some occasions, the checkList was failing to be initialized, so the modal had no checks in it. Upon investigation, this turned out to be because `platform` and `osver` were "" and 0, respectively. The underlying issue is that they were held in `useState`, which is updated asynchronously.

Jack pointed out that `useRef` is more appropriate here, as the `platform` and `osver` only need to be set once, and are not anticipated to change. The access changed, as useRef variables are read and set from "var.current". (see [docs](https://react.dev/learn/referencing-values-with-refs#differences-between-refs-and-state) on the difference between `useState` and `useRef`) 

Before these changes, roughly 1/3 of the time, the `platform` and `osver`were not set yet, and so the checks were not set up as the platform was interpreted as not valid. After the change, the `platform` and `osver` are always set properly in time to call createChecklist. I watched debug statements and refreshed about 10x before and after the change. 

I also updated the `useEffect` that calls `createChecklist` to do so if `checkList` does not exist or has a length of 0, rather than the else that I had before. I had observed that the `useEffect` was called twice as `appConfig` loaded, but this is unreliable behavior and it is much better to call the create function if it has not been created yet.

www/js/control/AppStatusModal.tsx
```diff
@@ -371,16 +370,16 @@ const AppStatusModal = ({permitVis, setPermitVis}) => {
      //load when ready
      useEffect(() => {
         if (appConfig && window['device']?.platform) {
-            setPlatform(window['device'].platform.toLowerCase());
-            setOsver(window['device'].version.split(".")[0]);
+            platform.current = window['device'].platform.toLowerCase();
+            osver.current = window['device'].version.split(".")[0];
```
shankari, line 374: I have no objections around this since it works, but for my own edification: I thought that it was not OK in react to set values directly - you needed to call a `set` function instead. But here, you are setting the values directly!?

Is this OK because it is a part of `useEffect`?

shankari (approved):

www/js/control/AppStatusModal.tsx
Abby-Wheelis: It's ok because they are `useRef` variables, a change from a `useState` variable because those are updated asynchronously. Since the `platform` and `osver` are set once, and not expected to change, I switched the way they were stored (at Jack's suggestion). The docs I linked in my comments above outline a bit about the difference between `useRef` and `useState`, but my basic understanding is that we use `useState` when the variable controls something in the UI, because we need to re-render when it changes, whereas `useRef` does not trigger a re-render (this is demonstrated in some examples on that docs page). The platform and osver here are not going to change, so we don't need them to be able to trigger a re-render. The asynchronous nature of `useState` was what was causing the blank checklist issue because the platform and osver were not always set before the checks were initialized.

PR #1042, opened 2023-09-28
Rewrite angular service - customURL

jiji14:
## Rewrite `customURL.js` angular service

Related Issue : https://github.com/e-mission/e-mission-docs/issues/987

While rewriting the `customURL.js`, I noticed that `CustomURL.onLaunch` is only called once in `handleOpenURL` in `join-ctrl.js`.  I removed the angular event handlers `$broadcast` and `$on`, and instead, I directly called the `onLaunchCustomURL` within `handleOpenURL`. As a result, I no longer needed the angular event handler in `customURL.ts`.

I added test cases with `valid url` and `invalid url`.


JGreenlee (approved):
Great work! LGTM!

I do think these changes are going to conflict with my changes in https://github.com/e-mission/e-mission-phone/pull/1040 because this PR modifies `join-ctrl.js`, and `join-ctrl.js` no longer exists in my PR.

I think it's better to first get this PR merged into `service_rewrite_2023` and then I can fix conflicts on my PR. In the future I will ensure that we use a better branch management strategy.

JGreenlee:
@jiji14 Since this PR has been open for 2 weeks and didn't get merged quickly, there are some fairly complex merge conflicts here.
Would you like me to address them?

jiji14:
> @jiji14 Since this PR has been open for 2 weeks and didn't get merged quickly, there are some fairly complex merge conflicts here.
> Would you like me to address them?

Yes, please 🆘

JGreenlee:
I addressed the merge conflicts. LGTM

(as a side note, I am not sure that we actually need customURL moving forward as we no longer use links to trigger the onboarding flow. However, I think it's still useful to keep this around)

jiji14:
> I addressed the merge conflicts. LGTM
> 
> (as a side note, I am not sure that we actually need customURL moving forward as we no longer use links to trigger the onboarding flow. However, I think it's still useful to keep this around)

Looks good to me too and I agree to keep it! 👍

www/js/splash/customURL.ts
```diff
@@ -0,0 +1,22 @@
+type UrlComponents = {
+    [key : string] : string
+}
+
+type OnLaunchCustomURL = (rawUrl: string, callback: (url: string, urlComponents: UrlComponents) => void ) => void;
+
+
+export const onLaunchCustomURL: OnLaunchCustomURL = (rawUrl, handler) => {
```
shankari, line 8: I'm confused. where is this called from? I only see code deleted from `ngApp` - I don't see anything added there.

shankari (changes requested):

www/js/splash/customURL.ts
jiji14: As Jack mentioned, `onLaunchCustomURL` is no longer called after the onboarding flow changed. Nevertheless, we think it's still valuable to keep this code.

shankari:
I am merging for now, but we should remove it after ~ 1 year if we are not using it.

shankari (approved):

PR #1047, opened 2023-10-02
✓ Updated "Verify Inference" Checkmark

the-bay-kay:
Work discussed in  [Issue 979](https://github.com/e-mission/e-mission-docs/issues/979) !
- The "verify trip" button now only only appears if both Mode and Purpose have verifiable inferences
- The button has been changed for clarity (centered, larger icon, color matches the verification color)

### DIFF
--- a/www/js/survey/multilabel/MultiLabelButtonGroup.tsx
+++ b/www/js/survey/multilabel/MultiLabelButtonGroup.tsx
@@ -101,7 +101,7 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
         })}
       </View>
       {trip.verifiability === 'can-verify' && (
-        <View style={{marginTop:'1rem'}}>
+        <View style={{marginTop: 16}}>
           <IconButton icon='check-bold' mode='outlined' size={18} onPress={verifyTrip}
                        style={{width: 24, height: 24, margin: 3, backgroundColor: colors.secondaryContainer}}/>
         </View>
--- a/www/js/survey/multilabel/MultiLabelButtonGroup.tsx
+++ b/www/js/survey/multilabel/MultiLabelButtonGroup.tsx
@@ -103,7 +103,8 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
       {trip.verifiability === 'can-verify' && (
         <View style={{marginTop: 16}}>
           <IconButton icon='check-bold' mode='outlined' size={18} onPress={verifyTrip}
-                       style={{width: 24, height: 24, margin: 3, backgroundColor: colors.secondaryContainer}}/>
+                      containerColor={colors.secondaryContainer}
+                      style={{width: 24, height: 24, margin: 3, borderColor: colors.secondary}} />
         </View>
       )}
     </View>
--- a/www/js/survey/multilabel/MultiLabelButtonGroup.tsx
+++ b/www/js/survey/multilabel/MultiLabelButtonGroup.tsx
@@ -100,10 +100,10 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
           )
         })}
       </View>
-      <View>
-        <IconButton icon='check-bold' mode='outlined' size={16} onPress={verifyTrip}
+      <View style={{marginTop:'1rem'}}>
+        <IconButton icon='check-bold' mode='outlined' size={18} onPress={verifyTrip}
                     disabled={trip.verifiability != 'can-verify'}
-                    style={{width: 20, height: 20, margin: 3}}/>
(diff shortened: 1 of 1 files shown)


www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -100,11 +100,12 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
           )
         })}
       </View>
-      <View>
-        <IconButton icon='check-bold' mode='outlined' size={16} onPress={verifyTrip}
-                    disabled={trip.verifiability != 'can-verify'}
-                    style={{width: 20, height: 20, margin: 3}}/>
-      </View>
+      {trip.verifiability === 'can-verify' && (
+        <View style={{marginTop:'1rem'}}>
```
JGreenlee, line 104:
```suggestion
        <View style={{marginTop: 16}}>
```
`rem` units are great for web apps but they don't actually exist in the React Native spec. If you saw a warning from the linter in VSCode, that is why.

Using `rem` does happen to work right now because we are using React Native Web, but our eventual goal is React Native (not web). So we should try to avoid these web-specific units as we move forward.

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -100,11 +100,12 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
...
+          <IconButton icon='check-bold' mode='outlined' size={18} onPress={verifyTrip}
+                       style={{width: 24, height: 24, margin: 3, backgroundColor: colors.secondaryContainer}}/>
```
JGreenlee, line 106:
```suggestion
          <IconButton icon='check-bold' mode='outlined' size={18} onPress={verifyTrip}
                      containerColor={colors.secondaryContainer}
                      style={{width: 24, height: 24, margin: 3, borderColor: colors.secondary}} />
```
I made 2 suggestions here:

1. There actually exists a dedicated property on `IconButton` called `containerColor` that we can use instead of `style.backgroundColor`
https://callstack.github.io/react-native-paper/docs/components/IconButton/#containercolor Although setting `backgroundColor` does the exact same thing in this instance, it's probably best practice to use the property since it's there. React Native Paper is well-planned component library and there might be a good reason that the developers included a property for this.
1. I recommend coloring the border of the checkmark button to match the border of the yellow labels.

JGreenlee (changes requested):
Great work! I hace a couple minor suggestions and then this looks good to me.

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
the-bay-kay: Ah, good to know!  Interesting it let me get away with using 'rem' at all - my thought was to use the rem to help show it was offsetting the text - but since it doesn't work, I'll stick with the straight numbers!

JGreenlee:
@the-bay-kay Can you move this into "Ready for review" on https://github.com/orgs/e-mission/projects/1/views/1 ?
If I hadn't been keeping an eye on the related issue, I might have missed this. And it will not get merged unless it's in "Ready for review".

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -100,11 +100,13 @@ const MultilabelButtonGroup = ({ trip, buttonsInline=false }) => {
           )
         })}
       </View>
-      <View>
-        <IconButton icon='check-bold' mode='outlined' size={16} onPress={verifyTrip}
-                    disabled={trip.verifiability != 'can-verify'}
-                    style={{width: 20, height: 20, margin: 3}}/>
-      </View>
+      {trip.verifiability === 'can-verify' && (
```
shankari, line 103: I see that this logic is unmodified, so I am going to merge it. But I would like to understand what it does, maybe illustrated with a short video.

From the code, it looks like if either button is yellow, we can "confirm" it. So if I had one yellow label and one blank label, I could press the confirm button and it would turn the yellow -> blue but leave the blank label unchanged.

Is that correct?

shankari (approved):

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
the-bay-kay: My apologies for the late reply! Here are two videos, comparing the old UI and the changes this added.

https://github.com/e-mission/e-mission-phone/assets/98350084/373773f5-4c55-43ac-94ad-0b225e26ba5a


https://github.com/e-mission/e-mission-phone/assets/98350084/952bfc8a-1e81-4dff-8975-eba55e8bf4a2

PR #1049, opened 2023-10-03
React Component Testing Setup

jiji14:
### testenvironment
To test React components, we need to change the environment from `node` to` jsdom`. Since we perform both non-UI function tests and UI function tests, it's better to change the environment in each component test file by adding a `@jest-environment` docblock at the beginning of the file.

Learn more: [Jest Test Environment Configuration](https://jestjs.io/docs/configuration#testenvironment-string)
<br>

### transformIgnorePatterns
During testing, I encountered multiple errors related to `node_modules`. This is because some react-native libraries ship uncompiled ES5 code, but Jest requires ES6 code for testing. It's important to ensure that `node_modules` doesn't have to be transpiled during testing. I resolved this issue by adding `transformIgnorePatterns`.

Learn more: [Jest Transform Ignore Patterns](https://jestjs.io/docs/configuration#transformignorepatterns-arraystring)
<br>

### Current issue with polyfills/error-guard.js
After adjusting `transformIgnorePatterns`, I resolved most of the `node_modules` errors. However, I still encountered an error related to `polyfills/error-guard.js`. This issue appears to be common, and several people have faced similar challenges. While there are suggestions available, nothing worked for my scenario.

The additional code in `polyfills/error-guard.js` is not Typescript, it is [flow](https://flow.org/) (flow is what FB uses instead of Typescript). We need to remove any flow syntax on the `error-guard polyfill` to run the test.

References:
[React Native GitHub Issue](https://github.com/facebook/react-native/issues/33426)
[Stack Overflow Question](https://stackoverflow.com/questions/71662538/syntaxerror-unexpected-token-error-guard-js)
[Jest GitHub Issue](https://github.com/jestjs/jest/issues/11236)
[Upgrade Support GitHub Issue](https://github.com/react-native-community/upgrade-support/issues/152)


JGreenlee:
It was necessary to use Babel with Jest because otherwise, Jest cannot understand the Flow code that comes from the core React and React Native packages. (Why FB doesn't transpile it themselves before publishing it, I do not know.)
The Babel presets and plugins now used are:  `@babel/preset-env`, `@babel/preset-typescript`, `@babel/preset-react`, and `@babel/plugin-transform-flow-strip-types`.

Since we are using Babel to TS transpilation, `ts-jest` is moot and can be removed.

jiji14:
I also added "jest-environment-jsdom" in configuration since "jest-environment-jsdom" is no longer shipped by default. 

Screenshot 2023-10-05 at 10 38 43 AM

Now we are ready for React Component testing! 🥳 🥳

jest.config.js
shankari, line 1: @JGreenlee this was modified as part of https://github.com/jiji14/e-mission-phone/commit/2990b7ba1e52b1a2eb7946d7c634f497c97b5a17 why did we need to rename the file?

shankari (changes requested):
@jiji14 great job taking this on and getting it to work! I am really excited to merge this and make sure that we can start testing components.

This is almost ready to review - here are a couple of high level comments:
- Can we document this some more?
        - Is this the standard option to test react components?
        - which other testing options did you consider?
        - what was the reason for picking this one?
- Can you also resolve the merge conflict so that I merge?

jest.config.js
JGreenlee: It worked with the JSON file, but most projects seems to use a JS file for Jest configuration. It's a little bit cleaner to work with and gets interpreted better by the IDE

jiji14:
> @jiji14 great job taking this on and getting it to work! I am really excited to merge this and make sure that we can start testing components.
> 
> This is almost ready to review - here are a couple of high level comments:
> 
> * Can we document this some more?
>   - Is this the standard option to test react components?
>   - which other testing options did you consider?
>   - what was the reason for picking this one?
> * Can you also resolve the merge conflict so that I merge?

Updated the related issue (https://github.com/e-mission/e-mission-docs/issues/996)

jiji14:
`jest.config.json` file was deleted, leading to a conflict error when attempting to resolve it. I tried adding the file with empty content and then removing it, but the conflict error still persists, and I find myself stuck with this issue 😢 

@shankari i, could you please delete the `jest.config.json`  file when you merge this?

I followed the steps outlined in this Stack Overflow post for reference: [Link to Stack Overflow](https://stackoverflow.com/questions/1380670/how-do-i-fix-a-merge-conflict-due-to-removal-of-a-file-in-a-branch/1380809#1380809)

jiji14:
@JGreenlee 
Thanks! How did you resolve the conflict with `jest.config.json` file?

JGreenlee:
Just to make sure there was no other weirdness going on, I resolved the conflict myself just now.
@jiji14 Below, you can see the process by which I did this. There are many ways you can work with Git, including `git` commands or Github desktop, but the built-in source control tab in VSCode is my personal preference.

image

image

image

image

jiji14:
> Just to make sure there was no other weirdness going on, I resolved the conflict myself just now. @jiji14 Below, you can see the process by which I did this. There are many ways you can work with Git, including `git` commands or Github desktop, but the built-in source control tab in VSCode is my personal preference.
> 
> image image image image

I used the command and kept getting the same issue. This helps a lot! Thanks again

shankari (approved):

PR #1050, opened 2023-10-04
🐛 Onboarding fixes 10/3

JGreenlee:
More fixes:

- Squashed bug (https://github.com/e-mission/e-mission-docs/issues/999)
- Made the summary page look nicer (from https://github.com/e-mission/e-mission-docs/issues/998)

### DIFF
--- a/www/js/onboarding/onboardingHelper.ts
+++ b/www/js/onboarding/onboardingHelper.ts
@@ -4,13 +4,13 @@ import { getConfig } from "../config/dynamicConfig";
 
 export const INTRO_DONE_KEY = 'intro_done';
 
-// state = null if onboarding is done
 // route = WELCOME if no config present
 // route = SUMMARY if config present, but not consented and summary not done
 // route = CONSENT if config present, but not consented and summary done
 // route = SAVE_QR if config present, consented, but save qr not done
 // route = SURVEY if config present, consented and save qr done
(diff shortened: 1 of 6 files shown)


www/js/onboarding/onboardingHelper.ts
```diff
@@ -27,9 +27,10 @@ export const setRegisterUserDone = (b) => registerUserDone = b;
 
 export function getPendingOnboardingState(): Promise<OnboardingState> {
   return Promise.all([getConfig(), readConsented(), readIntroDone()]).then(([config, isConsented, isIntroDone]) => {
-    if (isIntroDone) return null; // onboarding is done; no pending state
-    let route: OnboardingRoute = OnboardingRoute.NONE;
-    if (!config) {
```
shankari, line 32: just for my edification, we already had a route of `NONE`. 
It is apparently not used for anything else.
So couldn't you have just used it to distinguish between unknown and intro not done?

shankari (approved):

www/js/onboarding/onboardingHelper.ts
JGreenlee: Yes I considered that, but I reasoned that a state of `NONE` can be ambiguous to someone looking at the codebase with fresh eyes. (Is the state NONE because onboarding is over? Or because it hasn't started yet? Or we don't even know yet?)

But `DONE` is self-evident and unambiguous in its meaning. We also already use the word 'done' through this file.

PR #1052, opened 2023-10-04
🎛️ ControlHelper Rewrite

the-bay-kay:
See [Issue 994](https://github.com/e-mission/e-mission-docs/issues/994) for context!

In order to fix the 'Download JSON Dump' feature, we need to rework the ControlHelper service.  The goal is to separate this service into its own file (e.g., `controlHelper.ts`).  In the process of converting this from an angular service to a TS component, I'll be using [luxon](https://moment.github.io/luxon/api-docs/index.html) instead of [moment.js](https://momentjs.com/docs/#/-project-status/), as discussed in the issue.


the-bay-kay:
It seems that moment.js and luxon handle the `.endOf('day')` function differently - one ends on the minute (59:00), and the other the seconds (59:59).  This has been causing issues, since we need to get the exact unix number for the server call, I'm looking into how to emulate the old method.

the-bay-kay:
Update:
Using an [Epoch Converter](https://www.epochconverter.com/), I've found some odd things about the methods.
It seems moment.js's "End of day" function takes the given date, and "fast forwards" by a day, minus a second - e.g.,
Thursday, April 13, 2023 7:00:00 AM --> Friday, April 14, 2023 6:59:59 AM
(Local Time: '...' --> Thursday, April 13, 2023 11:59:59 PM GMT-07:00 DST)

Wheras luxon, oddly, is not changing the date at all:
Thursday, April 13, 2023 7:00:00 AM --> Friday, April 14, 2023 6:59:59 AM
(Local Time: '...' --> Thursday, April 13, 2023 12:00:00 AM GMT-07:00 DST)

My guess is that the issue lies within the local time - it seems Luxon uses that for the endOf() method.  I'm going to go skim through both of these libraries, and see if I can find the code itself.  If anyone has sample data of an overnight travel (e.g., spanning two days), I would really appreciate if I could borrow that for testing! The timestamps seem to be pretty finicky, so I'd like to test my code with as many types of data as possible.

the-bay-kay (commented):
In 98350084, I got the "startTime" and "endTime" values mixed.  To clarify: the startTime must be processed through the old conversion method `getUnixNum()` and the endTime is fetched by running luxon's `.toUnixInteger()` .The other way around does not work, I'll update the comment in my next commit.

the-bay-kay:
After reading up on the `requestFileSystem()` method within the `getMyData()` member function, it seems that Mozilla considers this to be a [deprecated feature](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestFileSystem).  I'm currently looking at rewriting this section with the more contemporary [Web File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API)!

the-bay-kay:
Update: After some wrangling with the androidSDK, the rewrite is confirmed to work on android!  Since the iOS version relies on the built-in app, which is not included on the simulator, further testing is needed.  As discussed in our meeting today, I'm going to see if we can utilize the "share" feature of iOS, rather than the mail app.  This would allow us to test with the simulator and have greater flexibility when sending our timelines.

(I'll update this comment later with a video of the email working - unfortunately, my usual screen recording options on Linux (Flameshot, OBS) aren't working with this fresh install, so I need to do a bit more troubleshooting before I have full videos of Android!)
image
image
image

the-bay-kay:
Update:  The code is not committed, but I've got a variation of the JSONDump written that utilizes the [Social Sharing](https://github.com/EddyVerbruggen/SocialSharing-PhoneGap-Plugin#1-description) plugin, similar to how the QR Code is shared.  Below are a few of my notes on the current status of this:
- The download is confirmed to work on iOS: I have not been able to test an email specifically, but as shown in the video below, the file downloads to 'files' successfully.
- An interesting side-effect of using the 'socialSharing' library, is that the "message" portion of the emails is also downloaded as a separate file.  I'm looking to see if this can be disabled for local downloads
- It seems that there are some issues on Android - upon my first few tests, the files downloaded returned empty. Luckily, at least "message" and "subject" lines work as intended when emailing.



https://github.com/e-mission/e-mission-phone/assets/98350084/523e0313-cbf0-4ab1-8a27-8dd13b6e6bc9

the-bay-kay:
After further digging, it seems that the [socialSharing](https://github.com/EddyVerbruggen/SocialSharing-PhoneGap-Plugin) library doesn't have a means to stop the extra "message" file from being downloaded.  Furthermore, I haven't made much progress on the android side of things.  Before continuing down this path further, I think it's worth asking for other folks' opinions! As I see it, we've got two possible ways of moving forward:
- I can continue to work with the socialSharing library, and make it worth with both Android and iOS.  If we _really_ want it working soon,  could case on the O.S., using socialSharing for iOS, and the openEmail() function for Android (this would mean controlHelper relies on two separate email plugins, which doesn't seem ideal).  Either way, this would give more flexibility for iOS sharing, with the caveat that the "direct-to-files" seems to dump an extra text file in certain contexts.
- We can stick to the current code - I still need to test on hardware, but the current version is confirmed to work with Android, and will presumably also work for iOS - however, this would mean the user still relies on the built-in mail app.

Any thoughts @shankari @JGreenlee ?  I think this boils down to a question of "Is the extra flexibility worth complicating the share process".  Would my time be better spent pivoting to other parts of the [rewrite](https://github.com/e-mission/e-mission-docs/issues/977), or is it worth changing this while I'm already updating it?

For the time being, I'll keep working on setting up socialSharing on Android!

shankari:
@the-bay-kay I don't have a strong preference wrt functionality. To my knowledge, the only people who use this button are:
- developers
- people testing out the app who don't yet have access to an environment with an admin dashboard aka people who have signed the MOU but are waiting for NREL to sign it as well

So I don't think it is a huge deal to them to configure the default mail app. But it would obviously be better if that wasn't required. So I would say that we spend a time-bounded effort on this. If you can get the android code to work in (say) one day, let us clean it up. If not, let us revisit.

the-bay-kay:
Success!! Got it working on Android.  First thing tomorrow I'll go ahead and test on a few other versions before committing and setting up the PR for review.
[Screencast from 2023-10-11 19-47-00.webm](https://github.com/e-mission/e-mission-phone/assets/98350084/b9af0bb3-81b2-4663-8d5b-5d0e109b9b47)

the-bay-kay:
A quick note on the force-push; I accidentally added code from master, and wanted to make sure this branch was as separate when merging.  I don't believe the commit would've changed much (it was just a prior update I did), but I figure better to keep them separate for now.

the-bay-kay:
Here's an email of the JSONDump working on both Android and iOS!!  I tested on a couple different versions, so this should be good to go!
(Also, for anyone else wanting to document on an XOrg Linux machine - I found vokoScreen to be the simplest and best screen recording software!)

https://github.com/e-mission/e-mission-phone/assets/98350084/66408257-23f2-4d03-be14-c10ad4b2344b

https://github.com/e-mission/e-mission-phone/assets/98350084/e7f07144-7d9a-4f37-8e9d-1349d9094d3c

JGreenlee:
Great work! Before I review this, can you change the target of this PR to some other branch, then change it back to `service_rewrite_2023`?

JGreenlee (changes requested):
The code looks clean - I just have a few substitutions I recommend.

I am unsure how, or even if, we would go about testing `controlHelper`.
I don't think we can test the email/share menu with unit tests. We might be able to write a test validating the contents of the dump file. But this would require us to mock the retrieval of entries, return some fake entries, and mock the filesystem write.
If you break `getMyData` into smaller functions you can probably validate the 'dump' without worrying about mocking filesystem.

www/js/controlHelper.ts
```diff
@@ -0,0 +1,128 @@
...
+                { type: "application/json" });
+              fileWriter.write(dataObj);
+            })
+
+          });
+        });
+      });
+    }
+
+    const emailData = function() {
```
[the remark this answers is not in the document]
the-bay-kay: I didn't think of that! Yeah, since we expanded the functionality,`shareData` would make more sense at this point.

I had the same apprehension towards writing tests as well - it seemed like a tremendous amount of work to mock and verify email inputs, for mixed returns - but, like you said, I think splitting the function is a great way to tackle it! I'll go ahead and start that now : )

the-bay-kay:
> ... If you break `getMyData` into smaller functions you can probably validate the 'dump' without worrying about mocking filesystem. Function A creates the dump and function B writes the dump to filesystem; function A calls function B. You would test only function A.

Quick update on this! I split apart the two functions in commit 92cd6fb, and they seem to (upon manual testing) be functioning fine!  My current issue, however, is with how `i18nextInit.ts` (used by`controlHelper`) interacts with Jest.  I've run into two small hiccups:

1. `i18nextInit.ts` relies on the global variable, `__DEV__`: This can be set in `jest.config.json`, or set up during the individual tests.  Is there any reason we shouldn't set this globally?  Jest will only be running in a dev environment, and I can't find a reason  why we'd want it disabled. I set it in the config file for now, but can move it into the test if we'd prefer.
2. I've run into an issue with the use of `react-native-paper-dates` .  When Jest reaches line 73 of `i18nextInit.ts`, it throws a syntax error as follows:
```
    Details:

    /Users/katier/Documents/e-mission-phone/node_modules/react-native/Libraries/Utilities/codegenNativeComponent.js:18
        paperComponentName ?  : string,
                              ^

    SyntaxError: Unexpected token ':'

      71 |
      72 | // Next, register the translations for react-native-paper-dates
    > 73 | import { en, es, fr, it, registerTranslation } from 'react-native-paper-dates';
         | ^
      74 | const rnpDatesLangs = {
      75 |   en, es, fr, it,
      76 |   lo: loJson['react-native-paper-dates'] /* Lao translations are not included in the library,
```
I still need to do further investigation as to why exactly this issue is happening!  As I discussed with Abby, some of the issues with the i18next_ utils stem from this branch being somewhat out of date, compared to the e-mission-translate repository.  Could this be part of the issue? 

The only function that we need to test is the `createWriteFile()` method, which doesn't rely on `i18nextInit.ts`.  I'd be curious to see if we could test the function in isoluation, without dealing with i18next.   That being said, I don't want to "kick the can down the road" and wait until something else breaks, so I'll go ahead and keep searching for fixes!

JGreenlee:
See @Abby-Wheelis's PR https://github.com/e-mission/e-mission-phone/pull/1063 for ideas on how to address both of these.

(1. You can define `__DEV__` as a global in the jest config and 2. You can pull in the changes from https://github.com/e-mission/e-mission-phone/pull/1049 and this will allow Jest to not get tripped up on the React code)

the-bay-kay:
For those curious, I've got a type interface written up for the objects returned by the `getRawEnteries()` function!  These objects are pretty large in the samples: the `phone_data` field will contain thousands of dataObj in any given instantiation.    This interface is also pretty large, too.  Since these are so bulky, I plan to generalize the interface a bit (add some "any"s), and then use this template to create some dummy values to  test `createWriteFile()`.

I don't remember which PR thread it was, but I remember there being talk of making some more generalized files to contain typescript interfaces - I thought it'd be useful to record this now, since implementing strict typing is a ways down the road.
```
interface dataObj {
  data: {
    name: string, 
    ts: number,
    reading: number
  },
  metadata: {
    key: string,
    platform: string, //e.g., server
    write_ts: number,
    time_zone: string, //e.x., America/Los_Angeles
    write_fmt_time: string, // 2023-04-14T01:10:01.XXX , etc.
    write_local_dt: {
      day: number,
      hour: number, 
      minute: number,
      month: number,
      second: number,
      timezone: string,
      weekday: number,
      year: number,
    }
  },
  user_id: {
    $uuid: string
  },
  _id: {
    $oid: string
  }
}

interface phoneObj {
  phone_data: Array<dataObj>
}
```

JGreenlee:
Great! You can coordinate with @jiji14 as she's been adding some types in https://github.com/e-mission/e-mission-phone/pull/1061 in a new folder `www/js/types`.

There is some overlap here - you should be able to use Jiji's `LocalDt` type at least

the-bay-kay:
I did have one final question for the review process - when setting up the types and templates, which of the two should we default to? It seems there are some [subtle differences](https://blog.logrocket.com/types-vs-interfaces-typescript/#differences-between-typescript-interfaces), and maybe it would make more sense to create 'types', unless interface functionality is needed (e.g., `FsWindow` utilizes the declaration merging).  I can go back and update `fileShareTypes.ts` if we'd like to default to type declarations!

shankari:
@the-bay-kay can you make sure to put the screenshots into tables to make them not overwhelm the PR?
See example at: https://github.com/e-mission/e-mission-phone/pull/1052#issuecomment-1756392350

Note that this doesn't work for video, so the impact may be a bit limited 😄

www/js/controlHelper.ts
```diff
@@ -0,0 +1,136 @@
...
+            logDebug(`Successfull file read with ${readResult.length} characters`);
+            const dataArray = JSON.parse(readResult);
+            logDebug(`Successfully read resultList of size ${dataArray.length}`);
+            let attachFile = fileEntry.nativeURL;
+              if (window['device'].platform === "android")
+                attachFile = "app://cache/" + fileName;
+              if (window['device'].platform === "ios")
+                alert(i18next.t("email-service.email-account-mail-app"));
+              const email = {
+                'files': [attachFile],
```
shankari, line 71: I'm surprised this works with socialsharing. IIRC, I had to create a PR to have the email plugin support the `app://` prefix. Have you tried actually emailing/saving the related file? Does it generate a non-blank file?

www/js/controlHelper.ts
```diff
@@ -0,0 +1,136 @@
...
+              }
+              window['plugins'].socialsharing.shareWithOptions(email, function (result) {
+                logDebug(`Share Completed? ${result.completed}`); // On Android, most likely returns false
+                logDebug(`Shared to app:  ${result.app}`);
+                resolve();
+              }, function (msg) {
+                logDebug(`Sharing failed with message ${msg}`);
+              });
+          }
+          reader.readAsText(file);
```
shankari, line 83: This is what is missing from the test, IIRC

www/js/controlHelper.ts
```diff
@@ -0,0 +1,136 @@
...
+    // Simulate old conversion to get correct UnixInteger for endMoment data
+    const getUnixNum = (dateData: DateTime) => {
+      const tempDate = dateData.toFormat("dd MMM yyyy");
+      return DateTime.fromFormat(tempDate, "dd MMM yyyy").toUnixInteger();
+    };
+
+    const writeDumpFile = createWriteFile(dumpFile);
+    const shareData = createShareData(dumpFile, startTimeString, endTimeString);
+
+    getRawEntries(null, getUnixNum(startTime), startTime.toUnixInteger())
```
shankari, line 119: you get a `startTs` as input into the function.
I'm not sure about why you need to convert back and forth to `startTime` and use `getUnixNum`
Also, why are both the start and end, the `startTime`?
Again, have you checked on whether you actually get a file with data?

shankari (changes requested):
@the-bay-kay this is a great start, but I am not sure that it actually works 😄 

Please indicate "Testing done" before I merge.

www/js/controlHelper.ts
the-bay-kay: Sorry for the confusion - yes, I've checked with a few different dates/timelines, on iOS and Android! I'll go ahead and do some re-tests after fixing these revisions.  

As for the reason we need `getUnixNum`, Thank you for asking!! I honestly struggled for a while to pick this problem apart.  Revisiting the issue, I'm not actually sure we need it.  To the best of my understanding, here's the problem:

- Before switching to ReactNativePaper, we used an Angular date-picker, and converted the input date to a `moment()` object (`dd MMM yyyy`).  After we switched to reactNativePaper, we started using a different time-string format (`dd-mm-yyyy`), which caused some problems (see: [issue](https://github.com/e-mission/e-mission-docs/issues/994#issuecomment-1745866273)).
- Different strings are converted to different Unix numbers differently: For example, If I have the following code...
```
    console.log(`\n\n==> Start Time : ${startTime} 
                 \nUnixNum: ${startTime.toUnixInteger()} \nConversionTime: ${getUnixNum(startTime)}`);
    console.log(`\n\n==> End Time : ${endTime} 
                 \nUnixNum: ${endTime.toUnixInteger()} \nConversionTime: ${getUnixNum(endTime)}`);
```
The resulting Unix values are different:
```
==> Start Time : 2023-04-14T23:59:59.000-07:00   
[phonegap] UnixNum: 1681541999 
[phonegap] ConversionTime: 1681455600

==> End Time : 2023-04-14T23:59:59.999-07:00 
UnixNum: 1681541999 
ConversionTime: 1681455600
```
- As the UnixNumber and Conversion Times show, using the "conversion" method simply rounds the time to the nearest hour.  I think this is what I was trying to figure out in [this](https://github.com/e-mission/e-mission-phone/pull/1052#issuecomment-1749381370) comment.  The issue may not be the date conversion, but rather, the fact that Luxon performs the "endOfDay" function differently than Moment.js. 

All of this to say - going back to the numbers, it appears I was using the old conversion as a _really_ hacky way to get the correct second precision, and just wasn't aware that was the issue.  This method works for all timelines I've tested, but now that I have more clarity on the issue, I'll go ahead and figure out a proper fix! ( I remember reading on a related issue that someone else was having problems with the moment/luxon precision - and, in that code, they used a "timezone" flag to ensure the correct "endOfDay" value.  I can't find the thread again, but I'll start by looking into this!)

the-bay-kay: Update: I was able to refactor the code so it doesn't need the 'getUnixTime' hack! I'll be adding a commit with a few of these revisions, plus the updated code, here.
As for why `startTs` and `endTs` both come from the startTs - since we can only fetch one day at at time, we only need start value, and can figure out the Unix Integers (used to fetch the data) from that single date.  That being said, the name "startTs" is a bit misleading, so I'll go ahead and rename that!

www/js/controlHelper.ts
the-bay-kay: Yes! As written, this works fine - but after a quick test, it also works without the `app://` prefix, so I'll go ahead and remove it.

www/__tests__/controlHelper.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    user_id: {
+      $uuid: '41t0l8e00s914tval1234567u9658699',
+    },
+    _id: {
+      $oid: '12341x123afe3fbf541524d8',
+    }
+  };
+
+  // The parse/stringify lets us "deep copy" the objects, to quickly populate/change the data
+  let values = Array.from({length: arraySize}, e => JSON.parse(JSON.stringify(sampleDataObj)));
```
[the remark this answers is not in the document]
the-bay-kay: From the bit of research I did, it doesn't seem that react-native has a builtin deepcopy.  At first, I thought [structuredClone()](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) could be a solution, but it seems that the HTML DOM API doesn't work with native code.  If / when we want to add cloning libraries in the future, it seems there are a few options available! (e.x., [clone](https://www.npmjs.com/package/clone) or [Lodash](https://lodash.com/))

www/__tests__/controlHelper.test.ts
```diff
@@ -0,0 +1,116 @@
...
+it('correctly writes the files', async () => {
+  const testPromise = generateFakeValues(1);
+  let dataCluster = null; 
+  testPromise.then((result) => {dataCluster = result});
+  const fileExists = confirmFileExists(fileName, dataCluster);
+  const temp = createWriteFile('badFile.test')
+
+  expect(testPromise.then(temp).then(fileExists)).resolves.toEqual(false);
+  expect(testPromise.then(writeFile).then(fileExists)).resolves.not.toThrow();
+  expect(testPromise.then(writeFile).then(fileExists)).resolves.toEqual(true);
```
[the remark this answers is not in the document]
the-bay-kay: My bad! Yeah, there were some issues with synchronization here - I'm currently rewriting these tests now.

the-bay-kay:
Some updates on where I'm at rewriting tests:
- The synchronization has been fixed; the helper test functions have been updated, and all methods resolve without timing out.
- It seems that using the window methods in Jest was trickier than I originally thought.  It isn't enough to declare Jest's `global.window = window`, as we need the extension `declare let window: FsWindow`. (Because of this, Jest cannot read constants like `window.LocalFileSystem.TEMPORARY`, or access other LocalFilesystem methods)
- I'm looking into using the `spyOn()` methods ([link](https://jestjs.io/docs/jest-object#jestspyonobject-methodname)) to mock the values of the FsWindow functions themselves.  That being said, this would require us to mock the other FsWindow methods, which somewhat defeats the purpose of testing (as the goal is to confirm whether or not we successfully wrote the files).
- Running through the codebase, I don't believe we've had tests interact with the `window` global before, and am open to any fixes!  I'll continue exploring how Jest sets globals, and see if there's any way to extend their typing.

the-bay-kay:
Another update:
Jiji and I confirmed the issue isn't with the `.test.ts` file itself, but rather that Jest does not have access to the window object or its extensions during function calls.  Below is the error, for those interested:

image


 It seems that common consensus is to mock the `window` object and its subsequent functions.  While I could mock the `window.requestFileSystem` method (and am looking into it), this method is the backbone of these functions.  If I were to mock the file system, I'm worried that I'd end up testing my own mock more than the function itself.

Another alternative is to pass the window object _itself_ into the mock functions.  Most of the blogposts and threads I've read on the issue recommend against this, as it can significantly slow down testing, and cause unexpected behavior if not properly contained.  I have yet to successfully pass the the `window` object to the mocks, but this does seem to be the most accurate method of testing, so I'll do my best to get it working.

the-bay-kay:
I've made a small bit of progress, in that I can now set the `window.LocalFileSystem.TEMPORARY` constants.  Unfortunately, that leaves us with with the same issue of `window.requestFileSystem` not being recognized.  If we attempt to pass the original function itself into the mock (I've seen this approach referred to as "injecting the dependency"), the tests throw a variety of errors, and ultimately breaks.  

Since we can't inject the `requestFileSystem` method, it seems our only option is to mock the function.  The `writeDumpFile` and `shareData` methods are basically wrappers for this method - wouldn't the tests then be checking the mock over the actual code? @shankari , I'd like to hear your thoughts on the issue : )

I'm going to continue doing a bit more reading to see if mocking is even viable - If I were to set Jest spys for every method (the requestFileSystem methods, the FileEntry objects that it utilizes,  etc), I may be able to get it running.  It seems the 'spy' method has been the most successful.  I've tried mocking the `window` object and setting its properties, passing the `window` directly to the test methods' higher-order functions.  I've read that there may be a way to avoid global entirely, by manual re-exporting the object from a  separate file.  I'll keep digging into alternatives for the time being!

#### The New Code
```
let windowSpy: jest.SpyInstance;
 
beforeEach(() => {
  windowSpy = jest.spyOn(globalThis, 'window', 'get');
});
afterEach(() => {
  windowSpy.mockRestore();
});

it('correctly writes the files', async () => {
  const testPromise = generateFakeValues(1);
  let dataCluster = null; 

  windowSpy.mockImplementation(() => ({
    LocalFileSystem: {
      TEMPORARY: 0,
      PERSISTENT: 1,
    },
    // requestFileSystem: window.requestFileSystem, // adding this breaks the test
  }));

  await testPromise.then((result) => {dataCluster = result});
  const fileExists = confirmFileExists(fileName, dataCluster);
  const callResult = await testPromise.then(writeFile);
});
```

#### Without Dependency Injection
image

#### The Error With Dependency  Injection
image

the-bay-kay:
> Since we can't inject the `requestFileSystem` method, it seems our only option is to mock the function. The `writeDumpFile` and `shareData` methods are basically wrappers for this method - wouldn't the tests then be checking the mock over the actual code? Shankari , I'd like to hear your thoughts on the issue : )

Moving this over to the "Ready For Review" section for feedback - The tests are currently broken, but I'd like some direction on how to proceed before polishing the code and getting it properly re-reviewed!

shankari:
@the-bay-kay good points:
1. @Abby-Wheelis also ran into the concept of mocking the filesystem when refactoring the upload service. And she did create a mock for it. It seems like you can re-use it for this test (https://github.com/e-mission/e-mission-phone/pull/1053/files#diff-7648996247a59dd3b370aa228db8e6ba5e47dd60edafa9bff1e48501ec609c17)
2. for the functions that are essentially wrappers for the native functions, integration testing is the solution. I believe the current plan is to use Appium for both integration testing and end-to-end testing. https://github.com/e-mission/e-mission-docs/issues/1005 Once that is done, you can/should also add tests that actually invoke the filesystem API

JGreenlee:
I can see there are some merge conflicts (not sure how bad because I can't see them).

Also, what is the status of this PR? Did we ever figure out how to test it? If not, can we get this out now and worry about that later?

www/js/services.js
```diff
@@ -119,6 +119,12 @@ angular
         },
       );
       return combinedPromise(localPromise, remotePromise, combineWithDedup);
+<<<<<<< HEAD
+    }
+})
+.factory('Chats', function() {
+  // Might use a resource here that returns a JSON array
+=======
```
JGreenlee, line 127: Is this from a bad merge conflict?

the-bay-kay:
> Also, what is the status of this PR? Did we ever figure out how to test it? If not, can we get this out now and worry about that later?

I'm pretty stumped on how to write tests for this.  The issue lies in the `fileEntry.onwriteend()` and `.onerror()` methods.  I've tried assigning them to a global `getFileWriter` object, and then calling the passed mocks from `.write(dataObj)`, but this didn't work.  

My understanding of the `fileEntry` object is that the the two "onend" methods are event listeners, but I'm unsure how to mock these.  A normal `jest.fn()` doesn't work, because we need to call the `resolve()` method that's in scope for these functions.  Normally I'd pass the resolve in as a higher order function, but none of the functions under ` window['resolveLocalFileSystemURL']` can accept a function like this. 

I'll go ahead and clean up the current PR, resolve the conflicts, and push what I have!  I've manually tested on iOS & Android, with a few emulators and OPCodes, and everything seems to run smoothly.

the-bay-kay:
In manual testing, I've run into another issue with this rewrite.  When loading the JSON files, all seems to be well when loading files without `usercache` data: however, if a given timeline file has usercache data, the timeline appears blank, and does not have any data loaded. 

If anyone has a `.timeline` file that contains usercache data, I would greatly appreciate a copy to test!  It seems that none of the files in the server's `emission/tests/data/real_examples` folder have `usercache` data.   I believe this issue occurs within the `intake-sinigle-user` phase of  loading the server data.  I generated my timeline file using this PR, using my open_survey opcode - I would be curious to see if older timelines still have this issue, that would help me narrow down the issue!

image
image

image

the-bay-kay:
Update: the UserCache data doesn't seem to be the issue.  After pre-processing the JSON dump and filtering out the userCache data, the timeline loads, with the following error.  Interestingly, the `intake-single-user` errors don't occur when using this data.  I'll spend some more time trying to hunt down the issue!
image

the-bay-kay:
> Update: the UserCache data doesn't seem to be the issue. After pre-processing the JSON dump and filtering out the userCache data, the timeline loads, with the following error. Interestingly, the `intake-single-user` errors don't occur when using this data. 


Interestingly, this issue doesn't occur when the file is loaded into the Android emulator.  Here is a side-by-side of the same timelines loaded on both iOS and android -  on iOS, the aforementioned error is thrown, and there are no detected modes.  On android, however, the file loads without any issue. Here is a copy of the timeline file ([download](https://github.com/e-mission/e-mission-phone/files/13395635/NoUsercache.txt)) - I would be curious to see if other folks are having this same android/ios discrepancy!

image

image

JGreenlee:
> If anyone has a `.timeline` file that contains usercache data, I would greatly appreciate a copy to test! It seems that none of the files in the server's `emission/tests/data/real_examples` folder have `usercache` data.

`emission/tests/data/real_examples/shankari_2016-06-20.user_inputs` has some mode and purpose labels, is that what you're looking for?

the-bay-kay:
> `emission/tests/data/real_examples/shankari_2016-06-20.user_inputs` has some mode and purpose labels, is that what you're looking for?

Not exactly!  When loading this data (using the `load_timeline_for_day_and_user.py`), I get the following output at the end:
```
Finished loading 0 entries into the usercache and 17 entries into the timeseries
```
When loading my personal data, the script shows that 9 "usercache" entries have been loaded:

```
Finished loading 9 entries into the usercache and 29231 entries into the timeseries
```
My current theory is that this `usercache` data may be the issue. I get multiple `AttributeError: 'NoneType' object has no attribute 'data' ` errors. (E.x. : [`analysis_timeseries_queries.py`](https://github.com/e-mission/e-mission-server/blob/master/emission/storage/decorations/analysis_timeseries_queries.py) throws an AttributeError on line 48, when called by line 200 of [`section_segmentation.py`](https://github.com/e-mission/e-mission-server/blob/81c4314a776eff5dee61b01f1ca16a85ee267a10/emission/analysis/intake/segmentation/section_segmentation.py#L200)).  When I go ahead and load the data, I am met by the error shown in the screenshots, and the timelines fail to load the predicted modes.

When running the `JSONDump` with timelines loaded from `emission/tests/data/real_examples/`, on iOS or Android, I haven't encountered this issue - this only occurs on my recent data, collected in the `open_access` survey. I've tested a few days from my opcode, and have gotten similar errors.  

I'm pretty stumped on this bug, so I'm putting this into review for a second set of eyes - I'm not terribly familiar with the server, so I may start by combing through the server errors and trying to understand what's missing.

image
image

the-bay-kay:
OK -- peculiarly, I loaded a timeline from Nov 24 from my live OPCode, and it seems to have loaded without any issue...? Looking into this further

the-bay-kay:
After some discussion in a meeting today, myself and others have determined this error may be similar to that documented in issue #936 . Below is a full copy of the errors found when running the intake on a faulty timeline file.
(Side Note: somewhere within this script, there is a [zen of python](https://peps.python.org/pep-0020/) somewhere in the script, which I've removed from this output.)

```shell
(emission) $ ./e-mission-py.bash bin/debug/intake_single_user.py -e nrelop_dev-emulator-program_katieBroken
storage not configured, falling back to sample, default configuration
URL not formatted, defaulting to "Stage_database"
Connecting to database URL localhost
analysis.debug.conf.json not configured, falling back to sample, default configuration
google maps key not configured, falling back to nominatim
nominatim not configured either, place decoding must happen on the client
overpass not configured, falling back to default overleaf.de
transit stops query not configured, falling back to default
[ Zen of Python Text ]
analysis.trip_model.conf.json not configured, falling back to sample, default configuration
expectations.conf.json not configured, falling back to sample, default configuration
ERROR:root:habitica not configured, game functions not supported
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/habitica/proxy.py", line 22, in <module>
    key_file = open('conf/net/ext_service/habitica.json')
FileNotFoundError: [Errno 2] No such file or directory: 'conf/net/ext_service/habitica.json'
2023-11-28T13:04:51.247143-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: moving to long term**********
New entry count == 0 and skip_if_no_new_data = False, continuing
2023-11-28T13:04:51.266319-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: updating incoming user inputs**********
2023-11-28T13:04:51.366549-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: filter accuracy if needed**********
2023-11-28T13:04:51.372261-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: segmenting into trips**********
2023-11-28T13:04:52.045295-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: segmenting into sections**********
Sectioning failed for user d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/segmentation/section_segmentation.py", line 52, in segment_current_sections
    segment_trip_into_sections(user_id, trip_entry, trip_entry.data.source)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/segmentation/section_segmentation.py", line 66, in segment_trip_into_sections
    distance_from_place = _get_distance_from_start_place_to_end(trip_entry)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/segmentation/section_segmentation.py", line 200, in _get_distance_from_start_place_to_end
    start_place = esda.get_object(esda.RAW_PLACE_KEY, start_place_id)
  File "/Users/jrischpa/Documents/e-mission-server/emission/storage/decorations/analysis_timeseries_queries.py", line 48, in get_object
    return get_entry(key, object_id).data
AttributeError: 'NoneType' object has no attribute 'data'
2023-11-28T13:04:52.059446-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: smoothing sections**********
Marking smoothing as failed
Traceback (most recent call last):
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3802, in get_loc
    return self._engine.get_loc(casted_key)
  File "pandas/_libs/index.pyx", line 138, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/index.pyx", line 165, in pandas._libs.index.IndexEngine.get_loc
  File "pandas/_libs/hashtable_class_helper.pxi", line 5745, in pandas._libs.hashtable.PyObjectHashTable.get_item
  File "pandas/_libs/hashtable_class_helper.pxi", line 5753, in pandas._libs.hashtable.PyObjectHashTable.get_item
KeyError: 'filter'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/location_smoothing.py", line 116, in filter_current_sections
    filter_jumps(user_id, section.get_id())
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/location_smoothing.py", line 142, in filter_jumps
    is_ios = section_points_df["filter"].dropna().unique().tolist() == ["distance"]
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/frame.py", line 3807, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 3804, in get_loc
    raise KeyError(key) from err
KeyError: 'filter'
2023-11-28T13:04:52.077582-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: cleaning and resampling timeline**********
Cleaning and resampling failed for user d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", line 88, in clean_and_resample
    last_raw_place = save_cleaned_segments_for_ts(user_id, time_query.startTs, time_query.endTs)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", line 114, in save_cleaned_segments_for_ts
    return save_cleaned_segments_for_timeline(user_id, tl)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", line 140, in save_cleaned_segments_for_timeline
    (last_cleaned_place, filtered_tl) = create_and_link_timeline(tl, user_id, trip_map)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", line 951, in create_and_link_timeline
    curr_cleaned_start_place = get_filtered_place(tl.first_place())
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/clean_and_resample.py", line 246, in get_filtered_place
    _copy_non_excluded(old_data=raw_place.data,
AttributeError: 'NoneType' object has no attribute 'data'
2023-11-28T13:04:52.628696-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: inferring transportation mode**********
Error while inferring modes, timestamp is unchanged
Traceback (most recent call last):
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 714, in urlopen
    httplib_response = self._make_request(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 403, in _make_request
    self._validate_conn(conn)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 1053, in _validate_conn
    conn.connect()
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connection.py", line 419, in connect
    self.sock = ssl_wrap_socket(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 449, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 493, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/ssl.py", line 501, in wrap_socket
    return self.sslsocket_class._create(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/ssl.py", line 1074, in _create
    self.do_handshake()
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/ssl.py", line 1343, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 798, in urlopen
    retries = retries.increment(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/util/retry.py", line 592, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='lz4.overpass-api.de', port=443): Max retries exceeded with url: /api/interpreter (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)')))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 24, in predict_mode
    mip.runPredictionPipeline(user_id, time_query)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 55, in runPredictionPipeline
    self.predictedProb = self.predictModesStep()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 74, in predictModesStep
    predictedProb.append(get_prediction(i, section_entry))
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 113, in get_prediction
    return get_motorized_prediction(i, section_entry)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 140, in get_motorized_prediction
    predicted_transit_mode = _get_transit_prediction(i, section_entry)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 175, in _get_transit_prediction
    start_transit_stops = enetm.get_stops_near(section_entry.data.start_loc, start_radius)
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/transit_matching/match_stops.py", line 124, in get_stops_near
    stops = get_public_transit_stops(lat - bbox_delta, lon - bbox_delta, lat + bbox_delta, lon + bbox_delta)
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/transit_matching/match_stops.py", line 81, in get_public_transit_stops
    call_return = make_request_and_catch(overpass_query)
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/transit_matching/match_stops.py", line 30, in make_request_and_catch
    response = requests.post(url + "api/interpreter", data=overpass_query)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/api.py", line 115, in post
    return request("post", url, data=data, json=json, **kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/adapters.py", line 563, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='lz4.overpass-api.de', port=443): Max retries exceeded with url: /api/interpreter (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)')))
2023-11-28T13:04:53.186132-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: inferring labels**********
2023-11-28T13:04:53.196621-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: populating expectations**********
2023-11-28T13:04:53.205540-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: creating confirmed objects **********
Error while creating confirmed objects, timestamp is unchanged
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/userinput/matcher.py", line 96, in create_confirmed_objects
    confirmed_tl = create_and_link_timeline(ts, timeline, last_confirmed_place)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/userinput/matcher.py", line 124, in create_and_link_timeline
    curr_confirmed_start_place = create_confirmed_entry(ts,
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/userinput/matcher.py", line 226, in create_confirmed_entry
    confirmed_object_data = copy.copy(tce["data"])
TypeError: 'NoneType' object is not subscriptable
2023-11-28T13:04:53.226488-08:00**********UUID d1f18eb8-8aed-4bcc-8b34-b6b589c78b1c: creating composite objects **********
```

the-bay-kay:
For reference, here's the output from running the intake pipeline on another JSON dump from my open survey, one that loads without errors.
```shell
(emission) $ ./e-mission-py.bash bin/debug/intake_single_user.py -e nrelop_dev-emulator-program_katieWorking
storage not configured, falling back to sample, default configuration
URL not formatted, defaulting to "Stage_database"
Connecting to database URL localhost
analysis.debug.conf.json not configured, falling back to sample, default configuration
google maps key not configured, falling back to nominatim
nominatim not configured either, place decoding must happen on the client
overpass not configured, falling back to default overleaf.de
transit stops query not configured, falling back to default
[ Zen of Python ]
analysis.trip_model.conf.json not configured, falling back to sample, default configuration
expectations.conf.json not configured, falling back to sample, default configuration
ERROR:root:habitica not configured, game functions not supported
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/habitica/proxy.py", line 22, in <module>
    key_file = open('conf/net/ext_service/habitica.json')
FileNotFoundError: [Errno 2] No such file or directory: 'conf/net/ext_service/habitica.json'
2023-11-28T13:43:48.249504-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: moving to long term**********
Got error None while saving entry AttrDict({'_id': ObjectId('65665f072bec2a55ca5f625c'), 'metadata': {'time_zone': 'America/Los_Angeles', 'plugin': 'none', 'write_ts': 1700876988.3018951, 'platform': 'ios', 'read_ts': 0, 'key': 'statemachine/transition', 'type': 'message'}, 'data': {'currState': 'STATE_ONGOING_TRIP', 'transition': None, 'ts': 1700876988.301783}, 'user_id': UUID('eada86fb-40ec-4cd1-8ca8-7862db147718')}) -> None
New entry count == 4991 and skip_if_no_new_data = False, continuing
2023-11-28T13:43:53.339275-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: updating incoming user inputs**********
2023-11-28T13:43:53.360571-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: filter accuracy if needed**********
2023-11-28T13:43:53.365240-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: segmenting into trips**********
2023-11-28T13:43:55.445780-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: segmenting into sections**********
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/segmentation/section_segmentation_methods/flip_flop_detection.py:59: RuntimeWarning: invalid value encountered in scalar divide
  sm.update({"trip_pct": (curr_section_time * 100)/total_trip_time})
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/segmentation/section_segmentation_methods/flip_flop_detection.py:59: RuntimeWarning: invalid value encountered in scalar divide
  sm.update({"trip_pct": (curr_section_time * 100)/total_trip_time})
2023-11-28T13:43:56.780488-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: smoothing sections**********
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
Caught error index -1 is out of bounds for axis 0 with size 0 while processing section, skipping...
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/location_smoothing.py", line 195, in get_points_to_filter
    filtering_algo.filter(with_speeds_df)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 240, in filter
    self.find_segments()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 127, in find_segments
    segmentation_points = self.get_segmentation_points_ios()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 155, in get_segmentation_points_ios
    jump_to = self.with_speeds_df[(self.with_speeds_df.index < jump) & (
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 5320, in __getitem__
    return getitem(key)
IndexError: index -1 is out of bounds for axis 0 with size 0
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
Caught error index -1 is out of bounds for axis 0 with size 0 while processing section, skipping...
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/location_smoothing.py", line 195, in get_points_to_filter
    filtering_algo.filter(with_speeds_df)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 240, in filter
    self.find_segments()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 127, in find_segments
    segmentation_points = self.get_segmentation_points_ios()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 155, in get_segmentation_points_ios
    jump_to = self.with_speeds_df[(self.with_speeds_df.index < jump) & (
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 5320, in __getitem__
    return getitem(key)
IndexError: index -1 is out of bounds for axis 0 with size 0
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
Caught error index -1 is out of bounds for axis 0 with size 0 while processing section, skipping...
Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/location_smoothing.py", line 195, in get_points_to_filter
    filtering_algo.filter(with_speeds_df)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 240, in filter
    self.find_segments()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 127, in find_segments
    segmentation_points = self.get_segmentation_points_ios()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/jump_smoothing.py", line 155, in get_segmentation_points_ios
    jump_to = self.with_speeds_df[(self.with_speeds_df.index < jump) & (
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/pandas/core/indexes/base.py", line 5320, in __getitem__
    return getitem(key)
IndexError: index -1 is out of bounds for axis 0 with size 0
2023-11-28T13:43:57.438092-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: cleaning and resampling timeline**********
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
/Users/jrischpa/Documents/e-mission-server/emission/analysis/intake/cleaning/cleaning_methods/speed_outlier_detection.py:27: FutureWarning: The default value of numeric_only in DataFrame.quantile is deprecated. In a future version, it will default to False. Select only valid columns or specify the value of numeric_only to silence this warning.
  quartile_vals = df_to_use.quantile([0.25, 0.75]).speed
2023-11-28T13:43:58.963839-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: inferring transportation mode**********
Error while inferring modes, timestamp is unchanged
Traceback (most recent call last):
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 714, in urlopen
    httplib_response = self._make_request(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 403, in _make_request
    self._validate_conn(conn)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 1053, in _validate_conn
    conn.connect()
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connection.py", line 419, in connect
    self.sock = ssl_wrap_socket(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 449, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/util/ssl_.py", line 493, in _ssl_wrap_socket_impl
    return ssl_context.wrap_socket(sock, server_hostname=server_hostname)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/ssl.py", line 501, in wrap_socket
    return self.sslsocket_class._create(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/ssl.py", line 1074, in _create
    self.do_handshake()
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/ssl.py", line 1343, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/connectionpool.py", line 798, in urlopen
    retries = retries.increment(
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/urllib3/util/retry.py", line 592, in increment
    raise MaxRetryError(_pool, url, error or ResponseError(cause))
urllib3.exceptions.MaxRetryError: HTTPSConnectionPool(host='lz4.overpass-api.de', port=443): Max retries exceeded with url: /api/interpreter (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)')))

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 24, in predict_mode
    mip.runPredictionPipeline(user_id, time_query)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 55, in runPredictionPipeline
    self.predictedProb = self.predictModesStep()
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 74, in predictModesStep
    predictedProb.append(get_prediction(i, section_entry))
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 113, in get_prediction
    return get_motorized_prediction(i, section_entry)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 140, in get_motorized_prediction
    predicted_transit_mode = _get_transit_prediction(i, section_entry)
  File "/Users/jrischpa/Documents/e-mission-server/emission/analysis/classification/inference/mode/rule_engine.py", line 175, in _get_transit_prediction
    start_transit_stops = enetm.get_stops_near(section_entry.data.start_loc, start_radius)
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/transit_matching/match_stops.py", line 124, in get_stops_near
    stops = get_public_transit_stops(lat - bbox_delta, lon - bbox_delta, lat + bbox_delta, lon + bbox_delta)
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/transit_matching/match_stops.py", line 81, in get_public_transit_stops
    call_return = make_request_and_catch(overpass_query)
  File "/Users/jrischpa/Documents/e-mission-server/emission/net/ext_service/transit_matching/match_stops.py", line 30, in make_request_and_catch
    response = requests.post(url + "api/interpreter", data=overpass_query)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/api.py", line 115, in post
    return request("post", url, data=data, json=json, **kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/Users/jrischpa/miniconda-23.1.0/envs/emission/lib/python3.9/site-packages/requests/adapters.py", line 563, in send
    raise SSLError(e, request=request)
requests.exceptions.SSLError: HTTPSConnectionPool(host='lz4.overpass-api.de', port=443): Max retries exceeded with url: /api/interpreter (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate in certificate chain (_ssl.c:1129)')))
2023-11-28T13:43:59.500719-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: inferring labels**********
2023-11-28T13:43:59.551722-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: populating expectations**********
2023-11-28T13:43:59.573080-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: creating confirmed objects **********
While getting section summary, section length = 0. This should never happen, but let's not crash if it does
While getting section summary, section length = 0. This should never happen, but let's not crash if it does
While getting section summary, section length = 0. This should never happen, but let's not crash if it does
While getting section summary, section length = 0. This should never happen, but let's not crash if it does
2023-11-28T13:43:59.978388-08:00**********UUID eada86fb-40ec-4cd1-8ca8-7862db147718: creating composite objects **********
```

JGreenlee:
Or, alternatively, would we expect the cleaned and confirmed objects to be re-created anyway when we `intake`? And only the raw locations would be used from the dump?

JGreenlee:
Also, I don't see any reason to think this is related to the migration.

@the-bay-kay can confirm whether the same thing happens on `master` for Nov 23, 24, and 25, but we have been observing issues with the JSON dump since before any of this was rewritten. If anything, the goal here was to make it functional again, which appears to have been at least mostly successful

If we can confirm that the issue affecting Nov 23 and 25 is not a regression, I suggest we open a separate issue for it and merge this in the meantime.

shankari:
Can @the-bay-kay first confirm that this is not a regression? I am happy to merge fairly soon after that.

the-bay-kay:
>I don't think @the-bay-kay was resetting the pipeline as part of this process, so there should have been no new timeline data to process and the pipeline shouldn't have done anything to the timeline.

> Can @the-bay-kay first confirm that this is not a regression? 

Can confirm both!  I just double checked, and loading this timeline onto a fresh clone of master has the same issue.  If anyone is interested getting a copy of my opcode or these timelines for testing, LMK and I can send them over!

www/js/services/controlHelper.ts
```diff
@@ -0,0 +1,159 @@
...
+          );
+        });
+      });
+    });
+  };
+
+  // window['cordova'].file.cacheDirectory is not guaranteed to free up memory,
+  // so it's good practice to remove the file right after it's used!
+  const localClearData = function () {
+    return new Promise<void>(function (resolve, reject) {
```
shankari, line 99: Woo! Good addition to the existing functionality!

www/js/services/controlHelper.ts
```diff
@@ -0,0 +1,159 @@
...
+declare let window: FsWindow;
+
+export const getMyDataHelpers = function (
+  fileName: string,
+  startTimeString: string,
+  endTimeString: string,
+) {
+  const localWriteFile = function (result: ServerResponse<any>) {
+    const resultList = result.phone_data;
+    return new Promise<void>(function (resolve, reject) {
```
shankari, line 18: can this be combined with the similar function that saves `loggerDB` and `userCacheDB`?
Maybe create a new `fileWriterService`?

www/js/types/diaryTypes.ts
```diff
@@ -39,7 +39,7 @@ export type CompositeTrip = {
   confirmed_trip: ObjectId;
   distance: number;
   duration: number;
-  end_confirmed_place: ConfirmedPlace;
+  end_confirmed_place: ServerData<ConfirmedPlace>;
```
shankari, line 42: dumb typescript question for my edification: what does this change in aid of?

shankari (approved):
I am merging this in the interests of unblocking the final set of changes for the rewrite. However, although I see some expanded mocks, I don't actually see any new tests here. I would suggest that you add tests similar to the ones proposed in the discussion starting at 
https://github.com/e-mission/e-mission-phone/pull/1097#issuecomment-1824819053 

I have also added some polishing, both large and small, for future fixes.

PR #1053, opened 2023-10-04
⬆️ ☁️ Upload Service -- rewrite and testing

Abby-Wheelis:
See: [upload service rewrite](https://github.com/e-mission/e-mission-docs/issues/988#top) for discussion.

While we do not use this anywhere in production now, we would like to restore it at some point and therefore still need tests and a `ts` version of this file. 

Writing the mocks for the file system was difficult, but I think I got something figured out that allows the code to run and to verify that this is working, at least behaving as expected on the phone side.


Abby-Wheelis:
This is now running, and passing the test I created for it. It relies heavily on mocks, specifically for the file reading, fetching, and logging. The test is simple, but proves that parameter data is being passed through the function in the expected manner. 

Window.alert is the current mechanism for notifying the user of progress or changes in the upload process, to be reconsidered based on [#992](https://github.com/e-mission/e-mission-docs/issues/992).

Abby-Wheelis:
This is working now!! Both in the emulator and in the test I wrote. The UI will need some re-vamping (window alerts are not a great way of notifying the user of progress) when we get this back into production, but it is now Reactified and functional.

Abby-Wheelis:
This PR is dependent on [✏️ Rewrite KVStore, ClientStats, and CommHelper](https://github.com/e-mission/e-mission-phone/pull/1040#top), that one should be merged first.

JGreenlee:

www/js/control/uploadService.ts
```diff
@@ -0,0 +1,131 @@
...
+                console.log("default uploadConfigString = " + JSON.stringify(uploadConfig['url']))
+                url.push(uploadConfig["url"]);
+                resolve(url);
+            } catch (err) {
+                logError("Error while reading default upload config" + err);
+                reject(err);
+            }
+        } 
+    })
+}
```
JGreenlee, line 31: To keep you in the loop- there is an ongoing discussion that may be relevant to this.
https://github.com/e-mission/e-mission-docs/issues/985#issuecomment-1760270148

I see that this looks for a file called `uploadConfig.json`. For the email service, there is similarly the `emailConfig.json` file. I am proposing that we move away from these individual JSON files and instead support them as optional fields in the dynamic config. And if not present in the dynamic config, we'll fallback the default which would be hardcoded here

www/__mocks__/fileSystemMocks.ts
JGreenlee, line 1: What's going on with the indentation here?
Would you mind tidying it up - either uniformly to 4-space indents, or preferably 2-space indents?

www/__mocks__/fileSystemMocks.ts
```diff
@@ -0,0 +1,18 @@
+export const mockFileSystem = () => {
+    window['resolveLocalFileSystemURL'] = function (parentDir, handleFS) {
+        const fs = {"filesystem":
```
JGreenlee, line 3: It doesn't matter a ton but just FYI, JS object declarations don't need quotes around property names (`filesystem:`) and it's somewhat preferred to not have them

www/js/control/ProfileSettings.jsx
```diff
@@ -219,8 +221,12 @@ const ProfileSettings = () => {
     
     //methods that control the settings
     const uploadLog = function () {
-        UploadHelper.uploadFile("loggerDB")
-    };
+        if(uploadReason != "") {
+            let reason = uploadReason.split('').join('');
```
JGreenlee, line 225: Out of curiosity, what is the reason for `split('').join('')`?
Can you add a comment explaining for others in the future?

www/js/plugin/logger.ts
```diff
@@ -33,6 +33,9 @@ export const logInfo = (message: string) =>
 export const logWarn = (message: string) =>
   window['Logger'].log(window['Logger'].LEVEL_WARN, message);
 
+export const logError = (message: string) =>
+  window['Logger'].log(window['Logger'].LEVEL_ERROR, message);
+
```
JGreenlee, line 38: I believe that as a rule, we want all caught errors to be user-facing (`displayError`/`displayErrorMsg`). That is why I had purposefully *not* implemented a `logError`.
Let me know if you've heard otherwise

JGreenlee (changes requested):
I know this code is not currently being used and we can't try it out end-to-end, but it looks clean and will surely come in handy when it's time.

I have a few questions and requests below.

www/__mocks__/fileSystemMocks.ts
Abby-Wheelis: Tidied up in [4f3d9fb](https://github.com/e-mission/e-mission-phone/pull/1053/commits/4f3d9fb1a1ecc2982ce0b81f2f80d80cc262d502), for some reason my VS code is very inconsistent about how it handles indentation, but I was able to fix it by just using spaces since the file is so short

JGreenlee: I like to make sure VSCode says "Spaces: 2" in the bottom right, then highlight a section of code and do `Cmd + K`, `Cmd + F`.

Abby-Wheelis: I did not know about that trick, thank you!

www/__mocks__/fileSystemMocks.ts
Abby-Wheelis: Good to know! I went ahead and resolved that since it was an easy fix.

www/js/control/ProfileSettings.jsx
Abby-Wheelis: That was part of the code that I migrated, it was just in an ionic on-the-fly popup before. I'll experiment with removing it and see if I can figure out why. The reason does get used for the filename on the server, so that might be it. I'll investigate and leave a comment.

www/js/plugin/logger.ts
Abby-Wheelis: That makes sense, I'll switch these over.

www/js/control/uploadService.ts
Abby-Wheelis: Good to know, I think it does make sense to configure this with all the other options. Right now it's just a single URL that is needed. If we decide to make that change now I can definitely include it here, else I can come back and clean it up when we do make the change.

www/js/control/ProfileSettings.jsx
Abby-Wheelis: Everything still works if I take that out, and I did not find it in the previous version of the file. It was probably something I was trying when I couldn't get things to upload, and I never took it out. I have removed it since it was unnecessary code.

JGreenlee (approved):
LGTM

www/js/control/uploadService.ts
shankari: Let's handle the app-level config separately, and include this in there
https://github.com/e-mission/e-mission-docs/issues/985#issuecomment-1769790309

This is fine for now.

www/__tests__/uploadService.test.ts
```diff
@@ -0,0 +1,46 @@
...
+mockCordova();
+
+mockLogger();
+mockFile(); //mocks the base directory 
+mockFileSystem(); //comnplex mock, allows the readDBFile to work in testing
+
+//use this message to verify that the post went through
+let message = "";
+
+global.fetch = (url: string, options: {method: string, headers: {}, body: string}) => new Promise((rs, rj) => {
```
shankari, line 21: can you clarify what this does? I understand that if there are options, you return "sent..." and otherwise, you return "mock data" after a timeout, but I don't understand why there is a timeout only in the `else` and not in the `if`

shankari (changes requested):
Almost there!

www/js/control/uploadService.ts
```diff
@@ -0,0 +1,131 @@
...
+            }, reject);
+        });
+    });
+}
+
+const sendToServer = function upload(url, binArray, params) {
+    //this was the best way I could find to contact the database, 
+    //had to modify the way it gets handled on the other side
+    //the original way it could not find "reason"
+    const urlParams = "?reason=" + params.reason + "&tz=" + params.tz;
```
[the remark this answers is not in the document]
Abby-Wheelis: This is an old comment, I will remove it. It took me a long time to figure out that I just needed to use url encoding because that's what the old method was doing under the hood. Once I figured out the right way to do it I was able to reset the code on [`em bug upload`](https://github.com/e-mission/em-bug-upload).

www/json/uploadConfig.json.sample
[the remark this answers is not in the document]
Abby-Wheelis: I was using [em-bug-upload](https://github.com/e-mission/em-bug-upload) for testing, which was why I made the change. I never figured out how to use `httpbin.org` but I'll try again today, it may have been that my code was broken for so long.

www/__tests__/uploadService.test.ts
Abby-Wheelis: I've added comments about what the difference between the if and the else are, for some more clarity. I also added a timeout to the if, you're right, it doesn't make sense to have it one place and not the other.

www/json/uploadConfig.json.sample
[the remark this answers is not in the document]
Abby-Wheelis: the httpbin seems to work (I'm not getting any errors) but I also haven't found any way to see what gets uploaded. I can change the sample file back though, since it does seem to work

www/__tests__/uploadService.test.ts
```diff
@@ -18,16 +18,22 @@ mockFileSystem(); //comnplex mock, allows the readDBFile to work in testing
...
+//each have a slight delay to mimic a real fetch request
+global.fetch = (url: string, options: { method: string, headers: {}, body: string }) => new Promise((rs, rj) => {
+  //if there's options, that means there is a post request
+  if (options) {
+    message = "sent " + options.method + options.body + " for " + url;
+    setTimeout(() => {
+      rs('sent ' + options.method + options.body + ' to ' + url);
+    }, 100);
+  }
+  //else it is a fetch request
```
shankari, line 30: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

```suggestion
  //else it is a get request
```

shankari (approved):

PR #1054, opened 2023-10-05
Change "no label_options in config" behavior to use local default lab…

sebastianbarry:
…el options

confirmHelper.ts
- Removed archaic behavior or using old Angular i18nUtils module to get the filename of the translations' trip_confirm_options.json file
- The "else" default "no label_options found in config" behavior is almost identical to the if(appConfig.label_options), but it just uses the default label options URL at label-options.json.sample
- Pulled the language-specific text handling behavior out of the if statement, because either way (label_options or no label_options) the JSON data model will look the same

label-options.json.sample
- Created new file
- Replaces trip_confirm_options.json.sample (this location in confirmHelper.ts was the only place in the codebase using this file)
- Modeled after https://github.com/e-mission/nrel-openpath-deploy-configs/blob/main/label_options/example-program-label-options.json
- Only has translations for EN and ES

trip_confirm_options.json.sample
- Removed this file


sebastianbarry:
Closing in favor of https://github.com/e-mission/e-mission-phone/pull/1055

PR #1043, opened 2023-09-29
Add Jest testing for useImperialConfig.ts

sebastianbarry:
useImperialConfig.ts
- Added exports to the pure functions to get them inside the test file

useImperialConfig.test.ts
- Added a Jest mock to mock the useAppConfig
- Tested use cases for the 4 functions

PR #1057, opened 2023-10-05
Use imperial config test

sebastianbarry:
useImperialConfig.ts
- Added exports to the functions, so that we can test them _(can't use `export function useImperialConfig()` as it gets caught in invalid React hooks with `useState`)_

useImperialConfig.test.ts
- Using mock of `useAppConfig` so that the testing doesn't terminate by following the chain of import statements within `useAppConfig`
- Added basic testing of the functions

**2 of the tests are failing, as they are not rounding properly**

1:

```
convertSpeed › should convert meters per second to miles per hour when imperial flag is true

    expect(received).toBe(expected) // Object.is equality

    Expected: 15
    Received: 14.99999535936
```

2:
```
convertDistance › should convert meters to miles when imperial flag is true

    expect(received).toBe(expected) // Object.is equality

    Expected: 1
    Received: 0.99999720514
```

### DIFF
--- a/www/__tests__/useImperialConfig.test.ts
+++ b/www/__tests__/useImperialConfig.test.ts
@@ -14,17 +14,24 @@ jest.mock('../js/useAppConfig', () => {
 describe('formatForDisplay', () => {
   it('should round to the nearest integer when value is >= 100', () => {
     expect(formatForDisplay(105)).toBe('105');
-    expect(formatForDisplay(119)).toBe('119');
+    expect(formatForDisplay(119.01)).toBe('119');
+    expect(formatForDisplay(119.91)).toBe('120');
   });
 
   it('should round to 3 significant digits when 1 <= value < 100', () => {
     expect(formatForDisplay(7.02)).toBe('7.02');
-    expect(formatForDisplay(11.3)).toBe('11.3');
+    expect(formatForDisplay(9.6262)).toBe('9.63');
+    expect(formatForDisplay(11.333)).toBe('11.3');
+    expect(formatForDisplay(99.99)).toBe('100');
   });
 
   it('should round to 2 decimal places when value < 1', () => {
-    expect(formatForDisplay(0.07)).toBe('0.07');
+    expect(formatForDisplay(0.07178)).toBe('0.07');
+    expect(formatForDisplay(0.08978)).toBe('0.09');
     expect(formatForDisplay(0.75)).toBe('0.75');
+    expect(formatForDisplay(0.001)).toBe('0');
+    expect(formatForDisplay(0.006)).toBe('0.01');
+    expect(formatForDisplay(0.00001)).toBe('0');
   });
 });
 
--- a/www/__tests__/useImperialConfig.test.ts
+++ b/www/__tests__/useImperialConfig.test.ts
@@ -1,0 +1,49 @@
+import { convertDistance, convertSpeed, formatForDisplay } from '../js/config/useImperialConfig';
+
+
+// This mock is required, or else the test will dive into the import chain of useAppConfig.ts and fail when it gets to the root
+jest.mock('../js/useAppConfig', () => {
+  return jest.fn(() => ({
+    appConfig: {
+      use_imperial: false
+    },
+    loading: false
+  }));
+});
+  
+describe('formatForDisplay', () => {
+  it('should round to the nearest integer when value is >= 100', () => {
+    expect(formatForDisplay(105)).toBe('105');
+    expect(formatForDisplay(119)).toBe('119');
+  });
+
+  it('should round to 3 significant digits when 1 <= value < 100', () => {
+    expect(formatForDisplay(7.02)).toBe('7.02');
+    expect(formatForDisplay(11.3)).toBe('11.3');
+  });
+
+  it('should round to 2 decimal places when value < 1', () => {
+    expect(formatForDisplay(0.07)).toBe('0.07');
+    expect(formatForDisplay(0.75)).toBe('0.75');
+  });
+});
+
+describe('convertDistance', () => {
+  it('should convert meters to kilometers by default', () => {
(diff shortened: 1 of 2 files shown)


sebastianbarry:
This PR replaces https://github.com/e-mission/e-mission-phone/pull/1043 as that PR was based on the wrong branch

> [3:01 PM] Barry, Sebastian
> Greenlee, Jack I have https://github.com/e-mission/e-mission-phone/pull/1043 which is based off master and going to master, but since I've made the PR there have been changes to useImperialConfig.ts that created merge conflicts so I will need to change my PR - I just want to make sure the PR is going to the right place
> ​[3:14 PM] Greenlee, Jack
> Yeah, you'll have to resolve those. We should have had you create that PR off of a newer branch because master was too old
> master still is too old to base anything on currently. I think our rate of development is currently outpacing our bandwidth for testing and we sort of have a backlog

sebastianbarry:
@JGreenlee Here's the new PR for `useImperialConfig.ts` testing

What do you want to do about the rounding discrepancy I encountered? (see above)

JGreenlee:
Oh yeah, that can sometimes happen with floating-point arithmetic

I think you can just let the check pass if the result is within some tolerance (like 0.01) of the expectation

(and for future reference, it's possible to change the target of a PR instead of making a new one)

sebastianbarry:
Used Jest's `toBeCloseTo` function which uses a tolerance of 0.005
https://jestjs.io/docs/expect#tobeclosetonumber-numdigits
> The optional numDigits argument limits the number of digits to check after the decimal point. For the default value 2, the test criterion is Math.abs(expected - received) < 0.005 (that is, 10 ** -2 / 2).

www/__tests__/useImperialConfig.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  it('should round to 3 significant digits when 1 <= value < 100', () => {
+    expect(formatForDisplay(7.02)).toBe('7.02');
+    expect(formatForDisplay(11.3)).toBe('11.3');
+  });
+
+  it('should round to 2 decimal places when value < 1', () => {
+    expect(formatForDisplay(0.07)).toBe('0.07');
+    expect(formatForDisplay(0.75)).toBe('0.75');
```
JGreenlee, line 27: It doesn't seem like we're actually putting the rounding to the test here. Unit tests should explore the boundaries of possible inputs. So don't be 'safe' with it.

I want to see some trickier inputs, like `105.01` or `99.99` or `0.001` or `0.00001` so we can actually see how the rounding is working in action and make sure it's desirable / as expected.

JGreenlee, line 27: It doesn't seem like we're actually putting the rounding to the test here. Unit tests should explore the boundaries of possible inputs. So don't be 'safe' with it.

I want to see some trickier inputs, like 105.01 or 99.99 or 0.001 or 0.00001 so we can actually see how the rounding is working in action and make sure it's desirable / as expected.

JGreenlee (changes requested):

www/__tests__/useImperialConfig.test.ts
JGreenlee: (I know you were probably just using the example inputs I put in the comments. Those are to help illustrate how the function works and these are to test and expose potential bugs. Different inputs for different purposes)

sebastianbarry: I created some hardier tests in my most recent commit:

```javascript
describe('formatForDisplay', () => {
  it('should round to the nearest integer when value is >= 100', () => {
    expect(formatForDisplay(105)).toBe('105');
    expect(formatForDisplay(119.01)).toBe('119');
    expect(formatForDisplay(119.91)).toBe('120');
  });

  it('should round to 3 significant digits when 1 <= value < 100', () => {
    expect(formatForDisplay(7.02)).toBe('7.02');
    expect(formatForDisplay(9.6262)).toBe('9.63');
    expect(formatForDisplay(11.333)).toBe('11.3');
    expect(formatForDisplay(99.99)).toBe('100');
  });

  it('should round to 2 decimal places when value < 1', () => {
    expect(formatForDisplay(0.07178)).toBe('0.07');
    expect(formatForDisplay(0.08978)).toBe('0.09');
    expect(formatForDisplay(0.75)).toBe('0.75');
    expect(formatForDisplay(0.001)).toBe('0');
    expect(formatForDisplay(0.006)).toBe('0.01');
    expect(formatForDisplay(0.00001)).toBe('0');
  });
});
```

JGreenlee (approved):
LGTM

www/__tests__/useImperialConfig.test.ts
```diff
@@ -0,0 +1,56 @@
+import { convertDistance, convertSpeed, formatForDisplay } from '../js/config/useImperialConfig';
+
+
+// This mock is required, or else the test will dive into the import chain of useAppConfig.ts and fail when it gets to the root
+jest.mock('../js/useAppConfig', () => {
+  return jest.fn(() => ({
+    appConfig: {
+      use_imperial: false
```
shankari, line 8: shouldn't we also test `use_imperial: true`?

www/__tests__/useImperialConfig.test.ts
```diff
@@ -0,0 +1,56 @@
+import { convertDistance, convertSpeed, formatForDisplay } from '../js/config/useImperialConfig';
+
+
+// This mock is required, or else the test will dive into the import chain of useAppConfig.ts and fail when it gets to the root
+jest.mock('../js/useAppConfig', () => {
```
shankari, line 5: this can be removed after all the angular code is removed, right?

shankari (changes requested):
I am fine with merging this for now, but would like a cleanup to address my comments.

shankari (approved):

PR #1058, opened 2023-10-06
More token log statements

Abby-Wheelis:
To help with [❗ Login on fresh install resulting in dashboard error and "not logged in"](https://github.com/e-mission/e-mission-docs/issues/1002#top), working to add additional log statements for the auth token, since that is what seems to be null in staging.


Abby-Wheelis:
I think this is enough log statements to help us identify the issue, but I'll be online another ~45 minutes if you see somewhere I missed

shankari:
Agreed. This seems like a reasonable first set of logs. Will get this out tonight; we can always add more if needed in a future RC

PR #1061, opened 2023-10-10
Rewrite angular service - input-matcher

jiji14:
## Rewrite `input-matcher.js` Angular service

**Related Issue:** [GitHub Issue #1001](https://github.com/e-mission/e-mission-docs/issues/1001)

`Input-matcher` is a function designed to check if the user's trip input is correct and print logs. I've kept the fundamental logic while updating the old JavaScript syntax and adding input/output types for each function.

Screenshot

<br/>

#### 1. Replacing `moment` with `Luxon`

Previously, the Moment format function was used to print logs. I have replaced `moment` with `Luxon`. 
<br />

Here's a comparison of the two outputs:
Comparison Screenshot

<br />

#### 2. Unit Testing

I have written test cases for both valid and invalid inputs for each function. To ensure proper testing, please add `"testEnvironment": "jsdom"` to your `jest.config.json` (I haven't pushed the update for `jest.config.json` to avoid conflicts). 

Here are the test results:
Test Results

<br />

#### I had to merge `onboarding_routing_sept_2023` branch because `service_rewrite_2023` is too out-dated.


www/js/survey/input-matcher.ts
```diff
@@ -0,0 +1,243 @@
...
+      /* if a place has no enter time, this is the first start_place of the first composite trip object
+      so we will set the start time to the start of the day of the end time for the purpose of comparison */
+      entryStart = DateTime.fromSeconds(entryEnd).startOf('day').toUnixInteger();
+    }
+
+    if (!entryEnd) {
+        /* if a place has no exit time, the user hasn't left there yet
+        so we will set the end time as high as possible for the purpose of comparison */
+        entryEnd = EPOCH_MAXIMUM;
+    }
```
jiji14, line 103: When `tlEntry.end_ts` and `tlEntry.exit_ts` are equal to `0`, `entryEnd` is `falsy` value so it goes into this `if statement` and set `entryEnd` as `EPOCH_MAXIMUM`, and we get the wrong result. Should I handle the case when `tlEntry.end_ts` and `tlEntry.exit_ts` are equal to `0`?

JGreenlee: It is really great that you're considering edge cases like this!
`end_ts` and `exit_ts` are measurements in "Unix time" aka "Epoch seconds" https://en.wikipedia.org/wiki/Unix_time.

In the modern era, this should be a 10-digit number (like 1697473049)

If either of these measurements were 0, it would definitely be an invalid input. So I think in this case, it is fair to treat a value of 0 the same as if it was undefined; all falsy values will be treated the same way. So it is safe the way it is.

www/js/survey/input-matcher.ts
```diff
@@ -0,0 +1,243 @@
...
+export type TlEntry = {
+    key: string,
+    origin_key: string,
+    start_ts: number,
+    end_ts: number,
+    enter_ts: number,
+    exit_ts: number,
+    duration: number,
+    getNextEntry: any
+}
```
JGreenlee, line 50: It is so great that you're adding type definitions on all of these!

Having types like these will be useful in a lot of places, so I was hoping to keep common type definitions in a central place. Could you refer to `js/diary/diaryTypes.ts`? The types there are still incomplete, but you can fill in what you have here, and then import those types into this file.

This way, we will not have type definitions cluttering up every file.

www/__tests__/input-matcher.test.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    beforeEach(() => {
+        // create valid userTrip and trip object before each test case.
+        userTrip  = { 
+            data: {
+                end_ts: 1437604764,
+                start_ts: 1437601247,
+                label: 'FOO',
+                status: 'ACTIVE'
+            },
```
JGreenlee, line 26: These tests look great! Could you add comments in the code explaining
- where the fake data comes from - are they based on real trips? Are they from `real_examples` on the server? Are they made-up values?
- how this testing strategy parallels the testing on the server (I think you referred to `TestUserInputFakeData.py` while making these?)

This way, anyone who updates these tests in the future will know where all of this came from.

JGreenlee (changes requested):

www/js/survey/input-matcher.ts
JGreenlee: That sounds good to me. Go ahead and move `js/diary/diaryTypes.ts` into `js/types` and then expand on `diaryTypes.ts` with the types you use here.

www/__tests__/input-matcher.test.ts
JGreenlee, line 1: Please rename to `inputMatcher.test.ts` (we want to use camelCase filenames)

www/js/survey/input-matcher.ts
JGreenlee, line 1: Please rename to `inputMatcher.ts` (we want to use camelCase filenames)

www/__tests__/input-matcher.test.ts
jiji14: Updated the comment. Please check this commit: b8ec27d

www/__tests__/input-matcher.test.ts
jiji14: Updated the file name. Please check this commit: b43b6a0

www/js/survey/input-matcher.ts
jiji14: Updated it. Please check this commit: ffcc871d

www/__tests__/inputMatcher.test.ts
JGreenlee, line 1: These tests look good, but I'm wondering if we want to match the server tests ([TestUserInputFakeData.py](https://github.com/e-mission/e-mission-server/blob/e9aae0289a78d79db1e38ca7506e619d6743291a/emission/tests/analysisTests/userInputTests/TestUserInputFakeData.py)) even more closely.
Maybe @shankari can answer: should we be striving to loosely parallel the server tests, or to match them exactly?

JGreenlee (changes requested):

www/js/types/diaryTypes.ts
```diff
@@ -67,6 +67,54 @@ export type PopulatedTrip = CompositeTrip & {
...
+export type TlEntry = {
+  key: string,
+  origin_key: string,
+  start_ts: number,
+  end_ts: number,
+  enter_ts: number,
+  exit_ts: number,
+  duration: number,
+  getNextEntry?: () => PopulatedTrip | ConfirmedPlace,
+}
```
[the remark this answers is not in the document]
jiji14: Thanks for correcting this! Then I will leave it as it is for now :)

www/__tests__/inputMatcher.test.ts
shankari: I think it would be best to, at a minimum, replicate the tests. In general, when we have these common code pieces (input matching/energy calculations), it is useful to think of them as a library with a standard test suite. In the future, we may also want to actually pull them out into libraries 📚 

Having said that, I believe these do include tests from `TestUserInputFakeData` I spot checked a bit and they seem fairly similar. There are some additional tests here (e.g. invalid input) that we should also probably include in the server code.

The server code also has `TestUserInput.py` that tests that the value is filled in correctly to the test. I don't see that in these tests.

JGreenlee: @shankari It seems to me that `TestUserInputFakeData.py` on the server uses completely different fake data. The `start_ts`, `end_ts`, and `write_ts` on the server are numbers between 1 and 15, not real-world timestamps.

The tests here on the phone use real timestamps like `1437601247` (copied from some of the `real_examples` in `e-mission-server` - Is that correct @jiji14?) Sounds like we want to replicate `TestUserInputFakeData.py`, so we should use the same fake trips and inputs.

If we *also* want to replicate `TestUserInput.py` (which uses real-world data), then we would want to use the trips and inputs from `real_examples/shankari_2016-06-20` and `real_examples/shankari_2016-06-20.user_inputs.manual_trip_user_input`. However, I think it would be fairly involved to do this and it might slow down our progress.

jiji14: @JGreenlee Yes, `1437601247` was copied from `real_examples`.

JGreenlee:
@shankari So what do we want to do with this PR? Are we going to change the tests to match the ones on the server (https://github.com/e-mission/e-mission-phone/pull/1061#discussion_r1374010647)? Are we going to do that now or later?

One other option is to let https://github.com/e-mission/e-mission-phone/pull/1086 absorb this PR since it includes these changes and builds on top of them. We could leave this PR open and come back to it when we figure out the testing strategy.

shankari:
@JGreenlee thanks for the clarification around `TestUserInputFakeData.py` and the different tests here.
I think we can go ahead and merge this now, but we should file a separate cleanup issue to pull out the javascript + python input matcher code into a separate module, and unify the tests in both languages.

This wasn't in "Ready for review", so I'm going to go and review https://github.com/e-mission/e-mission-phone/pull/1086 for now just in case there's anything else pending here. But the tests are passing, so if this is moved into the correct column, I am happy to merge ASAP

www/js/survey/enketo/enketo-add-note-button.js
```diff
@@ -3,131 +3,121 @@
...
+    enbs.initConfig = function (tripSurveyName, placeSurveyName) {
+      enbs.tripSurveyName = tripSurveyName;
+      if (tripSurveyName) {
+        enbs.MANUAL_KEYS.push('manual/trip_addition_input');
+      }
+      enbs.placeSurveyName = placeSurveyName;
+      if (placeSurveyName) {
+        enbs.MANUAL_KEYS.push('manual/place_addition_input');
+      }
+    };
```
shankari, line 32: wait, why do we have all these extraneous whitespace changes? I thought that once we used prettier, we wouldn't need this any more.

@jiji14 do you need to re-pull the branch?

www/js/survey/enketo/enketo-trip-button.js
shankari, line 1: Ditto

shankari (changes requested):
It is really hard to review this change given the extraneous whitespace.
@jiji14 @JGreenlee can one of you please fix the prettier issues so that I can focus on the actual changes?

jiji14:
> It is really hard to review this change given the extraneous whitespace. @jiji14 @JGreenlee can one of you please fix the prettier issues so that I can focus on the actual changes?

@shankari Let me fix this!

jiji14:
@shankari 

I deleted `inputMatcher` dependency in factory, and then Prettier changed the line break since it became shorter than the original one. That's why the white space got altered. The only solution I can think of to fix this issue is to use `prettier ignore` line so that you don't see the changes, but I don't think it's a good idea. What do you think?

[ before ] 
```js
  .factory(
    'EnketoNotesButtonService',
    function (InputMatcher, EnketoSurveyAnswer, Logger, $timeout) {
      var enbs = {};
```

[ after ]
```js
  .factory('EnketoNotesButtonService', function (EnketoSurveyAnswer, Logger, $timeout) {
    var enbs = {};
```

JGreenlee:
> I deleted `inputMatcher` dependency in factory, and then Prettier changed the line break since it became shorter than the original one. That's why the white space got altered.

Interesting that Prettier actually changes the indent level based on one line being longer or shorter.

I agree and think you should leave it how it is. When we look at the diff, we can just 'ignore whitespace' to make it easier to review

shankari:
> if there is a real change in the line, I am OK with associate whitespace changes.

Although it is really annoying that because of the difference in indentation, pretty much every single line was changed.
Particularly annoying since I have been very picky about whitespace changes to keep the commit history clean.
I wonder if it would be worthwhile to use `prettier ignore` when we have such large scale extraneous changes, and periodically remove them in a single commit labeled "prettier catch up" or something.

Let's see how often this happens before deciding...

shankari (approved):

shankari:
@JGreenlee I just merged this. If you pull it into #1086 I can review that tomorrow.

PR #1062, opened 2023-10-10
✍️📂 Rewrite Infinite Scroll Filters (Multilabel + Enketo)

sebastianbarry:
multilabel/infinite_scroll_filters.js
- Removed this file and replaced with .ts file

multilabel/infinite_scroll_filters.ts
- Rewrote this file in React

enketo/infinite_scroll_filters.js
- Removed this file and replaced with .ts file

enketo/infinite_scroll_filters.ts
- Rewrote this file in React

index.js
- Removed Angular reference to multilabel/infinite_scroll_filters.js
- Removed Angular reference to enketo/infinite_scroll_filters.js

diary.js
- Removed angular reference to multilabel.infscrollfilters
- Removed angular reference to enketo.infscrollfilters

LabelTab.tsx
- Changed the tripFilterFactory to reflect the new way of representing the surveyOpt object from survey.ts

survey.ts
- imported the multilabel and enketo filters the "React way", and changed the typescript to reflect that the filters property are arrays containing the label filters


sebastianbarry:
Testing:

| | Multilabel | Enketo |
|---|---|---|
| Labels appear | Simulator Screenshot - iPhone 13 - 2023-10-10 at 13 31 43 |  |
| Labeling stays | Simulator Screenshot - iPhone 13 - 2023-10-10 at 13 31 47 |  |
| Purpose labels work | Simulator Screenshot - iPhone 13 - 2023-10-10 at 13 31 51 |  |
| Labeled trip doesn't populate in to-label filter after labeling + refreshing | Simulator Screenshot - iPhone 13 - 2023-10-10 at 13 32 14 |  |
| Labeled trip appears in All Trips after labeling | Simulator Screenshot - iPhone 13 - 2023-10-10 at 13 33 33 |  |

sebastianbarry:
The EnketoTripButtonService is causing an issue that is stopping the app from starting when using ENKEO:
image

I had figured this could be circumvented by replacing the line
```javascript
return typeof t.userInput[getAngularService('EnketoTripButtonService').SINGLE_KEY] === 'undefined';
```
with
```javascript
return typeof t.userInput["SURVEY"] === 'undefined';
```
effectively cutting out the use of `getAngularService("EnketoTripButtonService");`. This however, did not fix the issue and the error is still coming up.

This is a regression with my commit, because this ENKETO setup is working on `onboarding_routing_sept_23`

www/js/survey/enketo/infinite_scroll_filters.ts
```diff
@@ -0,0 +1,31 @@
...
+ * For best performance, all data should be loaded into the in-memory timeline,
+ * and in addition to writing to storage, the data should be written to memory.
+ * All UI elements should only use $scope variables.
+ */
+
+import i18next from "i18next";
+import { getAngularService } from "../../angular-react-helper";
+
+const unlabeledCheck = (t) => {
+    return typeof t.userInput[getAngularService('EnketoTripButtonService').SINGLE_KEY] === 'undefined';
```
sebastianbarry, line 13: This is the line that is having problems

www/js/survey/enketo/infinite_scroll_filters.js
```diff
@@ -1,40 +0,0 @@
...
- * a particular day is retrieved from the Timeline factory and put into the scope.
- * For best performance, all data should be loaded into the in-memory timeline,
- * and in addition to writing to storage, the data should be written to memory.
- * All UI elements should only use $scope variables.
- */
-
-import angular from 'angular';
-
-angular.module('emission.survey.enketo.trip.infscrollfilters',[
-    'emission.survey.enketo.trip.button',
```
JGreenlee, line 14: @sebastianbarry Angular is no longer registering the `emission.survey.enketo.trip.button` module because this module, and the other `infinite_scroll_filters`, are no longer depending on it.

So when you try to `getAngularService('EnketoTripButtonService')`, Angular finds nothing.

Include `emission.survey.enketo.trip.button` in the dependency array of `diary.js` and I think your issue will go away.

sebastianbarry:
That fixed it, thanks @JGreenlee !

Final testing found in comment above, this PR is ready for review

JGreenlee:

(My PRs were merged, but my commits are still showing up in this PR. I've discovered that toggling the branch forces Github to update the diff)

www/js/survey/multilabel/infinite_scroll_filters.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const unlabeledCheck = (t) => {
+    return t.INPUTS
+         .map((inputType, index) => typeof t.userInput[inputType] === 'undefined')
+         .reduce((acc, val) => acc || val, false);
+}
+
+const invalidCheck = (t) => {
+    const retVal =
+        (typeof t.userInput['MODE'] !== 'undefined' && t.userInput['MODE'].value === 'pilot_ebike') &&
```
JGreenlee, line 19: This used to be `angular.isDefined(x)` and was replaced as `typeof x !== 'undefined'`

Instead, you can simply do `!x` (in all places this was done)

Why? Because if `x` was, for some reason, `null` or `false` or `NaN` or any other falsy value, `typeof x !== 'undefined'` would be true.

We should treat any falsy value, whether `undefined` or not, as invalid / 'not filled in'.

JGreenlee (changes requested):

www/js/survey/multilabel/infinite_scroll_filters.ts
sebastianbarry: Fixed this in: https://github.com/e-mission/e-mission-phone/pull/1062/commits/259778a0d994846693de0ec405e051f8ab09f040

JGreenlee (approved):
LGTM

www/js/diary.js
```diff
@@ -3,9 +3,8 @@ import LabelTab from './diary/LabelTab';
 
 angular.module('emission.main.diary',['emission.main.diary.services',
                                       'emission.survey.multilabel.buttons',
-                                      'emission.survey.multilabel.infscrollfilters',
                                       'emission.survey.enketo.add-note-button',
-                                      'emission.survey.enketo.trip.infscrollfilters',
```
shankari, line 8: I saw the comment related to this, but I also wonder where we are using `emission.main.diary` any more. It used to be in `main.js` for the routing, but we have also switched to react routing now, correct? can't we just remove this file?

shankari (approved):
I am fine with merging this for now, but we should really write tests for this functionality instead of relying on manual tests.

PR #1063, opened 2023-10-11
Rewrite answer.js into enketoHelper.ts

Abby-Wheelis:
This PR is based on Jack's rewrite PR so that I can work off his mocks and hopefully minimize merge conflicts once that branch can be merged. 

Addresses one of the rewrites in: [✍ Angular Services needing rewrite](https://github.com/e-mission/e-mission-docs/issues/977#top)


Abby-Wheelis:
I have discovered an issue here, when I try to add an activity to a trip or a place in the emulator, that activity does not show up in a list on the card as expected. Hopefully I can find the issue and resolve in the re-write and test process.

Abby-Wheelis:
>One of the things that's popping up with testing, is the need for a fake survey.

I was able to clear this for the test I was working on in that moment by just copying a part of the survey from the logs.


Now, I'm facing an issue with loading the config, in the process of trying to test `resolveLabel`. I think my next step is to test `_lazyLoadConfig` independently to make sure that is working, with the web of calls to storage that I might need to mock (some of them are already mocked). What's odd is that I have tests for `filterByNameAndVersion` already running, and those depend on `_lazyLoadConfig` so my problem might be elsewhere.

Abby-Wheelis:
> test _lazyLoadConfig independently to make sure that is working, with the web of calls to storage that I might need to mock (some of them are already mocked).

I did this by mocking `getDocument` in the cordova mocks in order to return a hard-coded of a config with just enough information to have one survey. This works for now, and can be expanded later as needed.

Abby-Wheelis:
I've moved on try trying to test `resolveLabels` and to do that I need `i18next` and `MessageFormat`. I have been trying to find a way to mock these and allow the tests to access the mocks, but so far have been unsuccessful. Here's some information I found about [Manual Mocks in Jest](https://jestjs.io/docs/manual-mocks). It seemed like a good place to start, but I have not been able to have the code running in the tests access the mocks, it just seems to fail as undefined for some reason.

Abby-Wheelis:
>I have discovered an issue here, when I try to add an activity to a trip or a place in the emulator, that activity does not show up in a list on the card as expected. Hopefully I can find the issue and resolve in the re-write and test process.

I have not found a way to fix this yet. I did decide to check on `master`, and it works there. The other difference between this branch and `master` is that the demographics survey does not save between logins on this branch. 

On `service_rewrite_2023`, the activities show up after I input them and persist after refreshing the label screen, but don't stay after logging out and back in. The demographic survey responses don't show up after a new login either. 

I also checked out [JGreenlee:rewrite-services-sept2023](https://github.com/JGreenlee/e-mission-phone/tree/rewrite-services-sept2023), since those changes are in this branch too. Not saving between logins is observed here, but when I input activities they show up, so there are no regressions between `service_rewrite_2023` and the end of Jack's changes.

To summarize:
- the demographic survey no longer saves between logins due to changes between `master` and `service_rewrite_2023`
- in my migration I broke saving the activity survey responses and having them show on the label screen

I can investigate further tomorrow, but I have a better idea now of when the regressions took place.

JGreenlee:
> On `service_rewrite_2023`, the activities show up after I input them and persist after refreshing the label screen, but don't stay after logging out and back in. The demographic survey responses don't show up after a new login either.

I think this is the expected behavior. The log out button should clear everything. If we do not "force sync" before logging out, any unsynced data is lost. This is why we have the "Are you sure?" popup explaining this to users before they log out

Abby-Wheelis:
Some updates after meeting with @JGreenlee this morning to discuss the problems I've been having:

Instead of trying to mock `i18next` and `messageformat`, it is better to use the actual packages. To do this we configured `i18next` in the test file, which should work after I incorporate the changes from [React Component Testing Setup](https://github.com/e-mission/e-mission-phone/pull/1049#top). Jack also helped me figure out that we were using a different `messageformat` package than we thought, and after updating to the new `@messageformat/core` the test is progressing further.

Next steps:

- [ ] incorporate the changes from #1049 
- [ ] update my mock config to resolve testing error with messageformat
- [ ] see if forcing sync causes the demographic survey to save
- [ ] investigate typings to make sure they are accurate and in typescript
- [ ] debug the activities not saving to restore
- [ ] finish writing tests

Abby-Wheelis:
I just learned something helpful for testing -- there were a couple of universal packages that were throwing errors in the tests. When I moused over them, they described, for example, that `URL` was an alias for `require('url').URL`. I was able to configure testing and get this to work by adding the following to my tests:

```
global.URL = require('url').URL;
global.Blob = require('node:buffer').Blob;
```

It was a quick and easy fix (which lots of testing configuration has not been) so documenting for future reference.

Abby-Wheelis:
>in my migration I broke saving the activity survey responses and having them show on the label screen

An update on this situation, in setting breakpoints, I noticed often that the `filterNameAndVersion` function often had an input of an empty `answer` array. When I set more breakpoints, and did more stepping through things, I ended up seeing some answers. Therefore, I think the issue has to do with timing.

Abby-Wheelis:
>I have discovered an issue here, when I try to add an activity to a trip or a place in the emulator, that activity does not show up in a list on the card as expected. 

I (finally) figured out what the problem was here. When migrating the files, I switched over to Luxon, but that caused the formatting of the timestamps to change a little bit. By default, the value of these timestamps were in milliseconds, but the calculations in `resolveTimestamps` were expecting precision to "seconds". There was an easy call of `dt.toSeconds()` that resolved this, the "downgrade" of precision is meaningless since the enketo times are only precise to the minute anyways.

Abby-Wheelis:
I just logged in, filling out the survey, added an activity, forced sync from Profile, and then logged out. When I logged back in after that, I had to fill out the survey again and my activity had not saved. My next step is to investigate the log statements related to force syncing to make sure that the data is actually being pushed to the server.

Abby-Wheelis:
After I was debugging for a while (and pressed the force sync button many times), the survey responses seem to have saved this time. So maybe it's fine? 

This is something that I would plan to keep a close eye on during staging in any case.

Abby-Wheelis:
This is reliant on #1049 (leveraged to get i18next working in testing) so it would be better for that branch to be merged first.

www/__mocks__/cordovaMocks.ts
```diff
@@ -87,6 +87,33 @@ export const mockBEMUserCache = () => {
...
+            TimeUseSurvey: { compatibleWith: 1, 
+              formPath: "https://raw.githubusercontent.com/sebastianbarry/nrel-openpath-deploy-configs/surveys-info-and-surveys-data/survey-resources/data-json/time-use-survey-form-v9.json", 
+              labelTemplate: {en: "{ erea, plural, =0 {} other {# Employment/Education, } }{ da, plural, =0 {} other {# Domestic, } }",
+                              es: "{ erea, plural, =0 {} other {# Empleo/Educación, } }{ da, plural, =0 {} other {# Actividades domesticas, }}"}, 
+              labelVars: {da: {key: "Domestic_activities", type: "length"},
+                          erea: {key: "Employment_related_a_Education_activities", type:"length"}}, 
+              version: 9}
+          }
+        }
+      }
```
JGreenlee, line 105: I do think it would be nice to extract this 'fake' config to a separate file, and also include more things in it besides just `survey_info`. Like you said, we'll likely want to expand on it later.

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,182 @@
...
+* We retrieve all the records every time instead of caching because of the
+* usage pattern. We assume that the demographic survey is edited fairly
+* rarely, so loading it every time will likely do a bunch of unnecessary work.
+* Loading it on demand seems like the way to go. If we choose to experiment
+* with incremental updates, we may want to revisit this.
+*/
+//   export function loadPreviousResponseForSurvey(dataKey: string) {
+it('loads the previous response to a given survey', () => {
+    //not really sure if I can test this yet given that it relies on an angular service...
+    // loadPreviousResponseForSurvey("manual/demographic_survey");
```
JGreenlee, line 123: Do you want to just wait for https://github.com/e-mission/e-mission-phone/pull/1064 ?

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -14,14 +17,137 @@ export type SurveyOptions = {
...
+    labelTemplate: {[lang: string] : string};
+    labelVars: {[activity: string]: {[key: string]: string, type:string}},
+    version: number;
+    compatibleWith: number;
+  }
+}
+
+/** @type {EnketoSurveyConfig} _config */
+//TODO find a more appropriate way to store this
+let _config: EnketoSurveyConfig;
```
JGreenlee, line 47: Actually, I don't think there's anything wrong with this. It's a pattern I've been using in the other rewrites.

Since this file is a module (ie it uses import/export syntax), `_config` is scoped to this file; it is not polluting the global scope.

I do usually put the variable (`_config` in this case) right above the function that uses it.

www/__mocks__/cordovaMocks.ts
Abby-Wheelis: Ok, I'll work on doing that. I'm a little worried about setting it up in a way that doesn't work with the next application that tries to use it, but we can always update. I got this section by reverse-engineering from log statements, but can base the rest on one of the `nrel-deploy-configs`.

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: yes, I think that will allow me to make some progress testing this function

www/js/survey/enketo/enketoHelper.ts
Abby-Wheelis: Ok, good to know! I didn't see a great way around it, since I knew `useState` wasn't the answer. I will move it though so the use case is clearer

Abby-Wheelis: Moved and removed the comment in [6ef2725](https://github.com/e-mission/e-mission-phone/pull/1063/commits/6ef2725b21a5bd613752904b93af346fd543702a)

www/__mocks__/cordovaMocks.ts
Abby-Wheelis: I moved this into it's own file and added the rest of the fields in [b480279](https://github.com/e-mission/e-mission-phone/pull/1063/commits/b48027958eb25a1e2f79f75dec53abd3f93a2ab8). I placed the new file in the `__mocks__` folder, but if it would be better somewhere else I can move it.

JGreenlee:
Since this PR is open now and it deals with Enketo, it seems like a good opportunity to add support for client-side XML -> JSON transforms (which we've been wanting to do for awhile now!)
This would allow us to supply the Enketo surveys as either JSON *or* XML and make it easier for new programs to have their own surveys.

I'm going to add a couple of commits here.

JGreenlee:
> Since this PR is open now and it deals with Enketo, it seems like a good opportunity to add support for client-side XML -> JSON transforms (which we've been wanting to do for awhile now!) This would allow us to supply the Enketo surveys as either JSON _or_ XML and make it easier for new programs to have their own surveys.
> 
> I'm going to add a couple of commits here.

This works!

I tested this by modifying the dev-emulator-timeuse config to have 2 of the surveys use an XML `formPath`.
Once it was working, I used `console.time` to see if performing the XML -> JS transformation added any significant delay to the launch of surveys.

On the time use survey, the transformation took about 30ms on average. On the demographics survey (which is a fairly long survey), it was about 60ms. I couldn't test the TripEndSurvey because the XML for that one is missing from https://github.com/e-mission/nrel-openpath-deploy-configs/tree/surveys-info-and-surveys-data/survey-resources/data-xml

I think this is worth it because it makes the process significantly easier partners making their own survey and us setting up config files. If the 30-60 ms delay is noticed in real-world use, we can find a way to do the transformation ahead of time.

JGreenlee:
I tried to test the XML-JS transformation in Jest to no avail. I think `enketo-transformer/web` relies on browser-specific APIs which can't be used in Jest.

Abby-Wheelis:
right now, Jest testing for `enketoHelper` is failing because a file is failing to be parsed - it seems like I'm going to need to update the config so that `node_modules/enketo-transformer` gets transformed? 
Screenshot 2023-11-10 at 10 39 03 AM

Abby-Wheelis:
>I'm wondering if it could be some issue that the variable is the promise and not just the array?

Comparing against what happens on my `metrics_services_rewrite` branch I think that's probably exactly the problem, since I do not get that error on that branch, and when I look at the variable in a breakpoint, it is an empty array and not a promise -- next to find the difference and work out a fix

Abby-Wheelis:
I was able to find the source of the promise being stored instead of the variable. In `enketoHelper`, `filterByNameAndVersion

www/__mocks__/cordovaMocks.ts
```diff
@@ -100,11 +101,22 @@ export const mockBEMUserCache = () => {
       );
     },
     getDocument: (key: string, withMetadata?: boolean) => {
-      return new Promise<any[]>((rs, rj) =>
-        setTimeout(() => {
-          rs(_storage[key]);
-        }, 100),
-      );
+      // this was mocked specifically for enketoHelper's use, could be expanded if needed
+      const fakeSurveyConfig = fakeConfig;
```
JGreenlee, line 105: One way of expanding the way we mock configs might be to accept the fakeConfig as an optional parameter of `mockCordova`. That way, we specify what config to use *from the test*, not from this shared code.

JGreenlee: Because the way it is right now, using fakeSurveyConfig affects all tests, even ones that this config might not make sense for.

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,245 @@
...
+import { mockLogger } from '../__mocks__/globalMocks';
+
+import initializedI18next from '../js/i18nextInit';
+window['i18next'] = initializedI18next;
+
+mockBEMUserCache();
+mockLogger();
+
+global.URL = require('url').URL;
+global.Blob = require('node:buffer').Blob;
```
JGreenlee, line 20: I'm curious about this - why was it needed?
Do Node environments not have these as globals but browser environments do?

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,245 @@
...
+  const badTimeDoc = xmlParser.parseFromString(badTimes, 'text/xml');
+  expect(resolveTimestamps(badTimeDoc, timelineEntry)).toBeUndefined();
+  //good info returns unix start and end timestamps -- TODO : address precise vs less precise?
+  const timeSurvey =
+    '<tag> <Start_date>2016-07-25</Start_date> <End_date>2016-07-25</End_date> <Start_time>17:24:32.928-06:00</Start_time> <End_time>17:30:31.000-06:00</End_time> </tag>';
+  const xmlDoc = xmlParser.parseFromString(timeSurvey, 'text/xml');
+  expect(resolveTimestamps(xmlDoc, timelineEntry)).toMatchObject({
+    start_ts: 1469492672.928242,
+    end_ts: 1469493031,
+  });
```
JGreenlee, line 96: I want to make sure one of the test cases is covered here:

When times from the survey response are within 1 minute of the timelineEntry's start_ts/end_ts, it should use the timelineEntry's start_ts/end_ts, otherwise it should use the times from the survey response.

Is that what is being tested here? It's a little hard to tell because start_ts, end_ts are in seconds and Enketo times are in HH:MM:SS

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,245 @@
...
+//resolve label
+it('resolves the label', async () => {
+  const xmlParser = new window.DOMParser();
+  const xmlString = '<tag> <Domestic_activities> option_1 </Domestic_activities> </tag>';
+  const xmlDoc = xmlParser.parseFromString(xmlString, 'text/html');
+  const xmlString2 =
+    '<tag> <Domestic_activities> option_1 </Domestic_activities> <Employment_related_a_Education_activities> option_3 </Employment_related_a_Education_activities> </tag>';
+  const xmlDoc2 = xmlParser.parseFromString(xmlString2, 'text/xml');
+
+  //if no template, returns "Answered" TODO: find a way to engineer this case
```
JGreenlee, line 108: To engineer this case you would need a different mock config

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,245 @@
...
+    },
+  };
+  const badForm = {
+    getDataStr: () => {
+      return '<aDxjD5f5KAghquhAvsormy xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" id="time_use_survey_form_v9_1"><start>2023-10-13T15:05:48.890-06:00</start><end>2023-10-13T15:05:48.892-06:00</end><group_hg4zz25><Start_date>2016-08-25</Start_date><Start_time>17:24:32.928-06:00</Start_time><End_date>2016-07-25</End_date><End_time>17:30:31.000-06:00</End_time><Activity_Type>personal_care_activities</Activity_Type><Personal_Care_activities>doing_sport</Personal_Care_activities><Employment_related_a_Education_activities/><Domestic_activities/><Recreation_and_leisure/><Voluntary_work_and_care_activities/><Other/></group_hg4zz25><meta><instanceID>uuid:dc16c287-08b2-4435-95aa-e4d7838b4225</instanceID><deprecatedID/></meta></aDxjD5f5KAghquhAvsormy>';
```
JGreenlee, line 133: Add comment explaining why this form is bad

www/__mocks__/cordovaMocks.ts
```diff
@@ -117,6 +129,16 @@ export const mockBEMUserCache = () => {
...
       }
     },
+    getAllTimeQuery: () => {
+      return { key: 'write_ts', startTs: 0, endTs: Date.now() / 1000 };
+    },
+    getSensorDataForInterval: (key, tq, withMetadata) => {
+      return new Promise<any>((rs, rj) =>
+        setTimeout(() => {
+          rs({ metadata: { write_ts: '1699897723' }, data: 'completed', time: '01/01/2001' });
+        }, 100),
```
JGreenlee, line 139: We should only resolve with this if the key is `manual/demographic_survey`

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,245 @@
...
+        version: '1', //survey version
+        xmlResponse: '<this is my xml>', //survey response XML string
+        jsonDocResponse: 'this is my json object', //survey response JSON object
+      },
+      metadata: {},
+    },
+  ];
+
+  //several responses -> only the one that has a name match
+  expect(filterByNameAndVersion('TimeUseSurvey', responses)).resolves.toStrictEqual(response);
```
JGreenlee, line 244: I see that responses are filtered if they aren't for TimeUseSurvey.
The more important thing to test here, though, is version compatibility

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -69,15 +245,14 @@ export function saveResponse(surveyName: string, enketoForm: Form, appConfig, op
...
         if (timestamps === undefined) {
           // timestamps were resolved, but they are invalid
           return new Error(i18next.t('survey.enketo-timestamps-invalid')); //"Timestamps are invalid. Please ensure that the start time is before the end time.");
         }
         // if timestamps were not resolved from the survey, we will use the trip or place timestamps
-        timestamps ||= opts.timelineEntry;
-        data.start_ts = timestamps.start_ts || timestamps.enter_ts;
-        data.end_ts = timestamps.end_ts || timestamps.exit_ts;
+        data.start_ts = timestamps.start_ts || opts.timelineEntry.enter_ts;
+        data.end_ts = timestamps.end_ts || opts.timelineEntry.exit_ts;
```
JGreenlee, line 255: I was curious about this change and briefly investigated to make sure it wouldn't cause issues.
I found that it was done while resolving merge conflicts when the Prettier reformat was merged in. `timestamps` never gets reassigned, but it is not used after this, so the behavior of the code will not change. It's fine like this (and cleaner anyway)

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -45,6 +168,60 @@ export function getInstanceStr(xmlModel: string, opts: SurveyOptions): string |
...
+  /* Enketo survey time inputs are only precise to the minute, while trips/places are precise to
+    the millisecond. To avoid precision issues, we will check if the start/end timestamps from
+    the survey response are within the same minute as the start/end or enter/exit timestamps.
+    If so, we will use the exact trip/place timestamps */
+  const entryStartTs = timelineEntry.start_ts || timelineEntry.enter_ts;
+  const entryEndTs = timelineEntry.end_ts || timelineEntry.exit_ts;
+  if (additionStartTs - (additionStartTs % 60) == entryStartTs - (entryStartTs % 60))
+    additionStartTs = entryStartTs;
+  if (additionEndTs - (additionEndTs % 60) == entryEndTs - (entryEndTs % 60))
+    additionEndTs = entryEndTs;
```
JGreenlee, line 216: This is the part I was talking about that I want to make sure is being tested

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -15,14 +19,133 @@ export type SurveyOptions = {
...
+export function _lazyLoadConfig() {
+  if (_config !== undefined) {
+    return Promise.resolve(_config);
+  }
+  return getConfig().then((newConfig) => {
+    logInfo('Resolved UI_CONFIG_READY promise in enketoHelper, filling in templates');
+    _config = newConfig.survey_info.surveys;
+    return _config;
+  });
+}
```
JGreenlee, line 110: After analyzing a bit further, I think this function is pointless.
`getConfig()` is already cached/lazyloaded, so we could just call it directly. The only difference is that `getConfig()` returns the whole config obj while this function only returns the `survey_info.surveys` sub-field of the config obj.

I think it would be clearer to remove this function. Anywhere it was used, call `getConfig()` and access the subfields from there.
Then move your `EnketoSurveyConfig` type def to `js/types/appConfigTypes.ts` (it doesn't exist yet on this branch but I started it on your other PR https://github.com/e-mission/e-mission-phone/pull/1098)
The reasoning is that eventually we're going to want a single source of truth for the entire config – not scattered typings for each sub-field of the config.

JGreenlee (changes requested):

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: I remember getting an error about this, I don't remember exactly what the error was, but I found that the best way to fix it was to set it like this. [comment about this](https://github.com/e-mission/e-mission-phone/pull/1063#issuecomment-1762249637). I think the issue had something to do with the testing environemnt

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: I'm not sure I knew there was two different cases for what timestamps to use, I did write these tests a while back though, I will verify that I test both cases (timelineEntry and survey response) and add comments to clarify the cases

www/__mocks__/cordovaMocks.ts
Abby-Wheelis: The parameter option would also help me test more cases, right? I think that sounds like a good plan

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: I had noted that, but had not tested it yet, this has now been tested (and commented as such) [8f7c674](https://github.com/e-mission/e-mission-phone/pull/1063/commits/8f7c6748ed161e248e247cc04514dacc46908317)

www/__mocks__/cordovaMocks.ts
Abby-Wheelis: done in [8b3ab76](https://github.com/e-mission/e-mission-phone/pull/1063/commits/8b3ab76327d93011e2a31cfa6409bcb23f710e1f)

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: done in [7ec575e](https://github.com/e-mission/e-mission-phone/pull/1063/commits/7ec575e41701725723609c9eb40654ef8d05385b)

www/js/survey/enketo/enketoHelper.ts
Abby-Wheelis: I made those changes in [d123b4f](https://github.com/e-mission/e-mission-phone/pull/1063/commits/d123b4f23f90537671e13e59ff672f7156e655c2), including moving the type. I did not work the type into the existing appConfig type yet though

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: I just added a case in [100eefb](https://github.com/e-mission/e-mission-phone/pull/1063/commits/100eefb76dd7bd118be1f78b8dd003a11c924ac2)

www/__mocks__/cordovaMocks.ts
Abby-Wheelis: I just added the parameter option in [22decbf](https://github.com/e-mission/e-mission-phone/pull/1063/commits/22decbfd0b58776450737ba536dc9bf42e375aa5), running smoothly for now! Going to see if I can leverage this to add some cases to `resolveLabel` tests

Abby-Wheelis: I was able to add more test cases in [80aa5ea](https://github.com/e-mission/e-mission-phone/pull/1063/commits/80aa5ea18081849819236d65dc5d07ea583aa8e8)!

www/__tests__/enketoHelper.test.ts
Abby-Wheelis: implemented the parameter in the mock, added test cases in [80aa5ea](https://github.com/e-mission/e-mission-phone/pull/1063/commits/80aa5ea18081849819236d65dc5d07ea583aa8e8)

www/js/types/appConfigTypes.ts
```diff
@@ -0,0 +1,22 @@
+// WIP: type definitions for the 'dynamic config' spec
+// examples of configs: https://github.com/e-mission/nrel-openpath-deploy-configs/tree/main/configs
+
+export type AppConfig = {
+  server: ServerConnConfig;
+  [k: string]: any; // TODO fill in all the other fields
+};
```
JGreenlee, line 7: Might as well add survey_info to the AppConfig type since you already typed the survey config
```suggestion
export type AppConfig = {
  server: ServerConnConfig;
  survey_info: {
    'trip-labels': 'MULTILABEL' | 'ENKETO';
    surveys: EnketoSurveyConfig;
  }
  [k: string]: any; // TODO fill in all the other fields
};
```

JGreenlee (changes requested):

www/__mocks__/cordovaMocks.ts
```diff
@@ -34,7 +35,7 @@ export const mockFile = () => {
 //for consent document
 const _storage = {};
 
-export const mockBEMUserCache = () => {
+export const mockBEMUserCache = (config) => {
```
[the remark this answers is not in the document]
Abby-Wheelis: That is a much better idea and definitely what I should have done in the first place

www/__mocks__/cordovaMocks.ts
```diff
@@ -100,11 +101,20 @@ export const mockBEMUserCache = () => {
...
-      return new Promise<any[]>((rs, rj) =>
-        setTimeout(() => {
-          rs(_storage[key]);
-        }, 100),
-      );
+      //returns the config provided as a paramenter to this mock!
+      if (key == 'config/app_ui_config') {
+        return new Promise<any>((rs, rj) =>
+          setTimeout(() => {
+            rs(config);
```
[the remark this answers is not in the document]
Abby-Wheelis: I went ahead and added the fakeConfig as the fallback, that way it can be in place and only need to be passed in if you want to test special cases (as I was wanting to do with enketoHelper)

www/__tests__/clientStats.test.ts
```diff
@@ -6,12 +6,13 @@ import {
...
   statKeys,
 } from '../js/plugin/clientStats';
+import fakeConfig from '../__mocks__/fakeConfig.json';
 
 mockDevice();
 // this mocks cordova-plugin-app-version, generating a "Mock App", version "1.2.3"
 mockGetAppVersion();
 // clientStats.ts uses BEMUserCache to store the stats, so we need to mock that too
-mockBEMUserCache();
+mockBEMUserCache(fakeConfig);
```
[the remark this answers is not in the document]
Abby-Wheelis: After adding the fallback, I took the parameter (and the import) out everywhere except `enketoHelper`

JGreenlee (approved):
LGTM!

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -45,21 +149,81 @@ export function getInstanceStr(xmlModel: string, opts: SurveyOptions): string |
...
+  let endTime = xmlDoc.getElementsByTagName('End_time')?.[0]?.innerHTML;
+
+  // if any of the fields are missing, return null
+  if (!startDate || !startTime || !endDate || !endTime) return null;
+
+  const timezone =
+    (timelineEntry as CompositeTrip).start_local_dt?.timezone ||
+    (timelineEntry as ConfirmedPlace).enter_local_dt?.timezone ||
+    (timelineEntry as CompositeTrip).end_local_dt?.timezone ||
+    (timelineEntry as ConfirmedPlace).exit_local_dt?.timezone;
```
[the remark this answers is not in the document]
Abby-Wheelis: we were, but we aren't now!

www/__tests__/enketoHelper.test.ts
```diff
@@ -0,0 +1,340 @@
...
+      Start_time: '17:24:32.928-06:00',
+      End_date: '2016-07-25',
+      End_time: '17:30:31.000-06:00',
+    },
+  };
+
+  //if no xmlModel, returns null
+  expect(getInstanceStr(null, opts)).toBe(null);
+  //if there is a prefilled survey, return it
+  expect(getInstanceStr(xmlModel, opts)).toBe(filled);
```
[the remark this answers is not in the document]
Abby-Wheelis: So, one that is already filled, but does not have the timestamps filled?

I'm a little confused what you mean, but maybe as I comb through the codecov report I can figure out where the gap you're referring to is.

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -109,7 +273,17 @@ export function loadPreviousResponseForSurvey(dataKey: string) {
...
   const getMethod = window['cordova'].plugins.BEMUserCache.getSensorDataForInterval;
-  return getUnifiedDataForInterval(dataKey, tq, getMethod).then((answers) =>
-    _getMostRecent(answers),
+  return getUnifiedDataForInterval(dataKey, tq, getMethod).then((responses) =>
+    _getMostRecent(responses),
   );
 }
+
+export async function fetchSurvey(url: string) {
+  const responseText = await fetchUrlCached(url);
```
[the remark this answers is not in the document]
Abby-Wheelis: This function fetches the survey, not the response (ie this is what is handling the xml/json differences) that @JGreenlee  added. Our previous response function [`loadPreviousResponseForSurvey`] does re-fetch every time.

PR #1064, opened 2023-10-12
🏗️🗃️ UnifiedDataLoader rewrite

the-bay-kay:
UnifiedDataLoader is the last remaining service in `/www/services.js` to be rewritten, as discussed in [Issue 977](https://github.com/e-mission/e-mission-docs/issues/977). The remaining services in this file (Chats and ReferHelper) are no longer being used, so my plan is to remove these and then convert `services.js` into `unifiedDataLoader.ts`.  

I did have an open question about the services.js rewrite - once all of the rewritten services are together, (commHelper, controlHelper, and unifiedData), should these be relocated into a directory other than `/www/js/`? I'm not sure whether these would be moved into a new `services/` directory, sorted into the current folders, or kept as they are.  I figured it was worth thinking about how we wanted to organize these going forward, since the `/www/js/` has expanded a bit!


the-bay-kay:
Just a quick observation about the behavior or `getUnifiedMessagesForInterval`: When called by `/www/js/survey/enketo/enketoHelper.ts`, the return value (serverResponse.phone_data) always seems to return blank.  I.e., the following snippet:
```
var remotePromise = getRawEntries([key], tq.startTs, tq.endTs)
          .then(function(serverResponse: serverData) {
            console.log('-->', JSON.stringify(serverResponse.phone_data));
            return serverResponse.phone_data;
          });
```
Results in the output:
```
[0] [phonegap] [console.log] --> []
```
Should this always be empty?  I've tried this on a few different emulators, timelines, and surveys: all of which had the same result.  If this isn't intentional, I can look into it further - regardless, I'll be committing and pushing my rewrite soon!

the-bay-kay:
From the manual tests I've done, the rewrite seems to be performing well!  I did have two more questions:
- The older code mentions refactoring the two `getUnified` methods into one generalized method.  Is it worth trying to do this, while I'm already re-writing the functions? I'll go ahead and try to rewrite these tomorrow morning.

- Under what context is `getUnifiedSensorDataForInterval()`  called? From what I can tell, it's used in `timelineHelper.ts` for the `getAllUnprocessedInputs()` method, which in turn is called in the `LabelTab` components.  After some time experimenting with different emulators and timelines, I could never get hit the breakpoints I put on these methods.  Any advice on testing these functions would be greatly appreciated!

the-bay-kay:
As discussed in PR #1052, switching the branches cleans the git logs to only contain my edits - I also updated the title, since I accidentally used a duplicate in icons : )

the-bay-kay:
Quick Update: Now that the `unifiedData` functions have been generalized, I'm working on generalizing `combinePromise` so that it can iterate over a list of promises (Like the original To-Do comment suggested).  

My general approach has been to recursively build up the promises.  For example, if we had an array of promises `[p1, p2, p3]`, they would be combined in the order `(p3+p2) + p1`.  Since this function takes two promises as input, and returns a promise, I think this is a relatively intuitive change - that being said, I've had some difficulty formatting the base case.  

I'm pretty sure I'm making some jumps in my logic and missing something obvious - I'll spend some more time on the recursive approach, and switch gears if that doesn't go anywhere.

the-bay-kay:
image
All tests have been passed successfully! Will be testing on Android before marking as Ready for Review.

shankari:
> I did have an open question about the services.js rewrite - once all of the rewritten services are together, (commHelper, controlHelper, and unifiedData), should these be relocated into a directory other than /www/js/? I'm not sure whether these would be moved into a new services/ directory, sorted into the current folders, or kept as they are. I figured it was worth thinking about how we wanted to organize these going forward, since the /www/js/ has expanded a bit!

Good thought! I vote for moving them into a new services/ directory. I think that the directories are within `www/js` (e.g. `www/js/diary`) though

shankari:
> When called by /www/js/survey/enketo/enketoHelper.ts, the return value (serverResponse.phone_data) always seems to return blank. I.e., the following snippet:

What is the key that the `enketoHelper` is trying to read? If it is the demographic result, it is not included in the fake data that we load, so it will always be empty. To check if it is non-empty, you need to:
- save the onboarding survey and "end trip and force sync" OR
- use a stage/production token

If the key is a trip or place confirm or an addition, it may just be that your data has not been pushed to the server. Pick a stage token that is related to time use (I think I have one in the wiki) and test; you should see some results.

Please report the results of this testing before I review further.

shankari:
> Edit: I may be wrong, but I believe these are called during the initial loading and refresh processes! I've also gone ahead and rewritten getUnified, in https://github.com/e-mission/e-mission-phone/commit/36cfff64668d819576c276a97f1c8052f332063f

Initial loading of the label screen, yes.

the-bay-kay:
> Good thought! I vote for moving them into a new services/ directory. I think that the directories are within `www/js` (e.g. `www/js/diary`) though

I went ahead and created a `www/js/services/` directory, and moved the rewrites into there! While I can't view the exact merge conflict that popped up, I'm hoping it should be an easy fix of changing the imports : )  I'll go ahead and move `controlHelper` in my other rewrite into the same directory!

shankari:
@the-bay-kay I just merged in the onboarding changes. You need to resolve merge conflicts

www/js/services.js
```diff
@@ -215,78 +117,4 @@ angular.module('emission.services', ['emission.plugin.logger'])
       return window.cordova.plugins.BEMConnectionSettings.getSettings();
     };
 
-})
-
-.factory('Chats', function() {
-  // Might use a resource here that returns a JSON array
-
-  // Some fake testing data
```
shankari, line 223: Thanks for doing this! FYI, this was part of the original ionic starter project that I used to create the the very first version of the app. I kept it around for a bit as an example, and then nobody ever had the time/chutzpah to clean it up 😄 

I'm glad we are finally getting around to it!

shankari (changes requested):
Looks good at the high level - just a few cleanup changes and a couple of fairly biggish ones where I am not 100% how the tests pass (or maybe I don't understand it myself).

www/js/services/unifiedDataLoader.ts
```diff
@@ -0,0 +1,103 @@
...
+      const firstPromise = promiseList[0];
+      const nextPromise = combinedPromises(promiseList.slice(1), combiner);
+     
+      firstPromise.then(function(currentFirstResult: Array<any>) {
+        firstResult = currentFirstResult;
+        firstPromiseDone = true;
+      }, function(error) {
+        firstResult = [];
+        firstError = error;
+        nextPromiseDone = true;
```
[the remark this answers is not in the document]
the-bay-kay: I believe you're right, thanks! I'll spend some time tweaking the tests to see why that didn't get caught.

www/js/services.js
```diff
@@ -215,78 +117,4 @@ angular.module('emission.services', ['emission.plugin.logger'])
...
-    id: 8,
-    name: 'Perry Governor',
-    lastText: 'Look at my mukluks again!',
-    face: 'img/perry.png'
-  }, {
-    id: 9,
-    name: 'Mike Harrington',
-    lastText: 'This is wicked good ice cream again.',
-    face: 'img/mike.png'
-  }];
```
JGreenlee, line 274: There are a bunch of unused images in `www/img` that might be appropriate to also remove at this point in time

JGreenlee: Not just the 'chats' images, but also remnants of Habitica, calorie-counting, and some random gifs.

I think the only image still needed is `enketo_bare_150x56.png`.

the-bay-kay:
Just noticed the failed check - I'm unsure why the`LoadMoreButton` tests are failing.  I'll go ahead and look into that, as the tests are also failing on my `controlHelper` branch.  Here's the error:
image

JGreenlee:
Is this still being worked on? Is it waiting for review?

the-bay-kay:
> Is this still being worked on? Is it waiting for review?

I had been waiting to run the `prettier` setup, but it'll be ready for review after I make one more small adjustment!  I'll go ahead and put it under your 'for-review' shortly : )

the-bay-kay:
At the time of writing, this does not pass the 'prettier' check because of `index.html` - please see PR #1096 for more info!

www/js/services/unifiedDataLoader.ts
```diff
@@ -0,0 +1,125 @@
...
+  });
+};
+
+/**
+ * combinedPromises is a recursive function that joins multiple promises
+ * @param promiseList 1 or more promises
+ * @param combiner a function that takes two arrays and joins them
+ * @returns A promise which evaluates to a combined list of values or errors
+ */
+export const combinedPromises = function (
```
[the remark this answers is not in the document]
JGreenlee: I don't think it's necessary to do this before merging this PR, but it might be good to keep around as a 'future change'

www/js/types/serverData.ts
JGreenlee, line 1: Great work here - these type defs will surely be helpful in the future

JGreenlee (changes requested):

www/js/services.js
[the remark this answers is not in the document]
the-bay-kay: Oh, oops - totally! I'll go ahead and remove it now.

www/js/services/unifiedDataLoader.ts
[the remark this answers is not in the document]
the-bay-kay: I didn't know `Promise.allSettled` was an option, cool! I'll spend a bit of time seeing if I can get that to work : )

www/js/services/unifiedDataLoader.ts
```diff
@@ -0,0 +1,68 @@
+import { getRawEntries } from './commHelper';
+import { ServerResponse, ServerData, TimeQuery } from '../types/serverData';
+
+/**
+ * removeDup is a helper function for combinedPromises
+ * @param list An array of values from a BEMUserCache promise
+ * @returns an array with duplicate values removed
+ */
+export const removeDup = function (list: Array<ServerData<any>>) {
```
JGreenlee, line 9: Note to self: we do the same deduplication here as we do in https://github.com/e-mission/e-mission-phone/pull/1086/files#diff-00cfb43594908885f4c2e9121a2224b2c080d9b96bff2e09ec6d4f8d8c6b0645R102-R104

Once both are merged it would be nice to use a common function in both places

www/js/services/unifiedDataLoader.ts
```diff
@@ -0,0 +1,68 @@
...
+) {
+  if (promiseList.length === 0) {
+    throw new RangeError('combinedPromises needs input array.length >= 1');
+  }
+  return new Promise(function (resolve, reject) {
+    Promise.allSettled(promiseList).then(
+      (results) => {
+        var allRej = true;
+        var values = [];
+        var rejections = [];
```
JGreenlee, line 30:
```suggestion
        let allRej = true;
        const values = [];
        const rejections = [];
```

the-bay-kay: I had no idea a const array could still be changed with its member functions, interesting!

JGreenlee: Yeah `const` just means it can't be reassigned. So you can modify `const` objects all you want as long as that object keeps the same address in memory.

JGreenlee (approved):
LGTM

shankari:
@the-bay-kay this now has merge conflicts *again*. I am going to try to resolve them to avoid an infinite cycle of changes.

shankari:
Ah, it failed because of prettier and I can't see anything obviously wrong with the file that it is flagging...

JGreenlee:
I would fix it tonight if I could, but I don't have write access to this one

JGreenlee:
@shankari Looks like the only file not passing Prettier is `diaryTypes.ts`. So if you wish to merge this tonight anyway, I can take care of `diaryTypes.ts` tomorrow in https://github.com/e-mission/e-mission-phone/pull/1086 and it will all even out

the-bay-kay:
>Looks like the only file not passing Prettier is `diaryTypes.ts`. So if you wish to merge this tonight anyway, I can take care of `diaryTypes.ts` tomorrow in #1086 and it will all even out

If you'd like, I can fetch the current code, pull from `service_rewrites_2023` to resolve the merge conflicts, and then run prettier on the whole thing before re-pushing the code!

shankari:
@the-bay-kay as you can see, there is a failing test and it is *not* `LoadMoreButton`. Can you please fix before I review?

the-bay-kay:
> @the-bay-kay as you can see, there is a failing test and it is _not_ `LoadMoreButton`. Can you please fix before I review?

Fixed! I missed a couple imports in the merge, all tests are passing now.

shankari (approved):
This touches 46 files, so I am merging it now. But I expect that this will be cleaned up in your next PR.

PR #1055, opened 2023-10-05
Deault label options

sebastianbarry:
# Added Default Label Options behavior

_Configs without `label_options` value now use a default locally stored `label-options.json.sample` file to generate label options - **no more `i81nUtils`**_

-------

confirmHelper.ts
- Removed archaic behavior or using old Angular i18nUtils module to get the filename of the translations' trip_confirm_options.json file
- The "else" default "no label_options found in config" behavior is almost identical to the if(appConfig.label_options), but it just uses the default label options URL at label-options.json.sample
- Pulled the language-specific text handling behavior out of the if statement, because either way (label_options or no label_options) the JSON data model will look the same

label-options.json.sample
- Created new file
- Replaces trip_confirm_options.json.sample (this location in confirmHelper.ts was the only place in the codebase using this file)
- Modeled after https://github.com/e-mission/nrel-openpath-deploy-configs/blob/main/label_options/example-program-label-options.json
- Only has translations for EN and ES

trip_confirm_options.json.sample
- Removed this file

PR #1065, opened 2023-10-12
🔖 [PR] New default label options and translations

sebastianbarry:
This PR is associated with https://github.com/e-mission/e-mission-docs/issues/1007

This PR is replacing https://github.com/e-mission/e-mission-phone/pull/1055, as that branch was having issues loading the app that I wasn't sure how to troubleshoot (likely something new changed since oct5 - oct9 commits from @JGreenlee and @shankari):
image

This new branch works, and includes everything from the old PR but adds the function of falling back to checking i18next translations in a new field named "multilabel" for label-options _if the translations are not defined in the provided dynamic (or default, technically) label-options file_.


sebastianbarry:
Testing:

| Scenario | App Screenshot | Code screenshot |
|---|---|
| I added a new `"example"` MODE which gathers its translation from label-options.json.sample by default | Simulator Screenshot - iPhone 13 - 2023-10-12 at 13 40 18 | image |
| | | image |
| I removed the translation from label-options.json.sample and it defaulted to the example translation I added to en.json | Simulator Screenshot - iPhone 13 - 2023-10-12 at 13 41 38 | image |

sebastianbarry:
Ready for review

JGreenlee:
This PR cannot be reviewed until `service_rewrite_2023` is in sync with `onboarding_routing_sept_2023`.
Once synced, we'll see if the merge conflicts are still there - if so we'll squash

www/js/survey/multilabel/confirmHelper.ts
```diff
@@ -36,29 +36,24 @@ export async function getLabelOptions(appConfigParam?) {
...
+    const defaultLabelOptionsJson = await fetchUrlCached(defaultLabelOptionsURL);
+    labelOptions = JSON.parse(defaultLabelOptionsJson) as LabelOptions;
+  }
+  /* fill in the translations to the 'text' fields of the labelOptions,
+    according to the current language */
+  const lang = i18next.language;
+  for (const opt in labelOptions) {
+    labelOptions[opt]?.forEach?.((o, i) => {
+      const translationKey = o.value;
+      // If translation exists in i18next, use that. Otherwise, use the one in the labelOptions.
```
JGreenlee, line 53: It's the other way around. labelOptions first, then i18next.
The code is correct but the comment is not

www/i18n/en.json
```diff
@@ -146,6 +146,41 @@
...
+        "not_a_trip": "Not a trip",
+        "no_travel": "No travel",
+        "home": "Home",
+        "work": "To Work",
+        "at_work": "At Work",
+        "school": "School",
+        "transit_transfer": "Transit transfer",
+        "shopping": "Shopping",
+        "meal": "Meal",
+        "pick_drop_person": "Pick-up/ Drop off Person",
```
JGreenlee, line 174: While we are rewriting and have this opportunity, can we make all the text with slashes look `"like / this"` instead of `"like/ this"`?
I believe they were made like that a while ago to force a line break. But we don't do that anymore in the UI so the slashes end up looking weird.

JGreenlee (changes requested):
The logic looks good. I am requesting a few textual changes.

We are going to need a PR to `e-mission-translate` adding the new keys to `es.json` and `lo.json`. Feel free to either do that yourself or coordinate with @Abby-Wheelis since she's been handling translations recently.

JGreenlee:

www/js/survey/multilabel/confirmHelper.ts
sebastianbarry, line 53:
```suggestion
      // If translation exists in labelOptions, use that. Otherwise, use the one in the i18next.
```

www/i18n/en.json
sebastianbarry: I changed the slashes to have spaces before / after in `i18n/en.json` and `json/label-options.json.sample` https://github.com/e-mission/e-mission-phone/pull/1065/commits/36b56daad2be26a6ce0ed1241797f6048707a65e

www/js/survey/multilabel/confirmHelper.ts
sebastianbarry: I fixed this comment: https://github.com/e-mission/e-mission-phone/pull/1065/commits/36b56daad2be26a6ce0ed1241797f6048707a65e
> // If translation exists in labelOptions, use that. Otherwise, use the one in the i18next.

sebastianbarry:
Last thing left to do is:

- [ ] PR to `e-mission-translate` adding the new keys to `es.json` and `lo.json`

JGreenlee (approved):
LGTM

sebastianbarry:
Created PR to `e-mission-translate` https://github.com/e-mission/e-mission-translate/pull/47

www/js/survey/multilabel/confirmHelper.ts
```diff
@@ -36,29 +36,24 @@ export async function getLabelOptions(appConfigParam?) {
...
+    labelOptions = JSON.parse(defaultLabelOptionsJson) as LabelOptions;
+  }
+  /* fill in the translations to the 'text' fields of the labelOptions,
+    according to the current language */
+  const lang = i18next.language;
+  for (const opt in labelOptions) {
+    labelOptions[opt]?.forEach?.((o, i) => {
+      const translationKey = o.value;
+      // If translation exists in labelOptions, use that. Otherwise, use the one in the i18next.
+      const translation = labelOptions.translations[lang][translationKey] || i18next.t(`multilabel.${translationKey}`);
```
shankari, line 54: This is a cool design, but I don't understand why you have the translations in `label-options.json.sample` in that case.

www/json/label-options.json.sample
```diff
@@ -0,0 +1,124 @@
...
+    {"value":"taxi"},
+    {"value":"bus"},
+    {"value":"train"},
+    {"value":"free_shuttle"},
+    {"value":"other"}
+  ],
+  "translations": {
+    "en": {
+      "walk": "Walk",
+      "e-bike": "E-bike",
```
shankari, line 57: here....

www/i18n/en.json
```diff
@@ -146,6 +146,41 @@
         "no-travel-hint": "To see more, change the filters above or go record some travel!"
     },
 
+    "multilabel":{
+        "walk": "Walk",
+        "e-bike": "E-bike",
+        "bike": "Regular Bike",
```
shankari, line 152: and here as well

shankari (changes requested):
@sebastianbarry this is a cool idea, but I don't understand why you need to have two copies of the translations in this case. If the whole point was to have one "source of truth" and not copy the information around, then there's no reason to have two copies...

sebastianbarry:
> this is a cool idea, but I don't understand why you need to have two copies of the translations in this case. If the whole point was to have one "source of truth" and not copy the information around, then there's no reason to have two copies...

The design principle for having 2 copies of the translations is as follows: There is an underlying layer of multilabel translations, found in the i18next files `en.json`, `es.json`, etc. These translations are the fallback in case any label-options file does not contain the translations portion for a given language. Then above that layer, are the translations from the label-options file. These translations will be the default. Having this default/backup translations is helpful for multiple reasons: 1. If there is a Spanish speaking user who is part of the Lao program that only has translations in Lao and English, they prefer Spanish on their device which does not have translations in the Lao label-options file. In this case, it would fall back onto the `es.json` file and get the translations from there. 2. In USA English we might call a mode of transportation a "Rickshaw", but in Lao English they might call a mode of transportation "Tuk tuk", so we have the ability to have area-specific translations of modes.

This method of implementation does have some (but minimal) copied translations, but it offers a higher level of flexibility because the label-options file only _needs_ to have **special translations**, and any other way of doing it would not offer this level of flexibility with this level of customization.

**That being said @shankari, I see what you're saying about this PR still having translations in the default sample label-options file, so I will remove those.**

> Also, where is the testing for default versus dynamic config, which was the original issue? This only tests for default translations from `label-options` and from `en.json`...

[The testing for default vs. dynamic config was in the old PR that I closed:](https://github.com/e-mission/e-mission-phone/pull/1055#issuecomment-1749303578)

Testing:

| | iPhone 13 iOS 15.4 |
|----|----|
| Labels being loaded locally | Screenshot 2023-10-05 at 10 48 04 AM |
| Labels working | Simulator Screenshot - iPhone 13 - 2023-10-05 at 10 45 20 |
| Labels working | Simulator Screenshot - iPhone 13 - 2023-10-05 at 10 45 23 |

sebastianbarry:
I removed the translations from label-options.json.sample and confirmed that it is still working - it required a change to the logic as well to get the translation, but it works for various purposes now
- label-options has NO translations (gets every translation from i18next)
- label-options has only some translations (first checks label-options, then checks i18next)
- label-options has all the translations (only checks label-options, but still has the backup of i18next in case one is missing or something)

@shankari Moving this back to ready for review

shankari:
> This method of implementation does have some (but minimal) copied translations, but it offers a higher level of flexibility because the label-options file only needs to have special translations, and any other way of doing it would not offer this level of flexibility with this level of customization.

Correct, but before your most recent change, the label-options file had all translations, not just special ones.

sebastianbarry:
> > This method of implementation does have some (but minimal) copied translations, but it offers a higher level of flexibility because the label-options file only needs to have special translations, and any other way of doing it would not offer this level of flexibility with this level of customization.
> 
> Correct, but before your most recent change, the label-options file had all translations, not just special ones.

I added some of my thoughts to this in https://github.com/e-mission/e-mission-translate/pull/47#discussion_r1367564018, adding it here for traction

PR #1069, opened 2023-10-17
🌐 i18n: fix merge translations if structure differs

JGreenlee:
If the structure is different between `lang` and `fallbackLang` such that a key is an object in `fallbackLang` and a string in `lang`, we experience an error when we attempt to assign a property on the string. This change will skip filling in if the key is not an object in *both* `lang` and `fallbackLang`, thus preventing the error.


shankari:
merging to service rewrite to make the prettier reformat easier
everybody should currently be working on the rewrite anyway, so there is no reason why this shouldn't work

PR #1071, opened 2023-10-18
🚑 Patch for showstopping ios popup issue

Abby-Wheelis:
We discovered that some ordering of processes in onboarding is what was causing [#1006](https://github.com/e-mission/e-mission-docs/issues/1006), and are fixing it by altering the ordering back to the original. See more discussion on the issue.


Abby-Wheelis:
I have implemented (I think) the changes Jack was talking about in [this comment](https://github.com/e-mission/e-mission-docs/issues/1006#issuecomment-1769073063). I have confirmed that this patch works by uninstalling the devapp and going through onboarding. The popup only comes up when I click fix, as expected, and looking at log statements confirm that we're stepping through the stages as expected.

Abby-Wheelis:
I just went back through to check, and I am confident that it does resolve the issue. I'm attaching a video that shows the popup waiting until the appropriate time to launch, and also resolving the "not refreshing" issue, as the check does turn green as soon as the permission is fixed. 

https://github.com/e-mission/e-mission-phone/assets/54848919/9a389d11-da08-4c63-ae2d-58d56f2c33bd

Also including the log statements here, you can see how the different held states vary, and how consent is marked at the same time as login. 

Screenshot 2023-10-18 at 4 18 16 PM

PR #1072, opened 2023-10-18
Rewrite StartPrefs

Abby-Wheelis:
This PR will rewrite the service in `splash/startprefs.js`, as noted in [this comment](https://github.com/e-mission/e-mission-docs/issues/977#issuecomment-1767242234). It might also be time to re-consider exactly how consent is handled, as discussed in [this issue](https://github.com/e-mission/e-mission-docs/issues/1016). After the patch in #1071, I anticipate some merge conflicts, since the files overlap, but since I've worked on both changes I'm hopeful to be able to resolve them quickly.


Abby-Wheelis:
The tests are currently failing because of a timeout issue with the React test (for LoadMoreButton) sometimes locally, and it looks like that's the case here as well. There is a way to expand the time limit, which I've done for one or two other tests.

Abby-Wheelis:
One small problem that I'm running into here is that one of the functions relies on `$emit` which I have not run into in the rewrite quite yet. I haven't figured out what this particular emit was desiged to do yet, but I can see where different `emit` events are received with `$on` so I think my next step is to try and find the corresponding `on` for this `emit`.

Abby-Wheelis:
> so I think my next step is to try and find the corresponding on for this emit.

Looking through the code base here, I can't really find where this event is received. What's emitted is "data_collection_consented" which I have been unable to find outside of this file on either `master` or `service_rewrite_2023`. I didn't find too many hints in the blame either. 

Ah! But when I look for `CONSENTED_EVENT` I find two `.on` listeners for it, one in `splash/storedevicesettings.js` and one in `splash/pushnotify.js`. We haven't rewritten either of these services yet, but it seems that this event is the catalyst to register push notifications and store device settings in a re-consent situation. I'm wondering if there would be a way to leverage our onboarding routes / states to set these processes off at the right times in the migrated code? There are a few other places where `emit` and `on` are used for similar purposes, so it would be good to figure out our migration strategy.

Abby-Wheelis:
I was just able to talk through this a bit with @sebastianbarry, and what we settled on was extracting the functionality from the `.on` receivers into functions, exporting them, and then calling them (in this case) when we mark consent. That way, the functions are called at the right time, and don't rely on the angular `emit` and `on`. Even though some of the "receiving" files are still in angular, we can leverage `getAngularService` to access those functions from our React contexts. 

I haven't quite implemented this yet, but it seems like a good migration strategy.

Abby-Wheelis:
This approach seems to be working based on logging in and out of the emulator and watching the log statements. The next step will be to write tests. This will not be very direct, since the file involves a lot of reading and writing from storage, but hopefully I can leverage existing mocks to help with that.

Abby-Wheelis:
After resolving merge conflicts and removing some of the old process, now I'm waiting until we make a final decision on how to handle the "check consent" from profile settings [this comment](https://github.com/e-mission/e-mission-docs/issues/1016#issuecomment-1775445364) before I think this is ready for merge.

Abby-Wheelis:
As decided in [#1016](https://github.com/e-mission/e-mission-docs/issues/1016):
>2) keep "check consent" and if no consent found, have a message to log out and log back in with the same code
>>But until we do that, the app may not be collecting any data because it can't find the consent. So we should go with (ii) until we implement the principled fix.

This is currently implemented:
| If Consented | If NOT Consented |
| ------------- | ------------- |
| Simulator Screenshot - iPhone 13 Pro - 2023-10-25 at 12 09 41|Simulator Screenshot - iPhone 13 Pro - 2023-10-25 at 12 11 03 |

www/js/splash/startprefs.ts
```diff
@@ -0,0 +1,121 @@
+import { storageGet, storageSet } from '../plugin/storage';
+import { logInfo, logDebug, displayErrorMsg } from '../plugin/logger';
+
+type StartPrefs = {
+  CONSENTED_EVENT: string,
+  INTRO_DONE_EVENT: string,
+}
```
JGreenlee, line 7: I don't think this type is necessary. Since `startPrefs` is declared outright right after this, Typescript already knows it's an obj with these 2 props and that the values of the props are `string`.

If you ever need to use that type signature for something else, you can do `typeof startPrefs`

www/js/splash/storedevicesettings.js
```diff
@@ -44,16 +45,17 @@ angular.module('emission.splash.storedevicesettings', ['emission.plugin.logger',
...
       }
-    });
+      })
+    }
 
-    $rootScope.$on(StartPrefs.INTRO_DONE_EVENT, function(event, data) {
+    $rootScope.$on(startPrefs.INTRO_DONE_EVENT, function(event, data) {
           console.log("intro is done -> original consent situation, we should have a token by now -> register");
        storedevicesettings.storeDeviceSettings();
     });
```
JGreenlee, line 61: It looks like `INTRO_DONE_EVENT` is no longer emitted from anywhere. So I don't think you need to listen for it anymore, but I *do* think you need to find a suitable replacement.

From analyzing these 2 events (CONSENTED_EVENT and INTRO_DONE_EVENT) and seeing how the old StartPrefs worked, here's the high-level takeaway I'm getting.
When either:
- a) the intro gets done
- b) the user reconsents (consent gets done and intro was already done)

We call  `storedevicesettings.storeDeviceSettings()` and `pushnotify.registerPush`.

So I think the `afterConsent` mechanism you made should be something more like `afterIntroOrReconsent`, and I think SaveQrPage is the wrong place to do it.
I'm not sure if reconsent is still a thing or will be a thing in the future (I saw there was recent discussion on this). If not, I suppose you'll only need to consider (a) and not (b)

www/js/onboarding/SaveQrPage.tsx
```diff
@@ -23,12 +24,22 @@ const SaveQrPage = ({  }) => {
...
+            setRegisterUserDone(true);
+            preloadDemoSurveyResponse();
+            refreshOnboardingState();
+
+            //fully consented, so can handle other aspects
+            //other plugins - previously used $emit
+            const PushNotify = getAngularService("PushNotify");
+            PushNotify.afterConsent();
+            const StoreSeviceSettings = getAngularService("StoreDeviceSettings");
+            StoreSeviceSettings.afterConsent();
```
JGreenlee, line 40: You can see my other comment below, but I think this is not the right place from which to make these calls

www/js/splash/startprefs.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export const startPrefs: StartPrefs = {
+  CONSENTED_EVENT: "data_collection_consented",
+  INTRO_DONE_EVENT: "intro_done",
+};
```
JGreenlee, line 12: At the end of this, I actually don't think you will need this `startPrefs` variable at all.
Because they're just for emitting angular events and we should be getting rid of those

JGreenlee (changes requested):
I have some suggestions for getting rid of the Angular emit events

www/js/splash/storedevicesettings.js
Abby-Wheelis: We're taking out reconsent for now, as decided in [this comment](https://github.com/e-mission/e-mission-docs/issues/1016#issuecomment-1778476728). Combining the mechanism makes sense, I'll work on combining things and finding a better place for it.

Abby-Wheelis: >a) the intro gets done

One place to catch this would be to add the functionality to the `markIntroDone` method in onboarding helper 

>b) the user reconsents (consent gets done and intro was already done)

I could add the after consent part to `markConsented` in `startPrefs`  - (if Intro is done) this would be a very similar place functionally, but maybe better organizationally? 

@JGreenlee do those places sound better?

JGreenlee: Yes, those 2 places seem appropriate to me

www/js/splash/startprefs.ts
Abby-Wheelis: removed in [a7a75d1](https://github.com/e-mission/e-mission-phone/pull/1072/commits/a7a75d1dde7f5621074a514b9ec9c0a71235e732)

www/js/splash/startprefs.ts
Abby-Wheelis: took that out in [3130741](https://github.com/e-mission/e-mission-phone/pull/1072/commits/313074123bf6c9e5b0470ef210eece70ebd97c41)

www/js/onboarding/SaveQrPage.tsx
Abby-Wheelis: I moved these calls into `markConsented` in `starPrefs`

www/js/splash/pushnotify.js
```diff
@@ -172,16 +173,18 @@ angular.module('emission.splash.pushnotify', ['emission.plugin.logger',
       Logger.log("pushnotify startup done");
     });
 
-    $rootScope.$on(StartPrefs.CONSENTED_EVENT, function(event, data) {
-      console.log("got consented event "+JSON.stringify(event.name)
-                      +" with data "+ JSON.stringify(data));
-      if (StartPrefs.isIntroDone()) {
+    //new way of handling this, called in startprefs by markConsent
```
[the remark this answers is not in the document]
Abby-Wheelis: I took this out, and was careful that comments I left in the new version were accurate

www/js/splash/storedevicesettings.js
Abby-Wheelis: Done! I tested by logging in and out of the emulator and seeing the log statement for the "after intro events" and not the log statement for the "reconsent events". This is expected behavior since reconsent is not what's happening on login, and we don't currently have a reconsent scenario. I'll add some git comments to the new code to flag the changes.

www/js/splash/startprefs.ts
```diff
@@ -0,0 +1,126 @@
...
+      // mark in local storage
+      storageSet(DATA_COLLECTION_CONSENTED_PROTOCOL,
+        _req_consent);
+      // mark in local variable as well
+      _curr_consented = { ..._req_consent };
+    })
+    //check for reconsent
+    .then(readIntroDone)
+    .then((isIntroDone) => {
+      if(isIntroDone) {
```
Abby-Wheelis, line 44: This is where the "reconsent" code is handled.

www/js/onboarding/onboardingHelper.ts
```diff
@@ -58,15 +62,22 @@ export function getPendingOnboardingState(): Promise<OnboardingState> {
...
+export async function readIntroDone() {
   return storageGet(INTRO_DONE_KEY).then((read_val) => !!read_val) as Promise<boolean>;
 }
 
 export async function markIntroDone() {
   const currDateTime = DateTime.now().toISO();
-  return storageSet(INTRO_DONE_KEY, currDateTime);
+  return storageSet(INTRO_DONE_KEY, currDateTime)
+    .then(() => {
+      //handle "on intro" events
```
Abby-Wheelis, line 76: This is where the "after intro" events are called

Abby-Wheelis (commented):
calling out the changes I just made wrt "intro done" and "reconsent"

I also ran the tests and verified that they still pass after my recent changes

www/js/plugin/storage.ts
```diff
@@ -30,7 +30,7 @@ const localStorageSet = (key: string, value: {[k: string]: any}) => {
 
 const localStorageGet = (key: string) => {
   const value = localStorage.getItem(key);
-  if (value) {
+  if (value && value != "undefined") {
```
JGreenlee, line 33: I read the commit for this and I see that the string `"undefined"` was getting stored for one of the tests (not sure which one)

What I want to know is why it got stored as such in the first place. Would it be cleaner to catch this before `"undefined"` ever gets stored and prevent this from happening?

www/js/splash/startprefs.ts
```diff
@@ -0,0 +1,126 @@
...
+        console.debug("reconsent scenario - marked consent after intro done - registering pushnoify and storing device settings")
+        const PushNotify = getAngularService("PushNotify");
+        const StoreSeviceSettings = getAngularService("StoreDeviceSettings");
+        PushNotify.registerPush();
+        StoreSeviceSettings.storeDeviceSettings();
+      }
+    });
+};
+
+let _is_consented;
```
JGreenlee, line 54: Is this variable needed? It is assigned below, but I don't see it being used for anything

JGreenlee (changes requested):

www/js/onboarding/onboardingHelper.ts
```diff
@@ -58,15 +62,22 @@ export function getPendingOnboardingState(): Promise<OnboardingState> {
...
+      console.log("intro done, calling registerPush and storeDeviceSettings");
```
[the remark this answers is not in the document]
Abby-Wheelis: I added all the `logDebug`s in [c54f939](https://github.com/e-mission/e-mission-phone/pull/1072/commits/c54f939f467ca43c3a2fc45796dd3bf5e2cc9f4b)

www/js/splash/startprefs.ts
Abby-Wheelis: You're right, that is assigned but never used, there were a lot of variables like that in the angular code, and I must have missed that this one isn't used anymore, thank you for pointing that out!

www/js/plugin/storage.ts
Abby-Wheelis: The second test is the one that was creating that error, I'll start working through why now

Abby-Wheelis: After digging through the log statements, the error is happening during the all to `markConsented`with the error `SyntaxError: "undefined" is not valid JSON`

Continuing to look at this, and I believe there is a timing issue - in `markConsented`, the first call is to `readConsent`, so that the required consent is read from the config. This, however, launches `checkNativeConsent` in the background, which sometimes times correctly to find the local consent (as it has now been marked) but not the native consent (as that is still in process). This causes an error to display (to the user) and could potentially be confusing. 

However, that is not the problem here, I'm guessing the issue is when `storageGet(DATA_COLLECTION_CONSENTED_PROTOCOL)` and the value is not defined it is for some reason finding the string "undefined" - though that makes very little sense. Maybe with the mark consented it is storing undefined?

Abby-Wheelis: This error: `SyntaxError: "undefined" is not valid JSON` happens specifically the second time `readConsentState` is called, and the error is thrown during the call for `storageGet` for  `data_collection_consented_protocol` -- the first time that `storageGet` call is made, I see log statements about `uc_stored_val should be non-null` and then a message that it is copying native to local ... maybe this is where `"undefined"` get's introduced?

Abby-Wheelis: This copying is happening in `getUnifiedValue` where the native value is undefined, and the local value in null -- I don't think either of them have been set yet, so I'm not sure why we're copying the native undefined over the local null, but maybe that's what we want to do?

Abby-Wheelis: I've now tried checking the types throughout the process, and can't determine why the `undefined` is a string. When it is ready, it's type is undefined, and when it is set, it's type is undefined. But then when we set it, we stringify for storage. So when it is read, the type is a string.

Could I add a check to the set to only set it if it exists? Or be checking in `getUnifiedValue` for undefined alongside null?

Abby-Wheelis: Added this code snippet, and all the tests now run, without the hacky string test.
```
  if (value) {
    localStorage.setItem(key, JSON.stringify(value));
  }
```

JGreenlee (approved):
Great!

www/js/onboarding/onboardingHelper.ts
```diff
@@ -58,15 +62,22 @@ export function getPendingOnboardingState(): Promise<OnboardingState> {
...
 
 export async function markIntroDone() {
   const currDateTime = DateTime.now().toISO();
-  return storageSet(INTRO_DONE_KEY, currDateTime);
+  return storageSet(INTRO_DONE_KEY, currDateTime)
+    .then(() => {
+      //handle "on intro" events
+      logDebug("intro done, calling registerPush and storeDeviceSettings");
+      const PushNotify = getAngularService("PushNotify");
+      const StoreSeviceSettings = getAngularService("StoreDeviceSettings");
```
shankari, line 79: Is this the recommended way to handle this in react? The reason that I had an event-driven implementation before this was to simplify the dependency chain. With this, we are adding dependencies between startPrefs and push notify and store device settings.

www/js/plugin/storage.ts
```diff
@@ -25,7 +25,9 @@ const unmungeValue = (key, retData) => {
 }
 
 const localStorageSet = (key: string, value: {[k: string]: any}) => {
-  localStorage.setItem(key, JSON.stringify(value));
+  if (value) {
+    localStorage.setItem(key, JSON.stringify(value));
+  }
```
shankari, line 30: would you add a comment and a link to the discussion around this here for clarity?

www/js/splash/pushnotify.js
```diff
@@ -15,12 +15,13 @@
 
 import angular from 'angular';
 import { updateUser } from '../commHelper';
+import { readConsentState, isConsented, startPrefs } from './startprefs';
+import { readIntroDone } from '../onboarding/onboardingHelper';
```
shankari, line 19: see, this is actually a circular dependency. onboardingHelper depends on pushnotify which depends on onboardingHelper. Is that not a problem/considered bad practice in react?

www/js/splash/pushnotify.js
shankari, line 1: nit: given that we also have `remotenotify`, we may want to rename this to `pushnotifysettings` and the other to `remotenotifyhandler` or something.

shankari (changes requested):
Great start, just a few comments...

www/js/App.tsx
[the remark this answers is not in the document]
Abby-Wheelis: I managed to get the diffs up to date after merging from upstream and toggling the target branch, so going to resolve these now

www/js/control/ProfileSettings.jsx
```diff
@@ -475,17 +477,13 @@ const ProfileSettings = () => {
...
-                            setNoConsentMessageVis(true)}}>
-                                {t('general-settings.cancel')}
-                        </Button>
-                        <Button onPress={()=>{
-                            setNoConsentVis(false);
-                            // $state.go("root.reconsent"); //don't know how to do this yet
-                            }}>
-                                {t('general-settings.confirm')}
+                            setNoConsentVis(false); }}>
+                                {t('general-settings.consented-ok')}
```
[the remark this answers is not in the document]
Abby-Wheelis: I'm not sure why we have two - right now "Confirm" is used as the action to agree and log out when we send out the "are you sure" prompt and it is presented for the user to agree to continue syncing data during force sync. 

"OK" is used to close the two "check consent result" popups, so the user is just saying "ok" they read the message.

After looking into the use cases, I think it makes sense to keep both - we use "Confirm" when the user is being asked to either "confirm" continuation of an action or cancel it, and "ok" when the user is just closing something/acknowledging that they read it.

www/js/plugin/storage.ts
Abby-Wheelis: done in [b2d6aa9](https://github.com/e-mission/e-mission-phone/pull/1072/commits/b2d6aa912f152515d791f136654b696acff125d1)

www/js/splash/pushnotify.js
Abby-Wheelis: I haven't done extensive research into the way react wants us to handle things like this, but the good news here is that is an old dependency, so line 19 can just be removed. 

First, we had the event handler - then I rewrote to call the same code (check for intro done then call `registerPush`

With Jack's help in review, this got cleaner https://github.com/e-mission/e-mission-phone/pull/1072#discussion_r1372339180. Now, the code that was called by the "intro done" event is called after intro is marked in `onboardingHelper` and the "reconsent" event code is called in from `startprefs` (I commented on these places above to flag them).

So there is no circular dependency, since we only call the "actionable" code from these other files once we've already checked the conditions, this was a major improvement during Jack's review.

I will do some more research about React's philosophy on tight dependencies and possibilities for event driven implementation.

Abby-Wheelis: There is still a good bit of circular dependency, as there is two-way dependency between:

`startPrefs` and each of `pushNotify`, `onboardingHelper`, and `storeDeviceSettings`

In each case, the dependency on `startPrefs` is for `readConsentState` and `isConsented`. I'm wondering if the more React way to handle that dependency would be to use `appContext`? But, at least the way we've used it so far it only applies to components. 

It might make some sense to collapse `startPrefs` into `onboardingHelper`, since both work some with consent, but that won't fix this problem.

Abby-Wheelis: Most of what I've been able to find about tight coupling or event emitting suggests passing the event handler into the child component, or using context as the "React" way of doing things. I have not found much pertaining specifically to modules like these. @JGreenlee do you know what the "React" way to navigate this kind of circular dependency should be of if we can use `appContext` to help with this?

Abby-Wheelis: I started an issue [here](https://github.com/e-mission/e-mission-docs/issues/1024), as I've kept looking into this and it seems like it will be a much bigger deal when it comes to the "notify" family of services

Abby-Wheelis: I think that we will be circling back to adding in events, but the consensus in our meeting seemed to be leaving this as is for now to prevent confusion and too much churn.

shankari (approved):

PR #1075, opened 2023-10-21
Add hybrid

Abby-Wheelis:
This PR will accomplish the changes proposed in [issue #1013](https://github.com/e-mission/e-mission-docs/issues/1013)


Abby-Wheelis:
I will start a translate PR for this after everything settles with the re-work of the label translations is wrapped up

Abby-Wheelis:
This was based on the feature branch (`onboarding_routing`), so that one may need to be merged into this branch first

shankari (changes requested):
@Abby-Wheelis I am fine with the two actual commits in the PR.
However, it looks like there are several other file changes here as well, and they are staying even after we have merged from main to service_rewrite_2023
e.g. https://github.com/e-mission/e-mission-phone/blob/7338c18fa5dde9321f4a40b27bf8acc2651645e0/www/js/App.tsx#L94 already includes the diff in `App.tsx` here.

If you can't figure out how this happened, you may want to just close this PR and create a new one to make sure that it is clean.

Abby-Wheelis:
I was able to resolve the diffs to just be the relevant changes by merging upstream and toggling the target branch

shankari (approved):

PR #1076, opened 2023-10-21
Edit tracking bug

Abby-Wheelis:
This is a temporary fix for [issue #995](https://github.com/e-mission/e-mission-docs/issues/995), until we can fully resolve the issue with a fix to the native code.

### DIFF
--- a/www/js/control/ProfileSettings.jsx
+++ b/www/js/control/ProfileSettings.jsx
@@ -156,7 +156,10 @@ const ProfileSettings = () => {
 
     //ensure ui table updated when editor closes
     useEffect(() => {
-        refreshCollectSettings();
+        if(editCollection == false) {
+            console.log("closed editor, time to refresh collect");
+            refreshCollectSettings();
+        }
     }, [editCollection])
 
     async function refreshNotificationSettings() {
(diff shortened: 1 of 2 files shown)


www/js/control/ProfileSettings.jsx
```diff
@@ -156,7 +156,12 @@ const ProfileSettings = () => {
...
     //ensure ui table updated when editor closes
     useEffect(() => {
-        refreshCollectSettings();
+        if(editCollection == false) {
+            setTimeout(function() {
+                console.log("closed editor, time to refresh collect");
+                refreshCollectSettings();
+              }, 1000);
+        }
     }, [editCollection])
```
shankari, line 165: I'm a bit confused. Where is `editCollection` changed? I only see `setEditVis` called here.

```
        try{
            let set = await setConfig(localConfig);
            setEditVis(false);
        } catch(err) {
            Logger.displayError("Error while setting collection config", err);
        }
```

Abby-Wheelis: The variable name is different in the different files, but I realize the name in `ProfileSettings` was confusing and did not match the rest of the file. I had been consistently using "Vis" in the state names for visibility, and should do that here. I've updated it to `editCollectionVis`.

shankari (approved):

PR #1079, opened 2023-10-23
Rewrite + test storedevicesettings.js

Abby-Wheelis:
This splash service is very similar to the `startprefs` that I have worked on most recently, it does not seem to be used in any other files, but I think that the device settings it stores are held in the user profile when `updateUser` from `commHelper` is called. So unless I discover that these settings are never retrieved, I think that the service does need to be re-written.


Abby-Wheelis:
I knew there would be merge conflicts, so I merged my #1072 branch into this one, that PR should be merged before this one.

Abby-Wheelis:
closing this as the changes are now included in [#1093 ](https://github.com/e-mission/e-mission-phone/pull/1093)

PR #1081, opened 2023-10-24
💪 Increase checks for scanned QR code

Abby-Wheelis:
In order to ensure that only valid users are created, we need to make sure that the qr codes that are scanned: are QR codes, have not been canceled, start with the appropriate prefix (either emission or nrelopenpath), and include a token parameter. 

This way, after we scan into the app, we are sure that the user has scanned the right thing and will have a valid opcode. Later in the login process, we check the opcode for validity. 

I still need to test this with a physical phone before it can be merged, which I plan to do tonight.

### DIFF
--- a/www/js/components/QrCode.tsx
+++ b/www/js/components/QrCode.tsx
@@ -35,6 +35,11 @@ export function shareQR(message) {
 }
 
 const QrCode = ({ value, ...rest }) => {
+  let hasLink = value.toString().includes("//");
+  if(!hasLink) {
+    value =  "nrelopenpath://login_token?token=" + value;
+  }
+  
   return <QRCode className="qr-code" value={value} style={[{ width: '100%', height: '100%' }, rest.style] as any} {...rest} />;
 };
 
--- a/www/js/components/QrCode.tsx
+++ b/www/js/components/QrCode.tsx
@@ -37,7 +37,7 @@ export function shareQR(message) {
 const QrCode = ({ value, ...rest }) => {
   let hasLink = value.toString().includes("//");
   if(!hasLink) {
-    value =  "nrelopenpath://login_token?token=" + value;
+    value =  "emission://login_token?token=" + value;
   }
   
   return <QRCode className="qr-code" value={value} style={[{ width: '100%', height: '100%' }, rest.style] as any} {...rest} />;
--- a/www/js/onboarding/WelcomePage.tsx
+++ b/www/js/onboarding/WelcomePage.tsx
@@ -20,12 +20,25 @@ const WelcomePage = () => {
   const [infoPopupVis, setInfoPopupVis] = useState(false);
   const [existingToken, setExistingToken] = useState('');
 
+  const checkURL = function (result) {
+    let notCancelled = result.cancelled == false;
+    let isQR = result.format == "QR_CODE";
+    let hasPrefix = false;
+    if (__DEV__) {
+      hasPrefix = result.text.startsWith("emission");
+    } else {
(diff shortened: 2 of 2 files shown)


Abby-Wheelis:
While testing this, I discovered that the QR code on the "save qr" page is not composed with a full URL, only the opcode. I'll work up a fix now.

Abby-Wheelis:
Fix up for the QR code being created with opcode and not whole link in onboarding. In the `QrCode` component, I added a check to see if the value was a url, and if not, prefix the value with "nrelopenpath://login_token?token=".

Checks have been added when the code is scanned. I chose to check for "nrelopenpath" or "emission" to accomodate the dev environment, but open to changing this. The reason I did not condition what I check for on the global `__DEV__` flag is that sometimes we login to the devapp with production opcodes for testing, but if we want I can add that condition, and we can just know to modify the code if we want to use the devapp to scan an qr code that starts with "nrelopenpath"

www/js/components/QrCode.tsx
```diff
@@ -35,6 +35,11 @@ export function shareQR(message) {
 }
 
 const QrCode = ({ value, ...rest }) => {
+  let hasLink = value.toString().includes("//");
+  if(!hasLink) {
+    value =  "nrelopenpath://login_token?token=" + value;
```
shankari, line 40:
```suggestion
    value =  "emission://login_token?token=" + value;
```

www/js/onboarding/WelcomePage.tsx
```diff
@@ -20,12 +20,22 @@ const WelcomePage = () => {
   const [infoPopupVis, setInfoPopupVis] = useState(false);
   const [existingToken, setExistingToken] = useState('');
 
+  const checkURL = function (result) {
+    let notCancelled = result.cancelled == false;
+    let isQR = result.format == "QR_CODE";
+    let hasPrefix = result.text.split(":")[0] == "nrelopenpath" || result.text.split(":")[0] == "emission";
```
shankari, line 26:
```suggestion
    let hasPrefix = result.text.split(":")[0] == "emission";
```

shankari (changes requested):
Let's keep this as emission in the public repo, I will change it to `nrelopenpath` in the internal one.
Long term, this should be part of the app config
https://github.com/e-mission/e-mission-docs/issues/985#issuecomment-1769790309

shankari (approved):

www/js/onboarding/WelcomePage.tsx
```diff
@@ -20,12 +20,22 @@ const WelcomePage = () => {
   const [infoPopupVis, setInfoPopupVis] = useState(false);
   const [existingToken, setExistingToken] = useState('');
 
+  const checkURL = function (result) {
+    let notCancelled = result.cancelled == false;
+    let isQR = result.format == "QR_CODE";
+    let hasPrefix = result.text.split(":")[0] == "emission";
+    let hasToken = result.text.includes("login_token?token");
+
+    logDebug("QR code " + result.text + " checks: cancel, format, prefix, params " + notCancelled + isQR + hasPrefix + hasToken);
```
shankari, line 29: @Abby-Wheelis instead of this working, but hacky solution, I think you should use a standard URL parser. 
https://developer.mozilla.org/en-US/docs/Web/API/URL
I have merged this for now, but I think you should fix it in a cleanup change.

Abby-Wheelis: started a cleanup at #1083, ready to go pending verification using a real phone -- aiming to do that tonight

PR #1083, opened 2023-10-25
Cleanup Pass for QR codes

Abby-Wheelis:
In a cleanup pass, we're working towards using URL methods rather than the "hacky" parts that I had before https://developer.mozilla.org/en-US/docs/Web/API/URL

Also re-work the method to pull out the token from the url components and return the code if it's good, or false if the url is bad

This is pending testing on a physical phone tonight, but some dummy testing I did seems to work


Abby-Wheelis:
Checked this on a real phone - and good to go!

shankari:
re-basing this to service rewrite since I don't want to push another release to staging
This fix can just go into the next release.

PR #1073, opened 2023-10-20
Shared Type & Interface Declarations

the-bay-kay:
The goal of this PR is to create a set of type interfaces & declarations that mimic the classes and objects found in [e-mission-server](https://github.com/e-mission/e-mission-server/tree/master/emission/core/wrapper).  This should help standardize typing across the angular service rewrites ([Issue 977](https://github.com/e-mission/e-mission-docs/issues/977)), and should help future function design.

This discussion originally started in PR #1061 - the overall design philosophy is to have higher-level interfaces (e.g., `ServerData`), which can then be extended to more specific "sub-classes" of server data in separate type files.

PR #1085, opened 2023-10-26
✏️📖 Diary Services Rewrite

the-bay-kay:
Before rewriting `www/js/diary/services.js`, I took some time to update `diaryHelper.ts` so that it no longer requires the legacy moment.js library.

The plan, as with the other rewrites, is to convert the Angular services into typescript. These functions will be moved from `services.js` into `timelineHelper.ts`.


www/js/diary/services.js
```diff
@@ -1,15 +1,13 @@
 'use strict';
 
 import angular from 'angular';
-import { getBaseModeByKey, getBaseModeOfLabeledTrip } from './diaryHelper';
 import { SurveyOptions } from '../survey/survey';
 import { getConfig } from '../config/dynamicConfig';
-import { getRawEntries } from '../commHelper';
 
 angular.module('emission.main.diary.services', ['emission.plugin.logger',
```
the-bay-kay, line 7: I am running into some issues when I remove this module.  I've deleted all references to the `Timeline` factory (e.g., the one in `LabelTab.tsx`), and have been able to remove it without any issues.  From what I can tell, there are only two remaining references to this module:
- the `emission.main.diary` service uses this module as a provider
- `index.js` imports the entire file

If I remove either of these references, the timeline breaks and freezes on a blank screen, without throwing an error or warning.  This made me thing one of these imports is lacking a provider higher up on the dependency chain - however, if I remove all of `emission.main.diary.services` providers, everything functions normally, despite the fact that the module now contributes no new code.  I figure I've missed something important, I'll update this thread with notes as needed!

the-bay-kay:
Update: All of the service rewrites seem to be working correctly, I'm now working on writing tests!  

The current issue I've run into is that `timelineHelper` still rely on the `$ionicLoading` service.  After doing some searching, it seems that Angular's [TestBed](https://angular.io/api/core/testing/TestBed)  is compatible with Jest, and can be used to mock services.  For the time being, I'll continue to work on this testing framework!  

That being said, should I be focusing on _replacing_ the `$ionicLoading` service instead?  @JGreenlee , I'd appreciate your two cents! I know the goal of the rewrites was to move away from angular, so working on mocking it seems counterproductive.

JGreenlee:
Yes! Replace if possible! We want to be done with Angular and move on to new things

(for the record we are using Angular 1.6.7 I believe, which is **not** what Ionic ships with but it was necessary to upgrade to be able to interop with React)

JGreenlee:
I would suggest handling the "loading" popup in the thing that calls `timelineHelper` (which I believe is just the `LabelTab` component).

`timelineHelper` is just plain Typescript and its basic purpose is just to reformat data into something the React components can use. It doesn't know anything about what's in the React DOM, so I think it's ok to decouple it from the user-facing loading popup.

At least for now until we (potentially) implement a 'global' popup mechanism

www/js/diary/timelineHelper.ts
```diff
@@ -217,3 +222,349 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
     };
   });
 };
+
+// Remaining functions from /diary/services.js
```
JGreenlee, line 226: I had left a comment here in the old file. Make sure to add it back (or something to that effect)

```
      // DB entries retrieved from the server have '_id', 'metadata', and 'data' fields.
      // This function returns a shallow copy of the obj, which flattens the
      // 'data' field into the top level, while also including '_id' and 'metadata.key'
```

the-bay-kay:
Tests and mocks are done - currently brainstorming more effective ways to test beyond `expect().resolves.not.toThrow()`.  Half of this file is test data - it seems cumbersome to hardcode in another 100+ lines of data to check with `.toEqual()`, but this also seems the most foolproof / futureproof way of testing.  Will try this approach first, just to set up some robust tests before moving forward.

www/js/diary/timelineHelper.ts
```diff
@@ -217,3 +222,349 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+  const locations = locationPoints.map((point, i) => ({
+    loc: {
+      coordinates: [point.data.longitude, point.data.latitude],
+    },
+    ts: point.data.ts,
+    speed: speeds[i],
+  }));
+
+  // used to mimic old momentJS moment.format()
+  const formatString = "yyyy-MM-dd'T'HH:mm:ssZZ";
```
JGreenlee, line 302: I'm pretty sure Luxon's `toISO()` is the same as moment's default `format()`.
If so, it would be a cleaner way of going about this than typing this out

the-bay-kay:
I had two questions about this PR's current unit tests:

- Currently, one of the tests `timelineHelper.test.ts` relies on a pretty large `readAllCheckOne` object, defined in the mock.  If I were to do this for the other 3 unit tests that require checks, I would need to add 200+ lines to the mock.  Is there a less bulky way to go about these checks? I'm not sure how to generate the "check object" more efficiently, without effectively re-writing the method that's being tested.  I could use `jest.spyOn()` to ensure each helper function is called, though that doesn't confirm the function is returning correct data. 

- Currently, I've only written tests for the functions moved from `/www/diary/services.js`.  While I'm already editing it, should I go ahead and design tests for the remaining functions in `timelineHelper.ts`?

www/__tests__/timelineHelper.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  jest.restoreAllMocks();
+});
+
+// Once we have end-to-end testing, we could utilize getRawEnteries.
+jest.mock('../js/commHelper', () => ({
+  getRawEntries: jest.fn((key, startTs, endTs, valTwo) => {
+    if (startTs === mockTLH.fakeStartTsOne) return mockTLH.mockData;
+    if (startTs == mockTLH.fakeStartTsTwo) return mockTLH.mockDataTwo;
+    return {};
+  }),
```
JGreenlee, line 25: The approach is good, but it seems like `mockDataTwo` is just 2 of the same trip?

www/__tests__/timelineHelper.test.ts
```diff
@@ -0,0 +1,60 @@
...
+}));
+
+it('works when there are no composite trip objects fetched', async () => {
+  expect(readAllCompositeTrips(-1, -1)).resolves.toEqual([]);
+});
+
+it('fetches a composite trip object and collapses it', async () => {
+  expect(readAllCompositeTrips(mockTLH.fakeStartTsOne, mockTLH.fakeEndTsOne)).resolves.toEqual(
+    mockTLH.readAllCheckOne,
```
JGreenlee, line 34: For something like this, I don't know that you need to compare the entire result against some 'check' object (this is commonly called a 'snapshot')

I think you can just look through the result and ensure that it has the values you would expect based on what you know the function you're testing is supposed to do.

In this instance, `readAllCompositeTrips` is supposed to input what's received from the server and output an array of unpacked `CompositeTrip`s.
So I would check that the result is an array of objects with a length of however many you expect from the mock. Each of these objects has `start_ts` / `end_ts` / any other distinctive properties that would define them as `CompositeTrip`s. The objects should not have `data` / `metadata` because they've been unpacked by now. They **should** have `_id`, `key`, and `origin_key` because these properties get mixed in during unpacking.
Then I would dig into the trips' `start_confirmed_place` and `end_confirmed_place` and make sure they've also been unpacked.

JGreenlee (changes requested):
This is a just surface-level initial review but I want to at least answer some of your questions so you can keep working.

>I'm not sure how to generate the "check object" more efficiently

I left a comment for this, but TLDR: don't make a bunch of massive check objects / snapshots - instead think critically about what the function is supposed to do and inspect the result for the specific properties that only a correct result would have

>Currently, I've only written tests for the functions moved from /www/diary/services.js. While I'm already editing it, should I go ahead and design tests for the remaining functions in timelineHelper.ts?

Ideally yes. I wrote those functions before we had a testing framework in place. Since you're working on `timelineHelper` in this PR, it is a convenient opportunity to go back and write those tests. I can help with this since I know `timelineHelper` is big.

the-bay-kay:
Currently getting WSOD after merge - spent quite some time trying to debug, without any success. Pushed changes despite being broken, so I can pull on my Linux machine and use a proper debugger!
image

the-bay-kay:
I'm a little confused as to the purpose of `updateLocalUnprocessedInputs` and `updateAllUnprocessedInputs` - after some experimentation with different OpCodes and OSs, I haven't found a time when `updateLocalUnprocessedInputs` is called, and `updateAllUnprocessedInputs` returns empty. (E.x., line 134 of `LabelTab.tsx` often prints `{"MODE":[], "PURPOSE":[], "REPLACED_MODE":[]}` and `[]`. ).  

Are there specific opcodes I should try to test these functions? I'm struggling to write test functions for these, as I haven't been able to see their return values.

JGreenlee:
They used to be called `getLocalUnprocessedInputs` and `getAllUnprocessedInputs` and have return values.
Now they are `updateLocalUnprocessedInputs` and `updateAllUnprocessedInputs` and they don't return anything; they just update the value of `unprocessedLabels` and `unprocessedNotes`.

`updateAllUnprocessedInputs` is called on the initial load of the Label tab. It queries the unified data loader to read unprocessed inputs from both the server and the local usercache.
`updateLocalUnprocessedInputs` is called once a new input is recorded + stored because it needs to re-match. It only queries the local usercache (because presumably, nothing on the server has changed)

Regardless of what configuration you are testing on, if you record some user input (ie mode, purpose, or survey response) and refresh the label tab you should see line 134 print out the input you just recorded.

www/js/diary/timelineHelper.ts
```diff
@@ -211,3 +215,348 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+    });
+};
+const dateTime2localdate = function (currtime: DateTime, tz: string) {
+  return {
+    timezone: tz,
+    year: currtime.get('year'),
+    //the months of the draft trips match the one format needed for
+    //moment function however now that is modified we need to also
+    //modify the months value here
+    month: currtime.get('month') + 1,
```
JGreenlee, line 256: You should remove the + 1 offset and this comment because https://github.com/moment/luxon/blob/master/docs/moment.md#major-functional-differences

www/js/diary/timelineHelper.ts
```diff
@@ -211,3 +215,348 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+    day: currtime.get('day'),
+    weekday: currtime.get('weekday'),
+    hour: currtime.get('hour'),
+    minute: currtime.get('minute'),
+    second: currtime.get('second'),
```
JGreenlee, line 261: I think you can access all of these directly (like `currtime.day`, `currtime.weekday`)

www/js/diary/timelineHelper.ts
```diff
@@ -211,3 +215,348 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+  var tq = { key: 'write_ts', startTs, endTs };
+  logDebug(
+    'about to query for unprocessed trips from ' +
+      DateTime.fromSeconds(tq.startTs).toLocaleString(DateTime.DATETIME_MED) +
+      DateTime.fromSeconds(tq.endTs).toLocaleString(DateTime.DATETIME_MED),
+  );
+
+  console.log('Testing...');
+  const getMessageMethod = window['cordova'].plugins.BEMUserCache.getMessagesForInterval;
+  console.log('Entering...');
```
JGreenlee, line 513: these log statements do not seem necessary

JGreenlee: if they are, make them logDebug and more descriptive

www/js/types/diaryTypes.ts
```diff
@@ -26,6 +24,21 @@ type ConfirmedPlace = {
...
+
+export type TripTransition = {
+  currstate: string;
+  transition: string;
+  ts: number;
+};
+
+export type LocationCoord = {
+  type: string; // e.x., "Point"
+  coordinates: [number, number] | Array<[number, number]>;
```
JGreenlee, line 41: What's the difference between `[number, number]` and `Array<[number, number]>`?
Why do you have to put both?

www/js/types/diaryTypes.ts
```diff
@@ -118,13 +131,56 @@ export type UserInputEntry = {
...
+export type GeoJSON = {
+  data: {
+    id: string;
+    type: string;
+    features: GjFeature[];
+    properties: {
+      start_ts: number;
+      end_ts: number;
+    };
+  };
```
JGreenlee, line 185: Thank you for doing this typing work — but I think we should be able to grab these types from a package. That way we can ensure our type defs align with the official GeoJSON spec.
I did some digging and I think Leaflet's types package includes the geojson type package too.

So, can you `npm i -D @types/leaflet` (updating `package.serve.json` and `package.cordovabuild.json` accordingly).
Then see if you can `import { Feature, FeatureCollection, GeoJsonObject } from 'geojson'` and use those types instead of declaring them ourselves

If you see something we're doing here that doesn't line up with the GeoJSON spec, feel free to fix it!

JGreenlee: I think we may end up with something like:
```ts
export type GeoJSON = {
  id: string, // the official spec doesn't have an 'id' field so we declare it ourselves
  data: GeoJsonObject, // everything that *is* in the official spec goes inside 'data'
}

Comment (JGreenlee) [reply www/js/types/diaryTypes.ts:185]: Actually, maybe something like this is better:
```ts
export type GeoJSON = GeoJsonObject & { id: string };
export type GjFeature = Feature & { style?: { color: string} };
```

JGreenlee: Try it and see what works!

www/js/diary/diaryHelper.ts
```diff
@@ -114,11 +114,15 @@ export function getFormattedDate(beginFmtTime: string, endFmtTime?: string) {
     return `${getFormattedDate(beginFmtTime)} - ${getFormattedDate(endFmtTime)}`;
   }
   // only one day given, or both are the same day
-  const t = moment.parseZone(beginFmtTime || endFmtTime);
-  // We use ddd LL to get Wed, May 3, 2023 or equivalent
-  // LL only has the date, month and year
-  // LLLL has the day of the week, but also the time
-  return t.format('ddd LL');
+  const t = DateTime.fromISO(beginFmtTime || endFmtTime);
```
JGreenlee, line 117: Please ensure that you are parsing ***with respect to timezone***
https://github.com/moment/luxon/issues/221#issuecomment-519194393

If not, we could end up on the wrong day sometimes.

www/js/diary/diaryHelper.ts
```diff
@@ -114,11 +114,15 @@ export function getFormattedDate(beginFmtTime: string, endFmtTime?: string) {
...
+  // We use toLocale to get Wed May 3, 2023 or equivalent,
```
JGreenlee, line 118: It changed from `Wed, May 3, 2023` to `Wed May 3, 2023`?

JGreenlee: If so, I think that's ok but we have to update the JS doc and some tests

www/js/diary/diaryHelper.ts
```diff
@@ -97,8 +97,8 @@ export function getBaseModeByText(text, labelOptions: LabelOptions) {
 export function isMultiDay(beginFmtTime: string, endFmtTime: string) {
   if (!beginFmtTime || !endFmtTime) return false;
   return (
-    moment.parseZone(beginFmtTime).format('YYYYMMDD') !=
-    moment.parseZone(endFmtTime).format('YYYYMMDD')
+    DateTime.fromISO(beginFmtTime).toFormat('YYYYMMDD') !=
+    DateTime.fromISO(endFmtTime).toFormat('YYYYMMDD')
```
JGreenlee, line 101: again, be careful with `parseZone` and make sure whatever you use from Luxon does the **exact** same thing

www/js/diary/diaryHelper.ts
```diff
@@ -144,9 +148,12 @@ export function getFormattedDateAbbr(beginFmtTime: string, endFmtTime?: string)
  */
 export function getFormattedTimeRange(beginFmtTime: string, endFmtTime: string) {
   if (!beginFmtTime || !endFmtTime) return;
-  const beginMoment = moment.parseZone(beginFmtTime);
-  const endMoment = moment.parseZone(endFmtTime);
-  return endMoment.to(beginMoment, true);
+  const beginTime = DateTime.fromISO(beginFmtTime);
+  const endTime = DateTime.fromISO(endFmtTime);
```
JGreenlee, line 152: this was `parseZone`, be careful

www/js/types/diaryTypes.ts
```diff
@@ -3,19 +3,18 @@
  As much as possible, these types parallel the types used in the server code. */
 
 import { BaseModeKey, MotionTypeKey } from '../diary/diaryHelper';
+import { ServerData, LocalDt } from './serverData';
+import { FeatureCollection, Feature, Geometry } from 'geojson';
```
JGreenlee, line 7: Looks like this is working well - but make sure that `@types/leaflet` (or whatever you did `npm i` for to be able to access `geojson`) is added to `package.serve.json` and `package.cordovabuild.json`.

In this repo we don't commit `package.json` directly, so we have this extra step whenever adding packages.

www/js/diary/diaryHelper.ts
```diff
@@ -144,9 +148,12 @@ export function getFormattedDateAbbr(beginFmtTime: string, endFmtTime?: string)
...
   if (!beginFmtTime || !endFmtTime) return;
-  const beginMoment = moment.parseZone(beginFmtTime);
-  const endMoment = moment.parseZone(endFmtTime);
-  return endMoment.to(beginMoment, true);
+  const beginTime = DateTime.fromISO(beginFmtTime, { setZone: true });
+  const endTime = DateTime.fromISO(endFmtTime, { setZone: true });
+  const range = endTime.diff(beginTime, ['hours']);
+  const roundedHours = Math.round(range.as('hours')); // Round up or down to nearest hour
+  const formattedRange = `${roundedHours} hour${roundedHours !== 1 ? 's' : ''}`;
+  return formattedRange;
```
JGreenlee, line 156: I'm guessing Luxon doesn't support the same kind of textual descriptions for time ranges, so you tried to replicate it like this.

Unfortunately, this will only work for English and only with hours as the unit.
We need a more robust solution that works across languages and on varying scales of time (e.g. "27 minutes", "3 hours", "1 day")

JGreenlee: Other people have noted that this is missing from Luxon: https://stackoverflow.com/questions/63526184/luxon-interval-human-readable

They suggest an external package [`humanize-duration`](https://www.npmjs.com/package/humanize-duration). I checked and it **does** support Lao.

JGreenlee: It seems like Luxon `Duration`s do have a `toHuman()` function, but it doesn't do what people expect it to do https://github.com/moment/luxon/issues/1134
Can you look into whether `toHuman()` would be good enough for our purposes? Otherwise, I think we should add the `humanize-duration` package

www/__tests__/timelineHelper.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    return Promise.resolve([]);
+  }),
+}));
+
+it('works when there are no unprocessed trips...', async () => {
+  expect(readUnprocessedTrips(-1, -1, null)).resolves.toEqual([]);
+});
+
+// In manual testing, it seems that `trip_gj_list` always returns
+// as an empty array - should find data where this is different...
```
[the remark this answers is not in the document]
JGreenlee: Let me know if you need further guidance or if you'd like me to show what I mean over a call

JGreenlee (changes requested):

the-bay-kay:
Some notes on future changes to the typing system I'd like to make (most likely in PR #1073):

I've been filling out the types return from `readUnprocessedTrips()`, and think there could be some future changes.  `ConfirmedPlace`, `UnprocessedTrip`, & `CompositeTrip` have several overlapping qualities. There are some subtle differences (e.x., `start_loc`, `starting_trip`) -- That being said, I think these could refactored such that they are intersections of a "core" type: this would help cut out some of the redundant types.  E.x.,

```typescript
type CoreType = {
  start_fmt_time: string;
  end_fmt_time: string;
  // More types...
};
export type UnprocessedTrip = CoreType & {/*...*/}
export type CompositeTrip = CoreType & {/*...*/}
```

the-bay-kay:
I'm debating how many tests to add for `readUnprocessedTrips()`, as adding more would mean adding more bulky test data to `timelineHelperMocks.ts`.

Regardless, I believe the only functions left that need tests are the `updateUnprocessedInputs()` methods!

the-bay-kay:
> Regardless of what configuration you are testing on, if you record some user input (ie mode, purpose, or survey response) and refresh the label tab you should see line 134 print out the input you just recorded.

After some time time experimenting with different opcodes, surveys, and labels, I haven't seen `unprocessedNotes` change from an empty array.  I have seen the `unprocessedLabels` change, however, so I'll go ahead and start writing tests for those!

image

the-bay-kay:
I'm running into a bizarre issue with Jest - when I run my test for `updateUnprocessedInputs`, the outside functions are running before the `expect()` test itself.  Below are some screenshots of the console logs and outputs.
 
image

image

JGreenlee:
> I'm running into a bizarre issue with Jest - when I run my test for `updateUnprocessedInputs`, the outside functions are running before the `expect()` test itself. Below are some screenshots of the console logs and outputs.

`updateAllUnprocessedInputs` is asynchronous code. When you call `expect(...).resolves` it just queues the promise resolution while execution continues.
If you wanted to `console.log('C:...')` AFTER `updateUnprocessedInputs` is done, you should either put it it `.then()` or use `async` / `await` syntax

JGreenlee:
Oh, I see now that you were wanting to test `updateUnprocessedInputs` by itself which is not async. In that case I think you could have put the function call by itself (without `expect` or `resolves`) and then *on the following line* done an `expect` statement to verify that `unprocessedLabels` and/or `unprocessedNotes` have been updated

the-bay-kay:
> Oh, I see now that you were wanting to test `updateUnprocessedInputs` by itself which is not async. In that case I think you could have put the function call by itself (without `expect` or `resolves`) and then _on the following line_ done an `expect` statement to verify that `unprocessedLabels` and/or `unprocessedNotes` have been updated

Gotcha -- I initially tried that, but the `unprocessedLabels` returns as an empty object.  It does seem to execute in order though (A B {crash before C}), so I may have been barking up the wrong tree.  Thanks for the help! I'll start looking for what else could be the issue.

JGreenlee:
> Gotcha -- I initially tried that, but the `unprocessedLabels` returns as an empty object. It does seem to execute in order though (A B {crash before C}), so I may have been barking up the wrong tree.

There might be something to fix here then - let's take another look at that function. Maybe it should be asynchronous and isn't

<hr>

Yeah, the function is currently this:

```ts
function updateUnprocessedInputs(labelsPromises, notesPromises, appConfig) {
  Promise.all([...labelsPromises, ...notesPromises]).then((comboResults) => {
    ...
  });
}
```

It should return the result of `Promise.all` like this:

```ts
function updateUnprocessedInputs(labelsPromises, notesPromises, appConfig) {
  return Promise.all([...labelsPromises, ...notesPromises]).then((comboResults) => {
    ...
  });
}
```

@the-bay-kay Would you mind making that change and then seeing if you can `await` or `.then()` the function?

the-bay-kay:
> ... Would you mind making that change and then seeing if you can `await` or `.then()` the function?

Oh awesome, that was it!! Thank you so much for the help : ) Let me clean up these tests, add a few more, and I'll push the changes.

the-bay-kay:
Tests that use `labelsPromises` have been added!  Through all of this, I still haven't seen `notesPromises` change from an empty array.  I assumed this would change when interacting with the enekto survey (see: line 108 of `timelineHelper.ts`), though that wasn't the case - when is this usually updated?  If we want to cover all of our bases for `updateUnprocessedInputs`, I think an example of the notes data is the last thing I need!  I'm workin on trying to get find these values on my android emulator, but ran into a hiccup with the survey.  I'll keep this thread updated!

www/js/diary/timelineHelper.ts
```diff
@@ -211,3 +223,345 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+        return undefined;
+      }
+      const sortedLocationList = locationList.sort(tsEntrySort);
+      const retainInRange = function (loc) {
+        return (
+          tripStartTransition.data.ts <= loc.data.ts && loc.data.ts <= tripEndTransition.data.ts
+        );
+      };
+
+      var filteredLocationList = sortedLocationList.filter(retainInRange);
```
JGreenlee, line 356:
```suggestion
      const filteredLocationList = sortedLocationList.filter(retainInRange);
```

www/js/diary/timelineHelper.ts
```diff
@@ -211,3 +223,345 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+
+  // start trip2
+  trip2.ending_trip = { $oid: trip1.id };
+  trip2.enter_fmt_time = trip1.exit_fmt_time;
+  trip2.enter_local_dt = trip1.exit_local_dt;
+  trip2.enter_ts = trip1.exit_ts;
+};
+
+export const readUnprocessedTrips = function (startTs, endTs, lastProcessedTrip) {
+  var tq = { key: 'write_ts', startTs, endTs };
```
JGreenlee, line 512:
```suggestion
  const tq = { key: 'write_ts', startTs, endTs };
```

www/js/diary/timelineHelper.ts
```diff
@@ -211,3 +223,345 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+ * See https://github.com/e-mission/e-mission-phone/issues/214#issuecomment-286279163
+ *
+ * Let's abstract this out into our own minor state machine.
+ */
+const transitions2Trips = function (transitionList: Array<ServerData<TripTransition>>) {
+  var inTrip = false;
+  var tripList = [];
+  var currStartTransitionIndex = -1;
+  var currEndTransitionIndex = -1;
+  var processedUntil = 0;
```
JGreenlee, line 446:
```suggestion
  const tripList = [];
  let inTrip = false;
  let currStartTransitionIndex = -1;
  let currEndTransitionIndex = -1;
  let processedUntil = 0;
```

JGreenlee:
It looks to me like tests passed but codecov failed. @jiji14 could you take a look?

jiji14:
@JGreenlee @shankari

The recent failure was due to a rate limit issue. Currently, **without the token**, the upload occasionally faces rate limitations, **resulting in failure**s. However, incorporating the token in the action **doesn't work for PRs from forks.**

This morning, out of 4 PRs and 8 GitHub actions, this specific PR, unfortunately, was the last one and encountered a failure.

Refer to the following related issues for more context:

[Codecov/feedback #126](https://github.com/codecov/feedback/issues/126)
[Community.Codecov - Public repositories unable to upload coverage](https://community.codecov.com/t/public-repositories-can-no-longer-upload-coverage-with-error-unable-to-locate-build-via-azure-api/3242/14)
[Community.Codecov - Upload issues via GitHub Actions API](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954)

<br />

I've identified two potential solutions to mitigate this issue:

1. Run Codecov report when PR is completed:
I verified that we can trigger a workflow for a specific event, such as when a pull request merges. [GitHub Actions Documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges-1)

2. Replace Codecov report with Jest report:
Considering the challenges with Codecov, a Jest Coverage Report appears to be a viable alternative.

Please let me know if you have other ideas!

shankari:
wrt the link that you posted https://community.codecov.com/t/public-repositories-can-no-longer-upload-coverage-with-error-unable-to-locate-build-via-azure-api/3242/14

It looks like the new uploader has been fixed

> [@johlju](https://community.codecov.com/u/johlju) [@samdoran](https://community.codecov.com/u/samdoran) I have patched the new uploader for public projects. It should not need the token. Would you be able to see if this has fixed your issues?

However, the issue indicates that this is broken as of Oct, so it does look like it is broken.

I think that, at least for now, we should run the Codecov report when we merge the PR. This is a bit sub-optimal, because the person contributing the PR cannot see code paths with missing tests and fix them before merge. As long as we are doing most of the contributions ourselves, we can set up a process to fix missing code coverage after merge and try it out.

Can you also investigate using Jest Coverage Report in parallel?

If that doesn't work, *and* we see an uptick in external contributions, we may have to resort to exposing the token instead of keeping it a secret.

JGreenlee:
I noticed that since @jiji14 enabled Jest `collectCoverage`, we can already see the Jest coverage report in the terminal, when we run tests locally.
So I think it's fine to use that while working on the PR -- and then once the PR is merged, Codecov will give a more detailed picture

JGreenlee:
> 1. I verified that we can trigger a workflow for a specific event, such as when a pull request merges. [GitHub Actions Documentation](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges-1)

My other suggestion would be to run the workflow when the PR is marked "Ready for Review". I think this should be possible based on the discussion in https://github.com/reviewdog/action-eslint/issues/29#issuecomment-985939887

This might be a good middle ground where the workflow runs before merging, but isn't running on every single commit.

shankari:
> This might be a good middle ground where the workflow runs before merging, but isn't running on every single commit.

This! There are other, similar options (at my old lab, we used to say "jenkins, test this please"), but "ready for review" seems to be a good flag for our process/flow

shankari:
@the-bay-kay this now has a merge conflict, can you please resolve?

shankari:
I just merged the changes, but ended up losing this change since it was overridden by the switch to luxon.
@Abby-Wheelis @the-bay-kay can you confirm that the new luxon code works properly here
https://github.com/e-mission/e-mission-phone/commit/6e3af4f2fdc73dd622cba8e25f4ddd1cbb8c58af

www/__mocks__/timelineHelperMocks.ts
```diff
@@ -0,0 +1,254 @@
...
+  additions: [],
+  user_input: {},
+  enter_fmt_time: '2015-07-22T08:14:53.881000-07:00',
+  exit_fmt_time: '2015-07-22T08:14:53.881000-07:00',
+  starting_trip: {
+    $oid: '6553c3a1f27f16fbf9d1df15',
+  },
+  ending_trip: {
+    $oid: '6553c3a1f27f16fbf9d1df15',
+  },
```
shankari, line 46: Do you need all this information? The oids will be different for every run, so I am not sure you can check against it. And if you can't check against it, do you still need it?

www/js/diary/timelineHelper.ts
```diff
@@ -212,3 +226,345 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+    month: currtime.month,
+    day: currtime.day,
+    weekday: currtime.weekday,
+    hour: currtime.hour,
+    minute: currtime.minute,
+    second: currtime.second,
+  };
+};
+
+const points2TripProps = function (locationPoints) {
```
shankari, line 276: given that this is typescript, shouldn't we have types for all the variables?

www/js/diary/timelineHelper.ts
```diff
@@ -212,3 +226,345 @@ const locations2GeojsonTrajectory = (trip, locationList, trajectoryColor?) => {
...
+  trip1.exit_ts = trip2.enter_ts;
+
+  // start trip2
+  trip2.ending_trip = { $oid: trip1.id };
+  trip2.enter_fmt_time = trip1.exit_fmt_time;
+  trip2.enter_local_dt = trip1.exit_local_dt;
+  trip2.enter_ts = trip1.exit_ts;
+};
+
+export const readUnprocessedTrips = function (startTs, endTs, lastProcessedTrip) {
```
shankari, line 514: Ditto. I think that most of the functions that have been copied over do not have a type defined

www/js/diary/timelineHelper.ts
shankari, line 1: I assume that most of the new code here is just copied over from the original js file so I have not looked through the implementation in detail.

www/__tests__/timelineHelper.test.ts
```diff
@@ -0,0 +1,313 @@
...
+  it('Works with a list of len >= 1, no flag', () => {
+    testValue = compositeTrips2TimelineMap(tripListTwo);
+    expect(testValue.size).toBe(2);
+    expect(testValue.get(keyTwo)).toEqual(tripListTwo[1]);
+    expect(testValue.get(keyThree)).toEqual(tripListTwo[0]);
+  });
+
+  it('Works with a list of len >= 1, with flag', () => {
+    testValue = compositeTrips2TimelineMap(tripListTwo, true);
+    expect(testValue.size).toBe(6);
```
shankari, line 93: Shouldn't you be checking that the values are consistent (at least wrt keys)?

shankari (changes requested):
Almost there, just some high-level questions and cleanup

shankari:
Given that the tests are failing, here's my proposed plan of action for finishing the merges:
- [ ] @sebastianbarry fixes the notifScheduler tests
- [ ] @the-bay-kay then pulls that change, addresses review comments, fixes prettier and updates
- [ ] I review the hopefully minor changes and merge
- [ ] @JGreenlee pulls these changes into the UI wrapup change and resolves merge conflicts
   - Hopefully, the remaining cleanup changes are fairly minor
- [ ] I review and merge the UI wrapup change

sebastianbarry:
@shankari  I am not sure how to proceed with fixing my bug; the jest test passes on my local machine, and it passed the github action tests while it was just my PR https://github.com/e-mission/e-mission-phone/pull/1092, but it fails in a strange unexpected way on this PR https://github.com/e-mission/e-mission-phone/actions/runs/7597279222/job/20691918950?pr=1085

It appears that the function to schedule the notifications works as expected, scheduling the 4 correct notifications. The issue is that after, I'm checking to see if the scheduledNotifs actually got scheduled by calling getScheduledNotifs. In here, the async behavior in getScheduledNotifs believes that the notifications are still in the process of being scheduled (which does not reproduce on my local machine), and then when it finishes scheduling, the notifications are not scheduled (however they are at this point on my local machine).

**Do you have any ideas for how I can fix this? Or perhaps test it in the github actions codecov environment so I at least have somewhere to go off of? At this point I'm not sure what the problem is and I don't know where to go from here.**


| | |
|---|---|
| Image of it working on my local machine | Screenshot 2024-01-22 at 9 25 42 AM |

shankari:
@sebastianbarry have you pulled from `service_rewrite_2023` after your PR was merged? I merged several other PRs at the same time, and one of them might have affected your tests.

I would:
- pull and test locally
- if it passes, test in the GH environment by creating a PR that you push additional logs to and wait for the run to complete. (we can squash merge this so it won't mess up the commit history)

There is no way to run GH actions locally.

jiji14:
> @sebastianbarry have you pulled from `service_rewrite_2023` after your PR was merged? I merged several other PRs at the same time, and one of them might have affected your tests.
> 
> I would:
> 
> * pull and test locally
> * if it passes, test in the GH environment by creating a PR that you push additional logs to and wait for the run to complete. (we can squash merge this so it won't mess up the commit history)
> 
> There is no way to run GH actions locally.

@sebastianbarry 

Also check if there is any difference between local environment and GH environment. For example, different version might cause the issue. It happened to me when I worked with GA so wanted to let you know!

shankari:
> Also check if there is any difference between local environment and GH environment. For example, different version might cause the issue. It happened to me when I worked with GA so wanted to let you know!

Was this "mismatched version on GA" on OpenPATH? This should not happen with OpenPATH because we have locked down our dependencies. When you run `setup` and `activate`, you use the version of the tools that is checked in to OpenPATH, and since we call setup and activate on GH (and we should in case we don't), it should run with the same environment.

The reason for locking down the dependencies was exactly this - we wanted to have a consistent build and test process, so that a test failure was meaningful

jiji14:
> > Also check if there is any difference between local environment and GH environment. For example, different version might cause the issue. It happened to me when I worked with GA so wanted to let you know!
> 
> Was this "mismatched version on GA" on OpenPATH? This should not happen with OpenPATH because we have locked down our dependencies. When you run `setup` and `activate`, you use the version of the tools that is checked in to OpenPATH, and since we call setup and activate on GH (and we should in case we don't), it should run with the same environment.
> 
> The reason for locking down the dependencies was exactly this - we wanted to have a consistent build and test process, so that a test failure was meaningful

It happened when setting up prettier, but it is fixed now!

sebastianbarry:
The updates are in my pull request https://github.com/e-mission/e-mission-phone/pull/1123

## Why was it failing?

The test was working correctly, but the problem was that the logDebug statement was not being ran.

## What did I do to fix it?

I corrected this by making the `expect` statement actually expect the output rather than the logDebug statement arguments.
However, when I did this I encountered a second error. I created an `expectedScheduledNotifs` function that is an array containing objects with keys `{ key: "November 19, 2023" val: "9:00 PM" }`. This is what the expected values look like, but the resulted values contain a strange unicode character U+202F in `"9:00 PM"`which fails the equivalence test. I am not sure how to go about testing these values, since my local computer fails the test if I try to expect these values. My solution was to only expect the `key` value for them. This is only a temporary soluiton, so I am open to suggestions, but it will pass the codecov test for now.

shankari (approved):

shankari:
@the-bay-kay I am merging this now to unblock @JGreenlee but it would be good if you could resolve the other comments as well when you have a chance.

PR #1091, opened 2023-10-31
Remove moped

sebastianbarry:
Removed moped from `en.json` and the default `label-options.json.sample` files

In response to @shankari and @Abby-Wheelis in https://github.com/e-mission/e-mission-docs/issues/1013#issuecomment-1787627109


Abby-Wheelis:
I think that's for the base mode moped, so we do want to keep that around for configs that do have that base mode. My understanding is that we support that underlying functionality (moped icon and color) for cases like `usaid-laos-ev` options, which does make use of that base mode. There's just no moped in the app by default, but we do support it. There's a few of those base modes that aren't in the default set - like "TROLLEYBUS" and "FERRY"

shankari (approved):

PR #1092, opened 2023-10-31
‼️📅 Rewrite - notifScheduler.ts

sebastianbarry:
Translated notifScheduler.js from Angular to React Typescript

commHelper.ts
- Added a temporary "any" return type for getUser because the TS needed it

ProfileSettings.jsx
- Imported the new useSchedulerHelper function (name in progress - need to find a good name for this hook)
- Replaced any instances of the old Angular NotificationScheduler module with the new schedulerHelper hook

notifScheduler.ts
- Translated the file from Angular to React TS
- This file exports a single function which initializes the appConfig and then returns an object containing the functions that the Notification Scheduler functions need (only ProfileSettings.jsx is using this file currently)
- Replaced any instances of moment with Luxon equivalent


sebastianbarry:
@JGreenlee Added this to your review channel and @ you for visibility

If you have a moment, could you take a look at this at a high level and let me know if the structure needs to be changed at all?

I also have another question: Is notifScheduler at a high level only being used by programs and studies that use `reminderSchemes

sebastianbarry:
Sorry, my finger slipped and it accidentally closed the PR 😖

sebastianbarry:
Testing:

| Description | Screenshot |
|---|---|
| Notifications populate in _"Developer Zone -> Upcoming Notifications"_ |  Simulator Screenshot - iPhone 13 - 2023-10-31 at 14 14 42 |
| Promise rejection errors popping up from outside files now | image |

sebastianbarry:
# To-Do:

- [ ] Address the promise rejection errors
- [ ] Come up with a good name for the hook
- [ ] Add stronger typing for the typescript (for example, the user returned in getUser should be typed)
- [ ] Consider whether we want to use a local variable for the appConfig instead of passing it as an argument down the chain of functions
- [ ] Test the notifications for every use case _**(this is something I'm confused about -- I'm not sure the best way to test this and know that what I've written doesn't have any regressions or issues; does this PR need Jest testing in it as well? How do you go about unit testing phone notifications?)**_

www/js/control/ProfileSettings.jsx
```diff
@@ -25,6 +25,7 @@ import { AppContext } from "../App";
 import { shareQR } from "../components/QrCode";
 import { storageClear } from "../plugin/storage";
 import { getAppVersion } from "../plugin/clientStats";
+import { useSchedulerHelper } from "../splash/notifScheduler";
```
JGreenlee, line 28: If you export `update`, `getScheduledNotifs`, `getReminderPrefs`, and `setReminderPrefs` directly from `notifScheduler.ts`, you can import them as individual functions instead of `useSchedulerHelper`

www/js/control/ProfileSettings.jsx
```diff
@@ -173,13 +175,13 @@ const ProfileSettings = () => {
         const newNotificationSettings ={};
 
         if (uiConfig?.reminderSchemes) {
-            const prefs = await  NotificationScheduler.getReminderPrefs();
+            const prefs = await schedulerHelper.getReminderPrefs();
```
JGreenlee, line 178: This will become `await getReminderPrefs(uiConfig.reminderSchemes)`

www/js/splash/notifScheduler.ts
```diff
@@ -0,0 +1,286 @@
...
+    useEffect(() => {
+        if (!appConfig) {
+            logDebug("No reminder schemes found in config, not scheduling notifications");
+            return;
+        }
+        setReminderSchemes(appConfig.reminderSchemes);
+    }, [appConfig]);
+
+    //setUpActions();
+    update(reminderSchemes);
```
JGreenlee, line 279: If you remove the React hook, you'll need to do this `update` call somewhere else. You can do it in `ProfileSettings` somewhere in the `whenReady` function.

www/js/splash/notifScheduler.ts
```diff
@@ -0,0 +1,286 @@
...
+const update = async (reminderSchemes) => {
+    const { reminder_assignment,
+            reminder_join_date,
+            reminder_time_of_day} = await getReminderPrefs(reminderSchemes);
+    var scheme = {};
+    try {
+        scheme = reminderSchemes[reminder_assignment];
+    } catch (e) {
+        console.log("ERROR: Could not find reminder scheme for assignment " + reminderSchemes + " - " + reminder_assignment);
+    }
```
JGreenlee, line 174: Appreciate the addition of this catch block - but instead of `console.log` let's `displayErrorMsg` (from logger.ts)

JGreenlee (changes requested):
I left my suggestions! In short, I think the React hook is unnecessary because `ProfileSettings` already has access to `appConfig`, so you can pass its `reminderSchemes` into the `notifScheduler` functions.

And here's my answer to that question:
> I also have another question: Is notifScheduler at a high level only being used by programs and studies that use `reminderSchemes` in their config?

Correct- if `reminderSchemes` is not present, nothing in `notifScheduler` should run

www/js/splash/notifScheduler.ts
JGreenlee, line 1: Did you actually remove the old file `notifScheduler.js`? I would be surprised if the new one works with the old one still present since they have the same name.

JGreenlee: You should remove it from `index.js` too, but nowhere else I don't think

www/js/control/ProfileSettings.jsx
```diff
@@ -187,13 +201,13 @@ const ProfileSettings = () => {
     const newNotificationSettings = {};
 
     if (uiConfig?.reminderSchemes) {
-      const prefs = await NotificationScheduler.getReminderPrefs();
+      const prefs = await getReminderPrefs();
```
JGreenlee, line 204: This function takes reminderSchemes as a parameter, doesn't it?

```suggestion
      const prefs = await getReminderPrefs(uiConfig.reminderSchemes);
```

sebastianbarry: I did forget this, good catch! It was working in my testing without this, because `reminderSchemes` only gets used in `getReminderPrefs()` if the user object does not already exist on the server and it needs to do the initialization of notification-scheduling for a new user

JGreenlee (approved):
LGTM

sebastianbarry:
@JGreenlee do we also want to add unit testing within our rewrite PRs?

JGreenlee (changes requested):
Yes, I almost forgot!

We should include unit tests for as much of this as possible.
Although unit tests can't actually verify that native notifications will display on a phone, you can verify that the correct notifications are getting scheduled for the right times.

You could have a mock implementation for the native plugin which keeps track of the scheduled notifications.

sebastianbarry:
I'm also having a significant problem: my initial testing after pushing the last commits was working fine, but now I have 1. Created a new OPcode, and 2. reloaded the app, and now the app is crashing when I so much as open the profile settings screen:

https://github.com/e-mission/e-mission-phone/assets/61334340/eaeb8890-5e11-4438-a624-ac0f8feedd61

@JGreenlee do you have any tips for debugging an issue that causes the app to crash? Maybe a log dump or something that can give me insight into what is happening to make the app crash? The phone serve does not crash, just the app itself

JGreenlee:
My first guess is that it's one of the calls to the native plugins. In this case, probably the 'local notifications' plugin since that's what notifScheduler works with

I would systematically comment out these calls, see if it stops crashing, and figure out which call was causing it.
Then I would add a breakpoint just before that call to see if there's any abnormality in the data.

sebastianbarry:
I think I've found the native plugin call that is causing the issue:

in `notifScheduler.ts`:
```javascript
//get scheduled notifications from cordova plugin and format them
const getNotifs = function () {
  return new Promise((resolve, reject) => {
    window['cordova'].plugins.notification.local.getScheduled((notifs) => {
      if (!notifs?.length) {
        console.log('there are no notifications');
        resolve([]); //if none, return empty array
      }

      const notifSubset = notifs.slice(0, 5); //prevent near-infinite listing
      let scheduledNotifs = [];
      scheduledNotifs = notifSubset.map((n) => {
        const time = DateTime.fromMillis(n.trigger.at).toFormat('t');
        const date = DateTime.fromMillis(n.trigger.at).toFormat('DDD');
        return {
          key: date,
          val: time,
        };
      });
      resolve(scheduledNotifs);
    });
  });
};
```

The line: `window['cordova'].plugins.notification.local.getScheduled((notifs) => {` is where this promise ends, and does not hit any of the other lines in this function. I believe that it's critical that we have this functionality, how could I proceed with this? Am I possibly calling the wrong native plugin?

sebastianbarry:
Current status:

* The devapp still crashes when you click the profile settings screen
* The crashing problem is definitely coming from `notifScheduler.ts` line 109 in `getNotifs()`, `window['cordova'].plugins.notification.local.getScheduled((notifs) => {`, because without this line and replace with a `resolve([]);`, the app no longer crashes
* With the `resolve([]);` temporary workaround, the notifications do not get scheduled but the console logs show that the notifications data is there and correct

JGreenlee:
I looked into this and here's what I found:

`getNotifs` is not the issue. Commenting out the native call and having it resolve with `[]` *did* make the issue disappear sometimes **but not always**.
I think the issue is actually in `scheduleNotifs` with the native call to `window['cordova'].plugins.notification.local.schedule(...)`.

I commented it out and added a debug statement like this:

```js
    window['cordova'].plugins.notification.local.cancelAll(() => {
      debugGetScheduled('After cancelling');
      console.debug('scheduling notifs', nots);
      // window['cordova'].plugins.notification.local.schedule(nots, () => {
      //   debugGetScheduled('After scheduling');
      //   isScheduling = false;
      //   rs(); //scheduling promise resolved here
      // });
    });
```

Here's what the console output of that was:
image

The `id` for each one is `NaN`, which the plugin probably doesn't know what to do with and crashes.
The `id`s are supposed to be timestamps (it doesn't actually matter what format they are in as long as all are unique).
So these issues must be related to the changes from Moment -> Luxon.

Another problem I noticed is that the `trigger.at` for each of these are ISO timestamps, when they should be JS `Date` objects. (The old version used Moment `toDate()` so the new version should use Luxon `toJSDate()` https://moment.github.io/luxon/#/moment?id=basics)

www/js/splash/notifScheduler.ts
```diff
@@ -0,0 +1,269 @@
...
+};
+
+// schedules the notifications using the cordova plugin
+const scheduleNotifs = (scheme, notifTimes) => {
+  return new Promise<void>((rs) => {
+    isScheduling = true;
+    const localeCode = i18next.resolvedLanguage;
+    const nots = notifTimes.map((n) => {
+      const nDate = n.toISO();
+      const seconds = nDate.ts / 1000;
```
JGreenlee, line 134:
```suggestion
      const nDate = n.toJsDate();
      const seconds = nDate.getTime() / 1000;
```

In fact, I think this simple change will fix everything.
Instead of being a string (which does not have a `ts` property), `nDate` will be a JS `Date` object, whose `getTime` will be in milliseconds. Dividing this by 1000 will give us a timestamp in seconds, which is what `id` was in the old version.

www/js/splash/notifScheduler.ts
```diff
@@ -0,0 +1,269 @@
...
+function range(start, stop, step) {
+  let a = [start],
+    b = start;
+  while (b < stop) a.push((b += step || 1));
+  return a;
+}
+
+// returns an array of moment objects, for all times that notifications should be sent
+const calcNotifTimes = (scheme, dayZeroDate, timeOfDay) => {
+  const notifTimes = [];
```
JGreenlee, line 23: Type this as an array of `DateTime`s. Then TypeScript will be able to infer that as the return type of this function.

www/js/splash/notifScheduler.ts
```diff
@@ -0,0 +1,269 @@
...
+          val: time,
+        };
+      });
+      resolve(scheduledNotifs);
+    });
+  });
+};
+
+// schedules the notifications using the cordova plugin
+const scheduleNotifs = (scheme, notifTimes) => {
```
JGreenlee, line 128: You can also type `notifTimes` as an array of `DateTime`s.
If you had done this before, Typescript could have helped you catch the issue that caused the crash (because Typescript would have known `DateTime.toISO` returns a string, which does not have a property called `ts`.)
That's why type definitions are worth the effort!

sebastianbarry:
@JGreenlee That did it, thanks so much!! That was really great detective work 🥳

Next steps for this PR:
- [ ] At this current point, the notifications don't fill in until you hit refresh, so I need to adjust the `updateScheduledNotifs` function (which only runs on the initial load of the profile settings page)
- [ ] Add in Jest testing
- [ ] (as we saw...) Add in type declarations!

sebastianbarry:
Testing:

| Step | Description | Image |
|---|---|---|
| 1 | Upon creating a new opcode, logging in, and seeing the scheduled notifs: they are there but for the wrong days | Simulator Screenshot - iPhone 13 - 2023-11-14 at 10 30 00 |
| 2 | Hitting refresh makes them be the correct days | Simulator Screenshot - iPhone 13 - 2023-11-14 at 10 30 06 |
| 3 | Closing the app and restarting with the same opcode, notifications do not populate | Simulator Screenshot - iPhone 13 - 2023-11-14 at 10 35 19 |
| 4 | Hitting refresh again populates the correct notifications | Simulator Screenshot - iPhone 13 - 2023-11-14 at 10 36 19 |

sebastianbarry:
This to-do item has been marked complete: 
> - [x] At this current point, the notifications don't fill in until you hit refresh, so I need to adjust the `updateScheduledNotifs` function (which only runs on the initial load of the profile settings page)

The problem was that `window['cordova'].plugins.notification.local.getScheduled((notifs) => {` was returning `notifs`, a list of objects - some of these objects were empty. This was causing errors which would end the promises prematurely and they did not have time to finish scheduling the notifications. Now that we remove the empty objects from notifs, notifs are scheduling properly.

Testing:

www/js/splash/notifScheduler.ts
```diff
@@ -0,0 +1,294 @@
...
+
+export const getReminderPrefs = async (
+  reminderSchemes: object,
+  isScheduling: boolean,
+  setIsScheduling: Function,
+  scheduledPromise: Promise<any>,
+): Promise<User> => {
+  const userPromise = getUser();
+  const user = (await userPromise) as User;
+  console.log('user', user);
```
sebastianbarry, line 246: This line is not necessary
```suggestion
```

Abby-Wheelis:
@sebastianbarry, I think I've worked on similar situations, and I would recommend checking how the user is stored (some things are cached locally, and so even if you overwrite the fetching process, it will use what was fetched in another test). Maybe you'd be able to get some ideas from the tests I wrote in #1063 or the discussion in that PR, I know changing the mocks out was something that Jack and I discussed/worked on in review. And in #1093 with the tests for `storeDeviceSettings` I think I had to clear storage to reset the user. Maybe your problem in similar to one of those fixes?

the-bay-kay:
>  It appears that Jest is either ignoring my re-mocking of this, I am not re-mocking correctly, or there is some logic that causes the outer shell of mocking to take precedence over the mocks within the it tests. Does anyone have any thoughts on this?

For re-mocking, you may want to check out the [.mockImplementationOnce()](https://jestjs.io/docs/mock-function-api#mockfnmockimplementationoncefn) function!  FWIW, I've personally had some issues making it run smoothly - but if I understand your problem correctly, this may be a solution.  You could try something like...
```typescript
jest.mock('../js/services/commHelper', () => ({
  ...jest.requireActual('../js/services/commHelper'),
  getUser: jest.fn(() =>
    Promise.resolve({ // This is the default mock
      // These values are **important**...
      //   reminder_assignment: must match a key from the reminder scheme above,
      //   reminder_join_date: must match the first day of the mocked notifs below in the tests,
      //   reminder_time_of_day: must match the defaultTime from the chosen reminder_assignment in the reminder scheme above
      reminder_assignment: 'weekly',
      reminder_join_date: '2023-11-14',
      reminder_time_of_day: '21:00',
    }) // The "Once" will get called before the default, so you may need to reorder the tests...
      .mockImplementationOnce(() => {Promise.resolve({})),
  ),
  updateUser: jest.fn(() => Promise.resolve()),
```
Obviously it doesn't feel the best mocking functions on a once-off basis, like you said - but this could be an answer!  If not this, I found a workaround by mocking the higher-order functions, so that different inputs returned different values (The code for those mocks is [here](https://github.com/the-bay-kay/e-mission-phone/blob/0d1717d38079a179c1b174a2522ab991729e31c4/www/__tests__/timelineHelper.test.ts#L127) for how I got around using `mockImplementationOnce` if you're curious!  I'm not sure it is the _exact_ same issue, but hopefully it helps : ))

JGreenlee:
Just resolved merge conflicts and fixed a bug.

JGreenlee (approved):
The rewrite itself looks good to me.

Not every function is tested yet, but the tests we *do* have are passing. I suggest we merge this as-is to unblock https://github.com/e-mission/e-mission-phone/pull/1106 and then fill in the rest of the tests in a later PR.

sebastianbarry:
After committing those final tests, I believe this PR is ready for review! 🎉

@JGreenlee the main things I'm interested in is:
- Are my tests actually testing the use of the functions? i.e. I'm not `expect`ing the wrong types of things (like testing logDebug arguments too much for what is appropriate)
- Are my tests readable?

www/__tests__/notifScheduler.test.ts
```diff
@@ -0,0 +1,453 @@
...
+    jest
+      .spyOn(window['cordova'].plugins.notification.local, 'schedule')
+      .mockImplementation((arg, callback) => callback(arg));
+    // call the function
+    await updateScheduledNotifs(reminderSchemes, isScheduling, setIsScheduling, scheduledPromise);
+
+    expect(setIsScheduling).toHaveBeenCalledWith(true);
+    expect(logDebug).toHaveBeenCalledWith('After cancelling, there are no scheduled notifications');
+    expect(logDebug).toHaveBeenCalledWith('After scheduling, there are no scheduled notifications');
+    expect(setIsScheduling).toHaveBeenCalledWith(false);
```
JGreenlee, line 258: The tests should be checking the **functionality** of `notifScheduler`. In this instance we want to see that the appropriate notifications are scheduled. The `logDebug` statements are good for dev purposes but I don't think they should be the focus of our testing strategy.

After calling `updateScheduledNotifs(...)` we should verify that the function did what it was supposed to do.
In this case, I think the best way to do that is to then call `getScheduledNotifs(...)` and check the result.

www/__tests__/notifScheduler.test.ts
```diff
@@ -0,0 +1,453 @@
...
+    // mock the cordova plugin
+    jest
+      .spyOn(window['cordova'].plugins.notification.local, 'getScheduled')
+      .mockImplementation((callback) => callback(mockNotifs));
+    // call the function
+    await updateScheduledNotifs(reminderSchemes, isScheduling, setIsScheduling, scheduledPromise);
+
+    expect(logDebug).toHaveBeenCalledWith(
+      'ERROR: Already scheduling notifications, not scheduling again',
+    );
```
JGreenlee, line 304: In this instance, though, I think checking the `logDebug` does make sense because we expect `updateScheduledNotifs` to have quit early and not scheduled any notifications.

www/__tests__/notifScheduler.test.ts
```diff
@@ -0,0 +1,453 @@
...
+    // updateScheduleNotifs arguments
+    let reminderSchemes: any = exampleReminderSchemes;
+    delete reminderSchemes.weekly; // delete the weekly reminder scheme, to create a missing reminder scheme error
+    let isScheduling: boolean = false;
+    const setIsScheduling: Function = jest.fn((val: boolean) => (isScheduling = val));
+    const scheduledPromise: Promise<any> = Promise.resolve();
+    // call the function
+    await updateScheduledNotifs(reminderSchemes, isScheduling, setIsScheduling, scheduledPromise);
+
+    expect(logDebug).toHaveBeenCalledWith('Error: Reminder scheme not found');
```
JGreenlee, line 317: This one is fine too

JGreenlee (changes requested):
The purpose of tests is to verify that when a function is called, the expected result is achieved. It looks to me like after calling `updateScheduledNotifs`, we don't actually make sure that those notifications were scheduled.

I think we need to revise the approach for testing that function, but I am still going to recommend merging this now and bulking up the tests in a later PR.

www/__tests__/notifScheduler.test.ts
sebastianbarry: I added the test you suggested, to test `getScheduledNotifs(...)` in https://github.com/e-mission/e-mission-phone/pull/1092/commits/ec9729de081000fcdc9bed288b8da5c0ee3b77c8

Great suggestion, thanks!

www/__tests__/notifScheduler.test.ts
shankari: I agree with:

> The logDebug statements are good for dev purposes but I don't think they should be the focus of our testing strategy.

In a future fix, I would suggest changing the underlying function to return a value that you can check, and then changing the test to check it

www/__tests__/notifScheduler.test.ts
shankari: Ditto

www/__tests__/notifScheduler.test.ts
```diff
@@ -0,0 +1,471 @@
...
```
shankari, line 335: I don't understand this. you are removing the `weekly` reminderScheme from the local in-memory variable.
There are no existing reminders, right? So how would `updateScheduledNotifs`  know that there was supposed to have been a weekly scheme?

shankari (approved):
I have some comments/questions, but am going to go ahead and merge this so that we can get the release out and clean this up later, consistent with 

> I suggest we merge this as-is to unblock https://github.com/e-mission/e-mission-phone/pull/1106 and then fill in the rest of the tests in a later PR.

shankari:
After this merge, `notifScheduler` tests are failing while running jest for code coverage
https://github.com/e-mission/e-mission-phone/actions/runs/7597279222/job/20691918950?pr=1085
while run as part of testing https://github.com/e-mission/e-mission-phone/pull/1085


```
Summary of all failing tests
FAIL www/__tests__/notifScheduler.test.ts
  ● updateScheduledNotifs › should resolve after scheduling notifications

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: "After scheduling, there are 4 scheduled notifications at 21:00 first is November 19, 2023 at 9:00 PM"
    Received
           1: "requesting fetch while still actively scheduling, waiting on scheduledPromise"
           2: "there are no notifications"
           3: "User already has reminder prefs, returning them: {\"reminder_assignment\":\"weekly\",\"reminder_join_date\":\"2023-11-14\",\"reminder_time_of_day\":\"21:00\"}"

    Number of calls: 5

      270 |     expect(setIsScheduling).toHaveBeenCalledWith(true);
      271 |     expect(logDebug).toHaveBeenCalledWith('After cancelling, there are no scheduled notifications');
    > 272 |     expect(logDebug).toHaveBeenCalledWith(
          |                      ^
      273 |       'After scheduling, there are 4 scheduled notifications at 21:00 first is November 19, 2023 at 9:00 PM',
      274 |     );
      275 |     expect(setIsScheduling).toHaveBeenCalledWith(false);

      at toHaveBeenCalledWith (www/__tests__/notifScheduler.test.ts:272:22)
      at call (www/__tests__/notifScheduler.test.ts:2:1)
      at Generator.tryCatch (www/__tests__/notifScheduler.test.ts:2:1)
      at Generator._invoke [as next] (www/__tests__/notifScheduler.test.ts:2:1)
      at asyncGeneratorStep (www/__tests__/notifScheduler.test.ts:2:1)
      at asyncGeneratorStep (www/__tests__/notifScheduler.test.ts:2:1)

```

PR #1090, opened 2023-10-30
Rewrite PushNotify.ts ➡️  PushNotifySettings.ts

Abby-Wheelis:
Rewrite the service that handles the push notification settings.

This PR will be a continuation of #1072, since it affects the "reconsent" and "intro done" actions. I will manage updates made during review to that service as I merge them into this service.

PR #1093, opened 2023-10-31
🎟️ 🔔 📱 Custom Events + pushNotify + storeDeviceSettings + remoteNotify rewrites

Abby-Wheelis:
This will replace #1090, after we re-evaluated and made a decision to continue forward with custom events in [#1024](https://github.com/e-mission/e-mission-docs/issues/1024). I elected to start a fresh PR to keep the blame as clean as I could. 

This introduces custom events, and migrates `pushNotify` to typescript. As this PR stands, there will be events "duplicated" because some applications still use `$emit` and `$on` while others use the new `publish` and `subscribe`. Where applicable, some events are broadcast using both methods, so they can be received by both methods.


Abby-Wheelis:
The custom events fit well into this paradigm, I've added the subscription calls to the "init" function that is replacing the `platform.ready()` call. The difficulty here is finding the best place to call this initialization script, I'm not sure we have a translation for `$ionicPlatform.ready()` yet, at least that I know of.

Abby-Wheelis:
One solution I found was to call the initialization logic using `componentDidMount` but that method did not seem to exist in `App.tsx` - I instead decided to add the call near other initialization logic: 
```
  useEffect(() => {
    if (!appConfig) return;
    setServerConnSettings(appConfig).then(() => {
      refreshOnboardingState();
    });
    initPushNotify();
  }, [appConfig]);
```
I'm not sure if this is the best place for this, but it does seem to be working

Abby-Wheelis:
This is going to have to be merged after #1072, because I pulled those changes into this PR, but the events are implemented and tested, and `pushNotify` has been rewritten and has tests. I'm glad we went with events, it made the code easier to test and somewhat simpler to implement.

Abby-Wheelis:
Ok, that's my test that's failing now, I can sometimes replicate the fail locally, let me see if I can find the cause.

Abby-Wheelis:
On checking for anything else listening for the events, I found that `remotenotify` listens for the notification event -- 
```
$rootScope.$on('cloud:push:notification', function (event, data) {
```
Since these changes alter the way we handle events, I'm probably going to move ahead with rewriting that file here as well. That will make this PR relatively big, but should take care of all the event-driven functionality in splash at once.

Abby-Wheelis:
This PR now encompasses all of the uses of the events for "intro done" "consent done" and "cloud notification". This includes `customEventHandler`, `pushNotify`, `storeDeviceSettings`, and `remoteNotify` -- along with the updates to my work in #1072 to ensure the events are handled in the new way. There are also tests for the custom events and each of the three `splash` services rewritten here!

www/js/splash/remoteNotifyHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+*/
+
+/**
+ * @function launches popup
+ * @param title string text for popup title
+ * @param text string text for popup bode
+ */
+const launchPopup = function (title, text) {
+  // THIS LINE FOR inAppBrowser
+  displayErrorMsg(text, title);
```
JGreenlee, line 42: I know we don't have a good popup mechanism right now so you used this temporarily, but this would actually log an error to Logger plugin and I don't think that's appropriate here

Either use `window.alert` directly or add a function to `logger.ts` like `displayPopup`, which could be called by `displayErrorMsg`

And I think you can do that call directly from line 74  -- this 1-line function serves no purpose

www/js/customEventHandler.ts
JGreenlee, line 1: This is really cool and not nearly as complex as I thought!

www/js/customEventHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+import { logDebug } from './plugin/logger';
+
+/**
+ * central source for event names
+ */
+export const EVENT_NAMES = {
+  CLOUD_NOTIFICATION_EVENT: 'cloud:push:notification',
+  CONSENTED_EVENT: 'data_collection_consented',
+  INTRO_DONE_EVENT: 'intro_done',
+};
```
JGreenlee, line 23: nitpicky, but might it be nice to rename and make these more concise?

Wherever these are used, we have to write out `EVENT_NAMES.CLOUD_NOTIFICATION_EVENT`. It seems redundant to repeat the word 'event' and I think it's self-evident that these are the names of the events
Could we rename to `EVENTS.CLOUD_NOTIFICATION`?

www/js/splash/pushNotifySettings.ts
```diff
@@ -0,0 +1,241 @@
...
+};
+
+/**
+ * @function registers for notifications and updates user
+ * currently called on reconsent and on intro done
+ */
+const registerPush = function () {
+  registerPromise()
+    .then(function (t) {
+      // alert("Token = "+JSON.stringify(t));
```
JGreenlee, line 93: remove?

www/js/splash/pushNotifySettings.ts
```diff
@@ -0,0 +1,241 @@
...
+      push.finish(function () {}, finishErrFn, notId);
+      displayError(error, 'Error while redirecting silent push');
+    });
+};
+
+/**
+ * @function shows debug notifications if simulating user interaction
+ * @param message string to display in the degug notif
+ */
+var showDebugLocalNotification = function (message) {
```
JGreenlee, line 165: `const` here

(or simply `function showDebugLocalNotification(message) {...}`)

www/js/splash/remoteNotifyHandler.ts
```diff
@@ -0,0 +1,88 @@
...
+  addStatEvent(statKeys.NOTIFICATION_OPEN).then(() => {
+    console.log('Added ' + statKeys.NOTIFICATION_OPEN + ' event. Data = ' + JSON.stringify(data));
+  });
+  logDebug('data = ' + JSON.stringify(data));
+  if (
+    data.additionalData &&
+    data.additionalData.payload &&
+    data.additionalData.payload.alert_type
+  ) {
+    if (data.additionalData.payload.alert_type == 'website') {
```
JGreenlee, line 60: @shankari Are we ever going to re-implement external web launches via push notification? If not, we can simplify

JGreenlee (changes requested):
First pass of review-
This looks really clean already and I just have a few comments right now. The event code is definitely useful and I can see how it all comes together in the places you've utilized it!

www/js/splash/pushNotifySettings.ts
Abby-Wheelis: That was there when I migrated over, I looked into the blame and didn't see any notes about it (and the change is 7 years old), so I think yes, safe to remove.

www/js/splash/remoteNotifyHandler.ts
Abby-Wheelis: done in [c25ebb6](https://github.com/e-mission/e-mission-phone/pull/1093/commits/c25ebb6f71f20d24f14b135186da66013b6806f8)

www/js/splash/pushNotifySettings.ts
Abby-Wheelis: done in [ad68e1b](https://github.com/e-mission/e-mission-phone/pull/1093/commits/ad68e1bb0e3a3387850ccc45912a3ce468184a37)

www/js/splash/pushNotifySettings.ts
Abby-Wheelis: Done in [85d2884](https://github.com/e-mission/e-mission-phone/pull/1093/commits/85d2884128aecc351ccaafd12a7ab553f8385e83) -- I decided to keep the const to stay consistent with the rest of the file

www/js/splash/remoteNotifyHandler.ts
shankari: I think we should re-implement all the different types of push notification. I have pitched the external web launches, in particular, as mechanism for providing incentives for deployments with autogenerated tokens. The program admin can send a link to an amazon gift card directly to the participant though a push notification, without having to know their email/phone number etc.
Follow on to https://github.com/e-mission/e-mission-docs/issues/976

shankari:
This has a merge conflict. Let me try to resolve if it is easy to avoid getting stuck in a circular loop...

www/__tests__/storeDeviceSettings.test.ts
```diff
@@ -0,0 +1,120 @@
...
+    client_app_version: '1.2.3',
+  });
+});
+
+it('stores device settings after consent if intro done', async () => {
+  initStoreDeviceSettings();
+  await new Promise((r) => setTimeout(r, 500)); //time to check consent and subscribe
+  markIntroDone();
+  await new Promise((r) => setTimeout(r, 500));
+  publish(EVENTS.CONSENTED_EVENT, 'test data');
```
shankari, line 104: **future fix**:  I think you also want to check that the user settings are not set before the CONSENTED_EVENT even though the intro is done.

shankari (approved):

shankari:
@JGreenlee merging this now - another potential source of merge conflicts

PR #1110, opened 2023-12-01
Customize Labels Step 1

jiji14:
### Related Issue
- System Design: [GitHub Issue 1012](https://github.com/e-mission/e-mission-docs/issues/1012)
- Server: [GitHub Server Pull Request 946](https://github.com/e-mission/e-mission-server/pull/946)

### What I added
1. Divided default label and custom label.
2. Saved user's custom mode.

### Next Step 
1. Manage "My Labels" section in the Profile tab.
2. Delete label function.

### Result

Screenshot 2024-01-05 at 12 34 49 PM

<br />

Screenshot 2024-01-05 at 12 34 59 PM

<br />

#### Please check out to `customize-modes` branch in the `Server` repository for testing.


jiji14:
@JGreenlee 
I wanted to add unit tests for `getModes` and `updateMode` functions but it seems like there's no point in just mocking`BEMServerComm`

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -151,16 +191,47 @@ const MultilabelButtonGroup = ({ trip, buttonsInline = false }) => {
             </Dialog.Title>
             <Dialog.Content style={{ maxHeight: windowHeight / 2, paddingBottom: 0 }}>
               <ScrollView style={{ paddingBottom: 24 }}>
+                <Text style={{ fontSize: 12, color: '#333333', paddingVertical: 4 }}>
```
JGreenlee, line 194: We have a dynamic theme system where all the colors are derived from the theme defined in `/js/appTheme.ts`.
Instead of inline hex codes please use the various properties of `colors`, like `colors.onBackground` or `colors.onSurface`
The reason for doing this is 1) it's tidy and 2) if we want to change something it's easier to modify or swap themes from one location rather than scattered across the codebase

More info on this: https://callstack.github.io/react-native-paper/docs/guides/theming/#theme-properties

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -151,16 +191,47 @@ const MultilabelButtonGroup = ({ trip, buttonsInline = false }) => {
...
                     />
                   ))}
+                  <Divider style={{ marginVertical: 10 }} />
+                  <RadioButton.Item
+                    key="Other"
+                    label={t('Other')}
+                    value="other"
+                    style={{ paddingTop: 10 }}
+                  />
                 </RadioButton.Group>
```
JGreenlee, line 235: Instead of skipping 'other' and tacking it back on at the end, here is what I would recommend.
Go through the list as normal and when you encounter 'other', just prepend it with the whole 'custom modes' section

```tsx
  <RadioButton.Group onValueChange={(val) => onChooseLabel(val)} value={chosenLabel}>
    {labelOptions?.[modalVisibleFor]?.map((o, i) => {
      const radioItemForOption = (
        <RadioButton.Item
          key={i}
          label={t(o.text)}
          value={o.value}
          style={{ paddingVertical: 2 }}
        />
      );
      /* if this is the 'other' option and there are some custom modes,
        show the custom modes section before 'other' */
      if (o.value == 'other' && customModes.length) {
        return (
          <>
            <Divider style={{ marginVertical: 10 }} />
            <Text style={{ fontSize: 12, color: '#333333', paddingVertical: 4 }}>
              Custom Mode
            </Text>
            {customModes.map((key, i) => (
              <RadioButton.Item
                key={i}
                label={getProcessedText(key)}
                value={key}
                style={{ paddingVertical: 2 }}
              />
            ))}
            <Divider style={{ marginVertical: 10 }} />
            {radioItemForOption}
          </>
        );
      }
      // otherwise, just show the radio item as normal
      return radioItemForOption;
    })}
  </RadioButton.Group>
```

JGreenlee: This will also work in the event that any future program opts not to have 'other' as an option

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -172,6 +243,7 @@ const MultilabelButtonGroup = ({ trip, buttonsInline = false }) => {
                   })}
                   value={otherLabel || ''}
                   onChangeText={(t) => setOtherLabel(t)}
+                  maxLength={25}
```
JGreenlee, line 246: good

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -151,16 +191,47 @@ const MultilabelButtonGroup = ({ trip, buttonsInline = false }) => {
...
+                        value={o.value}
+                        style={{ paddingVertical: 2 }}
+                      />
+                    );
+                  })}
+                  {customModes.length > 0 && (
+                    <>
+                      <Divider style={{ marginVertical: 10 }} />
+                      <Text style={{ fontSize: 12, color: '#333333', paddingVertical: 4 }}>
+                        Custom Mode
```
JGreenlee, line 215: This will need to be internationalized instead of using direct English text
Add a new translation entry to `en.json` and use the `t` function here.

www/__tests__/commHelper.test.ts
```diff
@@ -47,4 +47,6 @@ it('fetches text from a URL and caches it so the next call is faster', async ()
   * - updateUser
   * - getUser
   * - putOne
+  * - getModes
+  * - updateMode
```
JGreenlee, line 51: Fine to not having unit tests for these

www/js/services/commHelper.ts
```diff
@@ -229,3 +229,32 @@ export function putOne(key, data) {
...
+export function getModes() {
+  return new Promise((rs, rj) => {
+    window['cordova'].plugins.BEMServerComm.getUserPersonalData('/mode/get', rs, rj);
+  }).catch((error) => {
+    error = 'While getting modes, ' + error;
+    throw error;
+  });
+}
+
+export function updateMode(oldMode, newMode, isNewModeMustAdded) {
```
JGreenlee, line 242: I would make these function names more descriptive so it's clear these are for user-defined modes only.

Potentially `getCustomModes`+`updateCustomMode`, or even `getUserCustomModes`+`updateUserCustomMode`

JGreenlee: ~~I also see that the words here are `get`/`update` and I wonder if `get`/`put` would be more consistent with the terminology on the server~~

JGreenlee: ^ Nevermind, after looking I agree that `update` is better

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -28,24 +29,25 @@ import {
...
     'MODE' | 'PURPOSE' | 'REPLACED_MODE' | null
   >(null);
   const [otherLabel, setOtherLabel] = useState<string | null>(null);
+  const [customModes, setCustomModes] = useState<string[]>([]);
   const chosenLabel = useMemo<string>(() => {
     if (otherLabel != null) return 'other';
     return timelineLabelMap[trip._id.$oid]?.[modalVisibleFor]?.value;
   }, [modalVisibleFor, otherLabel]);
-
+  const initialLabel = chosenLabel;
```
JGreenlee, line 50: I don't see the purpose of this variable. It seems like `initialLabel` will always be the same as `chosenLabel`.

It is never reassigned (and can't be anyway because it is `const`)

JGreenlee (changes requested):
Good first steps. Interface looks great!
Left some comments for code cleanliness + an open question about performance

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -94,6 +111,29 @@ const MultilabelButtonGroup = ({ trip, buttonsInline = false }) => {
...
+
+  // ex) my_car -> My Car
+  const getProcessedText = (key: string): string =>
+    key
+      .split('_')
+      .map((k) => k.charAt(0).toUpperCase() + k.slice(1))
+      .join(' ');
+
+  useEffect(() => {
+    // Whenever the modal opens for Mode, retrieve updated modes from the server for synchronized data
```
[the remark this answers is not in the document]
jiji14: I was contemplating the choice between `consistency` and `efficiency`. If we only load custom modes on the initial load, there might be inconsistency. That's why I fetch the custom modes whenever the modal opens. However, I also had concerns about unnecessary API calls. Considering your comment, I think efficiency is a better trade-off. Therefore, I will change the logic as you recommended!

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
jiji14: @JGreenlee Thank you! I love this code! Brilliant 🥇

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
jiji14: The reason I used `initialLabel` was that I couldn't access the 'chosenLabel' value because the `store` function had a 'chosenLabel' parameter.

To access `chosenLabel`, I needed to either change the parameter name from `chosenLabel` to a different one or create a new variable that contains `chosenLabel`. That's why I declared 'initialLabel.'

I deleted line 50, changed `chosenLabel` to `initialLabel`, and modified the parameter name from `chosenLabel` to `newLabel`. I think it seems more intuitive.

JGreenlee (changes requested):
Great step forward! The logic in MultiLabelButtonGroup is cleaner now and I think this will be better for network performance.

Now, I have some questions about the design: Is this supposed to work for custom purposes too?

If so, it needs to be made more generic, and the feature may be more accurately described as "custom labels" rather than just "custom modes".

If not, we need to fix regressions. The current behavior is that the same list of "Custom modes" shows up for both "mode" and "purpose" popups, which is not what we want.

JGreenlee:
Here is what I mean by the same list of custom modes showing up in both popups

<table><td>

</td><td>

</td></table>

I think these should be kept as two separate lists, and the text should reflect that

jiji14:
> Here is what I mean by the same list of custom modes showing up in both popups
> 
> 	
> I think these should be kept as two separate lists, and the text should reflect that

Thanks for reporting this! I forgot to add to check `inputType === 'MODE'`. I will submit a different PR for `PURPOSE` later.

JGreenlee (commented):
If the expectation for this PR is to ***only*** support custom modes, and not custom purposes / custom replaced modes, it looks ready.

However, if the expectation is to have both custom modes and custom purposes supported (as was supported previously), I would still suggest revising this PR to a more generic "custom labels" implementation. It would be more efficient to do in one PR rather than two.

Sending this along to the next step of review so @shankari can weigh in.

jiji14:
@JGreenlee 
After rethinking the design, I agreed that it would be more readable and efficient to modify functions to accommodate all types of labels (mode, purpose, and replaced_mode) and dynamically change them with a key. I have also updated the server code, so please pull both branches if you want to run this. Thanks for your advice!

JGreenlee (changes requested):
I have a few other nitpicks for code style and tidieness left below

But when I tried to test this (I was using this branch `customize-modes` and the server branch `customize-modes`), I got the following peculiar result:

When I selected "Other" to add a custom mode, the text box appeared below, but the radio button next to "Other" did not fill in, making it look like "Other" was not selected


If I continued with trying to add a custom mode and clicked "Save", I got this message:

jiji14:
@JGreenlee 
Thank you for assisting me in writing cleaner code! Your comments are very helpful. I updated some variable names and fixed the issue with the 'other' radio button. While attempting to replicate the issue you mentioned, I encountered no errors. I tested with a new account and also devised a test case for scenarios where the label keys (mode, purpose, replaced_mode) are not present in the database. Considering the server code got changed, the error might have been due to outdated server code. Could you please check if the server code is up-to-date? If not, kindly pull the latest changes and retry. Thank you! 🙏

www/js/diary/LabelTab.tsx
```diff
@@ -27,19 +27,21 @@ import { fillLocationNamesOfTrip, resetNominatimLimiter } from './addressNamesHe
...
   TimelineLabelMap,
   TimelineMap,
   TimelineNotesMap,
+  CustomLabelKey,
 } from './LabelTabContext';
 
 let showPlaces;
 const ONE_DAY = 24 * 60 * 60; // seconds
 const ONE_WEEK = ONE_DAY * 7; // seconds
+const CUSTOM_LABEL_KEYS = ['mode', 'purpose', 'replaced_mode'];
```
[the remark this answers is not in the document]
JGreenlee: That's a good reason and makes sense. Thanks for clarifying!

JGreenlee (commented):
Great! I created a new test user and verified that the error did not occur, so that was my mistake.
I also verified that the other comments have been addressed.

All I have left are a couple design questions:

- Should the same list of custom modes be shared for "mode" and "replaced mode" ?
-- It was clear before that "mode" and "purpose" should be kept as separate lists. But after testing this branch and envisioning day-to-day use, I think it may be more logical for "replaced mode" to use the same list of custom labels as "mode".

- Is it necessary to have the section title "Default Mode", "Default Purpose", etc?
-- I think this takes up extra space and the layout would be clearer without it.

jiji14:
That's a very good point! I was also unsure in separating 'mode' and 'replaced_mode'. I agree with using the same list. Please check my updates!

JGreenlee (approved):
LGTM!

JGreenlee:
Resolved merge conflicts.

@jiji14 Can you please change the target of this PR to `master`?

jiji14:
> Resolved merge conflicts.
> 
> @jiji14 Can you please change the target of this PR to `master`?

Done!

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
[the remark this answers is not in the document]
shankari: Given that these are user-specific labels, I don't think there are any issues about consistency as long as you maintain a local cache and unify. You will only run into consistency issues if we maintain a shared set of modes. Which we may want to do eventually so that, for aggregate analysis, we may not have library, Library and liBray. That is much harder to design:
- what do we do about multiple languages, for example (e.g. church, Church and iglesia, Iglesia)
- even more complicated with non-roman scripts (e.g. temple versus கோவில் versus मंदिर)
- what about bad actors who use inappropriate language in the modes (no examples here to keep it SFW, but I think you can get the idea)

So we will stick with user specific modes for this initial step. The user can fill them in the language of their choice, and if they use inappropriate language, they are the only ones affected.

So definitely stick with performance for now. When we get to the global mode table, we can think about consistency (and will likely use eventual consistency for performance).

shankari:
@jiji14 I am reviewing this, but please note that the codecov tests are failing. The `run-codecov` task is passing, but the results of running the code coverage is that you don't have enough tests (the diff is not covered and the code coverage will go down thanks to this PR). Can you please add the appropriate tests before I merge?

jiji14:
> @jiji14 I am reviewing this, but please note that the codecov tests are failing. The `run-codecov` task is passing, but the results of running the code coverage is that you don't have enough tests (the diff is not covered and the code coverage will go down thanks to this PR). Can you please add the appropriate tests before I merge?

@shankari  We're currently facing challenges with testing API fetching functions, as they rely on native functions within BEMServerComm. While considering mocking these native functions for testing purposes, we initially deemed it unnecessary, as it would involve skipping tests for the actual API fetching functions. However, I will do some research about it!

Here are the comments about the test:
```
/* The following functions from commHelper.ts are not tested because they are just wrappers
    around the native functions in BEMServerComm.
  If we wanted to test them, we would need to mock the native functions in BEMServerComm.
    It would be better to do integration tests that actually call the native functions.

  * - getRawEntries
  * - getRawEntriesForLocalDate
  * - getPipelineRangeTs
  * - getPipelineCompleteTs
  * - getMetrics
  * - getAggregateData
  * - registerUser
  * - updateUser
  * - getUser
  * - putOne
  * - getModes
  * - updateMode
*/
```

shankari:
@jiji14 I saw that, but codecov didn't :) It is possible to mark these as integration wrapper functions so that codecov only considers them for integration tests?

JGreenlee:
I am going to suggest that we close this PR and focus on the Step 2 PR (https://github.com/e-mission/e-mission-phone/pull/1120), because:
1. The Step 2 PR already includes all of these Step 1 changes, plus more
1. There are bugs with getting Step 1 and Step 2 to work together, which should be resolved before we roll out Custom Labels
1. If @jiji14 adds tests / excludes files from coverage here, she will have to also pull that into Step 2, and it's simpler to just do it in one PR

So concretely, my suggestions for @jiji14:
1. Fix the bugs in https://github.com/e-mission/e-mission-phone/pull/1120
1. Then exclude `commHelper` from the Jest coverage report (https://stackoverflow.com/a/55415228)
1. Make sure #1120 works, mark it for review and make sure it passes all checks

jiji14:
> I am going to suggest that we close this PR and focus on the Step 2 PR (#1120), because:
> 
> 1. The Step 2 PR already includes all of these Step 1 changes, plus more
> 2. There are bugs with getting Step 1 and Step 2 to work together, which should be resolved before we roll out Custom Labels
> 3. If @jiji14 adds tests / excludes files from coverage here, she will have to also pull that into Step 2, and it's simpler to just do it in one PR
> 
> So concretely, my suggestions for @jiji14:
> 
> 1. Fix the bugs in [Custom Labels Step 2 #1120](https://github.com/e-mission/e-mission-phone/pull/1120)
> 2. Then exclude `commHelper` from the Jest coverage report (https://stackoverflow.com/a/55415228)
> 3. Make sure [Custom Labels Step 2 #1120](https://github.com/e-mission/e-mission-phone/pull/1120) works, mark it for review and make sure it passes all checks

@JGreenlee 
Good idea! Thanks for your suggestions :)

PR #1111, opened 2023-12-04
🧹 📊  Metrics cleanup

Abby-Wheelis:
Addressing code review from #1098 and codecov results from the same PR


Abby-Wheelis:
I made a discovery working through upping the codecov report - in both `metHelper` and `footprintHelper` we have some code that catches if the mode is 'ON_FOOT' and sets  it to `WALKING` -- 
in `metHelper` this is great -- we have the met-equivalents and fallbacks, so it turns out the same as 'WALKING`
but in 'footprintHelper` -- we are now always using custom modes, so we have things like `walk` and it does not recognize `WALKING` as a part of the list

For `footprintHelper` -- if the mode is `ON_FOOT` -- that means the energy value will always be 0, would it be bad to hardcode a return of 0 if the mode is `ON_FOOT`

Also, is there anywhere we still use `ON_FOOT`? I'm guessing this came from some previous iteration, as I don't think I've seen it in play as a base mode or label, is it still lurking somewhere? I suppose removing the code and allowing `ON_FOOT` to go to the default value is also an option.

Abby-Wheelis:
>For footprintHelper -- if the mode is ON_FOOT -- that means the energy value will always be 0, would it be bad to hardcode a return of 0 if the mode is ON_FOOT

This is actually even more complicated when it comes to the `IN_VEHICLE` tests... since we don't know what all the vehicle modes are...
`else if (mode == 'IN_VEHICLE') {
      const sum =
        footprint['CAR'] +
        footprint['BUS'] +
        footprint['LIGHT_RAIL'] +
        footprint['TRAIN'] +
        footprint['TRAM'] +
        footprint['SUBWAY'];
      result += (sum / 6) * mtokm(userMetrics[i].values);
    }`

What do we want to do there? My first instinct would be to cache some sort of calculation of all the modes with `IN_VEHCILE` as a met-equivalent, but even that starts mixing the mets into the carbon calculations and could be overturned as we add user-custom labels, grid data, etc

Abby-Wheelis:
Something else I've found while looking into the code coverage: the range - limits for modes
I can see where this WAS used in the code, but I don't think it's used anymore. I believe this was something that might have influenced things like mode recommendations, but we don't currently have those, should I just take it out? I could also invent some labels just for the tests to test it, but I don't think any of the labels in production have this, and the code base doesn't have the feature.

Abby-Wheelis:
Screenshot 2023-12-04 at 4 02 53 PM
The few uncovered sections in the metrics files are now pending discussion of the above issues:
- difference between default and fallback keys in the footprint calculations: how to handle `ON_FOOT` and `IN_VEHICLE`
- `range_limit_km` - I believe this is a holdover from a previous feature (recommendations?) as we don't seem to use it, so the best course of action is probably to remove it for now, though I think I see a relatively easy path to get it covered by tests if we decide we want that. 

Once we decide those two things, I think I can get the last few patches of lines tested.

Abby-Wheelis:
After discussion in our UI meeting today, I've gone ahead and removed the old code, which increased the code coverage, as that was most of what was not getting tested. I've tested in the emulator, and the dashboard is working well!

Abby-Wheelis:
After resolving merge conflicts and one issue the merge created this cleanup PR is ready again!

shankari:
> Also, is there anywhere we still use ON_FOOT? I'm guessing this came from some previous iteration, as I don't think I've seen it in play as a base mode or label, is it still lurking somewhere? I suppose removing the code and allowing ON_FOOT to go to the default value is also an option.

The android raw values include `ON_FOOT`. I believe that the server code converts to `WALKING` when we look at inferred sections, but the raw and cleaned values can still have ON_FOOT. If you look at the real examples in the server tests, you will see many examples of the expected result (aka "ground truth") having `ON_FOOT`

www/js/metrics/customMetricsHelper.ts
```diff
@@ -1,14 +1,23 @@
...
 let _customMETs: { [key: string]: { [key: string]: { range: number[]; met: number } } };
 let _customPerKmFootprint: { [key: string]: number };
-let _range_limited_motorized;
 let _labelOptions;
 
+/**
+ * ONLY USED IN TESTING
+ * @function clears the locally stored variables
+ */
+export function _test_clearCustomMetrics() {
```
shankari, line 15: I don't really like this, but also don't see a much better alternative.

www/js/metrics/footprintHelper.ts
```diff
@@ -47,21 +46,10 @@ export function getFootprintForMetrics(userMetrics, defaultIfMissing = 0) {
...
   for (let i in userMetrics) {
     let mode = userMetrics[i].key;
-    if (mode == 'ON_FOOT') {
-      mode = 'WALKING';
-    }
 
+    //either the mode is in our custom footprint or it is not
     if (mode in footprint) {
       result += footprint[mode] * mtokm(userMetrics[i].values);
-    } else if (mode == 'IN_VEHICLE') {
```
shankari, line 56: We can return `IN_VEHICLE` if we use cleaned modes, before the infer modes step of the pipeline.
That is why we see `IN_VEHICLE` in the public dashboard. There will not be an `IN_VEHICLE` mode in the custom labels, but it can still be returned if we set the server to use the cleaned modes instead of the inferred modes or the user labels (`conf/analysis/debug.conf.json.sample, `analysis.result.section.key`)

I would prefer to restore this until we decide how to handle anergy/emission calculations and/or count every trip.

**But** if it is too complicated, I would be open to leaving it off, since the fallback to the labeled trips is typically the inferred sections.

shankari (changes requested):
Almost there

www/__tests__/customMetricsHelper.test.ts
```diff
@@ -41,12 +52,13 @@ it('gets the custom mets', async () => {
...
-    walk: expect.any(Number),
-    bike: expect.any(Number),
-    bikeshare: expect.any(Number),
-    'e-bike': expect.any(Number),
-    scootershare: expect.any(Number),
-    drove_alone: expect.any(Number),
+    walk: 0,
+    bike: 0,
+    bikeshare: 0,
+    'e-bike': 0.00728,
```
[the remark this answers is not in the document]
Abby-Wheelis: If I recall correctly, that was because of the `-` -- when I get the chance to work on this and address the rest of review I'll double check it's needed!

www/js/metrics/footprintHelper.ts
Abby-Wheelis: I can add this back in, but I'm not sure I fully understand when this case would come up to be able to simulate it for testing. 

When we build the footprint in `customMetricsHelper` the keys that are used are the keys from the labels, and this snippet aims to average the numbers for 6 labels that might or might not be in that list. I think to do something like this we would need to refactor a bit to have base modes in the footprint

Maybe when the footprint is extracted from the labels we could average all the modes with a "vehicle" base mode and average them together and store it for use in this case?

Should I work on adding something like that to this PR or would it be better to wait?

www/__tests__/customMetricsHelper.test.ts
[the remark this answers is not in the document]
Abby-Wheelis: The quotes prevent a syntax error, I added them to the rest for consistency, but prettier removed them so I think it has to be formatted like this

www/js/metrics/footprintHelper.ts
shankari: @Abby-Wheelis just restore it. Let's decide what to do with it when we revisit the calculations. No tests (which means that test coverage will be a bit lower) but that's OK for now.

shankari (approved):

PR #1112, opened 2023-12-04
Accessibility support

niccolopaganini:
I repurposed the functions from one of the buttons in the Welcome Page and added on top of the scan and paste buttons. I will have to determine the size and everything.


niccolopaganini:
I will update it with prettier later after addressing `unpronounceable button`/ button issues

niccolopaganini:
- [x] Welcome Page now supports Accessibility 🎊

PR #1113, opened 2023-12-04
🧹 Enketo cleanup

Abby-Wheelis:
Addressing comments from #1063 and the codecov report for affected files


Abby-Wheelis:
I have added some more tests, and now the testing coverage is up to about 98% of the lines for `enketoHelper.ts`. This gap is accounted for by not having a test cases that tests out the xml->JSON transformer. @JGreenlee do you think we'd be able to test this easily? I tried briefly, but got an error from the transformer and wasn't sure what the best approach would be. 

I think I've also addressed the comments from #1063 except for the couple of places where I was confused. You can see those review comments here: [1063 review](https://github.com/e-mission/e-mission-phone/pull/1063#pullrequestreview-1761367005)

Abby-Wheelis:
Simulator Screenshot - iPhone 13 Pro - 2023-12-05 at 14 10 20

Double checked, working well in the emulator!

Pending my confusion over review comments on the previous PR, I think this is ready to start review!

www/js/survey/enketo/enketoHelper.ts
```diff
@@ -266,6 +267,13 @@ export function saveResponse(
               : opts.timelineEntry.exit_ts,
           };
         }
+        // if timestamps were not resolved from the survey, we will use the trip or place timestamps
```
JGreenlee, line 270: Why is this needed? Can you provide context?

JGreenlee: The if/else block above should already handle this

Abby-Wheelis: I think that was a bad merge conflict, Thank you for pointing it out! It's been so long since I first wrote this so I'm having a hard time cleaning it back up, moved it back to "in progress" until I can get the tests running again!

Abby-Wheelis:
>testing the case where invalid timestamps are entered is resolving with the whole form instead of rejecting with the expected error message, I haven't figured out why yet

The Promise is not rejecting because the `resolvedTimestamps` turn out to be `undefined` and the call to `Promise.reject` seems to only really apply to the `resolveTimestamps` function, rather than the parent Promise, which is fairly complicated. I can see in the console that the error is thrown, but I can't figure out how to keep the Promise from resolving, which I think is the expected behavior, since we want to reject invalid timestamps...

Abby-Wheelis:
>the test for loadPreviousResponseForSurvey is resolving with undefined instead of the expected object, I don't know why yet

The issue here is that there are no responses fetched, I'll have to check on the mocks to figure out why there are no responses, starting with looking at what might have changed since this test was running previously...

PR #1114, opened 2023-12-05
Run Codecov Workflow only when PR is ready

jiji14:
### Why Change the Workflow?

Sending a coverage report to `Codecov` with GitHub Action without a token encounters a **rate limit** issue. For more details, please refer to the comments on issue #1085.

In the current flow, the `Codecov` workflow runs whenever we push code to PR. To address the rate limit issue, we have decided to run the workflow only when we are **ready for review**.

### When PR is **draft**
Screenshot 2023-12-05 at 2 27 51 PM

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -1,0 +1,34 @@
+name: code coverage
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - ready_for_review
+
+jobs:
+  run-codecov:
+    if: github.event.pull_request.draft == false
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: "21"
+
+      - name: Install dependencies
+        run: npm install jest
+
+      - name: Run Jest tests
+        run: npm run test
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          files: ./coverage/coverage-final.json
+          flags: unit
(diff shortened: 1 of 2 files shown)


jiji14:
@JGreenlee 
Jest runs on both the `osx-serve-install` and `codecov` workflows. I think it's enough to execute Jest when the PR is ready, specifically in the code coverage workflow. What are your thoughts? If you agree, I will delete `Jest` execution in `osx-serve-install`

JGreenlee:
@jiji14 I think as long as the expectation is for contributors to run the tests locally ***before*** they mark a PR as ready, this makes sense.

JGreenlee:
If a contributor forgets to run tests locally, they would mark it ready for review causing the workflow to run, and checks might fail.

In this scenario, do they have a way to re-trigger the workflow after making fixes? Can they mark it ready for review again?
If so, I think it's fine to remove Jest from the `osx-serve-install` workflow.

jiji14:
@JGreenlee 

I've just verified that Codecov is executed whenever the PR status transitions from `Draft` to `Ready to Review` in this workflow. 

In your scenario,
1. If a contributor forgets to run tests and the initial attempt fails,
2. They can switch the PR to draft status to address the issues.
3. When the necessary fixes are made, they can mark the PR as `Ready to Review`, triggering Codecov.

Does this clarify the process?

JGreenlee:
> Does this clarify the process?

Yes it does!

> If so, I think it's fine to remove Jest from the `osx-serve-install` workflow.

Then I think we can remove Jest from the `osx-serve-install` workflow to avoid running the tests twice.

JGreenlee (approved):
LGTM!

shankari (approved):

PR #1115, opened 2023-12-08
🐛 Fix AppConfig `survey_info` backwards compat + add tests for dynamicConfig

JGreenlee:
Fixes https://github.com/e-mission/e-mission-docs/issues/1025

I reproduced the issue in the devapp by logging into `dev-emulator-study`, clearing localStorage and user cache, then manually storing into the usercache an app config that was missing `survey_info`.

Here is what that caused:

<hr>

Immediately after applying 8c31b655b6e867b61c5bd621a6249a95295aa560 and waiting for autoreload, it was fixed:

### DIFF
--- a/www/__tests__/footprintHelper.test.ts
+++ b/www/__tests__/footprintHelper.test.ts
@@ -9,8 +9,9 @@ import { getConfig } from '../js/config/dynamicConfig';
 import { mockBEMUserCache } from '../__mocks__/cordovaMocks';
 import { mockLogger } from '../__mocks__/globalMocks';
 import fakeLabels from '../__mocks__/fakeLabels.json';
+import fakeConfig from '../__mocks__/fakeConfig.json';
 
-mockBEMUserCache();
+mockBEMUserCache(fakeConfig);
 mockLogger();
 
 global.fetch = (url: string) =>
--- a/www/js/config/dynamicConfig.ts
+++ b/www/js/config/dynamicConfig.ts
@@ -167,8 +167,9 @@ async function fetchConfig(studyLabel: string, alreadyTriedLocal?: boolean) {
  */
 function extractStudyName(token: string): string {
   const tokenParts = token.split('_');
-  if (tokenParts.length < 3) {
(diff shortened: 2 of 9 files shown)


JGreenlee:
I didn't do anything for tests in this PR because it does not have any breaking changes for tests. But as I've realized, we even don't have tests yet for `dynamicConfig.ts` (it was written before we had Jest in place).

So, I'm marking this as a draft and will add tests before I mark it as ready again.

www/__tests__/footprintHelper.test.ts
```diff
@@ -9,8 +9,9 @@ import { getConfig } from '../js/config/dynamicConfig';
 import { mockBEMUserCache } from '../__mocks__/cordovaMocks';
 import { mockLogger } from '../__mocks__/globalMocks';
 import fakeLabels from '../__mocks__/fakeLabels.json';
+import fakeConfig from '../__mocks__/fakeConfig.json';
 
-mockBEMUserCache();
+mockBEMUserCache(fakeConfig);
```
shankari, line 14: This is not your change, but I would like to understand this better. Why do we pass in the config instead of calling `putRWDocument`

www/js/config/dynamicConfig.ts
```diff
@@ -224,47 +259,54 @@ function loadNewConfig(newToken, existingVersion = null) {
...
     }
     logDebug('No config found in KVStore, fetching from native storage');
     return window['cordova'].plugins.BEMUserCache.getDocument(CONFIG_PHONE_UI, false).then(
       (config) => {
         if (config && Object.keys(config).length) {
           logDebug('Got config from native storage: ' + JSON.stringify(config));
-          storedConfig = config;
-          return config;
+          storedConfig = _backwardsCompatFill(config);
+          return storedConfig;
```
shankari, line 309: I know we didn't do this before, but a future enhancement might be to also save the config after backwards filling it. That would ensure that we could eventually remove the backwards compat fill (or turn it into a NOP) instead of having it grow over time.

shankari (approved):

PR #1116, opened 2023-12-08
📟 🛠️ Update `setup_ios_native.sh` to support Apple Silicone

the-bay-kay:
[Homebrew](https://docs.brew.sh/Installation) installs packages differently depending on chip architecture.  As such, the native-build setup script needed a quick update to support M1-chip machines!

I have yet to try this on an Intel machine (I don't have another Mac on hand), so any help testing the adjusted setup script would be greatly appreciated!

### DIFF
--- a/setup/setup_ios_native.sh
+++ b/setup/setup_ios_native.sh
@@ -7,14 +7,32 @@ source setup/setup_shared.sh
 OSX_MAJOR_VERSION=`sw_vers | grep ProductVersion | cut -d ':' -f 2 | cut -d '.' -f 1`
 echo "Found OSX major version" $OSX_MAJOR_VERSION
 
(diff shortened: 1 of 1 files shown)


shankari:
@the-bay-kay we build as part of the CI (`osx-build-ios`). You can check to see which chip it is running on. It might be worthwhile looking at whether we can check both M1 and Intel as part of the CI by configuring it accordingly.

shankari:
checked it myself, found

```
Found x86_64 chip
```

So merging.

@the-bay-kay can you please submit a new PR that will run this on M1 chips as well?
https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

PR #1117, opened 2023-12-13
Use specific prettier version

jiji14:
### Background
`Prettier` failed on the [UI Migration Wrap-up PR](https://github.com/e-mission/e-mission-phone/pull/1106), and we discovered that it uses the latest Prettier version on GitHub Actions but version 3.0.3 in the project. Therefore, we have decided to stick with the specific version (3.0.3) to prevent this issue from occurring again.

<br />
<br />

Screenshot 2023-12-12 at 8 05 29 PM

### DIFF
--- a/.github/workflows/prettier.yml
+++ b/.github/workflows/prettier.yml
@@ -7,5 +7,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - run: npx prettier --check www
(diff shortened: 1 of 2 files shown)


JGreenlee (approved):
LGTM!

If we ever decide to update the Prettier version, we will change it both here and in `package.serve.json`.

shankari (approved):
LGTM!

PR #1120, opened 2023-12-29
Custom Labels Step 2

jiji14:
### Related Issue
- System Design: [GitHub Issue 1012](https://github.com/e-mission/e-mission-docs/issues/1012)
- Server: [GitHub Server Pull Request 946](https://github.com/e-mission/e-mission-server/pull/946)
- #1110 

### What I added
1. Manage Custom Labels on `Profile` Tab.
2. Add `Insert` / `Delete` Functions

### Result

<br />

#### Please check out to `customize-modes` branch in the `Server` repository for testing.


jiji14:
> @jiji14 Can you change target branch to `master`?

Done!

JGreenlee:
I noticed a few issues while testing this.

I don't know if they were present before; they could be new issues from merging in changes from upstream. I also merged in Step 1 here so we can be sure Step 1 and Step 2 are working together.

Either way, @jiji14 could you take a look?

### Custom Labels from label screen don't appear on profile screen 

Go to the Profile screen > Manage Custom Labels and see the list of current labels (if any). Then go to the Label screen and add a custom label (via "Other..."). Then go back to the Profile screen > Manage Custom Labels, the new label does not appear immediately. After reloading, the new label does appear there

### Custom labels don't appear on label screen until you interact with the "Other" input

On the label screen, I can go to "Other" and add a new custom label. Then if I go to label a new trip, the custom label shows (as expected). But if I kill the app and re-open the app, they appear to have disappeared. When I add a new custom label, I see the old ones again.

jiji14:
@JGreenlee 

The error occurred after merging. I fixed the bugs and also took care of the code coverage issue

jest.config.js
```diff
@@ -18,4 +18,5 @@ module.exports = {
   moduleDirectories: ["node_modules", "src"],
   globals: {"__DEV__": false},
   collectCoverage: true,
+  coveragePathIgnorePatterns: ['www/js/services/commHelper.ts']
```
JGreenlee, line 21: Great, this seems to work!!

Can we add a note here for explanation?

```suggestion
  // several functions in commHelper do not have unit tests; see note in commHelper.test.ts
  coveragePathIgnorePatterns: ['www/js/services/commHelper.ts'],
```

www/js/control/ProfileSettings.tsx
```diff
@@ -434,6 +435,7 @@ const ProfileSettings = () => {
           desc={authSettings.opcode}
           descStyle={settingStyles.monoDesc}></SettingRow>
         <DemographicsSettingRow></DemographicsSettingRow>
+        <CustomLabelSettingRow />
```
JGreenlee, line 438: This is a good place to put the `<CustomLabelSettingRow>`, but when testing I noticed that `<CustomLabelSettingRow>` shows up in all configurations, even if there are no multilabel choices. Some programs don't have "mode" and "purpose" labels - they just have surveys. If that's the case, custom labels don't apply.

We should show this row conditionally, like below (or a different way if you prefer)
```suggestion
        {appConfig?.survey_info?.['trip-labels'] == 'MULTILABEL' && <CustomLabelSettingRow />}
```

www/js/App.tsx
```diff
@@ -42,6 +43,16 @@ const defaultRoutes = (t) => [
...
+const CUSTOM_LABEL_KEYS_IN_DATABASE = ['mode', 'purpose'];
+type CustomLabelMap = {
+  [k: string]: string[];
+};
+
+const scenes = {
+  label: withErrorBoundary(LabelTab),
+  metrics: withErrorBoundary(MetricsTab),
+  control: withErrorBoundary(ProfileSettings),
+};
```
JGreenlee, line 55:
```suggestion
```

Duplicated code that can be removed.
I think this was also my mistake from bad merge conflict resolution

JGreenlee (changes requested):
I tested this again and it works really well! The bugs from merging in were probably my mistake (bad merge conflict resolution).

Below are 3 quick suggestions I have, once they are applied I will pass this along!

jiji14:
@JGreenlee 
Thanks for all the suggestions. It looks great so I updated the codes :) 

@shankari 
Please note that server code needs to merge first before this PR.

shankari:
@jiji14 or @JGreenlee can you resolve the merge conflict? I will likely target another staging release for this week so we can be testing next week.

JGreenlee:
resolved

www/js/App.tsx
```diff
@@ -13,16 +13,22 @@ import usePermissionStatus from './usePermissionStatus';
 import { initPushNotify } from './splash/pushNotifySettings';
 import { initStoreDeviceSettings } from './splash/storeDeviceSettings';
 import { initRemoteNotifyHandler } from './splash/remoteNotifyHandler';
+import { getUserCustomLabels } from './services/commHelper';
 import { initCustomDatasetHelper } from './metrics/customMetricsHelper';
 import AlertBar from './components/AlertBar';
 import Main from './Main';
 
 export const AppContext = createContext<any>({});
+const CUSTOM_LABEL_KEYS_IN_DATABASE = ['mode', 'purpose'];
```
shankari, line 22: @JGreenlee don't we need this for replaced mode as well? Or is the expectation that the replaced mode will also share the custom labels from the regular mode?

www/js/survey/multilabel/MultiLabelButtonGroup.tsx
```diff
@@ -164,16 +188,47 @@ const MultilabelButtonGroup = ({ trip, buttonsInline = false }) => {
...
+                      show the custom labels section before 'other' */
+                      if (o.value == 'other' && customLabelMap[customLabelKeyInDatabase]?.length) {
+                        return (
+                          <>
+                            <Divider style={{ marginVertical: 10 }} />
+                            <Text
+                              style={{ fontSize: 12, color: colors.onSurface, paddingVertical: 4 }}>
+                              {(modalVisibleFor === 'MODE' ||
+                                modalVisibleFor === 'REPLACED_MODE') &&
+                                t('trip-confirm.custom-mode')}
```
shankari, line 213: Ah I see that we do in fact use the same values for both `MODE` and `REPLACED_MODE`

shankari (approved):
I'm going to go ahead and merge this since it is correct, but I think it might need enhancements as part of the CO2 calculations. Right now, when the user adds a new custom mode, we don't store any CO2 information as part of it. It seems like we should, but it is not clear that the end user knows the energy/emissions intensity of the mode that they add.

I can put this up on staging/nrel-commute for now, but we should really decide how we are going to handle this before we move to production in general, since otherwise we have to deal with data migration.

PR #1123, opened 2024-01-22
❌🙁 Fix codecov-test failure - ‼️📅 Rewrite - notifScheduler.ts

sebastianbarry:


sebastianbarry:
## Why was it failing?

The test was working correctly, but the problem was that the logDebug statement was not being ran.

## What did I do to fix it?

I corrected this by making the `expect` statement actually expect the output rather than the logDebug statement arguments.
However, when I did this I encountered a second error. I created an `expectedScheduledNotifs` function that is an array containing objects with keys `{ key: "November 19, 2023" val: "9:00 PM" }`. This is what the expected values look like, but the resulted values contain a strange unicode character U+202F in `"9:00 PM"`which fails the equivalence test. I am not sure how to go about testing these values, since my local computer fails the test if I try to expect these values. My solution was to only expect the `key` value for them. This is only a temporary soluiton, so I am open to suggestions, but it will pass the codecov test for now.

**Do I need to do anything to squash merge my PR with this PR?**

sebastianbarry:
@shankari added to Ready for Review

www/js/splash/notifScheduler.ts
```diff
@@ -79,14 +79,20 @@ export const getScheduledNotifs = function (isScheduling: boolean, scheduledProm
         anywhere from 0-n of the scheduled notifs are displayed 
         if actively scheduling, wait for the scheduledPromise to resolve before fetching prevents such errors
         */
+    console.log('test log: isScheduling during getScheduledNotifs', isScheduling);
+    console.log('test log: scheduledPromise during getScheduledNotifs', scheduledPromise);
```
shankari, line 83: Do you need these `console.log` statements?

www/__tests__/notifScheduler.test.ts
```diff
@@ -267,13 +274,11 @@ describe('updateScheduledNotifs', () => {
...
-    expect(logDebug).toHaveBeenCalledWith('After cancelling, there are no scheduled notifications');
-    expect(logDebug).toHaveBeenCalledWith(
-      'After scheduling, there are 4 scheduled notifications at 21:00 first is November 19, 2023 at 9:00 PM',
-    );
-    expect(setIsScheduling).toHaveBeenCalledWith(false);
     expect(scheduledNotifs).toHaveLength(4);
+    expect(scheduledNotifs[0].key).toEqual(expectedResultcheduleNotifs[0].key);
+    expect(scheduledNotifs[1].key).toEqual(expectedResultcheduleNotifs[1].key);
+    expect(scheduledNotifs[2].key).toEqual(expectedResultcheduleNotifs[2].key);
+    expect(scheduledNotifs[3].key).toEqual(expectedResultcheduleNotifs[3].key);
```
shankari, line 281: This change removes the `setIsScheduling` checks. Why are they no longer required?

shankari (changes requested):
I am going to merge this to unblock @the-bay-kay but I have several questions about this.

> The test was working correctly, but the problem was that the logDebug statement was not being ran.

1. Why was the `logDebug` statement not being run?
   - Note also my cleanup comment for the original PR was that we shouldn't rely on `logDebug` to verify values but should add in introspection methods and check the values correctly. Please make sure to implement all those "future" changes.
2.  U+202F is a non-breaking space https://www.compart.com/en/unicode/U+202F It has nothing to do with the quotes. I'm guessing you copy-pasted something from a website, and introduced the NNBSP into the text by default. You will need to delete and type this out to remove it, or view the text in non-unicode to see the character that you remove.

Finally, I am very skeptical that you were not able to reproduce the error(s) while running locally after pulling from the `service_rewrite_2023` branch, since multiple other PRs that were based on the branch had failing tests.

I would encourage you to re-pull after this and try to clean up the tests, both in this PR and in the original one.

shankari:
Squash merging to avoid commit churn
@the-bay-kay please pull and then resolve conflicts and address my comments so I can merge your PR as well.

PR #1125, opened 2024-02-02
🚑 Fix Processed Trip Survey Responses Not Matching

JGreenlee:
Processed responses for user input surveys were not being matched because they were being looked up under "manual/trip_user_input", while it should have just been "trip_user_input". This is because we were only invoking a function to trim off "manual/" for MULTILABEL user input matching, and it was neglected for ENKETO.

Revise the functions in confirmHelper (keeping 2 versions, `removeManualPrefix` which just trims the key, and `inputType2retKey` which looks up the key and then trims), and use `removeManualPrefix` in inputMatcher appropriately

Some day we can unify the data model (https://github.com/e-mission/e-mission-docs/issues/1045) and not keep 2 versions of so many things


JGreenlee:
Added some basic tests for `mapInputsToTimelineEntries`.
Having coverage for that function should prevent another issue like this from popping up again.

JGreenlee:
@shankari for visibility

PR #1129, opened 2024-02-12
📝 Conditional Surveys depending on trip characteristics

JGreenlee:
We plan to support a conditional approach to surveys: depending on characteristics of the trip / timelineEntry, it may prompt for one survey or another. This involves rearranging UserInputButton so that the survey name is not hardcoded as "TripConfirmSurvey". Instead, survey names and button labels will be pulled from the config. Each survey that could show conditionally has a `showsIf` field, which is evaluated at runtime in conditionalSurveys.ts. The first survey whose condition matches is the one that is used for that timelineEntry.

This is a draft, and I don't expect it to work as-is, but it should serve as an outline to the planned approach

### DIFF
--- a/www/js/survey/enketo/UserInputButton.tsx
+++ b/www/js/survey/enketo/UserInputButton.tsx
@@ -15,17 +15,33 @@ import { useTheme } from 'react-native-paper';
 import { displayErrorMsg, logDebug } from '../../plugin/logger';
 import EnketoModal from './EnketoModal';
 import LabelTabContext from '../../diary/LabelTabContext';
+import useAppConfig from '../../useAppConfig';
+import { getSurveyForTimelineEntry } from './conditionalSurveys';
 
 type Props = {
   timelineEntry: any;
 };
 const UserInputButton = ({ timelineEntry }: Props) => {
   const { colors } = useTheme();
+  const appConfig = useAppConfig();
   const { t, i18n } = useTranslation();
 
   const [prevSurveyResponse, setPrevSurveyResponse] = useState<string | undefined>(undefined);
   const [modalVisible, setModalVisible] = useState(false);
   const { userInputFor, addUserInputToEntry } = useContext(LabelTabContext);
+
+  // which survey will this button launch?
+  const [surveyName, notFilledInLabel] = useMemo(() => {
+    const tripLabelConfig = appConfig?.survey_info?.buttons?.['trip-label'];
+    if (!tripLabelConfig) {
+      // config doesn't specify; use default
+      return ['TripConfirmSurvey', t('diary.choose-survey')];
+    }
+    // config lists one or more surveys; find which one to use
+    const s = getSurveyForTimelineEntry(tripLabelConfig, timelineEntry);
+    const lang = i18n.resolvedLanguage || 'en';
+    return [s?.surveyName, s?.['not-filled-in-label'][lang]];
+  }, [appConfig, timelineEntry, i18n.resolvedLanguage]);
 
   // the label resolved from the survey response, or null if there is no response yet
   const responseLabel = useMemo<string | undefined>(
@@ -52,23 +68,21 @@ const UserInputButton = ({ timelineEntry }: Props) => {
     }
   }
 
+  if (!surveyName) return <></>; // no survey to launch
   return (
     <>
       <DiaryButton
+        // if a response has been been recorded, the button is 'filled in'
         fillColor={responseLabel && colors.primary}
         onPress={() => launchUserInputSurvey()}>
-        {/* if no response yet, show the default label */}
-        {responseLabel || t('diary.choose-survey')}
+        {responseLabel || notFilledInLabel}
       </DiaryButton>
 
       <EnketoModal
         visible={modalVisible}
         onDismiss={() => setModalVisible(false)}
         onResponseSaved={onResponseSaved}
-        surveyName={'TripConfirmSurvey'} /* As of now, the survey name is hardcoded.
-                                        In the future, if we ever implement something like
-                                        a "Place Details" survey, we may want to make this
-                                        configurable. */
+        surveyName={surveyName}
         opts={{ timelineEntry, prefilledSurveyResponse: prevSurveyResponse }}
       />
     </>
--- a/www/js/survey/enketo/conditionalSurveys.ts
+++ b/www/js/survey/enketo/conditionalSurveys.ts
@@ -32,7 +32,7 @@ export function getSurveyForTimelineEntry(
   for (let surveyConfig of tripLabelConfig) {
     if (!surveyConfig.showsIf) return surveyConfig; // survey shows unconditionally
     const scope = {
-      tlEntry,
+      ...tlEntry,
       ...conditionalSurveyFunctions,
     };
     try {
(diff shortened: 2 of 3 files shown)


JGreenlee:
After discussing in https://github.com/e-mission/nrel-openpath-deploy-configs/pull/69, I am proposing that the config structure will look like...

```json
...
  "survey_info": {
    "surveys": {
      "NonEVTripSurvey": {
        "formPath": "<URL>/gpg-non-ev-trip-v1.xml",
        "version": 1,
        "compatibleWith": 1,
        "dataKey": "manual/trip_user_input",
        "labelTemplate": { "en": "Answered" }
      },
      "EVRoamingTripSurvey": {
        "formPath": "<URL>/gpg-ev-roaming-trip-v1.xml",
        "version": 1,
        "compatibleWith": 1,
        "dataKey": "manual/trip_user_input",
        "labelTemplate": { "en": "Answered" }
      },
      "EVReturnTripSurvey": {
        "formPath": "<URL>/gpg-ev-return-trip-v1.xml",
        "version": 1,
        "compatibleWith": 1,
        "dataKey": "manual/trip_user_input",
        "labelTemplate": { "en": "Answered" }
      },
      "UserProfileSurvey": {
        "formPath": "<URL>/gpg-onboarding-v1.xml",
        "version": 1,
        "compatibleWith": 1,
        "dataKey": "manual/demographic_survey",
        "labelTemplate": { "en": "Answered" }
      }
    },
    "buttons": {
      "trip-label": [
        {
          "surveyName": "NonEVTripSurvey",
          "not-filled-in-label": { "en": "Gas Car Survey" },
          "showsIf": "!bt_sensed_mode.startsWith('e_car')"
        },
        {
          "surveyName": "EVRoamingTripSurvey",
          "not-filled-in-label": { "en": "EV Survey" },
          "showsIf": "bt_sensed_mode.startsWith('e_car') && !pointIsWithinBounds(end_loc.coordinates, [[39.719, -105.118], [39.717, -105.115]])"
        },
        {
          "surveyName": "EVReturnTripSurvey",
          "not-filled-in-label": { "en": "EV Survey" },
          "showsIf": "bt_sensed_mode.startsWith('e_car') && pointIsWithinBounds(end_loc.coordinates, [[39.719, -105.118], [39.717, -105.115]])"
        }
      ]
    },
    "trip-labels": "ENKETO"
  },
...
```

...for a program that has a conditional trip survey: either NonEVTripSurvey, EVRoamingTripSurvey, or EVReturnTripSurvey depending on the Bluetooth sensed mode and end location of the trip

JGreenlee:
@jiji14 for visibility

JGreenlee:
<table>
<td>
trips >30 min have gas car survey

image
</td><td>
homebound trips have return survey

image
</td><td>
non-home trips have roaming survey

image
</td></table>

JGreenlee:
@jiji14 Can you review this before I pass it on to for Shankari to review? Let me know if you have suggestions or questions

jiji14:
@JGreenlee Thank you for the clear organization! Your documentation and comments are very helpful, and I can understand everything well. I just have a few small questions. I was wondering if there is no need to handle the edge case where `pt` (end_loc.coordinates) is a `null` value. 
```
  pointIsWithinBounds: (pt: Position, bounds: Position[]) => {
  },
```
Additionally, I'm curious about the inspiration behind the `scopedEval` function; it seems quite clever!

www/js/survey/enketo/conditionalSurveys.ts
```diff
@@ -0,0 +1,54 @@
...
+    };
+    try {
+      const evalResult = scopedEval(surveyConfig.showsIf, scope);
+      if (evalResult) return surveyConfig;
+    } catch (e) {
+      displayError(e, `Error evaluating survey condition "${surveyConfig.showsIf}"`);
+    }
+  }
+  // TODO if none of the surveys passed conditions?? should we return null, throw error, or return a default?
+  return null;
```
jiji14, line 53: In my opinion, we should display a default survey in this case.

JGreenlee:
> I was wondering if there is no need to handle the edge case where `pt` (end_loc.coordinates) is a `null` value.

Good thinking!

I did include a try/catch block around the call to `scopedEval`. So if `end_loc` or `end_loc.coordinates` were undefined/null/ /empty, the error will be caught and `displayError` will be called (which triggers a user-facing popup).

But I do wonder if it would be better to explicitly throw the error from the `pointIsWithinBounds` function.
Need input from shankari on this.

> Additionally, I'm curious about the inspiration behind the scopedEval function; it seems quite clever!

Here's the discussion between shankari and me: https://github.com/e-mission/nrel-openpath-deploy-configs/pull/69#issuecomment-1935259205
This was before I made the code changes, but was in the stage of planning

www/js/survey/enketo/conditionalSurveys.ts
JGreenlee: I was initially thinking that, but now I am thinking the current implementation (which returns `null` and shows no survey) is best. It could be useful to not show a survey for certain trips.

For example, what if a program wants to **only** prompt surveys for EV trips and not ask anything about gas car trips?

jiji14 (approved):
Thanks for your comment! Approved it :D

www/js/survey/enketo/conditionalSurveys.ts
jiji14: My thought was that we need to show at least one survey, regardless of the circumstances. However, I agree that returning `null` makes sense if we don't want to show any survey other than a specific one.

www/js/survey/enketo/UserInputButton.tsx
```diff
@@ -15,18 +15,34 @@ import { useTheme } from 'react-native-paper';
...
   const [modalVisible, setModalVisible] = useState(false);
   const { userInputFor, addUserInputToEntry } = useContext(LabelTabContext);
 
+  // which survey will this button launch?
+  const [surveyName, notFilledInLabel] = useMemo(() => {
+    const tripLabelConfig = appConfig?.survey_info?.buttons?.['trip-label'];
+    if (!tripLabelConfig) {
+      // config doesn't specify; use default
+      return ['TripConfirmSurvey', t('diary.choose-survey')];
+    }
```
shankari, line 39: Would it ever make sense to never have a trip-level survey? Similar to the argument for the `null` match, maybe we want to only show place surveys and not trip surveys? I see that this is what we do now, which is why the time use displays the `TripConfirmSurvey` although there is no button configured for it, but we might want to revisit that later

shankari (commented):
This is still a bit experimental, so I am merging this. I fully anticipate that we will fine-tune it later as we experiment on staging.

Also, for the record, we don't have the primary mode stored in the confirmed or composite trips. As I said, we use it in the public dashboard, but we determine it as part of the scaffolding code.
https://github.com/e-mission/em-public-dashboard/blob/434dec58186bad70877df3fd43b599738807a43b/viz_scripts/scaffolding.py#L205

```
        expanded_ct["primary_mode_non_other"] = participant_ct_df.cleaned_section_summary.apply(lambda md: max(md["distance"], key=md["distance"].get))
        expanded_ct.primary_mode_non_other.replace({"ON_FOOT": "WALKING"}, inplace=True)
        valid_sensed_modes = ["WALKING", "BICYCLING", "IN_VEHICLE", "AIR_OR_HSR", "UNKNOWN"]
        expanded_ct["primary_mode"] = expanded_ct.primary_mode_non_other.apply(lambda pm: "OTHER" if pm not in valid_sensed_modes else pm)
```

PR #1131, opened 2024-02-21
🔞 Update "short" demo survey so people born in 2006 can take it

JGreenlee:
The short version of the demographics survey asks for year of birth and we restrict the valid range of responses so as to not allow participants younger than 18. Last year when I uploaded this, I wrote in 2005 as the upper limit. But theoretically, people born in early 2006 can now participate in those studies.

I found a more generic solution using the ODK forms constraint syntax and some built-in functions they have:
`. > 1900 and . + 18 <= format-date(today(), '%Y')`

So now we won't have to update this every year; it actually takes into account the current year when validating the form.


JGreenlee:
<table>
<tr>
<td>

### Valid
</td>
<td>

### Not valid
</td>
</tr>

<tr><td>

image

</td><td>

image

</td></tr><tr><td>

image

</td><td>

image

</td></tr></table>

PR #1133, opened 2024-02-29
Moved raw_data_use backwards compat hack

sebastianbarry:
Moved backwards compat hack to fill in the raw_data_use for programs that don't have it to PrivacyPolicy.tsx, from ProfileSettings.tsx so that the privacy policy is updated for the login screen, and it carries over to the Profile settings screen.

| @Abby-Wheelis pointed out with the `washingtoncommons` program that their privacy policy had a textual error: | And @Abby-Wheelis phone on `nrel-commute` had this textual error: |
|---|---|
| MicrosoftTeams-image | MicrosoftTeams-image (1) |

They were both a result of a config not having the `raw_data_use` field filled in.

We had a backwards compat hack for programs without `raw_data_use`, but it was only in `ProfileSettings.tsx`, so now I removed it from there and placed it in PrivacyPolicy.tsx so that it checks for `raw_data_use` and gets filled in anytime the PrivacyPolicy is rendered.

Also, to fix the double "to to," I just removed the "to" and "para [ES]" from the backwards compat hack.

### DIFF
--- a/www/js/control/ProfileSettings.tsx
+++ b/www/js/control/ProfileSettings.tsx
@@ -131,15 +131,6 @@ const ProfileSettings = () => {
   function whenReady(newAppConfig: AppConfig) {
     const tempUiConfig = newAppConfig;
 
-    // backwards compat hack to fill in the raw_data_use for programs that don't have it
-    const default_raw_data_use = {
-      en: `to monitor the ${tempUiConfig.intro.program_or_study}, send personalized surveys or provide recommendations to participants`,
-      es: `para monitorear el ${tempUiConfig.intro.program_or_study}, enviar encuestas personalizadas o proporcionar recomendaciones a los participantes`,
-    };
-    Object.entries(tempUiConfig.intro.translated_text).forEach(([lang, val]) => {
-      val.raw_data_use = val.raw_data_use || default_raw_data_use[lang];
-    });
-
     // Backwards compat hack to fill in the `app_required` based on the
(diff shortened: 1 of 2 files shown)


sebastianbarry:
Testing:

| Simulator Screenshot - iPhone 13 - 2024-02-29 at 10 10 56 | Simulator Screenshot - iPhone 13 - 2024-02-29 at 10 45 06 |
|---|---|
| | |

sebastianbarry:
Also, the template survey for creating new configs does not have a `raw_data_use` field, should I add this?

Abby-Wheelis:
>pointed out with the washingtoncommons program that their privacy policy had a textual error:

Those screenshots are actually both from my phone on the `nrel-commute` configuration, but it was `washingtoncommons` that brought it to my attention because they wanted a few other changes to the intro and noted that this section of the policy was weird!

It also looks like in testing that the privacy policy in the intro and in the developer zone still have different text, with the default reason only being filled in the developer zone, is it possible to have it match in both places?

sebastianbarry:
@Abby-Wheelis My bad, I posted the wrong screenshot. I edited the comment to reflect the proper testing

shankari:
> Also, the template survey for creating new configs does not have a raw_data_use field, should I add this?

Yes, please do.

Also, I don't want to hold this up given that it is fixing something that is currently broken, but can we add some tests for the privacy policy formatting?

shankari (approved):

PR #1135, opened 2024-03-08
🛜 ᛒ  Adding BLE Beacon Scanner

the-bay-kay:
The intent of this PR is to add a BLE Beacon Scanner to the developer settings, building off of the work done in the previous dev-scanner PRs.

For further context, please see:
  - PR [https://github.com/e-mission/e-mission-phone/pull/1128/](1128)
  - Issue [1046](https://github.com/e-mission/e-mission-docs/issues/1046)


the-bay-kay:
## BLE Plugin Notes
Louis and I have done some reading into the BLE plugins available for Cordova, and have found that the [cordova-plugin-ibeacon](https://github.com/petermetz/cordova-plugin-ibeacon) plugin seems to be the best fit.  While we're looking at incorperating this into our preexisting scanner code, it seems to me that there are some subtle differences in how BLE is handled.  Below are some of my stream-of-thought notes on the plugin and this project!

- As Shankari, Louis and I have discussed, we will most likely be utilizing [CLLocationManager](https://developer.apple.com/documentation/corelocation/cllocationmanager?language=objc) for iOS.  Since this plugin utilizes this, it should be a great example for our future changes!
- Unlike our previous BluetoothClassicSerial plugin, or even the [cordova-plugin-ble-central](https://github.com/don/cordova-plugin-ble-central?tab=readme-ov-file), there is no single "scan / subscribe" function.  Rather, when monitoring beacons, it seems the workflow is:
  1. Store information surrounding the beacon's ID, name, etc.  It seems that this, in the plugin, is referred to as a "region"
  2. Enable Beacon logging via a "DOM" delegate
     -  From what I understand, the plugin itself acts as its own DOM
  3. Give the delegate a "region" to monitors
      -  Then, when broadcasts are detected, they will be logged via the delegate
  4.  Finally, you can "unsubscribe" from broadcasts by ending your monitoring for a specific region.
      - What does it look like to monitor multiple regions at once? That's something we'll have to consider down the road 
- As with the Bluetooth Classic Scanner, we will need to make some changes to the iOS app permissions.

## Going Forward
With all of this in mind,  Louis and I have come up with a rough game-plan for this project!
1.  Add the plugin to `e-mission-phone`, resolve any config conflicts within `em-data-collection`
2. Refactor the Bluetooth Scanner UI to better incorporate the new plugin
3. Implement our own "start scanning" and "stop scanning" functions
4. Incorporate the DOM logs into our scanner's `setLogs()`
5. Add our test beacons data into an external data-file (a `.json` file, a `.csv`, etc.)
    - After these tests are done, we will need to modify this to support whatever vehicle data is mapped to a given beacon!
    - For Louis and I's test beacons, the setup guide can be found [here](https://bluecharmbeacons.com/bc011-ibeacon-multibeacon-quick-start-guide/).  For the beacons that will be used in deployment, the configuration process will look slightly different.
7. Confirm the app recognizes the beacons
8. Begin running tests, which may include
    - Confirm the beacon appears / disappears as we move in and out of range
    - Finding the optimal output Db for the beacon (e.g., what range is detectable inside a car, but not outside? Looking to minimize noise
    - Confirm BLE is visible during transit (Please don't scan and drive!)

As we go forward with this plan, I'll keep this thread updates with our challenges / notes on the experience!

the-bay-kay:
It seems that only iOS builds after adding the plugin... For legibility / book keeping, these are the build errors that occurred:
```
> Configure project :app
Adding classpath: com.google.gms:google-services:4.3.3
Warning: The 'kotlin-android-extensions' Gradle plugin is deprecated. Please use this migration guide (https://goo.gle/kotlin-android-extensions-deprecation) to start working with View Binding (https://developer.android.com/topic/libraries/view-binding) and the 'kotlin-parcelize' plugin.

FAILURE: Build completed with 2 failures.


1: Task failed with an exception.
Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
-----------

* Where:
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
Script '/Users/runner/work/e-mission-phone/e-mission-phone/platforms/android/com.unarin.cordova.beacon/emission-cordova-plugin-ibeacon.gradle' line: 6


See https://docs.gradle.org/7.6/userguide/command_line_interface.html#sec:command_line_warnings
* What went wrong:
A problem occurred evaluating script.
> Could not find method compile() for arguments [org.altbeacon:android-beacon-library:2.16.1] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================

2: Task failed with an exception.
-----------
* What went wrong:
A problem occurred configuring project ':app'.
> compileSdkVersion is not specified. Please add it to build.gradle

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.
==============================================================================

* Get more help at https://help.gradle.org/

BUILD FAILED in 1m 2s
Command failed with exit code 1: /Users/runner/work/e-mission-phone/e-mission-phone/platforms/android/gradlew -b /Users/runner/work/e-mission-phone/e-mission-phone/platforms/android/build.gradle cdvBuildDebug
Error: Process completed with exit code 1.
```

louisg1337:
Touching on the Android build errors that Katie mentioned above, we managed to get those resolved. To fix it we had to fork our own version of this plugin and make some changes, which are shown in [this commit](https://github.com/petermetz/cordova-plugin-ibeacon/compare/v3.x...louisg1337:cordova-plugin-ibeacon:v3.x). The two problems in particular were...

```
Could not find method compile() for arguments [org.altbeacon:android-beacon-library:2.16.1] on object of 
type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

It seems like this plugin is a bit older, so it used a deprecated gradle method. To fix this we changed `compile` to `implementation `.

```
> Task :app:processDebugMainManifest FAILED
~/AndroidManifest.xml:21:9-27:20 Error: android:exported needs to be explicitly specified for 
element <receiver#org.altbeacon.beacon.startup.StartupBroadcastReceiver>. 
Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when
the corresponding component has an intent filter defined. 
```

This plugin defines a dependency in its gradle build file called `org.altbeacon:android-beacon-library:2.16.1`, and I'm assuming that it dynamically adds a receiver to the manifest from build, which causes us this error. To fix this, we instead define the receiver straight into the `plugin.xml` of `cordova-plugin-ibeacon` and give it the `android:exported` parameter.

Android now builds, but we have yet to test the beacon at all so we will see if it holds up!

louisg1337:
I was able to build for Android successfully with the plugin installed. I am now running into an issue, however, when trying to test the plugin. I am following [this](https://github.com/petermetz/cordova-plugin-ibeacon?tab=readme-ov-file#start-monitoring-a-single-ibeacon) part of the demo code the plugin provides, but whenever I try to invoke this code, I get this error in logcat.

```
--------- beginning of crash
03-09 17:55:03.613 19003 19003 E AndroidRuntime: FATAL EXCEPTION: main
03-09 17:55:03.613 19003 19003 E AndroidRuntime: Process: edu.berkeley.eecs.emission, PID: 19003
03-09 17:55:03.613 19003 19003 E AndroidRuntime: java.lang.RuntimeException: java.lang.IllegalArgumentException: edu.berkeley.eecs.emission: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
03-09 17:55:03.613 19003 19003 E AndroidRuntime: Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.job.JobServiceEngine$JobHandler.handleMessage(JobServiceEngine.java:109)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.os.Handler.dispatchMessage(Handler.java:106)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.os.Looper.loopOnce(Looper.java:201)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.os.Looper.loop(Looper.java:288)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.ActivityThread.main(ActivityThread.java:7839)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at java.lang.reflect.Method.invoke(Native Method)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: Caused by: java.lang.IllegalArgumentException: edu.berkeley.eecs.emission: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.
03-09 17:55:03.613 19003 19003 E AndroidRuntime: Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles.
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.PendingIntent.checkFlags(PendingIntent.java:375)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.PendingIntent.getBroadcastAsUser(PendingIntent.java:645)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.PendingIntent.getBroadcast(PendingIntent.java:632)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at org.altbeacon.beacon.service.scanner.CycledLeScanner.getWakeUpOperation(CycledLeScanner.java:497)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at org.altbeacon.beacon.service.scanner.CycledLeScanner.cancelWakeUpAlarm(CycledLeScanner.java:509)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at org.altbeacon.beacon.service.scanner.CycledLeScanner.setScanPeriods(CycledLeScanner.java:188)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at org.altbeacon.beacon.service.ScanJob.restartScanning(ScanJob.java:209)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at org.altbeacon.beacon.service.ScanJob.startScanning(ScanJob.java:250)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at org.altbeacon.beacon.service.ScanJob.onStartJob(ScanJob.java:84)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.job.JobService$1.onStartJob(JobService.java:62)
03-09 17:55:03.613 19003 19003 E AndroidRuntime: 	at android.app.job.JobServiceEngine$JobHandler.handleMessage(JobServiceEngine.java:105)
```

The main thing to consider here is this line, which is...

`AndroidRuntime: Caused by: java.lang.IllegalArgumentException: edu.berkeley.eecs.emission: Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent.`

This is caused by a `PendingIntent` in the package `org.altbeacon.beacon.service`, which is a dependency of the `cordova-plugin-ibeacon` plugin we just installed. I guess the plugin and its dependency are pretty old, which is why we are getting this error. 

To fix this, I am not quite sure how we do it. I am assuming to fix this, we will need to do these following things.

1. Find the package on github and create our own fork of it.
2. Modify it to make it up to date with Android development rules.
3. Redirect the dependency download in `cordova-plugin-ibeacon` from `org.altbeacon.beacon.service` to our new fork.

This seems straight forward, but I am a bit hazy on how to do step 3. In the next few days, hopefully we'll have that figured out and sorted so we can test the plugin and get it to work on Android!

the-bay-kay:
Quick observation: Looking at the config data for Louis and I's test beacons, it appears that the UUIDs of the beacons are _identical right out of the box_.  For this specific brand, we do have the capability to change these UUID's.

Whether or not we will need to change the UUIDs will depend on whether or not we can also access the systems-level beacon data: i.e., the System ID that appears on all Bluetooth devices (these are seen in the Classic Scanner, as a string of 8 hexidecimal(?) numbers in the format XX:XX:XX:XX:XX:XX).  We can gather this data when working with Bluetooth Classic; I have to imagine we will also be able to do so with BLE.  If we are able to have both the Sys_ID and UUID, we may be able to perform verification by checking both of these values.  Will update once I have more data!

the-bay-kay:
## iOS Build Issues
I'm noticing an discrepancy between the iOS DevApp build of this branch, and the iOS Hardware build.  Below are two videos: on the latter, I receive an error when attempting to run the test code:
### Serve Build
image

### Native / Full Build
image

### Error Code:
```
Objects are not valid as a React child
{found: object with keys {didDetermineStateForRegion, didStartMonitoringForRegion, didRangeBeaconsInRegion}}.  If you meant to render a collection instead of children, use an array instead. 
// Other ionic rendering errors follow...
```

### Notes
From my initial reading and assumptions, this issue has to do with the way React renders components; specifically, there's an issue with our set up methods.  Our code closely follows the example code from the plugin repo, but the React interface is a notable difference. 

Interestingly, we haven't had this issue on the Android Build; the components from the plugin render without error.  My assumption is, this has to do with how we're using the `testLogs()` method -- I'll do some more digging to figure out the cause!

louisg1337:
### **BLE Scanner Issue**

Katie and I have gotten the scanner set up and giving us some type of callback, but yet it doesn't pick up either of our beacons when we attempt to scan. We have been a bit stuck on this the past few days, but have made a little breakthrough as to what could be going wrong.

First and foremost, I never knew this but you can view the logcat of a physical Android device connected to your computer by following [this](https://stackoverflow.com/questions/3707880/using-adb-logcat-with-a-real-phone-and-not-the-emulator) StackOverflow post. I thought this was only a feature if you used Android Studio, so this is super nice to have. It also helped us find the error that we think is causing the scanning to not work.

**Error**

When following the sample code for [monitoring a single beacon](https://github.com/AltBeacon/android-beacon-library/blob/42eba232497ab550139080abd5097c333fc13472/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java#L1520), this error pops up whenever `didRangeBeaconsInRegion` gets called.
```
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: error reading beacon data
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: java.io.FileNotFoundException: https://s3.amazonaws.com/android-beacon-library/android-distance.json
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at com.android.okhttp.internal.huc.HttpURLConnectionImpl.getInputStream(HttpURLConnectionImpl.java:255)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.getInputStream(DelegatingHttpsURLConnection.java:211)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.getInputStream(HttpsURLConnectionImpl.java:30)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at org.altbeacon.beacon.distance.DistanceConfigFetcher.request(DistanceConfigFetcher.java:94)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at org.altbeacon.beacon.distance.ModelSpecificDistanceUpdater.doInBackground(ModelSpecificDistanceUpdater.java:28)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at org.altbeacon.beacon.distance.ModelSpecificDistanceUpdater.doInBackground(ModelSpecificDistanceUpdater.java:16)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at android.os.AsyncTask$3.call(AsyncTask.java:394)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
03-13 16:51:56.690 19163 19372 W DistanceConfigFetcher: 	at java.lang.Thread.run(Thread.java:920)
03-13 16:51:56.691 19163 19372 W ModelSpecificDistanceCalculator: Cannot updated distance models from online database at java.io.FileNotFoundException: 
```

**Thoughts**

As we are using an iBeacon plugin, distancing ranging is something that this plugin supports. I guess on the Android side of things, they have some type of distance json file that they use to compute the distances. If you try to go to [https://s3.amazonaws.com/android-beacon-library/android-distance.json](https://s3.amazonaws.com/android-beacon-library/android-distance.json) the file doesn't exist, which is where our problem lies. 

The code where this file gets called/used can be found [here](https://github.com/AltBeacon/android-beacon-library/blob/42eba232497ab550139080abd5097c333fc13472/lib/src/main/java/org/altbeacon/beacon/BeaconManager.java#L1520).

This issue gets discussed [here](https://github.com/AltBeacon/android-beacon-library/issues/967#issuecomment-620234768) in this Github issue, but the very next comment says it was a "permissions issue" that has been resolved already. Its been 4 years since he said that, so maybe the file either got taken down or the permissions issue came back.

I'm going to dig a bit deeper to see if I can find anything else about this issue along with potential solutions. If not that, maybe I can find that android distance file somewhere so we could upload it as a static JSON file in our project.

louisg1337:
### **Android Distance Missing File Update**

Looks like someone beat us to it and filed an issue about the missing file error we saw above. Here is the [issue](https://github.com/AltBeacon/android-beacon-library/issues/1183) that was just created.

It appears that this is unfortunately not the reason our beacons aren't scanning, as the main dev says you can ignore this error. We'll have to dig a bit deeper to figure out why our beacons aren't picking up!

the-bay-kay:
## iOS Updates
Success! It took some wrestling with XCode, but we are successfully detecting the BLE beacons on iOS.  Below is a screenshot of the scanner, as of the latest commit: as can be seen in the `state` field, the scanner both (i) successfully detects when the beacon is "inRegion", and (ii) terminates the scanning loop after the detection occurs.

Now that we know that this is working on iOS, I think we can formulate a plan for how we want to proceed:
- [ ] Fix the Android scanner, addressing the errors described above by Louis
- [ ] Adjust the Scanner UI, such that:
  - [ ] A "Scanning in Progress" bar is shown, and
  - [ ] The "Scan" button terminates an active scan
  - [ ] A successful callback is render within the `BluetoothCard` component
- [ ] Add the list of Bluetooth Beacons to the Dynamic Config

Ideally, I would also like to add the following functions to this scanner:
- [ ] Adjust the beacon TS type, such that it also includes the car make / model
- [ ] Display vehicle make / model upon a successful callback

 I'll make some of the cosmetic changes, since we should be able to knock those out pretty quickly.  In general, however, I believe getting the Android side of things up-and-running is priority #1. I'll update this thread's checklist as I make changes!

IMG_0003

the-bay-kay:
# Android Updates

I just had a chance to test the Android changes between meetings: on both my LGv30 and Nexus 6, it appears that **both** are scanning properly!! Please see the screenshots attached.

While exciting,  raises some questions: chief among them, why is the scanner not working on Louis' device?  At first, we thought it was an issue with the Android Version, as described in [this](https://github.com/petermetz/cordova-plugin-ibeacon/issues/456) thread.  Considering my LGv30 is running Android 9, I believe we can rule this out as the issue.  We're going to do some more testing, and report back!

## LGv30
LGv30Screenshot

## Nexus V6
NexusV6 Screenshot

louisg1337:
### **BLE Scan Update**

Seeing that Katie was able to scan, as shown above, I tried my personal iPhone and another Android test phone that I haven't used yet. I'm happy to report that they both work like shown above!

Piggybacking off of what Katie said above, it is bizarre that my Motorolla G6 test phone could not scan for the beacons. I was fiddling with it last night and decided to restart the phone to see if that would do anything, and of course it now works. Out of all the debugging I did, I never thought to try the old tried and true method of turning it on and off again.

With these issues now behind us, we can move forward and tidy this up!

the-bay-kay:
## BLE Scanner Update!
***Success!!*** Both Android and IOS are detecting the BLE Beacons!  Below is a (somewhat compressed) video of me testing on both Android (LGv30) and iOS (iPhone 8) simultaneously.  Some notes on the tests:
- As can be seen in the video, iOS updates _significantly_ slower than Android. This could be chalked up to the iPhone's older hardware / iOS, though I'm not entirely sure this is the reason.  It will be interesting to see how other iPhones perform!
- For these tests, I have the beacon on the lowest broadcasting power / range (for ease of testing, I can just toss the beacon out of frame).   Having tested with higher power settings, I can confirm this works with the larger regions (so, the beacon doesn't have to be within < 1 meter)
- With some testing, we can detect if we're already in range when we start scanning.  This means we won't have to rely on the "entrance into a zone" to start tracking! 

https://github.com/e-mission/e-mission-phone/assets/98350084/37552d91-c413-4f88-b733-48095ff9cba3

### Future Work
Currently, the scanner is searching for multiple beacons at once.  Unfortunately, I have not been able to test what happens when there are two beacons available; exiting one zone to another being in two zones at once, etc.  Once the other test beacons arrive, I'll update this thread accordingly!

shankari (approved):
I think we should add the ability to add new BLE entries, since iOS can only scan for known beacon UUIDs (@the-bay-kay @louisg1337 please confirm)

But I will probably want to add buttons to simulate enter and exit anyway, so will see if I can add it at the same time. This will be my first time writing react native UI code.

Whooo!!!

shankari:
> With some testing, we can detect if we're already in range when we start scanning. This means we won't have to rely on the "entrance into a zone" to start tracking!

Great news! This will allow us to make the FSM more robust. If we detect that we have entered the zone, we can start the trip, but if there is some noise/glitches in there, we can also use our regular geofence and start scanning

shankari:
Tried to remove all references to COARSE location from the beacon plugin

shankari:
However, I am not sure that this is the problem. That is a dynamic permission and should not affect the static checks.
Looking at `AndroidManifest.xml` I don't see the permission defined in `AndroidManifest.xml`

```
$ grep -r COARSE platforms/android/app/src/main/AndroidManifest.xml
$ 
```

However, I do see it in `platforms/android//app/build/intermediates/merged_manifest/release/AndroidManifest.xml`

```
platforms/android//app/build/intermediates/merged_manifest/release/AndroidManifest.xml:    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" /> <!-- Required by older versions of Google Play services to create IID tokens -->
```

why have we not encountered this before?

shankari:
The clue is in `platforms/android//app/build/outputs/logs/manifest-merger-debug-report.txt`
which indicates that it is added from altbeacon.

```
uses-permission#android.permission.ACCESS_COARSE_LOCATION
ADDED from [org.altbeacon:android-beacon-library:2.19] /Users/kshankar/.gradle/caches/transforms-3/eb47e011d1b70a4a657d3fb1caa7828e/transformed/jetified-android-beacon-library-2.19/AndroidManifest.xml:17:5-81
```

Anddddd here it is
https://github.com/AltBeacon/android-beacon-library/blob/42eba232497ab550139080abd5097c333fc13472/lib/src/main/AndroidManifest.xml#L6

altbeacon is the library that we want to use, so maybe the COARSE location permission is long-term after all.
but I don't want to ask for permissions that we don't need.
Let's try to remove the permission using https://stackoverflow.com/a/72988739

shankari:
That (at least with the `removeAll` variant from https://developer.android.com/build/manage-manifests seems to have removed the COARSE location permission request. I edited the `AndroidManifest.xml` manually this time

```
    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
    <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION" />
    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" tools:node="removeAll" />
    <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS" />
``` 

we should consider adding that line to the data collection `plugin.xml` when we finally incorporate this fucntionality directly into the plugin

PR #1137, opened 2024-03-19
[Bug Fix] Leaflet gets broken after updating labels 

jiji14:
### Overview
After optimizing the Leaflet map, a bug appears when updating labels.
We cache the map after rendering a Leaflet map and reset the `leafletMapRef`. However, when the map information changes, we need to re-render the map using `initMap`. Currently, this doesn't happen because we reset `leafletMapRef`, causing it to fail the conditional check.

### Before
Screenshot 2024-03-19 at 1 02 43 PM

### After
https://github.com/e-mission/e-mission-phone/assets/47590587/ae58c3c6-6f83-4b8b-bfb5-42ba07b425a9

### DIFF
--- a/www/js/components/LeafletView.tsx
+++ b/www/js/components/LeafletView.tsx
@@ -153,7 +153,9 @@ const LeafletView = ({ geojson, opts, downscaleTiles, cacheHtml, ...otherProps }
         dangerouslySetInnerHTML={
           /* this is not 'dangerous' here because the content is not user-generated;
           it's just an HTML string that we cached from a previous render */
-          cacheHtml && leafletCache?.has(mapElId) ? { __html: leafletCache.get(mapElId) } : undefined
+          cacheHtml && leafletCache?.has(mapElId)
+            ? { __html: leafletCache.get(mapElId) }
+            : undefined
         }
       />
     </View>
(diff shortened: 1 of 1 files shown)


jiji14:
@JGreenlee @shankari Please check this PR 🙏

shankari:
LGTM! Once Jack reviews, I will merge

shankari:
@JGreenlee tagging you for review

JGreenlee (approved):
It works well and looks good to me.

I tested this locally in the devapp because I wanted to make sure the color of the trajectory updates if the mode changes from one base mode to another (for example, bike -> car changes from green to red)

shankari:
@JGreenlee @jiji14 just to check, why was this not caught by the unit tests? It is a 100% reproducible bug for a fairly common function. Do we not have a test that pretends to click the button and label the trip?

JGreenlee:
@shankari I think we would have needed integration testing or end-to-end testing for this to have been caught. We started with Appium, but it is a significant effort. After encountering roadblocks with setup, we have not been able to prioritize it since there have been more urgent tasks lately

I have been loosely formulating a plan which I think would make integration testing more easily achievable, which is to support a version of the UI that runs as an Expo web app, detached from the native code.
Essentially, when Cordova is not available it would use a web compatibility layer for things the UI needs like usercache and servercomm. There would be no tracking; it would only display data that was already synced from another device.

Because it would be web-only, we could test the full extent of its capability, including integration testing with actual server calls.
And there's a couple bonus benefits, i) it will make it easier to transition to Expo, and ii) we will have a web version lying around if someone ever asks for it

shankari:
the web compatibility layer is an interesting idea and definitely the standard approach towards software based testing (e.g. creating mock/proxies/hardware abstraction layer). It would be interesting to see how thick the layer would need to be. If we are not testing the data collection, maybe only the usercache and the server calls. Would be great to have the web version "in our back pockets" in case there is a request for it.

PR #1140, opened 2024-03-29
Simulate BLE scans from the phone UI

shankari:
- Makes it easier to test how BLE works
- Makes it easier to test the plumbing end to end in the emulator

### DIFF
--- a/www/js/bluetooth/BluetoothCard.tsx
+++ b/www/js/bluetooth/BluetoothCard.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Card, List, useTheme } from 'react-native-paper';
+import { Card, List, Text, Button, useTheme } from 'react-native-paper';
 import { StyleSheet } from 'react-native';
 
 type Props = any;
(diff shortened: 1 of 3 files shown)


shankari:
@JGreenlee I would appreciate a review of this change since it is my first ever RN change. You are clearly more skilled than me in the e-mission UI at this point. However, in order to unblock us on the BLE work (notably to test https://github.com/e-mission/e-mission-docs/issues/1062#issuecomment-2026359038), I plan to merge this tonight and create a new staging release.

I will address your comments in a cleanup PR.

shankari:
The testing results are in https://github.com/e-mission/e-mission-docs/issues/1062
I should have put the screenshots in here but forgot. Sorry!

shankari:
| Before callbacks | Fake callback with monitor and range results (note that major/minor are different for range because of the hardcoded values | Trying with invalid UUID; first call succeeds but second call fails | Trying with a new random UUID and unset major and minor; range callback has them set to the hardcoded values | 
|-------|-------|------|-------|
| simulator_screenshot_C9B5EFD6-325C-4238-8C47-3757FB885660 | simulator_screenshot_308FDD34-0A45-415F-9C05-5CCC46EB537B | simulator_screenshot_3DCCB48D-1004-402A-9C4C-BF959CDCABF5 | simulator_screenshot_0C349050-DB4E-466F-A759-ACA544BF0B9B |

PR #1142, opened 2024-04-03
Make the Dashboard tab configurable, and implement new components

JGreenlee:
https://github.com/e-mission/e-mission-docs/issues/1055


jiji14:
[ Tasks ]
- [x] footprint_options . unlabeled_uncertainty
- [x] active_travel_options . modes_list
- [ ] summary_options . metrics_list

JGreenlee:
I resolved the merge conflicts on https://github.com/e-mission/e-mission-phone/pull/1138.
@jiji14 When you begin working on this again, you can pull from `JGreenlee:refactoring_timelinecontext` and the majority of the merge conflicts should go away

jiji14:
# Surveys UI screenshots 
Screenshot 2024-04-24 at 6 02 00 PM
Screenshot 2024-04-24 at 6 02 17 PM
Screenshot 2024-04-24 at 6 02 09 PM

jiji14:
@JGreenlee I have completed the frontend code for the survey dashboard. Currently, I am still working on the server code, and I will provide an explanation in the server pull request.

I have implemented a new API to fetch survey data every 24 hours because the survey data does not render based on the timeline. It should display the accumulated results. Therefore, I designed it to call the survey API when a user clicks on the dashboard tab and the last updated data was more than 24 hours ago.

Additionally, after retrieving survey data, I preprocess it for charts. Regarding the leaderboard, I have configured it to display only the top 5 members, myself, and the bottom 3 members.

Please review the code and comment if anything doesn't make sense to you. I will bring my personal laptop to Hawaii so I can communicate via GitHub. Thank you!

jiji14:
[ Upate ] 
Survey Comparison Card: Data depends on date range.
Survey Leaderboard Card: Data is accumulated.

To prevent confusion for users, we separate the two cards and inform them that the leaderboard data is accumulated.

Screenshot 2024-05-16 at 2 41 38 PM

Need to fix fetching data after the server side is done

JGreenlee:
Not bad actually. Looks like it's only `metricsHelper.test.ts`

JGreenlee:
We are lacking in code coverage, but tests passed

jiji14:
@JGreenlee 
Thanks for your wonderful work! I fixed some minor css issues.

I added 'no trip data' message for surveyComparison chart.
[ before ]
Screenshot 2024-05-25 at 12 12 32 PM

[ after ]
Screenshot 2024-05-25 at 12 12 41 PM


Also set the maxBarThickness to make it prettier!
Screenshot 2024-05-25 at 12 06 48 PM


@shankari 
Other than those CSS issues, everything went smoothly, so I think we are ready to go for staging! I will do more testing on the staging app.

shankari:
I am halfway through reviewing this change (15/31 files) but they were the easy 15. Will finish review of the changes to the  `metrics` folder tomorrow.

shankari:
@jiji14 you need to re-run prettier; the test is failing

jiji14:
@JGreenlee @shankari 
I added more unit tests in `metricsHelper` and `useImperialConfig`, as well as basic component tests for `Carousel` and `DateSelect`. I will continue to add more tests to achieve higher code coverage. Codecov seems to encounter a public key error again.. (The error log seems different from the previous one.) I will research what is causing it!

shankari:
@jiji14 the error appears to be the same, since it was successful on re-run. However, the patch check is still failing since only 16.23% of the newly added code. The project code coverage has gone up from 29% to 31%, but the patch (aka newly added code) is still not tested enough.

www/js/metrics/metricsHelper.ts
```diff
@@ -68,21 +75,21 @@ export function formatDateRangeOfDays(days: DayOfMetricData[]) {
 /* formatting data form carbon footprint calculations */
 
 //modes considered on foot for carbon calculation, expandable as needed
-const ON_FOOT_MODES = ['WALKING', 'RUNNING', 'ON_FOOT'] as const;
+export const ON_FOOT_MODES = ['WALKING', 'RUNNING', 'ON_FOOT'] as const;
```
shankari, line 78: I don't really think we need these any more. These were originally present because we thought we may want to return the `raw` section summaries. But we don't really use those any more. Even in the section summaries, we only have `cleaned` and `inferred` (and now `ble_sensed`). I think we should assume that we will not ever return raw section summaries and simplify this code.

www/js/metrics/metricsHelper.ts
```diff
@@ -151,14 +156,15 @@ export function parseDataFromMetrics(metrics, population) {
...
-  'date' in day ? day.date : day.fmt_time.substring(0, 10);
-
-export const tsForDayOfMetricData = (day: DayOfMetricData) =>
-  DateTime.fromISO(dateForDayOfMetricData(day)).toSeconds();
+const _datesTsCache = {};
+export const tsForDayOfMetricData = (day: DayOfMetricData) => {
+  if (_datesTsCache[day.date] == undefined)
+    _datesTsCache[day.date] = DateTime.fromISO(day.date).toSeconds();
+  return _datesTsCache[day.date];
+};
```
shankari, line 164: why is this a manually maintained cache instead of the existing `useMemo` function?
This is implemented in https://github.com/jiji14/e-mission-phone/commit/6ef2adf30caecdc249187ac02d2ec358facd8486 but I think that it is probably a performance optimization to avoid calling a function while rendering the charts.

www/__tests__/Carousel.test.tsx
shankari, line 1: I agree that this is a basic test, but it is really very basic, almost so basic as to be useless. I would assume that we at least try to click on the button and check that the views move.

www/__tests__/DateSelect.test.tsx
shankari, line 1: Ditto, I would expect that we click on the button and the date select is displayed with the correct date range, and that when the user selects a particular range, it is returned correctly. The key here is not just the button display, but the logic around the clamping IIRC

www/__tests__/metricsHelper.test.ts
```diff
@@ -1,51 +1,56 @@
...
       expect(getUniqueLabelsForDays(days1)).toEqual(['a', 'b', 'c', 'd']);
     });
-
-    const days2 = [
-      { mode_a: 1, mode_b: 2 },
-      { mode_c: 1, mode_d: 3 },
-    ] as any as DayOfClientMetricData[];
-    it("should return unique labels for days with 'mode_*'", () => {
-      expect(getUniqueLabelsForDays(days2)).toEqual(['a', 'b', 'c', 'd']);
-    });
```
shankari, line 31: @jiji14 please replace this with a test that actually have duplicate labels (see my comment in https://github.com/e-mission/e-mission-phone/pull/1138#discussion_r1607576184)

www/__tests__/metricsHelper.test.ts
```diff
@@ -70,59 +75,264 @@ describe('metricsHelper', () => {
...
-        [
-          { fmt_time: '2021-01-04T00:00:00Z' },
-          { fmt_time: '2021-01-08T00:00:00Z' },
-          { fmt_time: '2021-01-09T00:00:00Z' },
-          { fmt_time: '2021-01-10T00:00:00Z' },
-        ],
-        // prior week
-        [{ fmt_time: '2021-01-01T00:00:00Z' }, { fmt_time: '2021-01-02T00:00:00Z' }],
-      ]);
-    });
```
shankari, line 94: Again, please fix this test that was removed to get the tests to passs

www/__tests__/metricsHelper.test.ts
```diff
@@ -70,59 +75,264 @@ describe('metricsHelper', () => {
...
+            ['value3', 30],
+            ['value4', 40],
+          ],
+        },
+      ];
+      const result = isCustomLabels(modeMap);
+      expect(result).toBe(true);
+    });
+
+    it('returns true for all sensed labels', () => {
```
shankari, line 232: for sensed labels, we expect `isCustomLabels` to return **`false`**. Right now, this is identical to the previous test.

www/__tests__/metricsHelper.test.ts
```diff
@@ -70,59 +75,264 @@ describe('metricsHelper', () => {
...
+      expect(result[2](3600)).toBe('1 hours');
+    });
+
+    it('checks for count metric', () => {
+      const result = getUnitUtilsForMetric('count', imperialConfig);
+      expect(result).toEqual(['trips', expect.any(Function), expect.any(Function)]);
+      const mockTrip = { responded: 4, not_responded: 3 };
+      expect(result[1](mockTrip)).toBe(mockTrip);
+      expect(result[2](mockTrip)).toBe(mockTrip + ' trips');
+    });
```
shankari, line 328: This passes, but I don't think it is testing the right functionality. `count` is a metric to just count the number of trips. The metric that expects a structure with `responded` and `not_responded` is **`response_count`**

www/__tests__/useImperialConfig.test.ts
```diff
@@ -53,3 +61,15 @@ describe('convertSpeed', () => {
     expect(convertSpeed(6.7056, true)).toBeCloseTo(15); // Approximately 15 mph
   });
 });
+
+describe('useImperialConfig', () => {
+  it('returns ImperialConfig with imperial units', () => {
```
shankari, line 66: Imperial units are the old ones (popularized by the British Empire), so `miles`, `mph` etc. These are non-imperial or SI units. Shouldn't we also add a similar check where we expect to see `miles`, `mph` etc?

shankari (changes requested):
I am fine with the code, but have a lot of comments on the tests. I am merging this now so that we can get it onto staging, but @jiji14, please address the test comments, and improve the code coverage of this patch in a follow-up PR

shankari:
This is now on staging as version 1.7.9. The staging server has also been updated to incorporate 

- Nominatim test regression fix
    https://github.com/e-mission/e-mission-server/pull/967

- make localdate queries behave like canonical datetime range filters
    https://github.com/e-mission/e-mission-server/pull/968

- make push notifs configurable, i18n, & only for users missing labels
    https://github.com/e-mission/e-mission-server/pull/969

- Dashboard Survey Metrics
    https://github.com/e-mission/e-mission-server/pull/966

A couple of issues that encountered while testing before pushing to staging were

| Error while loading aggregate data | Only loads trips for the past 14 days, if there are no trips in that range, unlike before, nothing is loaded. How does the user know how far back they need to go? |
|------|-------|
| simulator_screenshot_BCD32087-A273-40FC-A169-C45C3D60AC8C |  simulator_screenshot_5B596C3A-4676-4C2F-9265-5D1544D629E5 |

shankari:
500 error has been resolved.

@JGreenlee I still see the following issues

| Only loads trips for the last 14 days | Group bar not displayed | 
|------|------|
| Screenshot_2024-05-31-13-37-54-51_5062a5234f99d532322a40d030e8949e | Screenshot_2024-05-31-13-41-01-23_5062a5234f99d532322a40d030e8949e | 

Note also that the active minutes table shows overlapping ranges (two weeks, then two one weeks and then per day). Not sure if that is intended, and it was a bit confusing when I first looked at it.

jiji14:
@shankari 
Thank you for all the comments on testing! I will create a new PR to address the issues you raised, adding more tests and fixing any issues identified. I agree that the component testing is too basic. I've been researching how to mock useState for click events, as component tests continue to fail due to this issue. I will keep digging into it!

JGreenlee:
> @JGreenlee I still see the following issues

I am looking into these issues this morning

PR #1145, opened 2024-04-12
🔵🦷 UI changes for BLE integration

JGreenlee:


JGreenlee:
New row in the Profile tab allows users to update the config without logging out and back in

Simulator Screen Recording - iPhone 8 - 2024-04-14 at 17 58 00

JGreenlee:
Once I finally started getting `background/bluetooth_ble` entries to show up, this error was triggered:

image


I looked at the data. Indeed, every single `background/bluetooth_ble` entry has `major` and `minor` of 0, despite those values being defined when they are passed to the `mockBLEObjects` function.

I looked through the plugin and I think I found the source of the issue.
https://github.com/e-mission/e-mission-data-collection/blob/a83ae18d84ad04946d50ab606022246f777fb692/src/android/wrapper/BluetoothBLE.java#L87

This code snippet uses `==` to compare strings when it should use `equals()`. That causes the major and minor (among other properties to never get filled in for the fake ble objects

www/index.html
```diff
@@ -3,7 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta name="viewport" content="initial-scale=1.0, maximum-scale=1.0, user-scalable=no, width=device-width, viewport-fit=cover">
-    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com https://nominatim.openstreetmap.org https://raw.githubusercontent.com; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' https://tile.openstreetmap.org 'unsafe-inline' data: 'unsafe-eval'">
+    <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: gap: https://ssl.gstatic.com https://nominatim.openstreetmap.org https://raw.githubusercontent.com http://10.0.2.2:9090; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-inline' 'unsafe-eval'; img-src 'self' https://tile.openstreetmap.org 'unsafe-inline' data: 'unsafe-eval'">
```
shankari, line 6: If we add this, we should also add `http://localhost:9090`. Although I really don't think we should change this in the `index.html` since it can be a (very farfetched) security hole in production. Ideally, we would add these (and the `edit-config` tag in `config.xml`  as part of `npm run build-dev` and remove them as part of `npm build`.

For now, I have the changes implemented locally to allow testing but have not checked them in.
I also made https://github.com/e-mission/e-mission-phone/pull/1145/files#diff-d7df3a5aedf99d611872de9affe54d6bd62531488711022ee9f33d1ac86baab6R148 locally but didn't check it in 😄

shankari:
@JGreenlee You should also be able to use the staging app to generate these now since you have real beacons
- use the geofence exit from the developer zone
- move the beacon near the device
- wait until you get real BLE scans
- log in to the same token from the emulator

JGreenlee:
Testing locally with "simulated" beacon scans. This was a fake trip (I used "Routes" in the android emulator) and it successfully matched to a vehicle identity ("Jack's Mazda 3").

This vehicle identity was configured with the major  and minor of the simulated beacon scans (they are hardcoded as 4567 and 1945 in decimal, which are 11d7 and 0799 in hexadecimal)

```
  "vehicle_identities": [
    {
      "value": "car_jacks_mazda3",
      "bluetooth_major_minor": ["11d7:0799"],
      "text": "Jack's Mazda 3",
      "baseMode":"CAR",
      "met_equivalent":"IN_VEHICLE",
      "kgCo2PerKm": 0.22031,
      "vehicle_info": {
        "type": "car",
        "license": "JHK ****",
        "make": "Mazda",
        "model": "3",
        "year": 2014,
        "color": "red",
        "engine": "ICE"
      }
    ...
```

image

JGreenlee:
Next, just need to update conditional surveys (along with the expressions in the config) so they can utilize the matched vehicle identity

JGreenlee:
With the updates to https://github.com/e-mission/nrel-openpath-deploy-configs/pull/91 (now using version 3 of that config):

Surveys are now shown conditionally based on the detected vehicle rather than by the sensed mode. Previously, this trip prompted the EV Roaming survey because the sensed mode was UNKNOWN. Now it prompts the Gas Car survey because the detected vehicle is Jack's Mazda 3, a gas car.



The January trips (test data) don't match a detected vehicle. So no survey is prompted.

JGreenlee:
@shankari I think it's ready. This PR includes #1141 so keep that in mind if you want to review it separately

shankari:
I don't have time for a review, and @JGreenlee knows the UI code better than me anyway. So I am going to merge this and create a new staging release, so we can test this afternoon/tonight and submit for app review tonight or tomorrow morning.
I will leave this in "Ready for review" to remind me to review for future cleanup

shankari:
It looks like code coverage is broken because of a codecov.io issue. We may have to revisit it @jiji14

PR #1146, opened 2024-04-17
Fermata fixes apr17

JGreenlee:
- UserInputButton: before appConfig is defined, don't show default survey

- in updateUnprocessedBleScans, await call to getUnifiedDataForInterval
    > Without the 'await' keyword here, execution would continue without the request being completed.
    > This would cause the BLE scans to sometimes not be considered while rendering and evaluating conditional surveys.
    > If the BLE scans took longer to load than the trips took to render, there would be no surveys prompted.
    > Thus, we must 'await' to make sure the BLE scans are loaded first.

- cacheResourcesFromConfig: reload the cache when config is downloaded
    > We recently added the ability to refresh the config from inside the app.
    > But whenever the config is downloaded for the first time, we cache the resources referenced in it by URL. Subsequent config downloads would still use the resources from the first time.
    > 
    > fetchUrlCached now accepts options to pass through to the fetch API; if cache is 'reload', we will skip checking our localStorage cache for a previously stored value. This option will also cause the fetch API will also skip its own internal cache
    > 
    > The result of this is that when we refresh the config, URL-referenced resources inside it will also be refreshed.

### DIFF
--- a/www/js/services/commHelper.ts
+++ b/www/js/services/commHelper.ts
@@ -5,17 +5,18 @@ import { TimestampRange } from '../types/diaryTypes';
 
 /**
  * @param url URL endpoint for the request
+ * @param fetchOpts (optional) options for the fetch request. If 'cache' is set to 'reload', the cache will be ignored
  * @returns Promise of the fetched response (as text) or cached text from local storage
  */
-export async function fetchUrlCached(url) {
+export async function fetchUrlCached(url: string, fetchOpts?: RequestInit) {
   const stored = localStorage.getItem(url);
-  if (stored) {
+  if (stored && fetchOpts?.cache != 'reload') {
     logDebug(`fetchUrlCached: found cached data for url ${url}, returning`);
     return Promise.resolve(stored);
   }
   try {
-    logDebug(`fetchUrlCached: found no cached data for url ${url}, fetching`);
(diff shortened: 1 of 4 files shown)


www/js/services/commHelper.ts
```diff
@@ -5,17 +5,18 @@ import { TimestampRange } from '../types/diaryTypes';
...
-  if (stored) {
+  if (stored && fetchOpts?.cache != 'reload') {
     logDebug(`fetchUrlCached: found cached data for url ${url}, returning`);
     return Promise.resolve(stored);
   }
   try {
-    logDebug(`fetchUrlCached: found no cached data for url ${url}, fetching`);
-    const response = await fetch(url);
+    logDebug(`fetchUrlCached: cache had ${stored} for url ${url}, not using; fetching`);
+    const response = await fetch(url, fetchOpts);
```
shankari, line 19: Just for my own understanding, we also cache separately in the localStorage. This is just to ensure that the HTTP cache (if any) also does not return obsolete values. The log statement prints the value from the `localStorage` cache.

shankari (approved):

www/js/services/commHelper.ts
JGreenlee: Correct

PR #1143, opened 2024-04-08
Add 'collectCoverageForm' configuration to track all files for Jest coverage

jiji14:
### Related Issue
https://github.com/e-mission/e-mission-docs/issues/1064

### Overview
We have been ignoring files not included in the test folder when calculating coverage. To improve this, I added the `collectCoverageFrom` configuration.

For more information, please refer to this document: https://jestjs.io/docs/configuration#collectcoveragefrom-array

### Code coverage report after adding `collectCoverageFrom`
Screenshot 2024-04-08 at 1 05 31 PM

PR #1148, opened 2024-04-26
Lock React version to ~18.2.0

JGreenlee:
React 19 and React 18.3 are out (https://react.dev/blog/2024/04/25/react-19-upgrade-guide#react-18-3), but the React Native ecosystem is still on React 18.2. We need to lock the versions down until the React Native packages update.


JGreenlee:
I fixed the error from https://github.com/e-mission/e-mission-phone/pull/1143#issuecomment-2078482343, but the `build` workflows are still failing. May be a separate issue

JGreenlee:
@shankari Any ideas? I fixed the React dependency issue, but CI still failed on both `osx-build-android` and `osx-build-ios`.

I was able to build Android locally

shankari:
The iOS failure is because the default iOS has been upgraded to 15, so unrelated to this

```
clang: error: SDK does not contain 'libarclite' at the path '/Applications/Xcode_15.0.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a'; try increasing the minimum deployment target
```

shankari:
The android error is

```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.
* What went wrong:

Could not open settings generic class cache for settings file '/Users/runner/work/e-mission-phone/e-mission-phone/platforms/android/settings.gradle' (/Users/runner/.gradle/caches/7.6/scripts/dqjrgsw17c62wbpoagod92e1e).
You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.
> BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 65
```

according to https://stackoverflow.com/questions/68597899/bug-exception-in-phase-semantic-analysis-in-source-unit-buildscript-unsup, this is due to an incompatible java version

shankari:
I think all these failures are because the github runners were recently upgraded and we hadn't locked down all the versions, or the way that we locked down the versions failed.

For android, we do attempt to lock down the version

```
  echo "Default java version"
  java -version
  echo "Setting to Java 11 instead"
  export JAVA_HOME=$JAVA_HOME_11_X64
  java -version
  echo "Checking gradle"
  which gradle
  gradle -version
```

but apparently it didn't work

```
Default java version
openjdk version "21.0.3" 2024-04-16 LTS
OpenJDK Runtime Environment Temurin-21.0.3+9 (build 21.0.3+9-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (build 21.0.3+9-LTS, mixed mode)
Setting to Java 11 instead
openjdk version "21.0.3" 2024-04-16 LTS
OpenJDK Runtime Environment Temurin-21.0.3+9 (build 21.0.3+9-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (build 21.0.3+9-LTS, mixed mode)
Checking gradle
/opt/homebrew/bin/gradle
```

I wonder if the GitHub runners have removed Java 11

shankari:
iOS is now using the correct version but is broken

```
ld: file not found: /Applications/Xcode_14.3.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/arc/libarclite_iphonesimulator.a
```

This is the same issue
https://stackoverflow.com/a/75729977

Falling back even further to 14.1, which i know works.

For the android issue, maybe I need to change the `PATH` as well and not just the `JAVA_HOME`
Let's try that too, and add some more printout statements.

shankari:
Failure on android seems to be because this is now an ARM machine. So instead of using  `$JAVA_HOME_11_X64`, we need to use `JAVA_HOME_11_arm64`

```
JAVA_HOME=/Users/runner/hostedtoolcache/Java_Temurin-Hotspot_jdk/21.0.3-9.0/arm64/Contents/Home/
JAVA_HOME_17_arm64=/Users/runner/hostedtoolcache/Java_Temurin-Hotspot_jdk/17.0.11-9/arm64/Contents/Home/
JAVA_HOME_11_arm64=/Users/runner/hostedtoolcache/Java_Temurin-Hotspot_jdk/11.0.23-9/arm64/Contents/Home/
```

The error in  iOS seems to be because 

```
  sudo xcode-select -s /Applications/Xcode_14.1.app
  echo "Version should now be 14.1, re-checking "
  xcode-select --print-path
  shell: /bin/bash -e {0}
xcode-select: error: invalid developer directory '/Applications/Xcode_14.1.app'
```

But it should be around:
https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode

Let's check the version of the image that we are using...

shankari:
Ah, we currently specify 

```
  build:
    # The type of runner that the job will run on
    runs-on: macos-latest
```

But `macos-latest` is now `macos-14`, which does indeed have
1. `JAVA_HOME_11_arm64`: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#java
2. `xcode 14.3.1`: https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md#xcode

So let's just use xcode 13 for now, at least in this PR for iOS and then fix it while fixing https://github.com/e-mission/e-mission-docs/issues/1060

For android, let's try to keep the most latest mac, but switch the environment variable

shankari:
This does fall back to android 11, but we still get a failure

```
Default java version
openjdk version "21.0.3" 2024-04-16 LTS
OpenJDK Runtime Environment Temurin-21.0.3+9 (build 21.0.3+9-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (build 21.0.3+9-LTS, mixed mode)
Java full path
/usr/bin/java
Setting to Java 11 instead
/usr/bin/java
openjdk version "11.0.23" 2024-04-16
OpenJDK Runtime Environment Temurin-11.0.23+9 (build 11.0.23+9)
OpenJDK 64-Bit Server VM Temurin-11.0.23+9 (build 11.0.23+9, mixed mode)
Checking gradle
/opt/homebrew/bin/gradle

------------------------------------------------------------
Gradle 8.7
------------------------------------------------------------


Checking Java JDK and Android SDK versions
ANDROID_HOME=/Users/runner/Library/Android/sdk (recommended setting)
ANDROID_SDK_ROOT=/Users/runner/Library/Android/sdk (DEPRECATED)
Using Android SDK: /Users/runner/Library/Android/sdk
Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* What went wrong:
Could not open settings generic class cache for settings file '/Users/runner/work/e-mission-phone/e-mission-phone/platforms/android/settings.gradle' (/Users/runner/.gradle/caches/7.6/scripts/dqjrgsw17c62wbpoagod92e1e).
> BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 65
```

Let's downgrade this to xcode 13 as well and confirm that it is fixed

shankari:
So this was all about the mac version.

https://github.com/e-mission/e-mission-phone/pull/1148/commits/0d0b21261c12fe2090045ba3e4b2aaee0551f04f
flipped the checks from failing to passing **although the java version was incorrect**

```
  echo "Default java version"
  java -version
  echo "Java full path"
  which java
  echo "Setting to Java 11 instead"
  export JAVA_HOME=$JAVA_HOME_11_arm64
  which java
  java -version
  echo "Checking gradle"
  which gradle
  gradle -version

Default java version
openjdk version "17.0.11" 2024-04-16
OpenJDK Runtime Environment Temurin-17.0.11+9 (build 17.0.11+9)
OpenJDK 64-Bit Server VM Temurin-17.0.11+9 (build 17.0.11+9, mixed mode)
Java full path
/usr/bin/java
Setting to Java 11 instead
/usr/bin/java
openjdk version "21.0.3" 2024-04-16 LTS
OpenJDK Runtime Environment Temurin-21.0.3+9 (build 21.0.3+9-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.3+9 (build 21.0.3+9-LTS, mixed mode)
Checking gradle
/usr/local/bin/gradle
```

Note that we still get the gradle warning but no error, and then we actually fix the gradle warning.
This may also be due to the fact that I removed the PATH modification, but for now, let's just pin to OSX 13 to get everything to work properly and then make the changes in a principled fashion


```
Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.6/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 5s
1 actionable task: 1 executed
Subproject Path: CordovaLib
Subproject Path: app
Downloading https://services.gradle.org/distributions/gradle-7.6-all.zip
...............10%................20%................30%................40%...............50%................60%................70%................80%...............90%................100%
```

shankari:
using macos 13 is not enough since the default xcode has changed. We do still need to pin xcode

```
Run xcode-select --print-path
/Applications/Xcode_15.0.1.app/Contents/Developer
```

shankari:
Now, everything passed except for the code coverage. I know it has worked in the past, so will try to rerun in a bit

shankari:
Yay! Everything is finally passing with minimal changes.
Squash merging to avoid commit churn....

jiji14:
@JGreenlee @shankari Thanks for taking care of the issue

PR #1150, opened 2024-05-05
🚑️ 🐛 Ensure that we do show the replaced mode when the `mode_of_inter…

shankari:
…est` matches

At some point, we changed the mode and purpose user inputs as objects that were stored to be full objects, with start and end timestamps, instead of just the labels. We changed all uses of the MODE and PURPOSE to match it, but apparently forgot to change this location, where the replaced mode button is conditionally displayed.

This is a quick change to make the usage here consistent with the rest of the code so that we can push this out ASAP.

@JGreenlee, please let me know if there is a more principled fix, it is late and I don't want to experiment any further.

Testing done:
Please see PR

### DIFF
--- a/www/__tests__/confirmHelper.test.ts
+++ b/www/__tests__/confirmHelper.test.ts
@@ -45,6 +45,16 @@ const fakeDefaultLabelOptions = {
     },
   },
 };
+const fakeInputs = {
+  MODE: [
+    { data: { label: 'walk', start_ts: 1245, end_ts: 5678}},
+    { data: { label: 'bike', start_ts: 1245, end_ts: 5678}},
+  ],
+  PURPOSE: [
+    { data: {label: 'home', start_ts: 1245, end_ts: 5678 }},
+    { data: {label: 'work', start_ts: 1245, end_ts: 5678 }}
+  ],
+};
 
 jest.mock('../js/services/commHelper', () => ({
   ...jest.requireActual('../js/services/commHelper'),
@@ -62,8 +72,8 @@ describe('confirmHelper', () => {
 
   it('returns base labelInputDetails for a labelUserInput which does not have mode of study', () => {
     const fakeLabelUserInput = {
-      MODE: fakeDefaultLabelOptions.MODE[1],
-      PURPOSE: fakeDefaultLabelOptions.PURPOSE[0],
+      MODE: fakeInputs.MODE[1],
+      PURPOSE: fakeInputs.PURPOSE[0],
     };
     const labelInputDetails = labelInputDetailsForTrip(
       fakeLabelUserInput,
@@ -74,8 +84,8 @@ describe('confirmHelper', () => {
 
   it('returns full labelInputDetails for a labelUserInput which has the mode of study', () => {
     const fakeLabelUserInput = {
-      MODE: fakeDefaultLabelOptions.MODE[0], // 'walk' is mode of study
-      PURPOSE: fakeDefaultLabelOptions.PURPOSE[0],
+      MODE: fakeInputs.MODE[0], // 'walk' is mode of study
+      PURPOSE: fakeInputs.PURPOSE[0],
     };
     const labelInputDetails = labelInputDetailsForTrip(
       fakeLabelUserInput,
--- a/www/__tests__/confirmHelper.test.ts
+++ b/www/__tests__/confirmHelper.test.ts
@@ -47,12 +47,12 @@ const fakeDefaultLabelOptions = {
 };
 const fakeInputs = {
   MODE: [
-    { data: { label: 'walk', start_ts: 1245, end_ts: 5678}},
-    { data: { label: 'bike', start_ts: 1245, end_ts: 5678}},
+    { data: { label: 'walk', start_ts: 1245, end_ts: 5678 } },
+    { data: { label: 'bike', start_ts: 1245, end_ts: 5678 } },
   ],
   PURPOSE: [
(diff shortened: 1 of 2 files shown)


shankari:
| On a godogo config | bike does not show replaced mode | double confirmation that there was only one "bike" in the labeloptions | After the fix, it is shown properly
|---------|----------|------------|---------|
| Simulator Screen Shot - iPhone 14 Pro - 2024-05-05 at 02 43 48 | Simulator Screen Shot - iPhone 14 Pro - 2024-05-05 at 02 43 35 | Simulator Screen Shot - iPhone 14 Pro - 2024-05-05 at 02 43 41 | Simulator Screen Shot - iPhone 14 Pro - 2024-05-05 at 02 45 38 |

shankari:
I am going to merge this tomorrow morning and submit for review, given that it is a showstopper.

shankari:
This is almost certainly already on production, so checking what happens when the fixed code is sent out by installing version 1.7.4 first and then upgrading.

| all trips shown as labeled | they were e-bike and bike | after upgrade, the bike trip showed up as needing to label the replaced mode (phew!) |
|---------------|----------------|----------------|
| simulator_screenshot_0D3C0158-48D5-45A0-B7BD-57430F3046AF | simulator_screenshot_93469123-55BA-405D-A6C0-AC382854B286 | simulator_screenshot_1AA573BA-605A-49AA-AD8B-870868FF343F |

shankari:
I looked at the git blame and it looked like this had not been changed since the before the rewrite - it still used `userInput.Mode.value` which was the old copy-over label value mapping way of accessing the input. So I installed binary versions of the app going back to the rewrite and it has been broken since February!!! We didn't notice, and none of the users noticed either. We haven't heard anything back about it, not even from the current e-bike deployers (DDOT/DOEE/ca-ebike). @JGreenlee am I correct?!

If so, then from https://github.com/e-mission/e-mission-phone/pull/1150#issuecomment-2094871569, people might get a whole slew of "To label" trips for trips that they had already labeled. That's good, because our data will be nice again, but we may want to send a heads-up, both as a notification, and to the current deployers in case they get a slew of questions. Maybe we can combine it with the upcoming stacked bar chart change. @JGreenlee can you draft this update?

| v1.6.9 is broken | v1.6.6 is broken | v1.6.3 broken | v1.5.6 works |
|----------|-------|--------|-------|
| simulator_screenshot_1F11310B-0AF4-4DB0-BA9B-C1508F4A695C | simulator_screenshot_8253F4B9-E9BC-4548-BD07-EE3845CB2D60 | simulator_screenshot_160425DB-26D2-4568-87B7-9F6BB0A0047E | simulator_screenshot_A0DCB339-141B-4A07-A573-68E2709EE519 |

shankari:
The tests are failing because of

```
  ● confirmHelper › returns full labelInputDetails for a labelUserInput which has the mode of study

    expect(received).toEqual(expected) // deep equality

    - Expected  - 6
    + Received  + 0

    @@ -9,12 +9,6 @@
          "choosetext": "diary.choose-purpose",
          "key": "manual/purpose_confirm",
          "labeltext": "diary.purpose",
          "name": "PURPOSE",
        },
    -   "REPLACED_MODE": Object {
    -     "choosetext": "diary.choose-replaced-mode",
    -     "key": "manual/replaced_mode",
    -     "labeltext": "diary.replaces",
    -     "name": "REPLACED_MODE",
    -   },
      }

      83 |     );
      84 |     const fullLabelInputDetails = getLabelInputDetails(fakeAppConfigWithModeOfStudy);
    > 85 |     expect(labelInputDetails).toEqual(fullLabelInputDetails);
         |                               ^
      86 |   });
      87 |
      88 |   it(`converts 'other' text to a label key`, () => {

      at Object.toEqual (www/__tests__/confirmHelper.test.ts:85:31)
```

shankari:
Ah, it looks like the text expects that the `MODE` will be filled in with the entry from the label option

```
    { value: 'walk', baseMode: 'WALKING', met_equivalent: 'WALKING', kgCo2PerKm: 0 },
```

But it is in fact filled in with the user input object, with data and metadata
Screenshot 2024-05-05 at 10 58 50 AM

JGreenlee:
I am glad you caught this. I believe your investigation is correct; when I unified the format of ENKETO vs MULTILABEL user inputs on the phone (which was the first step of https://github.com/e-mission/e-mission-docs/issues/1045), I missed this bit of code and the corresponding unit test.
Yes, I think all trips that were labeled as e-bike during that time will show back up in "To Label" with this fix.

<hr>

Since this was a pretty serious issue, I want to take a moment and reflect on it.

I think there are a couple things that could have helped prevent this issue from going undetected:
1. better type safety (if `labelInputDetailsForTrip` had typed arguments this would have been caught by the TS linter)
1. testing `REPLACED_MODE` in my own day-to-day manual testing
  (I don't own an ebike, so I don't usually test 'replaced mode'. I could fake it for the purpose of testing though)
1. integration + end-to-end testing

Type safety is an ongoing effort; we have been gradually adding types throughout the codebase wherever possible.

As OpenPATH grows to support more deployments with differing configurations, I think it is growing increasingly difficult to validate the full range of features with manual testing. We cover as much as we can with our small team

Integration and end-to-end testing have been long-term goals, but maybe one key takeaway from this is how crucial they are to the long-term success of the project. When we have time, we should really sit down and devote some time to those, or at least form a concrete plan for when and how we can implement them (like what can we do before the cordova migration vs. what should we wait until after to do?)

PR #1151, opened 2024-05-05
💩 Turn off the background checks for fleet deployments

shankari:
Since the fleet deployments are installed on work phones, and the background restriction is apparently grayed out on android work phones.
https://github.com/e-mission/e-mission-docs/issues/1070

This is a hack (hence the 💩 emoji) but it will allow us to make progress through the onboarding, get the logs and then resolve the issue the right way. Or if we can't figure out how to access it the right way, this can become a config option for deployments that plan to use work phones.

This should unblock
https://github.com/e-mission/e-mission-docs/issues/1070#issuecomment-2094354832

### DIFF
--- a/www/js/usePermissionStatus.ts
+++ b/www/js/usePermissionStatus.ts
@@ -446,9 +446,10 @@ const usePermissionStatus = () => {
       setupAndroidFitnessChecks();
       if (appConfig.tracking?.bluetooth_only) {
         setupAndroidBluetoothChecks();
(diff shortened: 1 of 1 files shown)


shankari:
There are no current unit tests for the permission screen.
Since this is hack code (note 💩), I am not inclined to set up the test framework just to test this.
All tests other than code coverage pass.

I am going to go ahead and merge this.

PR #1153, opened 2024-05-05
💚 ⬆️  Upgrade the pinned version of the runner to 14-large

shankari:
This should still have the correct environment variable because `14-large` runs x64 https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md

`macos-14` now runs amd, and we have to figure out how to run it. Maybe we can run both as part of a schedule instead of on every pull request

### DIFF
--- a/.github/workflows/android-build.yml
+++ b/.github/workflows/android-build.yml
@@ -53,13 +53,13 @@ jobs:
     - name: Setup the cordova environment
       shell: bash -l {0}
       run: |
-        export JAVA_HOME=$JAVA_HOME_11_X64
+        export JAVA_HOME=$JAVA_HOME_17_arm64
         bash setup/setup_android_native.sh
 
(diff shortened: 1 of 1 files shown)


shankari:
Failure is

```
Requested labels: macos-14-large
Job defined at: e-mission/e-mission-phone/.github/workflows/android-build.yml@refs/pull/1153/merge
Waiting for a runner to pick up this job...
Job is waiting for a hosted runner to come online.
```

Yup.
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Let's go back to `macos-14`, which is actually now an arm64 machine

```
macOS 14 Arm64 	macos-latest, macos-14, macos-latest-xlarge or macos-14-xlarge
```

PR #1154, opened 2024-05-07
🎆 Only do BLE matching on the phone for unprocessed BLE Scans

JGreenlee:
### fill in new BLE fields & update types

`ble_sensed_mode` and `ble_sensed_summary` were added to sections and confirmed trips in https://github.com/e-mission/e-mission-server/pull/965.

We are going to need `ble_sensed_mode` to determine vehicle info.

The section summaries (`ble_sensed_summary`, `cleaned_section_summary` and `inferred_section_summary`) are not used in unprocessed trips currently, but I am adding them so there is less of a gap between composite trips and unprocessed trips

<hr>

### use only unprocessed BLE scans in label tab

Since we have matching on the server, processed sections will already have BLE sensed modes filled in. We no longer need to query for all BLE scans; only unprocessed ones (ie newer than pipeline end ts).
Then while we are constructing unprocessed trips, the list of unprocessed BLE scans will be used to determine BLE sensed modes.

We no longer need timelineBleMap and won't treat BLE scans like user inputs anymore. For 'confirmedMode', instead of using timelineBleMap, we will use the ble_sensed_mode of the primary section. New simple function in diaryHelper to deternmine the primary section.

### DIFF
--- a/www/js/diary/LabelTab.tsx
+++ b/www/js/diary/LabelTab.tsx
@@ -29,11 +29,7 @@ import { getLabelOptions, labelOptionByValue } from '../survey/multilabel/confir
 import { displayError, displayErrorMsg, logDebug, logWarn } from '../plugin/logger';
 import { useTheme } from 'react-native-paper';
 import { getPipelineRangeTs } from '../services/commHelper';
-import {
-  getNotDeletedCandidates,
-  mapBleScansToTimelineEntries,
-  mapInputsToTimelineEntries,
-} from '../survey/inputMatcher';
+import { getNotDeletedCandidates, mapInputsToTimelineEntries } from '../survey/inputMatcher';
 import { configuredFilters as multilabelConfiguredFilters } from '../survey/multilabel/infinite_scroll_filters';
 import { configuredFilters as enketoConfiguredFilters } from '../survey/enketo/infinite_scroll_filters';
 import LabelTabContext, {
@@ -45,6 +41,7 @@ import LabelTabContext, {
 import { readAllCompositeTrips, readUnprocessedTrips } from './timelineHelper';
 import { LabelOptions, MultilabelKey } from '../types/labelTypes';
 import { CompositeTrip, TimelineEntry, TimestampRange, UserInputEntry } from '../types/diaryTypes';
+import { primarySectionForTrip } from './diaryHelper';
 
 let showPlaces;
 const ONE_DAY = 24 * 60 * 60; // seconds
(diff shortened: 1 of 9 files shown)


JGreenlee:
Unit tests passed but codecov failed

shankari:
We should just wait for ~ 30 mins and retry; I will do that
Is there a related server update to bump up common as well?

JGreenlee:
e-mission-server is fine to stay on 0.4.3 for now because it doesn't use `primary_ble_sensed_mode_for_trip` yet https://github.com/JGreenlee/e-mission-common/releases/tag/0.4.4.
In fact, neither does e-mission-phone right now.
I need to revise it anyway because of what we discussed on Tuesday

This can be pushed out to fix the slowness on dfc-fermata, while I work on the other changes in parallel

shankari:
The pipeline is reset to include `ble_sensed_mode` properly through the data model. We will need to reset once more after the `vehicle_summary_changes` are done

www/js/diary/LabelTab.tsx
```diff
@@ -171,7 +160,7 @@ const LabelTab = () => {
         unprocessedNotes = ${JSON.stringify(unprocessedNotes)}`);
       if (appConfig.vehicle_identities?.length) {
         await updateUnprocessedBleScans({
-          start_ts: pipelineRange.start_ts,
+          start_ts: pipelineRange.end_ts,
```
shankari, line 163: note to self, this is not an error, this is _the_ performance optimization!

www/js/diary/LabelTab.tsx
```diff
@@ -336,8 +330,8 @@ const LabelTab = () => {
    * @returns Confirmed mode, which could be a vehicle identity as determined by Bluetooth scans,
    *  or the label option from a user-given 'MODE' label, or undefined if neither exists.
    */
-  const confirmedModeFor = (tlEntry: TimelineEntry) =>
-    timelineBleMap?.[tlEntry._id.$oid] || labelFor(tlEntry, 'MODE');
+  const confirmedModeFor = (tlEntry: CompositeTrip) =>
+    primarySectionForTrip(tlEntry)?.ble_sensed_mode || labelFor(tlEntry, 'MODE');
```
shankari, line 334: nit, future fix: I think it would be worthwhile for you and Abby to sync up on how this should work across the public dashboard and the phone. The server side data does not have a `confirmedMode`; maybe it should add one. Also, I am not sure that `confirmed_mode` is the best name for this, since the ble_sensed_mode has not been confirmed by anybody.

shankari (approved):

www/js/diary/LabelTab.tsx
JGreenlee: Agreed, this is not the long term solution.

"Confirmed mode" was something temporary I threw together for the alpha. It actually represents the BLE sensed mode OR the labeled mode, depending on what is available.

For `dfc-fermata` there are no mode labels and I think it makes more sense to use something like `primary_ble_sensed_mode`. will coordinate this with Abby once I know what adjustments we are making to the section summaries 

I do think something like "confirmed mode" will be useful in the app dashboard and for analysis. I'm envisioning future programs that could theoretically use BLE sensing in addition to mode+purpsoe labels – the BLE sensed trips will not need manual user input since we already "confirmed" the mode when we matched it to a beacon.
When we do viz/ analysis we will want one property that encompasses both cases: i) determined by BLE or ii) manually confirmed

PR #1155, opened 2024-05-09
🐛 Bugfixes + Refactoring relating to User Inputs

JGreenlee:
### refactor ENKETO userinputs to use survey name as key instead of 'SURVEY'

Part of the initiative described in https://github.com/e-mission/e-mission-docs/issues/1045
This implements the "proposed" structure on the phone for Enketo user inputs.
When unprocessed user inputs are read and stored into unprocessedLabels in timelineHelper, responses will now be sorted by survey name instead of all being kept in one slot 'SURVEY'.

Also implement a new function, enketoHelper > resolveSurveyButtonConfig, which reads the config and determines what survey(s) are available for a button (like 'trip label', 'trip addition', 'place label', 'place addition').
Includes backwards compat so old configs (which don't have those fields) will just use the default 'TripConfirmSurvey'.
Including the translations for TripConfirmSurvey's 'not-filled-in-label' directly in the backwards compat to be consistent with new-style surveys where the translations are directly in the config.

This new function is used by <UserInputButton>, simplifying the code that was there previously. It now stores the returned SurveyButtonConfig object as 'survey' rather than keeping 'surveyName' and 'notFilledInLabel' separately.
This allows conditionalSurveys > getSurveyForTimelineEntry to be simplified too.
The new function is also used in timelineHelper > updateUnprocessedInputs. Instead of calling filterByNameAndVersion only for 'TripConfirmSurvey' and then storing those under the key of 'SURVEY', it first calls `resolveSurveyButtonConfig` to get *all* possible 'trip-label' surveys, iterates over those surveys calling filterByNameAndVersion for each, and stores unprocessed responses using the name of the survey as the key.

See types updated in LabelTabContext.ts.
appConfigTypes.ts updated to reflect that showsIf is optional (if not present, that survey shows unconditionally)

<hr>

### don't filter deleted additions twice

I found an edge case where if an old addition (ie one that was already processed) was edited, there would be duplicates; both the old and new version would show up (until the new one got processed).

This is because we were first applying the 'not deleted' filter to unprocessed entries. Then after they are merged with processed entreis we applied it to the merged copy. This didn't work for the edge case because the DELETED entry was already filtered out the first time, so the processed original entry just sticks around.

This fix ensures that the 'not deleted' filter ONLY happens once processed+unprocessed are merged together, this way the filter considers all entries at once

### DIFF
--- a/www/__tests__/inputMatcher.test.ts
+++ b/www/__tests__/inputMatcher.test.ts
@@ -1,6 +1,6 @@
 import { mockBEMUserCache } from '../__mocks__/cordovaMocks';
 import { mockLogger } from '../__mocks__/globalMocks';
-import { unprocessedLabels, updateLocalUnprocessedInputs } from '../js/diary/timelineHelper';
+import { updateLocalUnprocessedInputs } from '../js/diary/timelineHelper';
 import * as logger from '../js/plugin/logger';
 import { EnketoUserInputEntry } from '../js/survey/enketo/enketoHelper';
 import {
@@ -376,9 +376,9 @@ describe('mapInputsToTimelineEntries on an ENKETO configuration', () => {
       user_input: {
         trip_user_input: {
           data: {
-            name: 'TripConfirmSurvey',
+            name: 'MyCustomSurvey',
             version: 1,
-            xmlResponse: '<processed TripConfirmSurvey response>',
+            xmlResponse: '<processed MyCustomSurvey response>',
             start_ts: 1000,
             end_ts: 3000,
           },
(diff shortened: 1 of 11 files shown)


JGreenlee:
This fixes 2 bugs:

1. On trip surveys other than "TripConfirmSurvey", recording a survey response and refreshing before the response is processed causes it to temporarily disappear
    - This was due to line ~118 in timelineHelper; fixed by 02655ba28913a06caa80795aa3a07a4b15d06ef3
    - testing done: On `dfc-fermata`, record a GasTripSurvey response. It shows "Answered". Click the refresh button. Before applying patch, it showed "Gas Car Survey". After applying the patch, it still shows "Answered" as expected.
2. Processed additions that were edited/deleted still showed the original version until the edit/deletion was processed
    - fixed by cb1e45b2ccdafc04ae8b9ec567a7d09fba6e9f93
    - Testing done: On `stage-timeuse`, I had an old activity that was "2 Recreation/Leisure". I edited it to be "3 Recreation/Leisure" and saved. Then, before applying the patch, both "2 Reaction/Leisure" and "3 Recreation/Leisure" show on the same place. After applying the patch, only "3 Recreation/Leisure" shows.

shankari (approved):
Interesting! It always felt hacky to have a hardcoded value for the survey configuration `SURVEY`, so I am glad to see that the structure was actually prescient and forwards compatible!

PR #1159, opened 2024-05-21
TimelineContext fixes

JGreenlee:
#### fix address names not loading

'timelineMap' is a Map object, not just a plain { } object. So instead of Object.values() we should use Map.values() - which returns an iterator - but we want to reverse it.
So we have to convert to array, then reverse, and then call forEach to fillLocationNamesOfTrip.

Also, we should ensure that only trips, not places, get passed to fillLocationNamesOfTrip
(since places always have the same locations as the trips surrounding them, the place cards will automatically get their location names filled in)

<hr>

#### update TimelineScrollList only when the active tab is 'label'

Leaflet map encounters an error when prerendered, so we need to render the TimelineScrollList component when the active tab is 'label'
'shouldUpdateTimeline' state is used to determine whether to render the TimelineScrollList or not

<hr>

#### fix LabelTab filtering

The existing implementation used pub/sub events to trigger when filtering should happen. The events were triggered from TimelineContext.
However, this doesn't listen / wait for filterInputs to be defined because filterInputs is inside LabelTab, not TimelineContext. Filtering wasn't working because filterInputs was still empty when the initial filtering occured and LabelTab had no control over when the event was triggered.

A better solution is to listen for changes inside LabelTab. Every time the filters, the loaded trips, or the labels change, filtering occurs. If a trip had a user input in the last 30s, we skip it; but schedule a re-filter when its "immunity" expires.

Now LabelTab is responsible for all the filtering. TimelineContext doesn't have to worry about what LabelTab (its child) does. This is a cleaner, "React-recommended" unidirectional pattern

### DIFF
--- a/www/js/TimelineContext.ts
+++ b/www/js/TimelineContext.ts
@@ -50,6 +50,8 @@ type ContextProps = {
   loadMoreDays: (when: 'past' | 'future', nDays: number) => void;
   loadSpecificWeek: (d: string) => void;
   refreshTimeline: () => void;
+  shouldUpdateTimeline: Boolean;
+  setShouldUpdateTimeline: React.Dispatch<React.SetStateAction<boolean>>;
 };
 
 export const useTimelineContext = (): ContextProps => {
@@ -69,6 +71,9 @@ export const useTimelineContext = (): ContextProps => {
   const [timelineLabelMap, setTimelineLabelMap] = useState<TimelineLabelMap | null>(null);
   const [timelineNotesMap, setTimelineNotesMap] = useState<TimelineNotesMap | null>(null);
   const [refreshTime, setRefreshTime] = useState<Date | null>(null);
+  // Leaflet map encounters an error when prerendered, so we need to render the TimelineScrollList component when the active tab is 'label'
+  // 'shouldUpdateTimeline' gets updated based on the current tab index, and we can use it to determine whether to render the timeline or not
+  const [shouldUpdateTimeline, setShouldUpdateTimeline] = useState(true);
 
   // initialization, once the appConfig is loaded
   useEffect(() => {
@@ -356,6 +361,8 @@ export const useTimelineContext = (): ContextProps => {
     notesFor,
     confirmedModeFor,
     addUserInputToEntry,
+    shouldUpdateTimeline,
+    setShouldUpdateTimeline,
   };
 };
 
--- a/www/js/TimelineContext.ts
+++ b/www/js/TimelineContext.ts
@@ -21,7 +21,6 @@ import {
 } from './diary/timelineHelper';
 import { getPipelineRangeTs } from './services/commHelper';
 import { getNotDeletedCandidates, mapInputsToTimelineEntries } from './survey/inputMatcher';
-import { publish } from './customEventHandler';
 import { EnketoUserInputEntry } from './survey/enketo/enketoHelper';
 import { VehicleIdentity } from './types/appConfigTypes';
 import { primarySectionForTrip } from './diary/diaryHelper';
@@ -140,10 +139,6 @@ export const useTimelineContext = (): ContextProps => {
     );
     setTimelineLabelMap(newTimelineLabelMap);
     setTimelineNotesMap(newTimelineNotesMap);
-    publish('applyLabelTabFilters', {
-      timelineMap,
-      timelineLabelMap: newTimelineLabelMap,
-    });
     setTimelineIsLoading(false);
   }, [timelineMap]);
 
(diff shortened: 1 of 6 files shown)


JGreenlee:
Observed in testing

- Addresses loading correctly (from the bottom / most recent up). Also tested on `stage-timeuse` with places.

- When picking date range from the Dashboard, then switching back to LabelTab, the maps fill in rather than staying blank

- When loading trips with existing labels, "To Label" filter is applied. When switching back and forth between "To Label" and "All Trip", filters get re-applied. When a trip is labeled with both mode and purpose, the trip disappears from "To Label" 30 seconds after its most recent label

www/js/TimelineContext.ts
```diff
@@ -50,6 +50,8 @@ type ContextProps = {
   loadMoreDays: (when: 'past' | 'future', nDays: number) => void;
   loadSpecificWeek: (d: string) => void;
   refreshTimeline: () => void;
+  shouldUpdateTimeline: Boolean;
+  setShouldUpdateTimeline: React.Dispatch<React.SetStateAction<boolean>>;
```
shankari, line 53: For my understanding/edification: Why is only this wrapped in `React.Dispatch` when (say) `setDateRange` is not?

www/js/TimelineContext.ts
```diff
@@ -135,10 +139,6 @@ export const useTimelineContext = (): ContextProps => {
     );
     setTimelineLabelMap(newTimelineLabelMap);
     setTimelineNotesMap(newTimelineNotesMap);
-    publish('applyLabelTabFilters', {
-      timelineMap,
-      timelineLabelMap: newTimelineLabelMap,
-    });
```
shankari, line 141: Agreed that moving this into the `LabelTab` allows us to keep the presentation logic largely in the label which is the right thing to do.

shankari (approved):

shankari:
@JGreenlee one test is failing

shankari:
@JGreenlee it looked like that was a timing issue (you may want to look at the prior run to see what was going on).
It would also be helpful to add tests for the new functionality as part of the cleanup this week (if you have time).

JGreenlee:
I don't see a run where any of the Jest tests failed. All I see is codecov failing because of the token

shankari:
Maybe I don't fully understand what where we were supposed to be with the TimelineContext, but I thought from the code, that we would be passing in entries from the timeline to the metrics in the dashboard. However, even with the changes, I still see that recently labeled trips are not reflected in the dashboard

| Labeled trips from the dashboard | Not showing up in the dashboard | In the most recent version |
|-------|--------|--------|
| simulator_screenshot_34D1F1B0-AA59-4828-A0F1-BE50BDF3ADB6 | simulator_screenshot_A549FFA0-C5D8-4C2E-B6F1-D3A58FD602B9 | simulator_screenshot_5CA32EAE-07E0-4C4E-B895-5A7ADB387E6C |

shankari:
I don't see any obvious errors, though, so pushing out a release to staging anyway

shankari:
> I don't see a run where any of the Jest tests failed. All I see is codecov failing because of the token

It was the second attempt of the run that finally succeeded
https://github.com/e-mission/e-mission-phone/actions/runs/9179292350/attempts/2

shankari:
Update: After restarting the app, the dashboard was updated.

| Before update | Update | After update | After refresh (using button at top right) |
|--------------|-----------|-------------|----------|
| simulator_screenshot_04D09B28-F4E7-4343-9966-F75E416FAA5C | simulator_screenshot_B68E1EA8-22D4-4EAA-B6B6-5D993B969701 | simulator_screenshot_44A9B67A-E0D2-41DA-A050-6DFF218DA89E | simulator_screenshot_CF6AAC19-4696-45FD-839D-F1440EE11A37 |

shankari:
And even after refreshing, the charts seem to be inconsistent with the text in the tables.

| No unlabeled trips for the past week | Error in the dashboard |
|----------|-----------|
| simulator_screenshot_F2193258-3948-43AE-83D3-8D18B5393972 | simulator_screenshot_C7409145-0380-4742-A579-21B8C449E838 | 

| On restart, very little uncertainty | Significant unlabeled distance on the 14th (although no unlabeled trips | Total unlabeled distance is around that amount, but is it all on the 14th? and what what happened to the 5k in air | 
|---------|------|------|
| simulator_screenshot_B83594E1-99E1-4E7C-A204-1BB26059F1FE | simulator_screenshot_6E0E7456-8153-421A-B25C-55AC76C642D5 | simulator_screenshot_50E37679-26EB-47E1-A434-AD0A67439116 |

shankari:
@JGreenlee and @jiji14 for visibility

PR #1160, opened 2024-05-21
✉️📭 Refractored `emailService.ts` into `shareLocalDBFile.ts`

the-bay-kay:
## Purpose
Because of the several document problems with the previous email service, most recent being [Issue 1047](https://github.com/e-mission/e-mission-docs/issues/1074),  I am updating `emailServices.ts` to use the more flexible `socialSharing` plugin that is used in `controlHelper.ts`.  To reflect these updates, the service was renamed, and the English i18next entries updated. 

## Recent Changes:
- Moved into `services/` directory
- Refractored `emailService.ts` to now use sharePlugin, to match controlHelper
- To reflect this change, renamed the service to `shareLocalDBFile`
- Reworked 18next translations to reflect change
- Removed old email plugin from `package-cordovabuild.json`

## Remaining Tests
- [x] Build for iOS
- [x] Build for Android
- [x] Test on iOS Emulation
- [x] Test on iOS Hardware
- [x] Test on Android Hardware


the-bay-kay:
### Testing Updates
When running on both iOS and Android, the share function works, with a notable exception. 

Below are two screenshots: one sending logs via the Apple Mail app, and the second via GMail.  As can be seen, the `loggerDB` file cannot be attatched to a GMail message.  This error occurs on both iOS and Android.  We can, however, send the file through several other means - Google Drive, Apple Mail, etc.

My first theory as to why this occurs:  Gmail prohibits the sending of certain files ([link](https://support.google.com/mail/answer/6584?hl=en&co=GENIE.Platform%3DiOS#zippy=%2Cattachments-might-be-unavailable%2Cblocked-for-security-reasons)).  Because the `loggerDB` file has no extension, Gmail "plays it safe", and does not allow us to attach it.  

What is strange, however, is that _this error does not occur with the old email plugin_.  Why would Gmail prohibit the attachment in one circumstance, and not another?  From what I can gather from the working sharing (Google Drive, Apple Mail), the file itself is formed and sent correctly... I'll do a bit more digging

the-bay-kay:
## Quick Updates:
- It seems that this latest changed failed the prettier action, but running `npx prettier --write www/` does not change anything for me... Checking my prettier version to make sure I'm on the right version
- Because of the email issues described above, I reverted the android portion of this service to the older plugin.  Gmail now works as intended on Android (i.e., works as it did before)

shankari:
@the-bay-kay is a workaround to simply rename the file as .pdf before trying to attach it

the-bay-kay:
Really wrestling with the Cordova file plugin right now, stumped as to what's causing this issue.

When running the existing code:
```ts
if (window['cordova'].platformId == 'android') {
    filePath = 'app://databases/' + database;
  } else if (window['cordova'].platformId == 'ios') {
    logDebug(window['cordova'].file.dataDirectory);
    filePath = window['cordova'].file.dataDirectory + '../LocalDatabase/' + database;
  } else { // == 'unknown'
    displayErrorMsg('Error: Unknown OS!');
    return new Error('Unknown OS!');
  }


  logDebug(`FilePath is: ${filePath}`)

  const shareObj = {
    files: [filePath],
    message: i18next.t('shareFile-service.send-log.body-please-fill-in-what-is-wrong'),
    subject: i18next.t('shareFile-service.send-log.subject-logs'),
  };
```

The file path is printed as follows:
```
DEBUG: FIlePath is: file:///Users/{user}/Library/Developer/CoreSimulator/Devices/{ID}/data/Containers/Data/Application/{ID}/Library/NoCloud/../LocalDatabase/loggerDB

```

When attempting to access this URL the same way, using the [Cordova File Plugin](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-file/index.html):
```ts
async function localReadFile() {
    return new Promise<void>((resolve, reject) => {
      let pathToFile;
      if (window['cordova'].platformId == 'android') {
        pathToFile = 'app://databases/' + fileName;
      } else if (window['cordova'].platformId == 'ios') {
        pathToFile  = window['cordova'].file.dataDirectory + '../LocalDatabase/' + fileName;
        logDebug(`PathToFile: ${pathToFile}`);
      } else { // == 'unknown'
        displayErrorMsg('Error: Unknown OS!');
        reject();
      }
      window['resolveLocalFileSystemURL'](pathToFile, (fs) => {
        fs.filesystem.root.getFile(fileName, { create: false, exclusive: false }, (fileEntry) => {
          logDebug(`fileEntry ${fileEntry.nativeURL} is file? ${fileEntry.isFile.toString()}`);

          fileEntry.file(
            (file) => {
              const reader = new FileReader();
              logDebug(`fileEntry ${JSON.stringify(file, null, 2)}`);

              reader.onloadend = () => {
                const readResult = this.result;
                logDebug(`File Contents: ${JSON.stringify(readResult, null, 2)} `)
              };
              reader.readAsText(file);
              // The code continues...
```

We get the following output:
```
DEBUG:fileEntry file:///Users/{user}/Library/Developer/CoreSimulator/Devices/{ID}/data/Containers/Data/Application/{ID}/Library/loggerDB is file? true
```

It seems the relative path we're constructing is _not getting resolved_: 
-  `.../Application/{ID}/Library/NoCloud/../LocalDatabase/loggerDB`
- `.../Application/{ID}/Library/loggerDB`

Why would this happen? Trying to go through the plugin docs for an answer...

the-bay-kay:
I think there are some obvious holes in my logic here -- when this code executes:
```ts
      window['resolveLocalFileSystemURL'](pathToFile, (fs) => {
        fs.filesystem.root.getFile(fileName, { create: false, exclusive: false }, (fileEntry) => {
          logDebug(`fileEntry ${fileEntry.nativeURL} is file? ${fileEntry.isFile.toString()}`);
```

The "fs.filesystem.root" is set to the path of the file itself -- It'd be more sensical to instead point somewhere a bit higher, and then use the file name to navigate down.  This still doesn't solve answer (i) why we can't set the root using relative paths, and (ii) how we can use the cordova builtin directories to navigate to the file.  We need to set the "root" one above `cordova['window'].file.dataDirectory` -- how can we do so if we cannot navigate via a relative path?  Need to do more reading.

the-bay-kay:
Attempting a different approach, utilizing the Cordova File Plugin's `copyTo`.  Code is included below the cut: When running the code, I get the subsequent error...

<details>
<summary> TS For File Copy </summary>

```ts
  async function localCopyFile() {
    return new Promise<void>((resolve, reject) => {
      let pathToFile, parentDirectory;
      if (window['cordova'].platformId == 'android') {
        parentDirectory = window['cordova'].file.dataDirectory;
        pathToFile = fileName;
      } else if (window['cordova'].platformId == 'ios') {
        parentDirectory = window['cordova'].file.dataDirectory + '../';
        pathToFile = 'LocalDatabase/' + fileName;
      } else { // == 'unknown'
        displayErrorMsg('Error: Unknown OS!');
        throw new Error('Unknown OS!');
      }

      window['resolveLocalFileSystemURL'](parentDirectory, (fs) => {
        fs.filesystem.root.getFile(pathToFile, { create: false, exclusive: false }, (fileEntry) => {
          logDebug(`fileEntry ${fileEntry.nativeURL} is file? ${fileEntry.isFile.toString()}`);
          
          window['resolveLocalFileSystemURL'](window['cordova'].file.cacheDirectory, (copyDir) => {
            logDebug(`DirObject?: ${JSON.stringify(copyDir.filesystem.root, null, 2)}`)

            fileEntry.copyTo(copyDir.filesystem.root, fileName + '.txt', (res) => {
              logDebug(`Res: ${res}`)
            }, (rej)=> {
              logDebug(`Rej: ${JSON.stringify(rej, null, 2)}`)
            })
          });
          // Code follows...
```

</details>

```
[console.log] DEBUG:fileEntry file:///Users/{USER}/Library/Developer/CoreSimulator/Devices/{ID}/data/Containers/Data/Application/{ID}/Library/LocalDatabase/loggerDB is file? true
[console.log] DEBUG:DirObject?: {
  "isFile": false,
  "isDirectory": true,
  "name": "/",
  "fullPath": "/",
  "filesystem": "<FileSystem: cache>",
  "nativeURL": "file:///Users/{User}/Library/Developer/CoreSimulator/Devices/{ID}/data/Containers/Data/Application/{ID}/Library/Caches/"
}
[console.log] DEBUG:Rej: {
  "code": 12
}
```

I have confirmed that
-  The FileEntry is constructed correctly (file to copy)
- The DirectoryEntry is constructed correctly (copy destination)
- The name is a string, and both success and failure callbacks are being passed in

So why is this failing? The source code for the `copyTo` method can be found [here](https://github.com/apache/cordova-plugin-file/blob/03a4f4cb53e54de31344851cc5d01b5da96aa0d7/www/Entry.js#L148) -- I'm stumped as to why this error is getting thrown, will return to the problem after lunch.

the-bay-kay:
Bingo -- [here](https://github.com/apache/cordova-plugin-file/blob/03a4f4cb53e54de31344851cc5d01b5da96aa0d7/types/index.d.ts#L100) is my problem.
> A copy of a file on top of an existing file must attempt to delete and replace that file.

Manually navigating to this folder and deleting the file results in a successful resolution. Woot! Now to test the share functionality in combination with this.

the-bay-kay:
Finally!  Seems to be completely functional in the simulator.  Going to run builds for iOS & Android -- if tests go OK, will move to R4R.

https://github.com/e-mission/e-mission-phone/assets/98350084/ae92dda6-9367-4178-ba69-9bef871aad71

the-bay-kay:
Good news: This is now working 100% on iOS!  Adding the `.txt` extension solved our issues with GMail.

Bad news: As written, this PR _does not work on Android_.  I'm almost certain this is due to the way the directory has been set up -- I think I made a mistake in the initial file path.  Investigating now

the-bay-kay:
>  I'm almost certain this is due to the way the directory has been set up -- I think I made a mistake in the initial file path. Investigating now

Well, that wasn't the easy fix I thought it'd be -- time to break out my PC and run with `.adb` to see what's going on...

louisg1337:
## Android Update

I was thankfully able to fix a few bugs on the Android side of things which, in conjunction with the rest of @the-bay-kay's work, now makes it function as expected! I ran into a few issues, which I'll touch on below, along with a caveat to my solution that I wanted to mention. PR can be found [here](https://github.com/the-bay-kay/e-mission-phone/pull/2).

### Issues

1. Android `parentDirectory` was _slightly_ off

   The `parentDirectory` ended up being `file:///data/user/0/edu.berkeley.eecs.emission/databases/`, which we can represent as `window['cordova'].file.dataDirectory.replace('files', 'databases')`. To find this path, I learned that you can shell into a hardware device using `adb shell`, then "cd" into your app using `run-as package-name`, and from there you have access to all of the apps files. 

2. In the `localCopyFile()` function, `fs.filesystem.root.getFile()` could not find the `loggerDB` file

    Once I found the right directory we should be in, I was now getting a file not found error from `getFile()`. This was really strange because I was physically in the device and could confirm that `loggerDB` was in that directory. I then thought that maybe `window['resolveLocalFileSystemURL'](parentDirectory, (fs => {}))` was returning the wrong directory, but I took `fs` and created a [`DirectoryReader`](https://github.com/apache/cordova-plugin-file/blob/03a4f4cb53e54de31344851cc5d01b5da96aa0d7/types/index.d.ts#L216) object, which let me scan the directory, and it returned `loggerDB` as one of the files in it. I did notice in the [docs](https://github.com/apache/cordova-plugin-file/blob/03a4f4cb53e54de31344851cc5d01b5da96aa0d7/types/index.d.ts#L155) for `getFile()` that we can either pass in a relative path, which we do now, or an absolute path. When I pass in an absolute path, `getFile()` is able to find `loggerDB` and the rest of the functionality just works.

It is a bit annoying how I wasn't able to figure out why `getFile()` was not working with a relative path, but I am convinced it could _maybe_ be an issue with the Android code and I don't know if its worth our time to debug, especially if we have the fix above. This solution also circumvents the original implementation a bit as we don't need to access the `parentDirectory` anymore on Android, but I don't think its too big of an issue.

the-bay-kay:
@louisg1337 Looks good to me! I'd maybe add a video showing the fix works, just so Shankari can merge this quick.  Thanks again for finishing this off -- I had issues on iOS with relative file paths too, seems to be an issue when working with this plugin.  

Before I forget -- I'll quickly update the JSON Dump feature, to fix it alongside this (no need for an extra PR, should be a < 2 line change!)

louisg1337:
Thanks for reviewing it for me @the-bay-kay! As request, here is the video of it work on Android...


https://github.com/e-mission/e-mission-phone/assets/79764532/6392b640-b47e-48f4-bc1c-75a90b9302d5

the-bay-kay:
@shankari, As written, this PR is finished and fixes the log functionality.  Louis and I discussed fixing the JSON dump feature -- from an initial investigation, it appears that is broken for a different reason.  

My question is: would you prefer we take the time and include the JSON Dump fix in this PR, or should we open a new PR for that?  I would lean toward the latter, as this is a bottleneck for the iOS fixes (and is ready for review otherwise)

shankari:
new PR. Also, it looks like we didn't add tests for the new code, so the code coverage check is failing...

www/i18n/en.json
```diff
@@ -278,16 +278,14 @@
     "reading-unprocessed-data": "Reading unprocessed data..."
   },
 
-  "email-service": {
-    "email-account-not-configured": "Email account is not configured, cannot send email",
-    "email-account-mail-app": "You must have the mail app on your phone configured with an email address. Otherwise, this won't work",
-    "going-to-email": "Going to email database from {{parentDir}}",
-    "email-log": {
+  "shareFile-service": {
+    "send-to": "Please send this file to k.shankari@nrel.gov, or another OpenPATH team member.",
```
shankari, line 282:
```suggestion
    "send-to": "Please send this file to openpath@nrel.gov, or another OpenPATH team member.",
```

www/i18n/en.json
shankari, line 1: note that while making changes to this file, you also need to update the translations in https://github.com/e-mission/e-mission-translate

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,116 @@
+import i18next from 'i18next';
+import { displayError, displayErrorMsg, logDebug, logWarn } from '../plugin/logger';
+
+function localDBHelpers(fileName: string, fileExtension: string = '.txt') {
+  async function localCopyFile() {
+    return new Promise<void>((resolve, reject) => {
+      let pathToFile, parentDirectory;
+      if (window['cordova'].platformId == 'android') {
+        parentDirectory = window['cordova'].file.dataDirectory.replace('files', 'databases');
```
shankari, line 9: Can you please add in examples of what this manipulation looks like, either as comments here, or in a comment in the PR, which you add a link as a comment here? I see that @the-bay-kay had several comments around why this wasn't working and I want to make sure that future developers are able to see what is supposed to happen.

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,116 @@
...
+        pathToFile = parentDirectory.replace('file://', '') + fileName;
+      } else if (window['cordova'].platformId == 'ios') {
+        parentDirectory = window['cordova'].file.dataDirectory + '../';
```
shankari, line 12: Ditto

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,116 @@
...
+        pathToFile = 'LocalDatabase/' + fileName;
+      } else {
+        displayErrorMsg('Error: Unknown OS!');
+        throw new Error('Error: Unknown OS!');
+      }
+
+      window['resolveLocalFileSystemURL'](parentDirectory, (fs) => {
+        fs.filesystem.root.getFile(pathToFile, { create: false, exclusive: false }, (fileEntry) => {
+          // logDebug(`fileEntry ${fileEntry.nativeURL} is file? ${fileEntry.isFile.toString()}`);
+          logDebug(`fileEntry is: ${JSON.stringify(fileEntry, null, 2)}`);
```
shankari, line 22: Ditto

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,116 @@
...
+
+            fileEntry.copyTo(
+              copyDir.filesystem.root,
+              fileName + fileExtension,
+              (res) => {
+                logDebug(`Res: ${res}`);
+                resolve();
+              },
+              (rej) => {
+                logDebug(`Rej: ${JSON.stringify(rej, null, 2)}`);
```
shankari, line 34: instead of `logDebug`, where the error gets put into the log file which the user can't email because of the error, please use `displayError` so that they can at least send us a screenshot.

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,116 @@
...
+            };
+            window['plugins'].socialsharing.shareWithOptions(
+              shareObj,
+              (result) => {
+                logDebug(`Share Completed? ${result.completed}`); // On Android, most likely returns false
+                logDebug(`Shared to app:  ${result.app}`);
+                resolve();
+              },
+              (msg) => {
+                logDebug(`Sharing failed with message ${msg}`);
```
shankari, line 64: Ditto

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,116 @@
...
+  };
+}
+export async function sendLocalDBFile(database: string) {
+  alert(i18next.t('shareFile-service.send-to'));
+
+  const dataMethods = localDBHelpers(database);
+  dataMethods
+    .copyFile()
+    .then(dataMethods.shareData)
+    .then(dataMethods.clearData)
```
shankari, line 109: why are we clearing the data here after sharing? That is not what we do now. If we clear the data after sharing, then sharing the userCacheDB will result in collected and locally cached data getting deleted. Similarly, logs will be deleted after sharing and will not wait until they age out

shankari (changes requested):
@louisg1337 @the-bay-kay please fix these issues ASAP so that I can merge before the weekend

www/i18n/en.json
louisg1337: Updated all of the translations in my PR [here](https://github.com/e-mission/e-mission-translate/pull/55).

www/js/services/shareLocalDBFile.ts
the-bay-kay: TL;DR: because we need to copy the file to add the `.txt` extension, we need to delete the duplicate file.

---

Full Explanation: Because we can't rename the file, we instead need to copy it (adding the `.txt`), and share that new file.  Per the plugin documentation ([link](https://github.com/apache/cordova-plugin-file/blob/03a4f4cb53e54de31344851cc5d01b5da96aa0d7/types/index.d.ts#L100))...
>  A copy of a file on top of an existing file must attempt to delete and replace that file.

This was an issue I ran into during development (see [this](https://github.com/e-mission/e-mission-phone/pull/1160#issuecomment-2192454813) comment) - if we do not clear space for this temporary file, the copy function will fail.  

> Similarly, logs will be deleted after sharing and will not wait until they age out

If I understand your concern correctly, this shouldn't be an issue.  We never delete the "main" logs -- we create a copy at the time of sharing, and then clear that copy.  (FWIW, Louis tested sending the logs twice in a row, and we had no issue with old logs being cleared before aging out).  

Perhaps the `clearData` function could be renamed to better explain its purpose, `clearTempFile`, etc.

shankari: Yes, please rename to clarify

www/js/services/shareLocalDBFile.ts
```diff
@@ -0,0 +1,121 @@
...
+            window['plugins'].socialsharing.shareWithOptions(
+              shareObj,
+              (result) => {
+                logDebug(`Share Completed? ${result.completed}`); // On Android, most likely returns false
+                logDebug(`Shared to app:  ${result.app}`);
+                resolve();
+              },
+              (msg) => {
+                displayErrorMsg(`Sharing failed with message ${msg}`);
+              },
```
shankari, line 70: @the-bay-kay @louisg1337 this should give you much better formatting

```suggestion
              (error) => {
                displayError(error, `Sharing failed with error`);
              },
```

shankari (approved):

shankari (approved):

PR #1162, opened 2024-06-07
fix timeline initial date range regression

JGreenlee:
When refactoring into TimelineContext, there was a regression regarding the date range that is initially loaded. https://github.com/e-mission/e-mission-phone/pull/1142#issuecomment-2141345034

It used to load 7 days before the pipelineEnd up to today. After the change, it would load 7 days before today up to today. If the user has travel everyday this ends up being equivalent. But if there isn't recent travel, this becomes a problem.

So in TimelineContext, the initial dateRange needs to depend on pipelineRange. instead of giving an initial value upfront, we'll let dateRange be null until pipelineRange is set, at which point we'll also set dateRange. This means we have to consider some extra cases where dateRange can be null.

### DIFF
--- a/www/js/TimelineContext.ts
+++ b/www/js/TimelineContext.ts
@@ -25,9 +25,7 @@ import { EnketoUserInputEntry } from './survey/enketo/enketoHelper';
 import { VehicleIdentity } from './types/appConfigTypes';
 import { primarySectionForTrip } from './diary/diaryHelper';
 
-// initial query range is the past 7 days, including today
-const today = DateTime.now().toISODate().substring(0, 10);
-const initialQueryRange: [string, string] = [isoDateWithOffset(today, -6), today];
+const TODAY_DATE = DateTime.now().toISODate();
 
 type ContextProps = {
   labelOptions: LabelOptions | null;
@@ -43,7 +41,7 @@ type ContextProps = {
   addUserInputToEntry: (oid: string, userInput: any, inputType: 'label' | 'note') => void;
   pipelineRange: TimestampRange | null;
   queriedDateRange: [string, string] | null; // YYYY-MM-DD format
-  dateRange: [string, string]; // YYYY-MM-DD format
+  dateRange: [string, string] | null; // YYYY-MM-DD format
   timelineIsLoading: string | false;
   loadMoreDays: (when: 'past' | 'future', nDays: number) => boolean | void;
   loadDateRange: (d: [string, string]) => boolean | void;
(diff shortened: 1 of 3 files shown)


shankari (approved):

shankari:
After this fix, things look at lot better. The regressions from https://github.com/e-mission/e-mission-phone/pull/1142 seem to be fixed, there are no errors with aggregate trips. I still don't see the group bar and now I don't see the "worst case". The lack of group bar may just because we don't have any other data for that week, but I don't understand why the worst case is not showing up.

@JGreenlee last fix (if needed!) and then we are ready to move this to production!

shankari:
Note also that bikeshare doesn't seem to count against active minutes

| Walk and bikeshare | Walk and bike | Now bike shows up | In the per-day as well |
|------------|----------|------------|-------------|
| Simulator Screenshot - iphone 17 Pro - 2024-07-09 at 14 39 35 | Simulator Screenshot - iphone 17 Pro - 2024-07-09 at 14 40 54 |. Simulator Screenshot - iphone 17 Pro - 2024-07-09 at 14 40 58 | Simulator Screenshot - iphone 17 Pro - 2024-07-09 at 14 41 05 |

shankari:
I checked this on my personal phone, and I see the worst case, but not the group average. I also got a timeout on retrieving the aggregate value the first time I went to the dashboard. @JGreenlee

shankari:
@JGreenlee example of dashboard (intermittently) not updating

| Active modes filled in | Not showing in graph | or in raw data | 
|----------------|-------------|----------|
| Screenshot_2024-07-10-10-09-32-19_5062a5234f99d532322a40d030e8949e | Screenshot_2024-07-10-10-10-02-98_5062a5234f99d532322a40d030e8949e | Screenshot_2024-07-10-10-10-21-93_5062a5234f99d532322a40d030e8949e |

shankari:
My personal phone is on nrel_commute, so I also upgraded it and then checked the logs. We do get the correct `yyyy_mm_dd` and it returns 242 results. But I still don't see the comparison in the UI

```
2024-07-11 05:05:22,499:DEBUG:140319168288320:START POST /result/metrics/yyyy_mm_dd
2024-07-11 05:05:22,499:DEBUG:140319168288320:Aggregate call, checking user_only policy
...
2024-07-11 05:05:22,502:DEBUG:140319168288320:metric_list = {'distance': ['mode_confirm'], 'duration': ['mode_confirm'], 'count': ['mode_confirm']}
2024-07-11 05:05:22,502:DEBUG:140319168288320:for user None, returning timeseries <emission.storage.timeseries.aggregate_timeseries.AggregateTimeSeries object at 0x7f9e835a27f0>
2024-07-11 05:05:22,502:DEBUG:140319168288320:curr_query = {'invalid': {'$exists': False}, '$or': [{'metadata.key': 'analysis/composite_trip'}], 'data.start_fmt_time': {'$lte': '2024-07-10Z', '$gte': '2024-06-26'}}, sort_key = None
2024-07-11 05:05:22,503:DEBUG:140319168288320:orig_ts_db_keys = [], analysis_ts_db_keys = ['analysis/composite_trip']
2024-07-11 05:05:22,503:DEBUG:140319168288320:finished querying values for [], count = 0
2024-07-11 05:06:19,580:DEBUG:140319168288320:finished querying values for ['analysis/composite_trip'], count = 242
2024-07-11 05:06:19,580:DEBUG:140319168288320:orig_ts_db_matches = 0, analysis_ts_db_matches = 242
2024-07-11 05:06:20,806:DEBUG:140319168288320:Returning entry with length 242 result
2024-07-11 05:06:20,854:DEBUG:140319168288320:END POST /result/metrics/yyyy_mm_dd ... 58.35479283332825 
```

PR #1164, opened 2024-07-11
Integrate unified base mode color and helper function

louisg1337:
Integrated `e-mission-common` [PR](https://github.com/JGreenlee/e-mission-common/pull/1) which unifies the base mode colors and a helper function. Went through the code base and swapped all instances of `getBaseModeByKey()` with `base_mode_colors.get_base_mode_by_key()`. I also converted `modeColors` and `BaseModes` in `diaryHelper.ts` to now use our unified functions.  In `diaryHelper.ts` I also had to add in types for one of the functions as I was getting a typescript "type unknown of distA/distB" error. I guess when I removed `getBaseModeByKey()`, the type inference couldn't figure out that they were numbers so it was throwing that error.

**Testing Done**

I went through each component that I converted `getBaseModeByKey()` for `base_mode_colors.get_base_mode_by_key()` and tested it out to make sure it still worked properly.

### DIFF
--- a/www/js/diary/details/TripSectionsDescriptives.tsx
+++ b/www/js/diary/details/TripSectionsDescriptives.tsx
@@ -2,7 +2,6 @@ import React, { useContext } from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Icon, Text, useTheme } from 'react-native-paper';
 import useDerivedProperties from '../useDerivedProperties';
-import { getBaseModeByValue } from '../diaryHelper';
 import TimelineContext from '../../TimelineContext';
 import { base_modes } from 'e-mission-common';
 
--- a/www/js/diary/details/TripSectionsDescriptives.tsx
+++ b/www/js/diary/details/TripSectionsDescriptives.tsx
@@ -4,7 +4,7 @@ import { Icon, Text, useTheme } from 'react-native-paper';
 import useDerivedProperties from '../useDerivedProperties';
 import { getBaseModeByValue } from '../diaryHelper';
 import TimelineContext from '../../TimelineContext';
-import { base_mode_colors } from 'e-mission-common';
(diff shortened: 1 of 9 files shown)


JGreenlee:
@louisg1337

I made some changes in e-mission-common, can you update this PR?
- bump the version of e-mission-common that is used to [0.5.4](https://github.com/JGreenlee/e-mission-common/releases/tag/0.5.4)
- fix the imports that will be broken because I renamed `base_mode_colors` to `base_modes`

www/js/diary/details/TripSectionsDescriptives.tsx
```diff
@@ -2,7 +2,6 @@ import React, { useContext } from 'react';
 import { View, StyleSheet } from 'react-native';
 import { Icon, Text, useTheme } from 'react-native-paper';
 import useDerivedProperties from '../useDerivedProperties';
-import { getBaseModeByValue } from '../diaryHelper';
```
shankari: This function still exists in the diaryHelper, right? Are you removing the usages now so that you can remove it from the helper later?

shankari (changes requested):
Can you also include some screenshots to show that there are no regressions before I merge?

www/js/diary/details/TripSectionsDescriptives.tsx
JGreenlee: Yes, that's what I was thinking – but I now see that there were no usages left so I have removed it altogether

JGreenlee:
`<ModesIndicator>` is showing the correct colors and icons of base modes



<hr>

Also seen on LabelDetailsScreen with `<TripSectionsDescriptives>` showing colors & icons of modes on multimodal trip

shankari (approved):

PR #1165, opened 2024-08-16
Tidy excessive logs

JGreenlee:
Logs have gotten very cluttered, which is causing issues with the size of the LoggerDB, and is likely inhibiting performance

I am removing logs that happen repeatedly; e.g.
- logs on every single function call on a function that is called dozens / hundreds of times 
- logs on every UI/component re-render
- redundant logs that convey information that is already being logged by another statement

I am being careful not to remove logs that would be useful for debugging, since logs are critical to ensuring we can resolve bugs affecting end users.


shankari:
@JGreenlee should I go ahead and merge this along with this weekend's changes?

JGreenlee:
Yes. There might be more log statements to clean up later, but this is what I had time for this week

PR #1166, opened 2024-08-19
📍 Use the server-side geocoded addresses for place names

JGreenlee:
Since the server now fills in geocoded_address (we may later change this to reverse_geocode), we will only need to perform a client-side nominatim request for unprocessed trips. If either field is present, we'll use that; otherwise schedule a request like before.

Updated types:

-Since the CompositeTrip type definition is based on the return type of Timeline.readAllCompositeTrips, which is after "unpacking", `start_confirmed_place` and `end_confirmed_place` are simply `ConfirmedPlace`, not `BEMData<ConfirmedPlace>`
-ConfirmedPlace.location is a `Point`, (having `coordinates`) not just a generic `Geometry`
-added `geocoded_address` and `reverse_geocode`, both optional since it depends on https://github.com/e-mission/e-mission-server/pull/973

### DIFF
--- a/www/js/diary/addressNamesHelper.ts
+++ b/www/js/diary/addressNamesHelper.ts
@@ -77,6 +77,23 @@ import Bottleneck from 'bottleneck';
 import { displayError, logDebug } from '../plugin/logger';
 import { CompositeTrip } from '../types/diaryTypes';
 
+export type NominatimResponse = {
+  address: {
+    road?: string;
+    pedestrian?: string;
+    suburb?: string;
+    neighbourhood?: string;
+    hamlet?: string;
+    city?: string;
+    town?: string;
+    county?: string;
+    state?: string;
+    postcode?: string;
+    country?: string;
+    country_code?: string;
+  };
+};
+
 let nominatimLimiter = new Bottleneck({ maxConcurrent: 2, minTime: 500 });
 export const resetNominatimLimiter = () => {
   const newLimiter = new Bottleneck({ maxConcurrent: 2, minTime: 500 });
@@ -128,7 +145,7 @@ async function fetchNominatimLocName(loc_geojson) {
       status = ${response.status}; 
       data = ${JSON.stringify(data)}`);
     localStorage.setItem(coordsStr, JSON.stringify(data));
-    publish(coordsStr, data);
+    publish(coordsStr, JSON.stringify(data));
   } catch (error) {
     if (!nominatimError) {
       nominatimError = error;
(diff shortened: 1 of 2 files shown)


JGreenlee:
This PR works with either `geocoded_address` or `reverse_geojson`.
If this is merged before https://github.com/e-mission/e-mission-server/pull/973 is merged, it will switch over seamlessly.

If https://github.com/e-mission/e-mission-server/pull/973 is merged first, I can update this PR to only use `reverse_geojson`

shankari:
I can merge https://github.com/e-mission/e-mission-server/pull/973 soon. I am just waiting to merge the automatic build PRs (e.g. https://github.com/e-mission/op-admin-dashboard/pull/113) so that I could see it work! All the public PRs are merged, the internal ones need to be fixed @mukuflash03

shankari:
Note that even after https://github.com/e-mission/e-mission-server/pull/973 is merged, the older trips will still have only the `geocoded_address` field. You need to update this PR to check for `reverse_geocode` and then for `geocoded_address` and then fall back to looking it up.

We should then also write a migration script to copy over all of the `geocoded_address` into `reverse_geocode` so that we have all the info we need for the footprint calculations. It may be a bit tricky because the `geocoded_address` doesn't actually have all the info we need for the carbon footprint calculations, and so we probably have to look it up again. Hopefully we get the same result as before, but if we don't, we have to likely use the new version. I don't think it will affect the calculation outputs because the census tract is unlikely to have changed in the interim.

JGreenlee:
> You need to update this PR to check for reverse_geocode and then for geocoded_address and then fall back to looking it up.

That is what this PR already does because I anticipated that we may need to support both fields.


> We should then also write a migration script to copy over all of the geocoded_address into reverse_geocode so that we have all the info we need for the footprint calculations.

Actually, this no longer has anything to do with the footprint calculations. I wanted to use the zip codes, but they aren't reliable enough. Sometimes they are missing from the response, and sometimes there isn't a clear mapping to the urban area or eGRID region.

I do think we should write the migration script, but it isn't blocking footprint calculations if we are ok with having both fields temporarily.

shankari:
Ah I am clearly not on top of the changes to the carbon footprint calculations.

www/js/diary/addressNamesHelper.ts
```diff
@@ -139,8 +156,16 @@ async function fetchNominatimLocName(loc_geojson) {
...
 // Schedules nominatim fetches for the start and end locations of a trip
 export function fillLocationNamesOfTrip(trip: CompositeTrip) {
-  nominatimLimiter.schedule(() => fetchNominatimLocName(trip.end_loc));
-  nominatimLimiter.schedule(() => fetchNominatimLocName(trip.start_loc));
+  [trip.start_confirmed_place, trip.end_confirmed_place].forEach((place) => {
+    if (place?.geocoded_address || place?.reverse_geocode) {
+      const coordsStr = place.location.coordinates.toString();
+      const data = place.reverse_geocode || { address: place.geocoded_address };
+      localStorage.setItem(coordsStr, JSON.stringify(data));
+      publish(coordsStr, JSON.stringify(data));
```
shankari, line 164: nice! great idea to reuse the data from prior mappings even for trips that don't have the address instead of looking it up.

www/js/types/diaryTypes.ts
```diff
@@ -96,7 +102,7 @@ export type CompositeTrip = {
   confirmed_trip: ObjectId;
   distance: number;
   duration: number;
-  end_confirmed_place: BEMData<ConfirmedPlace>;
+  end_confirmed_place: ConfirmedPlace;
```
shankari, line 105: Curious about what this change does. Why did we have BEMData before and why don't we need it now?

shankari (approved):
My comments are just requests for clarification to help me better understand this.

www/js/types/diaryTypes.ts
JGreenlee: It was just incorrect. I noticed because Typescript complained when I accessed properties on the place objects while making the changes

> Updated types:
> 
> -Since the CompositeTrip type definition is based on the return type of Timeline.readAllCompositeTrips, which is after "unpacking", start_confirmed_place and end_confirmed_place are simply ConfirmedPlace, not BEMData<ConfirmedPlace>
> -ConfirmedPlace.location is a `Point`, (having `coordinates`) not just a generic `Geometry`
> -added geocoded_address and reverse_geocode, both as optional since it depends on e-mission/e-mission-server#973

PR #1169, opened 2024-08-28
⚛️ Switch to Expo Web

JGreenlee:
Switches the project from React Native Web (using Webpack bundler) to Expo Web (using Metro bundler) and also introduces a minimal web version of the app, which can be used as a dev tool to make UI changes easier (the devapp is not reliable anymore as the cordova/phonegap toolset is too outdated)

🔄 Remove global SCSS; use scoped CSS for leaflet and enketo components

🔄 Use directly imported JSON files for local demo surveys

> We have two versions of the demographics survey kept locally in e-mission-phone.
> Statically importing them with ES import syntax instead of using fetch() will make our lives
> easier wrt bundler behavior

🤡 Expand Cordova mocks

> Gradually, we have filled in more of the Cordova plugins' behavior with mocks for testing.
> This extends the functionality to mock BEMConnectionSettings and BEMServerComm
> Most importantly for the current task, this mocks pushGetJSON with fetch, which allows a
> minimal web version of the app to function without the native code

✨ Add Expo to the project

> Switches the project from React Native web project using Webpack to an Expo web project using Metro bundler
> 
> We will now use Metro, the standard bundler for React Native / Expo projects, instead of webpack. Metro is less flexible but more tailored to the use case of a react native / expo project, and switching to it will make our lives easier for future native code migration.
> 
> It also allows us to add a small compatibility layer (expoCompat.ts) which will allow us to run a minimal version of the app in a browser with our native code mocks. If nothing else, this will be useful as a dev tool
> 
> Cordova assumes that index.html and its required resources will exist in www/.
> By default, Expo export generates the index.html and bundled resources and outputs to dist/.
> We can change the output dir to www/, but we should keep our source files separate, so I have moved them to src/.
> So www/ now functions as an "output" or "dist" directory to which Metro exports, and from which Cordova builds.
> public/index.html is the "template" that the Expo export uses for www/index.html which ensures that the output index.html actually has the cordova.js included
> 
> We will no longer have a "serve" configuration, so the cordovabuild config.xml is now the de facto config.xml.

⬆️ Upgrade React/RN packages and remove unneeded packages

🗑️️ emove unneeded config files and scripts

> import label-options.default.json
> 
> instead of using fetch() we can use ES import here. Need to rename to *.json, so calling it label-options.default.json

🧪 Fix tests under new project structure

🌐 Adjust CSP

👷 Refine setup/teardown scripts and workflows

🎨 Upgrade prettier; fix formatting


JGreenlee:
Screenshot 2024-08-27 at 10 52 25 PM

JGreenlee:
@copilot didn't leave any comments, but it looks like it meant to. I grabbed these from the session:

> {
>   "comment_content": "The current Promise chaining calls `login(...)` immediately and passes a Promise into `markConsented().then(...)` instead of a callback. This means consent may not be recorded before login/registerUser runs, and the `.then(...)` handler on `markConsented` is effectively ignored. Refactor this effect to await `markConsented()` and then call/await `login(...)` (and have `login` return the `registerUser()` promise) so sequencing and error handling are correct.",
>   "comment_type": "bug",
>   "corrected_text": "",
>   "end_line": 33,
>   "file_location": "src/js/onboarding/SaveQrPage.tsx",
>   "fixed": false,
>   "original_text": "",
>   "severity": "critical",
>   "start_line": 22
> }
> Stored bug comment [ID: 001] for lines 22-33 in src/js/onboarding/SaveQrPage.tsx
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "`scopedEval` uses the `Function` constructor to evaluate `survey.showsIf`. Since `showsIf` comes from deployment config, this creates an arbitrary code execution surface in the webview (and can access the provided scope and globals). Replace this with a safe expression evaluator (e.g., a small parser/whitelisted operators/functions such as JSONLogic, jsep+custom interpreter, expr-eval with disabled member access) and validate/limit what the condition can do.",
>   "comment_type": "security",
>   "corrected_text": "",
>   "end_line": 29,
>   "file_location": "src/js/survey/enketo/conditionalSurveys.ts",
>   "fixed": false,
>   "original_text": "",
>   "severity": "critical",
>   "start_line": 23
> }
> Stored security comment [ID: 002] for lines 23-29 in src/js/survey/enketo/conditionalSurveys.ts
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "`addStatReading()`/`addStatError()` call `displayErrorMsg()` when `BEMUserCache` is missing. But `displayErrorMsg()` always calls `addStatError()`, so if the DB isn't available this can recurse indefinitely and crash (stack overflow). Make the fallback path DB-independent (e.g., log to console and return) and/or guard `displayErrorMsg` so it doesn't attempt to record stats when stats storage isn't available.",
>   "comment_type": "bug",
>   "corrected_text": "",
>   "end_line": 61,
>   "file_location": "src/js/plugin/clientStats.ts",
>   "fixed": false,
>   "original_text": "",
>   "severity": "critical",
>   "start_line": 47
> }
> Stored bug comment [ID: 003] for lines 47-61 in src/js/plugin/clientStats.ts
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "`displayErrorMsg()` unconditionally calls `addStatError()`. If stats storage isn't available yet, `addStatError()` currently falls back to `displayErrorMsg()`, which can lead to infinite recursion/crash. Add a guard to avoid calling `addStatError` from within error display when stats are unavailable (or allow `addStatError` to fail silently without calling back into `displayErrorMsg`).",
>   "comment_type": "bug",
>   "corrected_text": "",
>   "end_line": 26,
>   "file_location": "src/js/plugin/logger.ts",
>   "fixed": false,
>   "original_text": "",
>   "severity": "critical",
>   "start_line": 17
> }
> Stored bug comment [ID: 004] for lines 17-26 in src/js/plugin/logger.ts
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "`react-qr-code` expects `style` to be a single style object, but this passes an array (`[{...}, rest.style]`) which can break rendering in React DOM. Merge styles into one object (e.g., `{ width: '100%', height: '100%', ...(rest.style || {}) }`). Also, the `logDebug` template literal includes an extra `'+` fragment/newline, so the logged message is malformed—build the string purely via template interpolation.",
>   "comment_type": "bug",
>   "corrected_text": "",
>   "end_line": 61,
>   "file_location": "src/js/components/QrCode.tsx",
>   "fixed": false,
>   "original_text": "",
>   "severity": "moderate",
>   "start_line": 37
> }
> Stored bug comment [ID: 005] for lines 37-61 in src/js/components/QrCode.tsx
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "The `set` implementation mutates React state (`prev.set(...)`) before wrapping it in a new Map. This can cause subtle bugs with state updates and makes updates harder to reason about. Create a new Map first (e.g., `const next = new Map(prev); next.set(key, value); return next;`) to avoid mutating prior state.",
>   "comment_type": "best_practices",
>   "corrected_text": "",
>   "end_line": 9,
>   "file_location": "src/js/components/useLeafletCache.ts",
>   "fixed": false,
>   "original_text": "",
>   "severity": "moderate",
>   "start_line": 6
> }
> Stored best_practices comment [ID: 006] for lines 6-9 in src/js/components/useLeafletCache.ts
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "`combinedPromises` collects fulfilled values as arrays, but types `values` as `Promise<any>[]` and then calls `values.flat(1)`. This is inconsistent and can hide real typing mistakes. Change the type to something like `any[]` / `Array<any[]>` (or `Array<BEMData<any>[]>`) and consider renaming `test` to something meaningful (or inline the boolean) so the `localGetMethod` flag is self-documenting.",
>   "comment_type": "maintainability",
>   "corrected_text": "",
>   "end_line": 39,
>   "file_location": "src/js/services/unifiedDataLoader.ts",
>   "fixed": false,
>   "original_text": "",
>   "severity": "moderate",
>   "start_line": 18
> }
> Stored maintainability comment [ID: 007] for lines 18-39 in src/js/services/unifiedDataLoader.ts
> 
> 
> Call to store_comment
> Input
> {
>   "comment_content": "In dev mode (`__DEV__`), missing keys whose fallback value is an object won't be merged because this branch checks `typeof lang[key] === 'object'` while `lang[key]` is `undefined`. That can leave nested translation namespaces missing in dev and potentially break lookups. When `value` is an object and `lang[key]` is missing, initialize `lang[key] = {}` and recurse unconditionally.",
>   "comment_type": "bug",
>   "corrected_text": "",
>   "end_line": 31,
>   "file_location": "src/js/i18nextInit.ts",
>   "fixed": false,
>   "original_text": "",
>   "severity": "moderate",
>   "start_line": 18
> }
> Stored bug comment [ID: 008] for lines 18-31 in src/js/i18nextInit.ts

These are unrelated to the Expo web changes, addressing them in later PR...

PR #1172, opened 2024-08-30
⬆️ Upgrade the phone native code to the latest

shankari:
This fixes https://github.com/e-mission/e-mission-docs/issues/1079


shankari:
The failure in the automated install is again due to running out of space.

```
Warning: An error occurred while preparing SDK package Google Play Intel x86_64 Atom System Image: No space left on device.
```

We see this in the existing codebase as well so it doesn't need to hold up the merge.
Screenshot 2024-08-30 at 9 04 39 AM

But let's do a quick check on how it worked earlier but is broken now
Screenshot 2024-08-30 at 9 05 28 AM

PR #1174, opened 2024-09-06
🐛 Re-enable "Download JSON dump"

shankari:
After the giant refactor, both "email log" and "download JSON dump" were broken. We fixed "email log" in
https://github.com/e-mission/e-mission-phone/pull/1160 but didn't fix "Download JSON dump then, because of lack of time.

This fixes the "Download JSON dump" as well by making it similar to implementation in "email log".

It was already fairly similar, but for some reason, was reading the data before sharing the file

```
             const reader = new FileReader();

             reader.onloadend = () => {
             const readResult = this.result as string;
             logDebug(`Successfull file read with ${readResult.length} characters`);
```

and "this" doesn't exist, resulting in an undefined error. Since the shareObj takes in a file name anyway, I just made this similar to the emailLog changes by passing in the filename directly.

Also, similar ot the "Email Log", added a `.txt` extension so that the file can be sent on iOS.

Testing done:
- Before this change: clicking on "Download JSON dump" did not do anything
- After this change: clicking on "Download JSON dump" launched the share menu

I haven't actually tried sharing yet because gmail is not configured on the emulator.

### DIFF
--- a/package.cordovabuild.json
+++ b/package.cordovabuild.json
@@ -120,7 +120,7 @@
     "cordova-plugin-app-version": "0.1.14",
     "cordova-plugin-customurlscheme": "5.0.2",
     "cordova-plugin-device": "2.1.0",
-    "cordova-plugin-em-datacollection": "git+https://github.com/e-mission/e-mission-data-collection.git#v1.8.9",
+    "cordova-plugin-em-datacollection": "git+https://github.com/e-mission/e-mission-data-collection.git#v1.9.0",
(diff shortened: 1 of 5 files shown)


shankari:
@JGreenlee  after this change, the email log and download JSON dump code is almost the same. Would be good to cleanup the next time we touch this code.

shankari:
| Retrieving file | Share menu |
|-----|----|
| Screenshot 2024-09-06 at 3 30 25 PM | Screenshot 2024-09-06 at 3 31 19 PM |

PR #1175, opened 2024-09-09
Make it easier to debug issues with draft trips

JGreenlee:
Add log statement to debug filtered unprocessed trips

add InspectableWebview = true to making debugging in production easier

### DIFF
--- a/config.cordovabuild.xml
+++ b/config.cordovabuild.xml
@@ -10,6 +10,7 @@
     <content src="index.html" />
     <access origin="*" />
     <hook src="hooks/before_prepare/download_translation.js" type="before_prepare" />
+    <preference name="InspectableWebview" value="true" />
     <preference name="webviewbounce" value="false" />
     <preference name="UIWebViewBounce" value="false" />
     <preference name="DisallowOverscroll" value="true" />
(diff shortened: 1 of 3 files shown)


shankari (approved):

shankari:
@JGreenlee I have pulled this and rebuilt, but I don't think that the best option is to push it to staging, because the staging version is typically compiled with production webpack so I think you won't be able to connect to it anyway. I have uploaded an app zip here; you can download and test in the simulator with the fake locations.

If you would like to test differently, let me know what you would like me to do to generate a debug build for you to use.

[NREL-OpenPATH-build-85.dev.test.app.zip](https://github.com/user-attachments/files/16989162/NREL-OpenPATH-build-85.dev.test.app.zip)

JGreenlee:
@shankari 
Since it seems to only affect draft trips recorded on the phone, but not yet pushed to the server, I don't think that the Simulator will be of any help, unless you know a way to mock a trip on the Simulator.
I was hoping to just reproduce the issue on a test iPhone.

I still think that having these change on staging will help me debug. Besides inspectability, there is a new log statement allow me to view draft trips that are being filtered (which prior logs indicated was the case)

And although staging has Webpack-minified JS, the log statements will still be there to tell me where I am in the code and there's a chance I might find something that way. I have debugged minified JS before. (Remember when we had that Angular annotations bug that only affected prod?)

PR #1177, opened 2024-09-10
Unify android and ios setup scripts

catarial:
General cleanup suggested by @shankari

Android and ios builds can be toggled with SETUP_ANDROID and SETUP_IOS environment variables, both are setup if they aren't set.


catarial:
Oh right I guess this conflicts with our current CI

shankari:
@catarial can you edit the current CI as well? The files are in `.github/workflows`. I want to move from working -> working.

setup/setup_native.sh
```diff
@@ -0,0 +1,141 @@
...
+    if [ $CURR_RUBY_VERSION == $RUBY_VERSION ]; then
+        echo "Found ruby version "$CURR_RUBY_VERSION" expected "$RUBY_VERSION" no need to upgrade"
+    else
+        if [ -x "${WORKING_DIR}/bin/brew" ]; then
+            echo "Found brew installation with version" ` brew --version`
+            echo "Installing ruby version to brew" $RUBY_VERSION
+            brew install ruby@$RUBY_VERSION
+        else
+            if [ $OSX_MAJOR_VERSION -ge $OSX_EXP_VERSION ]; then
+                echo "No brew installation found, but OSX major version "$OSX_MAJOR_VERSION" and expected version "$OSX_EXP_VERSION" so CocoaPods should work"
```
shankari, line 42: nit; future fix: I think we should remove this check. Per @iantei, it doesn't actually work on recent versions of android. We should just say that brew is a requirement.

setup/setup_native.sh
```diff
@@ -0,0 +1,141 @@
...
+            else
+                echo "No brew installation found, but OSX major version "$OSX_MAJOR_VERSION" != expected version "$OSX_EXP_VERSION" CocoaPods install will likely fail"
```
shankari, line 44: nit, future fix: we might as well remove this as well, if we are going to make `brew` required

shankari (approved):
LGTM! I have some nits/suggestions for future polishing

PR #1178, opened 2024-09-12
remove unused survey-resources directory

JGreenlee:


shankari:
🥳 Yay for your first merge to master, exercising that write access!! 🎉

shankari:
@JGreenlee now that this is removed, I don't think we have a copy of the default demographic survey in editable (xls) form anywhere. Could we check in the editable versions somewhere so that people who want to customize the survey(s) can find them?

JGreenlee:
We do have them checked into `nrel-openpath-deploy-configs`, but they are only on the `surveys-info-and-surveys-data` branch.
https://github.com/e-mission/nrel-openpath-deploy-configs/tree/surveys-info-and-surveys-data/survey-resources/data-xls

I think we ought to PR those files from `surveys-info-and-surveys-data` so they don't get lost or forgotten

PR #1179, opened 2024-09-13
cleanup appstate hook, refreshTimeline on app resume, revamp clientstats

JGreenlee:
TimelineContext
When appstate resumes from background, refresh the timeline to avoid stale state

useAppStateChange:
Added stat reading and removed log statement (because the stat reading itself performs a log) Clean up the subscription on teardown (the return of useEffect)

clientStats:
Updated the stat keys to be typed strings, eliminating the need for statKeys. `'my_stat_key'` is less verbose than importing statKeys and calling`statKeys.MY_STAT_KEY`, and this approach utilizes Typescript. Updated existing clientstats readings and removed unused clientstats keys. Added new clientstats readings in useAppStateChange MultiLabelButtonGroup. Unified CLIENT_NAV_EVENT and CLIENT_TIME since they are fundamentally the same and only differ by whether the 'reading' field has a value or is null


JGreenlee:
Not sure what is going on with the checks. Failed once and it seemed like some tests were not able to run.
Now it is stuck on "Setup the serve environment".

shankari:
@JGreenlee tests are now running, but all dynamic config tests are failing because we are trying to read some cordova value

```
    TypeError: Cannot read properties of undefined (reading 'getVersionNumber')

      19 | export function getAppVersion() {
      20 |   if (appVersion) return Promise.resolve(appVersion);
    > 21 |   return window['cordova']?.getAppVersion.getVersionNumber().then((version) => {
         |                                           ^
      22 |     appVersion = version;
      23 |     return version;
      24 |   });

      at getVersionNumber (www/js/plugin/clientStats.ts:21:43)
      at getAppVersion (www/js/plugin/clientStats.ts:29:36)
      at call (www/js/plugin/clientStats.ts:2:1)
      at Generator.tryCatch (www/js/plugin/clientStats.ts:2:1)
      at Generator._invoke [as next] (www/js/plugin/clientStats.ts:2:1)
      at asyncGeneratorStep (www/js/plugin/clientStats.ts:2:1)
      at asyncGeneratorStep (www/js/plugin/clientStats.ts:2:1)
      at _next (www/js/plugin/clientStats.ts:2:1)
      at www/js/plugin/clientStats.ts:2:1
      at apply (www/js/plugin/clientStats.ts:27:29)
      at getStatsEvent (www/js/plugin/clientStats.ts:45:23)
      at call (www/js/plugin/clientStats.ts:2:1)
      at Generator.tryCatch (www/js/plugin/clientStats.ts:2:1)
      at Generator._invoke [as next] (www/js/plugin/clientStats.ts:2:1)
      at asyncGeneratorStep (www/js/plugin/clientStats.ts:2:1)
      at asyncGeneratorStep (www/js/plugin/clientStats.ts:2:1)
      at _next (www/js/plugin/clientStats.ts:2:1)
      at www/js/plugin/clientStats.ts:2:1
      at apply (www/js/plugin/clientStats.ts:43:35)
      at displayErrorMsg (www/js/plugin/logger.ts:24:15)
      at displayErrorMsg (www/js/plugin/logger.ts:14:3)
      at _callee$ (www/js/services/commHelper.ts:30:17)
      at call (www/js/services/commHelper.ts:2:1)
      at Generator.tryCatch (www/js/services/commHelper.ts:2:1)
      at Generator._invoke [as throw] (www/js/services/commHelper.ts:2:1)
      at asyncGeneratorStep (www/js/services/commHelper.ts:2:1)
      at asyncGeneratorStep (www/js/services/commHelper.ts:2:1)

```

It is related to `ClientStats`, so does seem to be related to your changes.

JGreenlee:
### fix tests, record client stats for missing native consent

in dynamicConfig.test.ts, there are some tests where an error is expected to be thrown and error popup is triggered. clientstats is invoked because we are now recording a ui_error stat on all error popups.
Client stats needs mockDevice and mockGetAppVersion to work

in startprefs.test.ts, a similar error popup occurs so the mocks are needed there too.
However, while inspecting startprefs.ts, I noticied the popup occurs when consent document exists in localstorage but not in native storage.
I do not think we need an error popup for this; instead let's do what we do in storage.ts, which is record a 'missing_keys" client stat, and give it a type of 'document_mismatch'.

<hr>

### simplify mocking by specifying jest setup file

Instead of having each test file call the mocks it uses, we can have jest set up all the mocks beforehand.
in jest.config.js, setupFiles points to setupJestEnv.js where all the mocks are called.

This is way less verbose, and may also speed up the execution of the test suites because the mocks are applied once rather than duplicated for every individual test.

We still have flexibility for individual tests to call mocks on their own; for example in enketoHelper.test.ts there are several calls to mockBEMUserCache(<fakeconfig>) because it is testing out various configs

shankari:
>  However, while inspecting startprefs.ts, I noticied the popup occurs when consent document exists in localstorage but not in native storage.
> I do not think we need an error popup for this; instead let's do what we do in storage.ts, which is record a 'missing_keys" client stat, and give it a type of 'document_mismatch'.

Fair. We should record both mismatches (localstorage but not native storage, and native storage but not localstorage) as errors because that is one of the hardest issues to debug.

shankari (approved):
Nice! Much cleaner to put all the init into one place.
I am now merging this, and will push out to staging soon!

PR #1180, opened 2024-09-16
Minor fixes for setup_native.sh

catarial:
The current setup script does not allow only performing a setup for ios or android. This change would fix that by just requiring them to be explicitly set instead of trying to do them both if none are set.

The script now requires brew if the correct ruby installation is not found, regardless of OSX version.

### DIFF
--- a/setup/setup_native.sh
+++ b/setup/setup_native.sh
@@ -3,93 +3,85 @@ set -e
 
 source setup/setup_shared.sh
 
-export PLATFORMS=""
 
-if [ $SETUP_IOS ]; then
-    # The Homebrew pac-man is installed in different locations, depending on whether the processor
-    # is an Intel or Apple Silicone chip.  Intel uses x86_64, Apple chips are amd64, so we can
-    # check the chip type using these hardware platforms.
-    CHIP_ARC=`uname -m`
-    INTEL="x86_64"
-    APPLE_SILICONE="arm64"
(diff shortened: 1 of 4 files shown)


setup/setup_native.sh
```diff
@@ -5,10 +5,7 @@ source setup/setup_shared.sh
 
 export PLATFORMS=""
 
-if [ $SETUP_IOS ] || ( [ !$SETUP_ANDROID ] && [ !$SETUP_IOS ] ); then
-    OSX_MAJOR_VERSION=`sw_vers | grep ProductVersion | cut -d ':' -f 2 | cut -d '.' -f 1`
-    echo "Found OSX major version" $OSX_MAJOR_VERSION
-
+if [ $SETUP_IOS ]; then
```
shankari, line 8: Doesn't this just go back to the behavior we had before, where ruby (and by extension, cocoapods) are only installed on iOS builds? This just happens to work on GitHub actions because the default image has ruby installed. But it will not work on a brand new laptop, for example.

shankari (changes requested):
I would suggest simplifying further and just always installing both iOS and android dependencies. We can also unify the two separate CI/CD builds into one - e.g. `osx-native-install`.

shankari (approved):
LGTM! And I can test it out now on my new laptop as well!

PR #1181, opened 2024-09-25
Sep 2024 Fixes

JGreenlee:
### remove StatusBar plugin

I attempted to unify the appearance of the statusbar between platforms by adding the StatusBar plugin (which had already been referenced in the code).
However, this caused major issues on some Android devices with the entire appbar no longer being visible.
This was not a high priority; rather than spend a lot of time trying to get the right appearance, I am just going to revert it for now

### Don't override fraction digits on SummaryCard

This allows formatForDisplay to determine digits appropriately (e.g. instead of 300.1, just shows 300)

### fix redundant dates showing on Label tab

using isoDatesDifference here (which wraps Luxon's diff) yields decimal values when a full datetime string is provided. If on the same day we are expecting the diff to be 0 (falsy) but get 0.2 or similar (truthy)
All we actually need here is to know if the start and end fall on different dates, so we can simply take the YYYY-MM-DD part of the strings and test for string equality
KISS

### charting: use emcommon color fns; remove unused fns

uses e-mission-common's scale_lightness and dedupe_colors, so no longer need darkenOrLighten and dedupColors
Also remove getMeteredBackgroundColor and createDiagonalPattern which were both unused


www/js/components/charting.ts
```diff
@@ -59,44 +59,44 @@ function getBarHeight(stacks) {
...
-  c.stroke();
-  c.beginPath();
-  c.moveTo(0, 8);
-  c.lineTo(2, 10);
-  c.stroke();
-  return c.createPattern(shape, 'repeat');
-}
+// //fill pattern creation
+// //https://stackoverflow.com/questions/28569667/fill-chart-js-bar-chart-with-diagonal-stripes-or-other-patterns
+// function createDiagonalPattern(color = 'black') {
```
shankari, line 64: nit: why is this commented out? will it be restored or removed in the future?

shankari (approved):
LGTM!

PR #1188, opened 2024-10-25
🐛 Fixes for onboarding initiation if already logged in, UI loading in the background

JGreenlee:
### prevent onboarding initiation if already logged in

Bug:
If already logged in and an nrelopenpath:// link is launched (either via "JOIN" button or external QR scan), onboarding will re-initiate.
Storage from the first login is not cleared before attempting a second login and that could cause major issues.

Fix:
Prevent onboarding if onboarding has already been initiated. When handling an external URL launch we first check the onboarding state to see if the user has already initiated onboarding (i.e. gotten past "WELCOME"). If so, we'll popup with a message.

I also renamed `handleOpenURL` in the AppContext. cordova-plugin-customurlscheme requires a function with this name on the window object, which is why I named it that before. However we use that function to handle raw tokens too; not just URLs, so it is a misnomer.
Renamed it to `handleTokenOrUrl`; then defined window.handleOpenURL as a function that calls `handleTokenOrUrl` with a joinMethod of 'external'. This allows customurlscheme to work while keeping our code descriptive + readable.

<hr>

### prevent UI loading in background; refactor useAppState

```
    // On iOS, the UI can run if the app is launched by the OS in response to a notification,
    // in which case the appState will be 'background'. In this case, we definitely do not want
    // to load the UI because it is not visible.
```

`useAppStateChange` only accepted a callback for onResume. Refactored this into a more versatile hook `useAppState`, which still allows optional onResume or onChange callbacks and also returns the underlying appState.

This allows App.tsx to useAppState and return null if appState is not 'active'. in this case the App.tsx itself will still initialize but it will not have any children

TimelineContext will no longer be rendered while the app is in the background so it no longer needs to listen for app state changes.
usePermissionStatus will still listen for onResume via the new interface, since it is invoked in App.tsx.

### DIFF
--- a/www/js/useAppState.ts
+++ b/www/js/useAppState.ts
@@ -3,26 +3,30 @@
 //the executes "onResume" function that is passed in
 //https://reactnative.dev/docs/appstate based on react's example of detecting becoming active
 
-import { useEffect, useRef } from 'react';
+import { useEffect, useState } from 'react';
 import { AppState } from 'react-native';
 import { addStatReading } from './plugin/clientStats';
 
-const useAppStateChange = (onResume) => {
-  const appState = useRef(AppState.currentState);
+type Props = {
+  onResume?: () => void;
+  onChange?: (nextAppState: string) => void;
+};
+const useAppState = ({ onResume, onChange }: Props) => {
+  const [appState, setAppState] = useState(AppState.currentState);
 
   useEffect(() => {
     const subscription = AppState.addEventListener('change', (nextAppState) => {
-      if (appState.current != 'active' && nextAppState === 'active') {
-        onResume();
+      addStatReading('app_state_change', nextAppState);
+      onChange?.(nextAppState);
+      if (nextAppState == 'active' && appState != 'active') {
+        onResume?.();
       }
-
-      appState.current = nextAppState;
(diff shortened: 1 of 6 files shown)


JGreenlee:
Tests are consistently failing in the workflow but passing locally. It is not due to this PR; I will have to look into it separately

shankari:
This resolves https://github.com/e-mission/e-mission-docs/issues/1095#issuecomment-2436052842

www/js/useAppState.ts
```diff
@@ -3,26 +3,30 @@
...
+  onChange?: (nextAppState: string) => void;
+};
+const useAppState = ({ onResume, onChange }: Props) => {
+  const [appState, setAppState] = useState(AppState.currentState);
 
   useEffect(() => {
     const subscription = AppState.addEventListener('change', (nextAppState) => {
-      if (appState.current != 'active' && nextAppState === 'active') {
-        onResume();
+      addStatReading('app_state_change', nextAppState);
```
shankari, line 19: Great! This should finally allow us to distinguish between the app restarting and the user actually launching the app into the foreground.

shankari (approved):

PR #1190, opened 2024-12-19
register for push notifications on iOS even if UserNotifications denied

JGreenlee:


JGreenlee:
Submitted a PR to https://github.com/havesource/cordova-plugin-push. The repo seems active so let's see if they accept quickly

PR #1192, opened 2024-12-21
Fix the automated sdk install build

shankari:
In https://github.com/e-mission/e-mission-phone/pull/1180 we unified the setup scripts so that we don't have separate android and iOS setup. This is because the plugins include both, so installing the plugins will fail unless the dependencies are installed, which, in turn, depends on having cocoapods installed

6dd71f18321c77cfaf3f67ba19911b4bc282396d fixed all the other github action scripts, but left this one out since it wasn't triggered


shankari:
@JGreenlee  for visibility

PR #1193, opened 2024-12-21
💚 Enable babel transpiling of react-native as well

shankari:
This fixes the code coverage github action.

Without this change, we were getting the error

```
i FAIL  www/__tests__/DateSelect.test.tsx
  ● Test suite failed to run

    SyntaxError: /private/tmp/e-mission-phone/node_modules/react-native/Libraries/vendor/emitter/EventEmitter.js: Unexpected token, expected "]" (39:5)
```

The last successful run was from Oct 17.
The first failed run was from Oct 23.

We have not made any changes to the react-native components that we directly import since Sept 2024. However, we use a carat semver, so we will pull newly released minor versions. One of our dependencies bumped up the version of RN that they use, and released a new minor version, which broke this.

After fruitlessly searching around for a root cause (as documented in the PR comments), since we import a lot of components and it was laborious to check when each of them have been updated, I did a brute force search in GitHub for this error and found both the same issue
https://github.com/facebook/react-native/issues/48228 and a fix that worked
https://github.com/facebook/react-native/issues/48228#issuecomment-2547160218

I am surprised this has not been reported widely, but I guess it is fairly new! Thanks to @Basil-Code for the suggestion!

### DIFF
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -26,6 +26,14 @@ jobs:
           source setup/activate_serve.sh
           npm run test
 
+      - name: Upload exact version(s) of packages used
+        uses: actions/upload-artifact@v4
+        with:
+          name: versions-results
+          path: |
+            package-lock.json
+            ./coverage/coverage-final.json
+
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v4
         with:
--- a/.github/workflows/code-coverage.yml
+++ b/.github/workflows/code-coverage.yml
@@ -24,7 +24,7 @@ jobs:
         shell: bash -l {0}
         run: |
           source setup/activate_serve.sh
-          npx jest
+          npm run test
 
       - name: Upload coverage reports to Codecov
         uses: codecov/codecov-action@v4
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,4 +1,4 @@
 module.exports = {
-  presets: ['@babel/preset-env', '@babel/preset-typescript', '@babel/preset-react'],
+  presets: ['@babel/preset-env', '@babel/preset-typescript', '@babel/preset-react', 'module:@react-native/babel-preset'],
   plugins: ['@babel/plugin-transform-flow-strip-types'],
 }
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,7 @@
+coverage:
+  status:
+    project:
+      default: false  # disable the default status that measures entire project
 comment:                  
   layout: " diff, flags, files"
   behavior: default


shankari:
I thought this would be a simple fix so I didn't file an issue to track progress.
It is a simple fix (one-line), but the investigation was long and complex, so documenting it here for the record.

shankari:
Similar issues were apparently caused due to not using the correct version of node
https://github.com/puppeteer/puppeteer/issues/10509

But bumping up node to 22.12.0 did not fix anything

shankari:
We have not made any changes to the react-native components since Sept 2024. Did this break around that
time? 

https://github.com/e-mission/e-mission-phone/commit/2eab6c54158f9b786d76f19e21555c0d92958483

No, the last successful run was from Oct 17.
The first failed run was from Oct 23.

The most likely cause would be an underlying library change, but we have pinned dependencies.

Where does EventEmitter.js come from anyway?  It is (c) Meta, and is in the
`node_modules/react-native/Libraries/vendor/emitter` directory so appears to
be part of the core react-native. Wait, core react-native implements the interface `IEventEmitter` through `NativeEventEmitter`. Since we use RNweb, I wonder if this is coming from that package.

Nope, RNWeb just uses the EventEmitter.
https://github.com/search?q=repo%3Anecolas%2Freact-native-web%20EventEmitter&type=code

Aha! the `vendor` directory means that it comes from somewhere else
https://stricker.digital/posts/how-to-vendor-npm-dependencies/

So we do have to focus on the react-native package. 
However, I could not figure out where this vendor package was specified, although I checked the `package.json`

We do note that the versions are specified with a carat `^`, so there could have been an upgrade of the version as long as it was a minor version.

Alas, since I don't have an older instance of the "serve" so I can't compare to see what the pinned versions were earlier. But I do note that we have `node_modules/@react-native-community/cli`, version `11.3.10` in the native repos and `node_modules/@react-native/community-cli-plugin`, version `0.76.5` locally.

shankari:
I do see the error

```
There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
    Please check if you are using compatible versions of React Native and React Native Testing Library.
```

RNtesting library comes directly from this project

```
edu.berkeley.eecs.emission@2.5.0 /private/tmp/e-mission-phone
├── @testing-library/react-native@12.9.0
```

We get `react-native` from

```
├─┬ react-native-paper-dates@0.21.8
│ ├── react-native@0.76.5 deduped
├─┬ react-native-paper-tabs@0.10.4
│ ├─┬ react-native-pager-view@6.6.1
│ │ ├── react-native@0.76.5 deduped
│ ├── react-native@0.76.5 deduped
├─┬ react-native-paper@5.12.5
│ ├── react-native@0.76.5 deduped
├─┬ react-native-safe-area-context@4.14.1
│ ├── react-native@0.76.5 deduped
├─┬ @react-navigation/native@6.1.18
│ ├─┬ react-native@0.76.5
│ │ ├─┬ @react-native/virtualized-lists@0.76.5
│ │ │ ├── react-native@0.76.5 deduped
├─┬ @react-navigation/stack@6.4.1
│ │ ├── react-native@0.76.5 deduped
```

React native testing library was released Nov 27. What if we forcibly downgrade it to 12.3.0?

downgraded, still failing

```
edu.berkeley.eecs.emission@2.5.0 /private/tmp/e-mission-phone
├── @testing-library/react-native@12.3.0
```

shankari:
Also found https://dev.to/bytebodger/how-i-fixed-the-unexpected-token-error-in-jest-4o1j and tried to change the command to `jest --transformIgnorePatterns \"node_modules/(?!react-native)/\" --env=jsdom` but that didn't work.

shankari:
@JGreenlee for visibility

PR #1199, opened 2025-02-20
⬆️ 📌 ➕ Fix the build process

shankari:
This fixes several issues from https://github.com/e-mission/e-mission-docs/issues/1113


shankari:
My fix to the WSOD (https://github.com/e-mission/e-mission-phone/pull/1199/commits/7429fff989801c8f817877e9e40b64a658c17647) doesn't actually work because `process.env.ENV` is not actually set to anything.

So this still works for dev/unit tests, since the extra preset is added, but it still generates a WSOD on production because the preset is **always** added. I also note that `webpack.prod.js` has a bunch of `presets` that does not include `module:@react-native/babel-preset`

```
          presets: ['@babel/preset-env', '@babel/preset-react', '@babel/preset-typescript'],
```

so why isn't that overriding the presets defined in `babel.config.js`?

I can verify that if I add the bad preset to `webpack.prod.js` but remove it from `babel.config.js`, then I still get a WSOD, so it looks like it should override the defaults.

@JGreenlee this is not a super high priority, but it would be good to take another look at webpack when you get a chance

PR #1201, opened 2025-03-11
Support mode_studied being an array of multiple modes

JGreenlee:
mode_studied is currently a string, but we want to support multiple modes of interest. We can allow it to be `string | string[]` and handle both cases at runtime

Updated app config typing accordingly

### DIFF
--- a/www/js/survey/multilabel/confirmHelper.ts
+++ b/www/js/survey/multilabel/confirmHelper.ts
@@ -5,8 +5,9 @@ import { LabelOption, LabelOptions, MultilabelKey, InputDetails } from '../../ty
 import { CompositeTrip, InferredLabels, TimelineEntry } from '../../types/diaryTypes';
 import { UserInputMap } from '../../TimelineContext';
 import DEFAULT_LABEL_OPTIONS from 'e-mission-common/src/emcommon/resources/label-options.default.json';
+import AppConfig from '../../types/appConfigTypes';
 
-let appConfig;
+let appConfig: AppConfig;
 export let labelOptions: LabelOptions;
 export let inputDetails: InputDetails<MultilabelKey>;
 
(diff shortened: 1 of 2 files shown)


JGreenlee:
I tested this by locally hosting a modified `stage-program` config to have `["e-bike", "bike"]`

image

<hr>

Then I signed in on the Simulator with one of my real opcodes and added a few fake mode labels.
The walk trip does not prompt for replaced mode, but the bike and e-bike trips do

www/js/survey/multilabel/confirmHelper.ts
```diff
@@ -71,9 +72,13 @@ export function getLabelInputDetails(appConfigParam?) {
 export function labelInputDetailsForTrip(userInputForTrip, appConfigParam?) {
   if (appConfigParam) appConfig = appConfigParam;
   if (appConfig.intro.mode_studied) {
-    if (userInputForTrip?.['MODE']?.data?.label == appConfig.intro.mode_studied) {
-      logDebug(`Found trip labeled with ${userInputForTrip?.['MODE']?.data?.label}, mode of study = ${appConfig.intro.mode_studied}.
-        Needs REPLACED_MODE`);
+    const modesStudied =
+      typeof appConfig.intro.mode_studied == 'string'
+        ? [appConfig.intro.mode_studied]
+        : appConfig.intro.mode_studied;
```
shankari, line 78: we can remove this code once all the existing programs with a single mode of interest have been archived, right?

Can we add a TODO: here to flag that?

shankari (approved):

www/js/survey/multilabel/confirmHelper.ts
JGreenlee: If the long-term strategy is for this to always be an array, then yes. But in that case, it may be more appropriately named as `modes_studied`
And we'll need to change https://github.com/e-mission/nrel-openpath-deploy-configs/blob/38b12c8f8334d3156e70390b0f15ab92ef4eb427/.github/actions/convertIssue/parse-issue-body.js#L299

My thought was that we'd just keep `mode_studied` as `string | string[]` since the canonical program has one mode of study, and supporting multiple modes of study is a secondary use case

PR #1204, opened 2025-06-30
Fix dependency resolution issues with React 18

JGreenlee:


shankari:
It looks like the test still failed because of an e-mission-common install issue.

JGreenlee:
The issue is not e-mission-common, it is that we are only on React 18 and several of our dependencies define React `*`, which resolves to React 19 if we do not override

PR #1250, opened 2026-04-20
💚 Fix malformed "build"and "sdk install" workflow yml files

JGreenlee:
In https://github.com/e-mission/e-mission-phone/pull/1238 we made changes that were supposed to fix the "build" and "sdk install" workflows by adding a step to upgrade bash. However, these files are malformed (not valid yaml syntax) and the workflows do not run. The build and sdk install workflows didn't even run for that commit/PR, which gave the false appearance of "all checks passing" and I think is the reason we didn't see this.

Fixed syntax


JGreenlee:
All the workflows ran this time and passed. Merging

PR #1263, opened 2026-06-28
⬆️  Support xcode 26

shankari:
All builds submitted to the app store after Apr 2026 must be built by xcode 26+
This is consistent with https://github.com/e-mission/e-mission-docs/issues/1144#issuecomment-4824716695

Image

### DIFF
--- a/.github/workflows/ios-build.yml
+++ b/.github/workflows/ios-build.yml
@@ -22,7 +22,7 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
-    runs-on: macos-15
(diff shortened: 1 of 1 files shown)


shankari:
Argh! the checks passed!

Screenshot 2026-06-27 at 9 18 59 PM

But that it because it looks like the default xcode for that environment was 16.4

```
Run xcode-select --print-path
/Applications/Xcode_16.4.app/Contents/Developer
```
Trying `-26` instead...

shankari:
Subsumed by https://github.com/e-mission/e-mission-phone/pull/1264

PR #1273, opened 2026-08-18
Initial implementation of the client-side vehicle library

shankari:


shankari:
@JGreenlee this is a good checkpoint for you to sync and work in parallel.
At this point, the basic e2e flow should work except that I get a 400 error from bikeep while trying to unlock the dock.
- Setup the card (works)
- Setup, checkout, checkin buttons are all enabled/disabled correctly
- Refresh button pulls the setup state and rental history, similar to the `onActive` callback
- Current rental state is populated from the rental history so the server is the source of truth
- there is a button to pull the current set of bikeep locations

Couple of high-level TODOs:
1. This doesn't have the "ticking clock" effect that we had earlier. Is that important? How do we restore that/reconcile with the server?
2. How do we handle simulation mode? (related to the first question)
3. The current status check call is always the "heavyweight" one that goes all the way to the stripe servers. We need this when we are in the setup flow, but maybe we can switch to the lighter weight call that only looks at the version on our server when we are out of that flow.

| Control buttons | List of locations | Rental history |
|-----|----|-----|
simulator_screenshot_DEBBFC3B-5B46-45F7-804A-AEE29BB0737E | Simulator Screenshot - emission - 2026-08-21 at 22 38 13 | Simulator Screenshot - emission - 2026-08-21 at 22 39 56 |

shankari:
The modal dialog tests were consistently failing, so rewrote them to be consistent with the DateModal so that they could be tested the same way, and because @JGreenlee is going to make all this be pretty anyway. The functionality still works and now the tests work too.

| return dialog | checkout initial accessory question | checkout dialog | 
|-----|-----|-----|
| Simulator Screenshot - emission - 2026-08-21 at 23 49 33 | Simulator Screenshot - emission - 2026-08-21 at 23 49 25 | Simulator Screenshot - emission - 2026-08-21 at 23 47 14 |

JGreenlee:
I tested this e2e on Friday with https://github.com/e-mission/e-mission-server/pull/1099 and find that payment setup -> checkout and unlock -> return and lock flow works well.

I have rebased this branch to incorporate the changes from https://github.com/e-mission/e-mission-phone/pull/1169, which will make it easier to incorporate the UI I have been building into this PR.
Awaiting checks before merging https://github.com/e-mission/e-mission-phone/pull/1169 and will then push the rebase of this branch

PR #1286, opened 2026-09-08
🐛 Fix logic to only render timeline when 'label' tab is active, rename

JGreenlee:
We added a hack a couple years ago (c38c35362c40f0ac42c8bcb8aea6dfe36e0462b7) to work around an issue with Leaflet maps rendering by skipping rendering of the Label tab altogether unless it was active. Using `!index` to check this implicitly assumed that the index of label tab was 0 (i.e. the first tab) which is no longer true of all our configurations, causing a blank page to render with no errors logged. Changing this to explicitly check the key of the active route fixes the issue (reproduced + tested manually in Simulator)

Also renamed 'setShouldUpdateTimeline' to 'setShouldRenderTimeline' since that better describes what it does.

### DIFF
--- a/src/js/Main.tsx
+++ b/src/js/Main.tsx
@@ -79,9 +79,10 @@ const Main = () => {
   );
 
   useEffect(() => {
-    const { setShouldUpdateTimeline } = timelineContext;
+    const { setShouldRenderTimeline } = timelineContext;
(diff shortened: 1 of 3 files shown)


JGreenlee:
Ignoring codecov for this quick bugfix, merging

PR #1287, opened 2026-09-09
💄 UI tweaks for bike library demo

JGreenlee:
allow text to break lines and not be cut off, show deployment name on Active Rental page subtitle, adjust vertical margin of rental history section

Testing: manually in browser, tried small window size and long vehicle names


JGreenlee:
image

Text now wraps even if the display is very narrow

Subtitle now shows deployment name: "Smart Commute eBike Library"

